Issue #5, opened 2026-03-09
[Bug] Postgres Insert/Update not working with auto-generated columns


BtXin:
Related: https://github.com/datafusion-contrib/datafusion-table-providers/issues/538

Issue #3, opened 2026-03-09
[Feature Request] Support sqlite


shd101wyy:
It would be great if skardi can support sqlite so developers can test the tool locally more easily.

PR #34, opened 2026-03-29
Add catalog provider for postgres

reluctanthero104:
Adds the support for catalog provider on postgres. Note that datafusion doesn't support schema provider registration, so we will only keep the hierarchy into table & catalog for now.

### DIFF
--- a/crates/skardi/src/sources/providers/sqlx/pg/postgres.rs
+++ b/crates/skardi/src/sources/providers/sqlx/pg/postgres.rs
@@ -1228,6 +1228,7 @@ mod tests {
                 None,
                 false,
                 None,
+                None,
             )
             .await;
 
@@ -1641,6 +1642,7 @@ mod tests {
             "postgresql://127.0.0.1:19999/nonexistent",
             None,
             false,
+            None,
             Some(crate::sources::HierarchyLevel::Catalog),
         )
         .await;
@@ -1661,6 +1663,7 @@ mod tests {
             "postgresql://localhost:5432/db",
             None,
             false,
+            None,
             Some(crate::sources::HierarchyLevel::Table),
         )
         .await;
@@ -1684,6 +1687,7 @@ mod tests {
             "postgresql://localhost:5432/db",
             None,
             false,
+            None,
             None, // None → Table
         )
         .await;
@@ -1765,6 +1769,7 @@ mod tests {
             "postgresql://127.0.0.1:5432/mydb?sslmode=disable",
             Some(&options),
             true,
+            None,
             None,
         )
         .await
--- a/crates/skardi/src/sources/providers/sqlx/pg/postgres.rs
+++ b/crates/skardi/src/sources/providers/sqlx/pg/postgres.rs
@@ -1,4 +1,5 @@
-use crate::sources::providers::DatasetRegistry;
+use crate::sources::providers::sqlx::pg::knn_table_function::{PgKnnEntry, fetch_table_columns};
+use crate::sources::providers::{DatasetEntry, DatasetRegistry};
 use anyhow::{Context, Result};
 use arrow::array::{RecordBatch, UInt64Array};
 use arrow::datatypes::{DataType, Field, Schema, SchemaRef};
@@ -65,8 +66,15 @@ pub async fn register_postgres_tables(
     let hierarchy_level = hierarchy_level.unwrap_or_default();
     match hierarchy_level {
         crate::sources::HierarchyLevel::Catalog => {
-            register_postgres_catalog(session_ctx, name, connection_string, options, read_write)
-                .await
+            register_postgres_catalog(
+                session_ctx,
+                name,
+                connection_string,
+                options,
+                read_write,
+                pg_knn_registry,
+            )
+            .await
         }
         crate::sources::HierarchyLevel::Table => {
             register_single_postgres_table(
@@ -75,6 +83,7 @@ pub async fn register_postgres_tables(
                 connection_string,
                 options,
                 read_write,
+                pg_knn_registry,
             )
             .await
         }
@@ -88,6 +97,7 @@ async fn register_single_postgres_table(
     connection_string: &str,
     options: Option<&HashMap<String, String>>,
     read_write: bool,
+    pg_knn_registry: Option<&DatasetRegistry>,
 ) -> Result<()> {
     let mode_str = if read_write {
         "read-write"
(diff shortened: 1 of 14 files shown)


crates/skardi/src/sources/providers/sqlx/pg/postgres.rs
```diff
@@ -58,6 +62,99 @@ pub async fn register_postgres_tables(
     options: Option<&HashMap<String, String>>,
     read_write: bool,
     pg_knn_registry: Option<&DatasetRegistry>,
+    hierarchy_level: Option<crate::sources::HierarchyLevel>,
+) -> Result<()> {
+    let hierarchy_level = hierarchy_level.unwrap_or_default();
+    match hierarchy_level {
+        crate::sources::HierarchyLevel::Catalog => {
```
BtXin: **#1 — Project guideline violation: inline `crate::` paths (must fix)**

`crate::sources::HierarchyLevel` is used inline ~19 times throughout this file (function signatures, match arms, tests) instead of being imported at the top.

The project `.claude/CLAUDE.md` states: *"When using types from crates, import them at the top of the file with `use` statements. Never use full crate paths inline in function bodies."*

The `cli` and `server` crates already do this correctly. Fix:
```rust
use crate::sources::HierarchyLevel;
```
Then replace all `crate::sources::HierarchyLevel` occurrences with `HierarchyLevel`.

crates/skardi/src/sources/providers/sqlx/pg/postgres.rs
```diff
@@ -71,30 +168,181 @@ pub async fn register_postgres_tables(
...
+    options: Option<&HashMap<String, String>>,
+    read_write: bool,
+    pg_knn_registry: Option<&DatasetRegistry>,
+) -> Result<()> {
+    let mode_str = if read_write {
+        "read-write"
+    } else {
+        "read-only"
+    };
+    tracing::info!(
```
BtXin, line 220: **#3 — Connection string logged in plaintext (security)**

Both `register_single_postgres_table` (line 165) and `register_postgres_catalog` (line 219) log `connection_string` directly. If the connection string contains embedded credentials (e.g. `postgresql://user:password@host/db`), they will appear in logs.

Consider redacting the password portion or logging only the host/database.

crates/skardi/src/sources/providers/sqlx/pg/postgres.rs
```diff
@@ -114,90 +362,119 @@ pub async fn register_postgres_tables(
...
+        "SELECT table_schema, table_name
+         FROM information_schema.tables
+         WHERE table_type = 'BASE TABLE'
+           AND table_schema NOT IN ('pg_catalog', 'information_schema', 'pg_toast')
+           AND table_schema NOT LIKE 'pg_temp_%'
+         ORDER BY table_schema, table_name",
+    )
+    .fetch_all(pool)
+    .await?;
+    if let Some(allowed) = allowed_schemas {
```
BtXin: **#2 — `allowed_schemas` filtering happens client-side (suggestion)**

All non-system tables are fetched from Postgres, then filtered in Rust here. For databases with many schemas/tables, it would be more efficient to push the filter into the SQL with a `WHERE table_schema = ANY($1)` bind parameter, avoiding transfer of rows that are immediately discarded.

Not a blocker for most use cases, but worth considering for large Postgres instances.

crates/skardi/src/sources/providers/sqlx/pg/postgres.rs
```diff
@@ -114,90 +362,119 @@ pub async fn register_postgres_tables(
...
 
-        tracing::info!("Registered '{}' in pg_knn registry for vector search", name);
+async fn list_postgres_tables_in_catalog(
+    pool: &PgPool,
+    allowed_schemas: Option<&Vec<String>>,
+) -> Result<Vec<(String, String)>> {
+    let mut rows: Vec<(String, String)> = sqlx::query_as(
+        "SELECT table_schema, table_name
+         FROM information_schema.tables
+         WHERE table_type = 'BASE TABLE'
```
BtXin: **#5 — Only `BASE TABLE` is loaded, but README says "every table and view"**

This query filters `table_type = 'BASE TABLE'`, which excludes views. The README (line 554) says "every table and view." Either update the README to say "every table" or add `OR table_type = 'VIEW'` to the query.

crates/skardi/src/sources/providers/sqlx/pg/postgres.rs
```diff
@@ -71,30 +168,181 @@ pub async fn register_postgres_tables(
         mode_str,
     );
 
-    let table_name = options.and_then(|opts| opts.get("table")).ok_or_else(|| {
-        anyhow::anyhow!("PostgreSQL data source '{}' requires 'table' option", name)
-    })?;
-
     let schema_name = options
         .and_then(|opts| opts.get("schema"))
         .unwrap_or(&"public".to_string())
```
BtXin: **#8 — Minor: unnecessary allocation on default path**

`.unwrap_or(&"public".to_string())` allocates a `String` on every call, even when `options` contains a `schema` key. Consider:
```rust
let schema_name = options
    .and_then(|opts| opts.get("schema"))
    .map(|s| s.clone())
    .unwrap_or_else(|| "public".to_string());
```

demo/postgres/README.md
```diff
@@ -548,6 +548,220 @@ lsof -i :5432
 ```
 Either stop the conflicting service or use a different port for the Docker container.
 
+## Catalog Mode: Load an Entire Schema at Once
+
+Instead of adding one entry per table, omit the `table` option to let Skardi
+discover and register **every table and view** in the target schema
```
BtXin, line 723: **#5 (cont.) — Doc/code inconsistency**

The code in `list_postgres_tables_in_catalog` only loads `BASE TABLE` (not views). This line should say "every table" instead of "every table and view", or the query should be updated to also include views.

BtXin (commented):
## Code Review: Add catalog provider for Postgres

### Overall

Well-structured PR. The decomposition of the monolithic `register_postgres_tables` into focused helpers (`init_postgres_pools`, `build_postgres_table_provider`, `build_and_register_table`, `register_table_in_knn_registry`) is clean and makes both code paths easy to follow. Config validation is solid, test coverage is thorough, and the README additions are excellent.

See inline comments for specific suggestions. Items **#1** (import style) and **#3** (credential logging) should be addressed before merge; the rest are improvements that could be follow-ups.

PR #39, opened 2026-03-30
Add website generation from Docusaurus

reluctanthero104:
PR to generate website doc using Docusaurus


BtXin (commented):
How about let's put the docs web into a separate repo.

reluctanthero104:
@BtXin It's a bit debatable, as managing it separately means we cannot benefit from any automation for doc gen based on features. Although eventually we should have a doc writer to take care of doc updates, probably could build an agent for doing that.

reluctanthero104:
Closing as we are building the docs separately here: https://github.com/SkardiLabs/skardi-docs

PR #67, opened 2026-04-09
Created RAG demo for fastGPT

BtXin:
Created a RAG demo for fastGPT


reluctanthero104 (commented):
The demo has been verified and looks good.

demo/rag/pipelines/ingest.yaml
```diff
@@ -0,0 +1,18 @@
+metadata:
```
reluctanthero104, line 1: No merge, just demo for now.

PR #69, opened 2026-04-10
Add pg_fts table function for PostgreSQL full-text search

BtXin:
Implements `pg_fts(table, text_col, query, limit)` using `websearch_to_tsquery` for query parsing and `ts_rank` for relevance scoring. Supports web-search-style syntax (phrases, OR, negation) and WHERE clause filter pushdown.

Also refactors `build_column`/`rows_to_batch` from knn_exec into shared utils.rs so both pg_knn and pg_fts use the same row-to-Arrow conversion without a cross-dependency.


crates/skardi/src/sources/providers/sqlx/pg/fts_table_function.rs
```diff
@@ -0,0 +1,654 @@
...
+        // when real values are provided.
+        if !query.is_empty() && limit > MAX_FTS_LIMIT {
+            return plan_err!(
+                "pg_fts: limit must be between 1 and {}, got {}",
+                MAX_FTS_LIMIT,
+                limit
+            );
+        }
+        // Use a safe default when limit is a NULL placeholder (0).
+        let limit = if limit == 0 { 1 } else { limit };
```
reluctanthero104, line 97: extract_int accepts 0 for Int64(0) (it only rejects negatives), so a user who writes pg_fts('articles',
  'body', 'query', 0) bypasses the validation and silently gets 1 result instead of an error. The
  NULL-placeholder-to-0 path (ScalarValue::Null → Ok(0)) is intentional, but a literal 0 should be rejected.

  Fix: change the negative check in extract_int to *v <= 0 (or equivalently, add a 0 check before the
  null-to-1 coercion). Same issue exists for Int32.

crates/skardi/src/sources/providers/sqlx/pg/fts_table_function.rs
```diff
@@ -0,0 +1,654 @@
...
+fn extract_int(expr: &Expr, name: &str) -> DFResult<usize> {
+    match expr {
+        Expr::Literal(ScalarValue::Int64(Some(v)), _) => {
+            if *v < 0 {
+                return plan_err!("pg_fts: '{}' must be non-negative, got {}", name, v);
+            }
+            Ok(*v as usize)
+        }
+        Expr::Literal(ScalarValue::Int32(Some(v)), _) => {
+            if *v < 0 {
```
reluctanthero104: should consider blocking 0 here too

crates/skardi/src/sources/providers/sqlx/pg/fts_table_function.rs
```diff
@@ -0,0 +1,654 @@
...
+            }
+        };
+
+        // Build output schema: all columns + _score.
+        let mut fields: Vec<Field> = entry
+            .columns
+            .iter()
+            .map(|(name, dtype)| Field::new(name.clone(), dtype.clone(), true))
+            .collect();
+        fields.push(Field::new("_score", DataType::Float64, true));
```
reluctanthero104, line 126: If the source table has a column literally named _score, select_columns() silently drops it (the filter(|f|
   f.name() != "_score") guard), so the output schema advertises _score but it resolves entirely to the
  computed ts_rank value — the original column's data is silently lost. This is the same issue that exists in
   pg_knn so it's at least consistent, but it's worth noting since pg_fts includes all columns (not just
  non-vector ones), which raises the collision probability.

crates/skardi/src/sources/providers/sqlx/pg/fts_exec.rs
```diff
@@ -0,0 +1,442 @@
...
+    }
+
+    /// Build the full-text search SELECT query.
+    ///
+    /// Uses `$1` as the bind parameter for the search query string.
+    fn build_query(&self) -> String {
+        let cols = self.select_columns();
+        let text_col = format!("\"{}\"", self.text_col.replace('"', "\"\""));
+        let lang = &self.language;
+        let tsvec = format!("to_tsvector('{lang}', {text_col})");
```
reluctanthero104, line 133: The generated SQL has to_tsvector(...) in both the WHERE condition and ts_rank() in the SELECT. PostgreSQL
  won't CSE these. On tables without a GIN/GIST index on to_tsvector('lang', col), every matched row pays the
   conversion cost twice. The standard fix is to use a generated tsvector column or a CTE, but a simpler
  option is ts_rank(to_tsvector(...), q) > 0 — though the real mitigation is just documenting that users
  should create a tsvector-indexed column for large tables.

demo/postgres/README.md
```diff
@@ -494,6 +494,136 @@ JOIN csv_orders o ON o.user_id = v.id
...
+curl -X POST http://localhost:8080/fts-search-with-filter/execute \
+  -H "Content-Type: application/json" \
+  -d '{"query": "neural network", "category": "ai", "limit": 5}' | jq .
+```
+
+**Basic search** — results ranked by `ts_rank` (higher is more relevant):
+```json
+{
+  "data": [
+    {"id": 1, "title": "Intro to Machine Learning",  "category": "ai", "_score": 0.075990885},
```
reluctanthero104: My computed score is different:

```
{
  "success": true,
  "data": [
    {
      "id": 1,
      "title": "Intro to Machine Learning",
      "category": "ai",
      "_score": 0.09910321980714798
    },
    {
      "id": 4,
      "title": "Deep Learning Advances",
      "category": "research",
      "_score": 0.09910321980714798
    }
  ],
  "rows": 2,
  "execution_time_ms": 252,
  "timestamp": "2026-04-10T18:39:38.994683+00:00"
}
```

demo/postgres/README.md
```diff
@@ -494,6 +494,136 @@ JOIN csv_orders o ON o.user_id = v.id
...
+  --pipeline demo/postgres/pipelines/fts_demo/ \
+  --port 8080
+
+# Basic search
+curl -X POST http://localhost:8080/fts-search/execute \
+  -H "Content-Type: application/json" \
+  -d '{"query": "machine learning", "limit": 5}' | jq .
+
+# With category filter
+curl -X POST http://localhost:8080/fts-search-with-filter/execute \
```
reluctanthero104, line 604: Probably add the sample result too?
```
{
  "success": true,
  "data": [
    {
      "id": 1,
      "title": "Intro to Machine Learning",
      "category": "ai",
      "_score": 0.09910321980714798
    },
    {
      "id": 5,
      "title": "Neural Network Architectures",
      "category": "ai",
      "_score": 0.09910321980714798
    }
  ],
  "rows": 2,
  "execution_time_ms": 295,
  "timestamp": "2026-04-10T18:39:47.725679+00:00"
}
```

PR #71, opened 2026-04-11
Add sqlite_fts and sqlite_knn table functions for SQLite 

BtXin:
- sqlite_fts(table, text_col, query, limit): Full-text search using FTS5 inverted index with BM25 scoring via bm25(). Higher _score = more relevant.

- sqlite_knn(table, vector_col, query_vec, k): KNN vector search using sqlite-vec vec0 virtual tables. Supports literal vectors, candle() UDF embeddings, and scalar subqueries for seed-based similarity search.

- vec_to_binary() scalar UDF: Converts List<Float32> (from candle/gguf/ remote_embed) to packed f32 BLOB for vec0 INSERT.

- Refactored sqlite.rs into sqlite/ directory module with separate files for FTS, KNN, and vec_to_binary implementations.

- Extended knn_utils with List<Float32>, BinaryArray (packed f32 BLOB), and parse_f32_blob() support for cross-backend vector extraction.

- Added extensions/extensions_env options for loading SQLite extensions (e.g. sqlite-vec) at connection time.

- Empty-type virtual table columns (vec0) correctly mapped: pk → Int64, non-pk → Binary.

- 35 unit tests + 40 integration tests covering FTS search, KNN search, subquery vectors, read-your-own-write, and empty-type column handling.

- Demo pipelines and setup script for both FTS5 and sqlite-vec workflows.


crates/skardi/src/sources/providers/sqlite/fts_table_function.rs
```diff
@@ -0,0 +1,698 @@
...
+
+        // Accept NULL/empty placeholders during pipeline validation.
+        if !query.is_empty() && limit > MAX_FTS_LIMIT {
+            return plan_err!(
+                "sqlite_fts: limit must be between 1 and {}, got {}",
+                MAX_FTS_LIMIT,
+                limit
+            );
+        }
+        let limit = if limit == 0 { 1 } else { limit };
```
reluctanthero104: Lines 68–74 reject non-positive integers with a clear error, but L75 then converts 0 to 1
  silently. The zero case is never reached by that guard only if the Expr::Literal path is hit. If
  limit arrives as 0 via a different path, the user gets limit=1 with no indication. Either the
  guard should cover it or the fallback should be removed.

crates/skardi/src/sources/providers/sqlite/fts_exec.rs
```diff
@@ -0,0 +1,401 @@
...
+                move |conn| -> std::result::Result<_, tokio_rusqlite::rusqlite::Error> {
+                    let mut stmt = conn.prepare(&sql)?;
+                    let mut col_values: Vec<Vec<tokio_rusqlite::rusqlite::types::Value>> =
+                        (0..num_cols).map(|_| Vec::new()).collect();
+
+                    let mut rows = stmt.query([&query])?;
+                    while let Some(row) = rows.next()? {
+                        for col_idx in 0..num_cols {
+                            let val: tokio_rusqlite::rusqlite::types::Value =
+                                row.get_unwrap(col_idx);
```
reluctanthero104: get_unwrap() panics on schema mismatch:
  Three cases: `fts_exec.rs L168, knn_exec.rs L219, mod.rs L636` row.get_unwrap(col_idx). If the SQLite schema doesn't match what the code expects
  (e.g. a column is NULL where the code doesn't expect it, or the index is out of range), this
  panics at runtime instead of returning a DataFusion error. Should use row.get(col_idx)? or
  .map_err(...).

crates/skardi/src/sources/providers/sqlite/knn_exec.rs
```diff
@@ -0,0 +1,464 @@
...
+            .collect();
+
+        let col_values: Vec<Vec<tokio_rusqlite::rusqlite::types::Value>> = conn
+            .call(
+                move |conn| -> std::result::Result<_, tokio_rusqlite::rusqlite::Error> {
+                    let mut stmt = conn.prepare(&sql)?;
+
+                    // Convert query vector to packed f32 bytes for sqlite-vec MATCH.
+                    let vec_bytes: Vec<u8> =
+                        query_vector.iter().flat_map(|f| f.to_le_bytes()).collect();
```
reluctanthero104, line 220: query_vector values are packed directly to bytes with f.to_le_bytes(). NaN or Inf passed to
  sqlite-vec will produce silent incorrect results or undefined behaviour from the C extension. A
  simple query_vector.iter().all(|f| f.is_finite()) guard would catch this.

crates/skardi/src/sources/providers/sqlite/mod.rs
```diff
@@ -239,13 +292,30 @@ impl SqliteTableProvider {
...
     conn.call(
         move |conn| -> std::result::Result<(), tokio_rusqlite::rusqlite::Error> {
             conn.pragma_update(None, "journal_mode", "WAL")?;
             conn.pragma_update(None, "busy_timeout", timeout)?;
+
+            // Load dynamic extensions (e.g. sqlite-vec for vec0 virtual tables).
+            if !exts.is_empty() {
+                // SAFETY: extension loading is intentionally enabled here based on
+                // user-configured `extensions` option. We disable it again after loading.
+                unsafe { conn.load_extension_enable()? };
```
reluctanthero104, line 338: `load_extension` loads whatever path the user supplies via the extensions_env option. There's no
  check that the path is absolute, doesn't traverse directories, or points to a trusted location. A
  misconfigured or malicious value could load an arbitrary .so/.dylib. Consider validating the path
  (absolute only, no .. components) before loading

crates/skardi/src/sources/providers/sqlite/mod.rs
```diff
@@ -1,12 +1,22 @@
+pub mod fts_exec;
```
reluctanthero104, line 1: SqliteTableProvider follows the project convention, but SQLiteScanExec, SQLiteInsertExec, and
  SQLiteDmlExec use the all-caps SQLite prefix. Should be SqliteScanExec, SqliteInsertExec,
  SqliteDmlExec.

crates/skardi/src/sources/providers/sqlite/vec_to_binary.rs
```diff
@@ -0,0 +1,202 @@
...
+
+    fn name(&self) -> &str {
+        "vec_to_binary"
+    }
+
+    fn signature(&self) -> &Signature {
+        &self.signature
+    }
+
+    fn return_type(&self, _arg_types: &[DataType]) -> DFResult<DataType> {
```
reluctanthero104, line 54: the pass-through behaviour for an already-Binary input is correct, but a short comment would save the next reader from wondering why there's no transformation.

demo/sqlite/README.md
```diff
@@ -286,9 +286,276 @@ options:
...
+```
+
+## Vector Similarity Search (sqlite-vec)
+
+KNN vector search using [sqlite-vec](https://github.com/asg017/sqlite-vec) `vec0` virtual tables with indexed vector search.
+
+### Prerequisites
+
+```bash
+pip install sqlite-vec fastembed
```
reluctanthero104, line 451: should include lancedb too

demo/sqlite/README.md
```diff
@@ -286,9 +286,276 @@ options:
...
+The `candle()` UDF embeds the query text on the fly using a local BERT model. No external API needed.
+
+```bash
+cargo run --bin skardi-server --features candle -- \
+  --ctx demo/sqlite/ctx_sqlite_knn_demo.yaml \
+  --pipeline demo/sqlite/pipelines/knn_demo/ \
+  --port 8080
+
+# Semantic KNN search — query is embedded on the fly
+curl -X POST http://localhost:8080/sqlite-knn-search/execute \
```
reluctanthero104, line 482: Add expected output for the following queries?
```
{
  "success": true,
  "data": [
    {
      "item_id": 1,
      "_score": 0.6727945804595947
    },
    {
      "item_id": 4,
      "_score": 0.7535505890846252
    },
    {
      "item_id": 5,
      "_score": 0.8204019665718079
    }
  ],
  "rows": 3,
  "execution_time_ms": 1167,
  "timestamp": "2026-04-11T20:25:33.397918+00:00"
}
```

crates/skardi/src/sources/providers/sqlite/knn_exec.rs
```diff
@@ -0,0 +1,464 @@
...
+            .filter(|f| f.name() != "_score")
+            .map(|f| quote_sqlite_ident(f.name()))
+            .collect::<Vec<_>>()
+            .join(", ")
+    }
+
+    /// Build the vec0 KNN SELECT query.
+    ///
+    /// Uses `?1` as the bind parameter for the query vector (packed f32 bytes).
+    pub(crate) fn build_query(&self) -> String {
```
reluctanthero104, line 140: Any chance this could be shared with pg knn_exec?

crates/cli/src/main.rs
```diff
@@ -576,6 +576,7 @@ async fn register_source(
                 &resolved,
                 source.options.as_ref(),
                 false,
+                None,
```
reluctanthero104: Is it intentional to not register anything for CLI?

reluctanthero104 (commented):
Overall LG, with a few comments

demo/sqlite/README.md
```diff
@@ -286,9 +286,276 @@ options:
...
 
+## Full-Text Search (FTS5)
+
+SQLite FTS5 provides indexed full-text search with BM25 relevance scoring — no external extension needed.
+
+### Setup
+
+```bash
+# Create an FTS5 virtual table
+sqlite3 demo/sqlite/fts_demo.db << 'EOF'
```
reluctanthero104, line 297: Make sure to add a cleanup step by the end of the demo, and also update .gitignore to avoid committing any sqlite db files

crates/skardi/src/sources/providers/sqlite/knn_exec.rs
BtXin: dialect between pg and sqlite are quiet different

crates/skardi/src/sources/providers/sqlite/knn_exec.rs
BtXin: The finite-value guard is in place at lines 191-199 in run(), just before the bytes are packed at line 220. It runs after the query vector is resolved (literal or subquery path) and rejects NaN/Inf with a DataFusionError::Plan before any bytes reach sqlite-vec.

PR #75, opened 2026-04-12
latest feature update on README

reluctanthero104:


demo/rag/README.md
```diff
@@ -0,0 +1,254 @@
+# Hybrid Search / RAG Demo
+
+This demo shows a complete FastGPT-style hybrid search pipeline using Skardi,
```
reluctanthero104: Should we need to stress FastGPT here?

BtXin: Let's get rid of that

BtXin (commented):
Some followup:
1. let's rename `demo` directory to docs, and put `llm_wiki`, `rag` into a separate `demo` directory. Let CC re-organize the rest docs to more doc oriented style. Also let's rename the  current `onnx_predict` to `movie_recommendation` and put it into `demo` directory and let CC create a doc on `onnx_predict`. You can also add a example simple backend demo into `demo` directory. 
2. For RAG and llm_wiki demo, let's add a skardi-cli local version with sqlite_knn and sqlite_fts (Optional)
3. Let's make the main README.md in root directory more concise, we can put more details in docs
4. For our github.io doc site, let's make it to support different version, we can start from v0.2.0. And we can make it utilize our newly created docs and demo section.

reluctanthero104:
Addressed 1,3, and put 2,4 into issues for future tracking.

PR #76, opened 2026-04-13
Allow pipeline parameter binding for k in lance_knn and pg_knn

BtXin:
lance_knn and pg_knn rejected non-literal k at planning time, so pipeline `{k}` parameters (which surface as NULL during schema inference) could not be used — unlike sqlite_knn which already had a NULL fallback. Hoist the parsing into a shared knn_utils::extract_k helper that handles the NULL inference placeholder and the MAX_KNN_K bound check, and adopt it in all three KNN table functions. Update the lance/pg demo pipelines and their READMEs to take k (and keep limit where it was already a parameter).

### DIFF
--- a/demo/embeddings/gguf/README.md
+++ b/demo/embeddings/gguf/README.md
@@ -104,7 +104,8 @@ cargo run --bin skardi-server --features gguf -- \
 curl -X POST http://localhost:8080/semantic-search-gguf/execute \
   -H "Content-Type: application/json" \
   -d '{
-    "query": "how does similarity search work in vector databases?"
+    "query": "how does similarity search work in vector databases?",
+    "k": 10
   }' | jq .
 ```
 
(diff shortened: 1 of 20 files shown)


demo/embeddings/gguf/README.md
```diff
@@ -186,24 +187,25 @@ curl -X POST http://localhost:8080/semantic-search-gguf/execute \
 # Retrieval-Augmented Generation
 curl -X POST http://localhost:8080/semantic-search-gguf/execute \
   -H "Content-Type: application/json" \
-  -d '{"query": "how to ground LLM responses with retrieved documents"}' | jq .
+  -d '{"query": "how to ground LLM responses with retrieved documents", "k": 10}' | jq .
```
reluctanthero104, line 190: Would this create uncertainty in the runtime perf of embedding calls? 10 - 500 has a large variation, and I think we should only support hard-coded k, or make it a configuration variable to be enforced on global/per pipeline

PR #88, opened 2026-04-15
fix(handlers): extract substitute_sql_params and add unit tests

reluctanthero104:
Investigated a bug report claiming that {user} could corrupt {user_id} via str::replace when HashMap iteration produces short-before-long ordering. The report's premise turned out to be incorrect: the closing `}` in `{param}` notation means the shorter placeholder is never a literal substring of the longer one, so no corruption occurs.

Changes:
- Extract the SQL placeholder substitution loop into a standalone `substitute_sql_params` function (pure, deterministic, testable).
- Keep a longest-first sort on expected_params as a defensive measure against future format changes where the closing delimiter might not be present.
- Add 4 unit tests for substitute_sql_params covering: brace-notation ordering safety, unrelated params, missing params, and quote escaping.

### DIFF
--- a/crates/server/src/handlers.rs
+++ b/crates/server/src/handlers.rs
@@ -456,6 +456,62 @@ pub async fn get_data_sources(
     })))
 }
 
+/// Substitute `{param}` placeholders in `sql` with their SQL-safe values.
+///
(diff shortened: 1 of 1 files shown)


reluctanthero104:
Closing in favor of the long term fix https://github.com/SkardiLabs/skardi/pull/89

PR #89, opened 2026-04-16
fix: use DataFusion parameterized queries to prevent SQL placeholder corruption

BtXin:
Replace string-based parameter substitution in execute_pipeline_by_name with DataFusion's native $param binding via DataFrame::with_param_values. The old str::replace approach silently corrupted SQL when HashMap iteration placed a shorter parameter name (e.g. {user}) before a longer one sharing the same prefix ({user_id}), causing ~50% of affected requests to fail non-deterministically.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8091,6 +8091,7 @@ dependencies = [
  "clap",
  "cookie",
  "datafusion",
+ "datafusion-common",
  "datafusion-federation",
  "datafusion-tracing",
  "lance",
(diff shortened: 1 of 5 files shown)


reluctanthero104 (commented):
How did you catch this error? Would you mind adding some unit tests?

BtXin:
> How did you catch this error? Would you mind adding some unit tests?

From the bug report you sent me

reluctanthero104 (approved):

PR #91, opened 2026-04-20
feat(providers): add SeekDB data source with FTS and KNN support

BtXin:
## Summary

- Adds a first-class SeekDB ([oceanbase/seekdb](https://github.com/oceanbase/seekdb)) provider. SeekDB is MySQL wire-protocol compatible, so the CRUD path reuses `mysql_async` connection machinery; the FTS and KNN pieces are new because the existing MySQL provider doesn't cover either.
- `seekdb_fts(table, text_col, query, limit)` UDTF backed by `MATCH() AGAINST () IN NATURAL LANGUAGE MODE`, with `_score` projected and WHERE filter pushdown via DataFusion's `MySqlDialect` unparser.
- `seekdb_knn(table, vec_col, query_vec, metric, k)` UDTF backed by SeekDB's native `L2_DISTANCE` / `COSINE_DISTANCE` / `INNER_PRODUCT` over `VECTOR` columns with HNSW indexes. Accepts literal vectors and scalar subquery placeholders.

## Why

Teams running SeekDB as their unified search backend were blocked from Skardi because the existing MySQL provider doesn't understand SeekDB's vector and FULLTEXT semantics. Adding a first-class provider lets pipelines federate SeekDB with the rest of the catalog and keeps FTS/KNN symmetric with Lance, Postgres, and SQLite.

## Wiring

- `DataSourceType::Seekdb` variant.
- `DatasetEntry::Seekdb(SeekDbKnnEntry)` in the shared registry.
- `register_seekdb_tables` dispatcher arm in `crates/server/src/config.rs` + `Seekdb` entries in `CATALOG_SUPPORTED_SOURCES`, `WRITABLE_SOURCE_TYPES`, and the two validation match arms.
- `register_seekdb_knn_udtf` / `register_seekdb_fts_udtf` wired through `OptimizerRegistry`.
- SeekDB service container + SQL seed step in `.github/workflows/ci.yml` (ports 2881/2886, creates `users` / `orders` / `articles` with `FULLTEXT` / `docs` with `VECTOR`).

## ⚠️ Caveats to verify on a live SeekDB before merging

This PR was prepared with access to the SeekDB docs but without a live instance to hit, so the following should be verified end-to-end before we merge:

1. **VECTOR literal syntax** — assumes pgvector-style `'[0.1, 0.2, 0.3]'`. If SeekDB requires `CAST('[...]' AS VECTOR(N))`, it's a one-line fix in `format_vector_literal`.
2. **FULLTEXT parser hint** — CI seed uses `WITH PARSER IK`. Some OceanBase builds use `PARSER_NAME=ik`. Adjust CI seed + docs Quick Start if DDL fails.
3. **Default credentials** — assumed `root@sys` with empty password on port 2881 as shipped by `oceanbase/seekdb:latest`.
4. **mysql_async handshake** — SeekDB is MySQL 5.7-compatible; if the 0.36 client emits an unsupported capability flag we may need a handshake tweak.

## Test plan

- [x] `cargo check --workspace --all-targets` passes locally.
- [x] `cargo fmt --all` clean.
- [x] `cargo clippy --workspace --all-targets` — no new warnings in seekdb files.
- [x] `cargo test -p skardi --lib seekdb` — 54 unit tests pass (connection-string parsing, argument validation, SQL-builder shape, metric parsing).
- [x] CI integration tests pass against the new `oceanbase/seekdb:latest` service.
- [x] `docs/seekdb/README.md` Quick Start runs end-to-end against a live SeekDB (spin container → `cargo run --bin skardi-server` → `curl` every pipeline).
- [x] Caveats 1–4 above verified or fixed.

## Docs

- `docs/seekdb/README.md` — Quick Start, data model, 11 pipelines.
- `docs/seekdb/ctx_seekdb_demo.yaml` — demo data source config.
- `docs/seekdb/pipelines/` — point lookup, insert/update/delete, FTS (with/without filter), KNN (literal and by-seed), federated join with CSV, user-order summary.
- Root `README.md` supported-sources table now includes SeekDB.


crates/skardi/src/sources/providers/seekdb/knn_exec.rs
```diff
@@ -0,0 +1,592 @@
...
+    InnerProduct,
+}
+
+impl DistanceMetric {
+    /// The SeekDB distance function name for this metric.
+    pub fn function(self) -> &'static str {
+        match self {
+            DistanceMetric::L2 => "L2_DISTANCE",
+            DistanceMetric::Cosine => "COSINE_DISTANCE",
+            DistanceMetric::InnerProduct => "NEGATIVE_INNER_PRODUCT",
```
reluctanthero104, line 90: The commit message says INNER_PRODUCT in two places, the caveats section says nothing 
  about this, and the SeekDB/OceanBase docs mention INNER_PRODUCT — not               
  NEGATIVE_INNER_PRODUCT. If this function doesn't exist in SeekDB, every inner-product 
  KNN query silently fails at runtime. Either verify against a live instance before
  merge or add an explicit TODO with the SeekDB docs reference.

crates/skardi/src/sources/providers/seekdb/mod.rs
```diff
@@ -0,0 +1,1101 @@
...
+        catalog_name,
+        schema_tables,
+        |schema, table_name| {
+            let pool = Arc::clone(&pool);
+            let registry_c = registry_shared.clone();
+            let catalog_c = catalog_name_owned.clone();
+            let schema_c = schema.clone();
+            let table_c = table_name.clone();
+            async move {
+                let table_reference = TableReference::bare(table_c.as_str());
```
reluctanthero104: For catalog mode the pool is created against a single connection string (one DB), but 
  tables are enumerated across multiple schemas via information_schema. A bare reference
   only works if the table happens to be in the connection's default database. Tables in
   other schemas need TableReference::partial(schema, table), otherwise                 
  MySQLTableFactory will look in the wrong schema.

crates/skardi/src/sources/providers/seekdb/mod.rs
```diff
@@ -0,0 +1,1101 @@
...
+        &self,
+        _partition: usize,
+        _context: Arc<TaskContext>,
+    ) -> DataFusionResult<SendableRecordBatchStream> {
+        let pool = Arc::clone(&self.pool);
+        let sql = self.sql.clone();
+        let schema = Arc::clone(&self.schema);
+
+        let future = async move {
+            let conn_obj = pool.connect_direct().await.map_err(|e| {
```
reluctanthero104: FTS and KNN use pool.connect() (pooled). DML uses connect_direct(), which bypasses the
   pool — one new connection per DELETE/UPDATE statement. Under any write-heavy workload
   this will leak connections up to the OS limit. Should use pool.connect() for         
  consistency unless there's a documented reason for a direct connection here.

PR #98, opened 2026-04-21
feat(jobs): add `kind: job` batch primitive with Lance + SQL DML destinations

BtXin:
Pipelines answer a query; jobs commit the answer somewhere queryable. A job YAML is a parameterized SELECT plus a destination — Lance on disk for lake targets, any read-write DB table for DB targets — run in the background on `skardi-server`, persisted in a SQLite run ledger, and polled via `skardi job run|status|list|cancel` (HTTP endpoints at `/jobs/*`). Submit-time pre-flight diffs the query output schema against the destination; lake destinations commit atomically; crashed runs are reconciled to `failed` on server restart.


crates/cli/src/jobs_cli.rs
```diff
@@ -0,0 +1,258 @@
...
+        #[arg(long = "server", value_name = "URL")]
+        server: Option<String>,
+    },
+}
+
+fn server_base_url(override_url: Option<&str>) -> String {
+    override_url
+        .map(|s| s.to_string())
+        .or_else(|| std::env::var("SKARDI_SERVER_URL").ok())
+        .unwrap_or_else(|| "http://127.0.0.1:8080".to_string())
```
BtXin: Put the default URL into a const

crates/skardi/src/jobs/executor.rs
```diff
@@ -0,0 +1,1000 @@
...
+
+    let cancel_flag = Arc::new(CancelFlag::default());
+    {
+        let mut flags = exec.cancel_flags.lock().unwrap_or_else(|p| p.into_inner());
+        flags.insert(run_id.clone(), Arc::clone(&cancel_flag));
+    }
+
+    let store = Arc::clone(&exec.store);
+    let session_ctx = Arc::clone(&exec.session_ctx);
+    let flags_map = Arc::clone(&exec.cancel_flags);
```
BtXin: Wouldn't it better to use DashMap here than `Arc<Mutex<HashMap>>`?

crates/skardi/src/jobs/store.rs
```diff
@@ -0,0 +1,530 @@
...
+    pub fn path(&self) -> &Path {
+        &self.path
+    }
+
+    async fn ensure_schema(&self) -> Result<()> {
+        self.conn
+            .call(
+                |conn| -> std::result::Result<(), tokio_rusqlite::rusqlite::Error> {
+                    conn.execute_batch(
+                        "CREATE TABLE IF NOT EXISTS job_runs (
```
BtXin: Put this DDL into a const var

crates/skardi/src/jobs/executor.rs
```diff
@@ -0,0 +1,1000 @@
...
+        if let Some(name) = bad_types {
+            return Err(JobSubmitError::UnsupportedParameter(name));
+        }
+
+        // Destination pre-flight.
+        let destination = self.resolve_destination(&job.destination).await?;
+
+        let query_schema = self
+            .session_ctx
+            .sql(&rendered_sql)
```
BtXin, line 237: Does this do the sql read altogether, like read all the data at once and load all of them into a memTable, or does datafusion has the mechanism to do batch internally, cause if the data is too large, then we will have OOM issue.

crates/skardi/src/jobs/destination.rs
```diff
@@ -0,0 +1,402 @@
...
+                    })?
+                    .collect()
+                    .await
+                    .with_context(|| {
+                        format!("Failed to execute overwrite DELETE for '{}'", self.table)
+                    })?;
+            }
+            let insert_sql = format!("INSERT INTO {destination_sql} SELECT * FROM {staging_sql}");
+            self.ctx
+                .sql(&insert_sql)
```
BtXin: Same as read, does this part do it in one shot? What maybe better with small batches?

crates/skardi/src/jobs/destination.rs
```diff
@@ -0,0 +1,402 @@
...
+                } else {
+                    WriteMode::Create
+                }
+            }
+            DestinationMode::Overwrite => WriteMode::Overwrite,
+        };
+        let LanceWriteOutcome {
+            version,
+            rows_written,
+        } = write_lance_dataset(&self.path, batches, write_mode).await?;
```
BtXin: Does write_lance_dataset do small batch write internally?

crates/skardi/src/jobs/destination.rs
```diff
@@ -0,0 +1,402 @@
+//! Destination writers — where a job's rows go.
```
BtXin: Is it better to create a directory on destination and each destination with its own .rs file, and put the trait in mod.rs? So this file will not grow enormously.

crates/skardi/src/jobs/store.rs
```diff
@@ -0,0 +1,530 @@
+//! Run ledger for the jobs primitive.
```
BtXin, line 1: Maybe same with destination.rs, let's put it into a sub-directory and use for each impl let's put it into different rs file and put trait in mod.rs

crates/skardi/src/jobs/executor.rs
```diff
@@ -0,0 +1,1000 @@
...
+/// `cancel` call reach the spawned task before it commits to the
+/// destination.
+#[derive(Debug, Default)]
+struct CancelFlag {
+    cancelled: Mutex<bool>,
+}
+
+impl CancelFlag {
+    fn cancel(&self) {
+        *self.cancelled.lock().unwrap_or_else(|p| p.into_inner()) = true;
```
BtXin: Is it possible to refrain use of deference?

crates/skardi/src/jobs/executor.rs
```diff
@@ -0,0 +1,1000 @@
...
+    }
+
+    fn is_cancelled(&self) -> bool {
+        *self.cancelled.lock().unwrap_or_else(|p| p.into_inner())
```
BtXin: Is it possible to refrain use of deference?

crates/skardi/src/jobs/executor.rs
```diff
@@ -0,0 +1,1000 @@
...
+// ---------------------------------------------------------------------------
+// A more complete executor impl that honours lance paths provided externally.
+// ---------------------------------------------------------------------------
+
+/// Extension trait so callers that have a `lance_paths: HashMap<String,
+/// String>` can submit jobs whose Lance destinations resolve against
+/// those paths. Used by the server.
+#[async_trait::async_trait]
+pub trait JobExecutorExt {
+    async fn submit_with_lance_paths(
```
BtXin: I don't think this is a good name, lance is not an exception, we are going to add more datalake files in the future, if we also added Iceberg, Deltalake, Paimon, etc, are we going to put a submit_with_ for every datalake files?

crates/server/src/jobs_handlers.rs
```diff
@@ -0,0 +1,256 @@
+//! Axum handlers for the jobs primitive.
```
BtXin, line 1: If we call this jobs_handlers, should we rename `handler` to `pipeline_handlers` ?

PR #112, opened 2026-04-26
Bt xin/cohore/readme enhance and llm wiki

BtXin:
Change readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 
 <img src="asset/logo.png" alt="Skardi Logo" width="700">
 
-**Skardi is an agent data engine that gives AI agents data autonomy.**
+**Skardi is an agent data plane that gives AI agents data autonomy.**
 
 <a href="https://skardilabs.github.io/skardi-docs/">Documentation</a> •
 <a href="#roadmap">Roadmap</a> •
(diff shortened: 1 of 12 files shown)


docs/agent_data_plane.md
```diff
@@ -1,21 +1,23 @@
-# Spark for Agents
```
abbccdda, line 1: Consider adding a generated graph on this page to be more declarative on what data engine is, right now it's a bit hard to read and get the knowledge essence quickly

README.md
```diff
@@ -35,9 +35,9 @@
 
 ## What is Skardi?
 
-Skardi is an **open-source data platform for AI agents** — Pick any data in your stack (CSV, Parquet, S3, Postgres, MySQL, SQLite, MongoDB, Redis, Iceberg, Lance, SeekDB) and Skardi turns it into something an agent can query, join, write to, and operate on autonomously — through SQL, REST, shell, and (soon) MCP.
```
abbccdda, line 38: I'm concerned that agent data engine is a vague term, and introduced additional mental burden for readers to understand what Skardi is. `Spark` could clearly state the contract between agent and data, which is clear to the old world engineers. If we name it data engine, people would think we are building a new DB to be placed under Spark API or SQL query. 

Although I do agree that `Spark for agent` is not perfect either as it misses the important details such like Skardi serves SQL queries as endpoints, but it has a good and strong memory anchor to think "ok, Skardis is an engine for agent to query data", instead of misplacing us into a physical DB.

abbccdda: Also, just double checked, Spark is also claimed as an engine, so I think the metaphor here is correct
image

PR #117, opened 2026-04-27
feat(server): support array-of-arrays parameter shape for multi-row VALUES

reluctanthero104:
## Summary

- New tuple-list rendering branch in `substitute_sql_params`: an array whose every element is itself an array becomes `(c1, c2, …), (c1, c2, …)` so a pipeline like `INSERT INTO docs (id, embedding) VALUES {rows}` handles caller-chosen batch sizes without hardcoding N.
- Inner arrays inside a row tuple (e.g. an `embedding` cell) render as the bracketed scalar form `[v1, v2, v3]` so `VECTOR` / pgvector columns accept them as text literals — same shape used by single-row inserts.
- Mixed-shape arrays (some scalar, some array elements) are rejected as `Unsupported` so a misshapen payload can't silently produce malformed SQL.
- The previous flat-scalar-array path (vector literal form) is preserved unchanged — backward compatible.

## Why

Today an INSERT pipeline can only insert one row per call unless the YAML hardcodes a fixed number of named parameters. That blocks batched WAL replay (consumer-side amortisation of the per-row HTTP + parse + plan overhead) for any caller that wants to choose batch size dynamically — concretely, the skardi-cloud CDC consumer's Phase 2 work. With this shape supported, a single pipeline serves both single-row and N-row inserts.

## Changes

- `crates/server/src/pipeline_handlers.rs` — new `scalar_to_sql` / `row_cell_to_sql` helpers + tuple-list branch in `substitute_sql_params`. 6 new unit tests.
- `docs/pipelines.md` — new "Parameter shapes" section documenting the JSON value → SQL literal mapping with a runnable curl example.
- `docs/postgres/pipelines/batch_insert_users.yaml` — runnable example, scalar columns.
- `docs/seekdb/pipelines/batch_insert_users.yaml` — SeekDB counterpart.
- `docs/seekdb/pipelines/batch_insert_docs_with_embeddings.yaml` — runnable example with a VECTOR cell inside the row tuple.

## Test plan

- [x] `cargo build -p skardi-server` — clean
- [x] `cargo test -p skardi-server --lib pipeline_handlers::tests` — **17/17 pass** (6 new + 11 existing)
- [x] `cargo test -p skardi-server --lib` — **109/109 pass**, no regressions
- [ ] Reviewer to spot-check the docs example renders to the expected SQL by running it through any of the pipelines on a live skardi-server

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,12 @@ jobs:
           --health-retries=5
 
       postgres:
-        image: postgres:16
+        # `pgvector/pgvector:pg16` is the upstream Postgres 16 image with
+        # the pgvector extension pre-built in. Used so the `docs` table
+        # can declare an `embedding vector(N)` column and the integration
+        # tests can round-trip vector cells through the multi-row VALUES
+        # renderer without a separate extension build step.
+        image: pgvector/pgvector:pg16
         env:
           POSTGRES_DB: mydb
           POSTGRES_USER: skardi_user
@@ -184,6 +189,7 @@ jobs:
       - name: Seed PostgreSQL data
         run: |
           PGPASSWORD=skardi_pass psql -h 127.0.0.1 -U skardi_user -d mydb <<'EOF'
+          CREATE EXTENSION IF NOT EXISTS vector;
           CREATE TABLE users (
               id SERIAL PRIMARY KEY,
               name VARCHAR(100) NOT NULL,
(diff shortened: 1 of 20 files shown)


BtXin (commented):
Make sure all existing data sources support such grammar: 
```
INSERT INTO docs (id, title, category, embedding) VALUES
        ('d1', 'doc-a', 'books', [1.0, 0.0, 0.0, 0.0]),
        ('d2', 'doc-b', 'music', [0.0, 1.0, 0.0, 0.0]),
        ('d3', 'doc-c', 'books', [0.0, 0.0, 1.0, 0.0])
```

PR #141, opened 2026-06-21
feat(sources): add DynamoDB table provider

abbccdda:
## What

Adds **Amazon DynamoDB** as a read-write data source. A DynamoDB table maps to one Skardi table (item → row, top-level attribute → column).

- Full `TableProvider`: scan, `insert_into`, `delete_from`, `update`
- Binary-filter **pushdown** as DynamoDB `FilterExpression` (`=`, `<>`, `<`, `<=`, `>`, `>=` on column vs. literal)
- INSERT via `PutItem`; UPDATE/DELETE resolve matching keys via a filtered scan then mutate per key (DynamoDB can't mutate by predicate); key columns are immutable
- Schema inferred by sampling one item; `N`→Int64/Float64, `S`→Utf8, `BOOL`→Boolean; missing attribute → SQL NULL
- No FTS/KNN — matches the backend's native capabilities
- `connection_string` is the endpoint URL (DynamoDB Local or AWS regional endpoint); credentials via default AWS chain or `*_env` options

## Wiring

`DataSourceType::Dynamodb` + every exhaustive match site: `config.rs` (dispatch, validation, `WRITABLE_SOURCE_TYPES`, error string), `pipeline_handlers.rs`, `gui.rs`, `jobs/executor.rs` (classified non-transactional), CLI.

## Tests & verification

- 12 unit tests + 4 `#[ignore]`-gated integration tests (scan, filter pushdown, count(*), INSERT→UPDATE→DELETE round-trip) — pass against **DynamoDB Local**
- CI: `amazon/dynamodb-local` service + AWS-CLI seed step
- `docs/dynamodb/` (README + ctx + 7 pipelines); **every doc command was run end-to-end** against `skardi-server`, including a federated DynamoDB+CSV join
- `cargo fmt --check`, `cargo check --workspace --all-targets`, `cargo test -p skardi --lib` (420) and `-p skardi-server` (113) all green

## New dependencies

`aws-sdk-dynamodb`, `aws-config` (no prior AWS SDK in the tree).

### DIFF
--- a/crates/server/src/config.rs
+++ b/crates/server/src/config.rs
@@ -2206,4 +2206,100 @@ options:
             "Duplicate data source name: duplicate_source"
         );
     }
+
+    fn dynamodb_source(
+        name: &str,
+        connection_string: Option<&str>,
+        options: Option<HashMap<String, String>>,
+        access_mode: AccessMode,
+    ) -> DataSource {
+        DataSource {
+            name: name.to_string(),
+            source_type: DataSourceType::Dynamodb,
+            path: PathBuf::new(),
+            connection_string: connection_string.map(String::from),
+            schema: None,
+            options,
+            hierarchy_level: HierarchyLevel::default(),
+            access_mode,
+            enable_cache: false,
+            description: None,
+        }
+    }
+
+    #[test]
+    fn validate_accepts_dynamodb_read_write() {
+        let source = dynamodb_source(
+            "products",
+            Some("http://localhost:8000"),
+            None,
+            AccessMode::ReadWrite,
+        );
+        validate_data_sources(&[source]).expect("dynamodb supports read_write");
+    }
+
+    #[test]
+    fn validate_rejects_dynamodb_without_connection_string() {
+        let source = dynamodb_source("products", None, None, AccessMode::ReadOnly);
+        let err = validate_data_sources(&[source]).unwrap_err();
+        let config_err = err.downcast_ref::<ConfigError>().unwrap();
+        assert!(
+            matches!(
+                config_err,
+                ConfigError::MissingConnectionString { name } if name == "products"
+            ),
+            "got {config_err}"
+        );
+    }
+
+    #[test]
+    fn unsupported_write_mode_error_lists_dynamodb() {
+        let err = ConfigError::UnsupportedWriteMode {
+            name: "events".to_string(),
+            source_type: DataSourceType::Csv,
+        };
+        assert!(err.to_string().contains("'dynamodb'"), "got {err}");
+    }
+
+    #[tokio::test]
+    async fn register_dynamodb_without_connection_string_errors() {
+        let mut session_ctx = SessionContext::new();
+        let source = dynamodb_source("products", None, None, AccessMode::ReadOnly);
+        let err = register_data_sources(&mut session_ctx, &[source])
+            .await
+            .unwrap_err();
+        let config_err = err.downcast_ref::<ConfigError>().unwrap();
+        assert!(
+            matches!(
+                config_err,
+                ConfigError::MissingConnectionString { name } if name == "products"
+            ),
+            "got {config_err}"
+        );
+    }
+
+    #[tokio::test]
+    async fn register_dynamodb_without_options_reports_registration_failure() {
+        let mut session_ctx = SessionContext::new();
+        // `register_dynamodb_tables` fails before any network call when no
+        // options are provided, so this exercises the registration arm and
+        // its error mapping without a live endpoint.
+        let source = dynamodb_source(
+            "products",
+            Some("http://localhost:8000"),
+            None,
+            AccessMode::ReadOnly,
+        );
+        let err = register_data_sources(&mut session_ctx, &[source])
+            .await
+            .unwrap_err();
+        let config_err = err.downcast_ref::<ConfigError>().unwrap();
+        match config_err {
+            ConfigError::DataSourceRegistrationFailed { name, error } => {
+                assert_eq!(name, "products");
+                assert!(error.contains("requires options"), "got {error}");
+            }
+            other => panic!("expected DataSourceRegistrationFailed, got {other}"),
+        }
+    }
 }
--- a/crates/server/src/config.rs
+++ b/crates/server/src/config.rs
@@ -1291,6 +1291,7 @@ async fn register_data_source(
                 &source.name,
                 connection_string,
                 source.options.as_ref(),
+                source.access_mode.is_read_write(),
             )
             .await
             .map_err(|e| {
--- a/crates/server/src/config.rs
+++ b/crates/server/src/config.rs
@@ -5,6 +5,7 @@ use datafusion::prelude::*;
 use serde::{Deserialize, Serialize};
 use skardi::jobs::JobDefinition;
 use skardi::pipeline::pipeline::{Pipeline, StandardPipeline};
+use skardi::sources::providers::dynamodb::register_dynamodb_tables;
 use skardi::sources::providers::iceberg::register_iceberg_table;
 use skardi::sources::providers::influxdb::register_influxdb_tables;
 use skardi::sources::providers::lance::register_lance_table;
@@ -209,7 +210,7 @@ pub enum ConfigError {
     S3ObjectStoreRegistrationFailed { name: String, error: String },
 
     #[error(
-        "Data source '{name}' has access_mode 'read_write' but type '{source_type:?}' does not support write operations. Only 'postgres', 'mysql', 'sqlite', 'mongo', 'redis', and 'seekdb' sources support read_write mode."
+        "Data source '{name}' has access_mode 'read_write' but type '{source_type:?}' does not support write operations. Only 'postgres', 'mysql', 'sqlite', 'mongo', 'redis', 'seekdb', and 'dynamodb' sources support read_write mode."
     )]
     UnsupportedWriteMode {
         name: String,
@@ -717,6 +718,7 @@ const WRITABLE_SOURCE_TYPES: &[DataSourceType] = &[
     DataSourceType::Mongo,
     DataSourceType::Redis,
     DataSourceType::Seekdb,
+    DataSourceType::Dynamodb,
 ];
 
 /// Validate data source configurations
@@ -796,7 +798,8 @@ fn validate_data_sources(data_sources: &[DataSource]) -> Result<()> {
                 | DataSourceType::Mongo
                 | DataSourceType::Redis
                 | DataSourceType::Seekdb
-                | DataSourceType::Influxdb,
+                | DataSourceType::Influxdb
+                | DataSourceType::Dynamodb,
                 false,
             ) => {
                 // For database connections, ensure connection string is provided
@@ -969,7 +972,8 @@ async fn register_data_source(
             | DataSourceType::Mongo
             | DataSourceType::Redis
             | DataSourceType::Seekdb
-            | DataSourceType::Influxdb,
+            | DataSourceType::Influxdb
+            | DataSourceType::Dynamodb,
             _,
         ) => {
             // Database sources don't need file path validation
@@ -1260,6 +1264,41 @@ async fn register_data_source(
             .await
             .map_err(|e| {
                 tracing::error!("MongoDB registration failed for '{}': {:?}", source.name, e);
+                ConfigError::DataSourceRegistrationFailed {
+                    name: source.name.clone(),
+                    error: format!("{:?}", e),
+                }
+            })?;
+        }
+        DataSourceType::Dynamodb => {
+            tracing::info!("Registering DynamoDB table: {}", source.name);
+
+            let connection_string = source.connection_string.as_ref().ok_or_else(|| {
+                ConfigError::MissingConnectionString {
+                    name: source.name.clone(),
+                }
+            })?;
+
+            tracing::debug!(
+                "Endpoint for {}: {} (options: {:?})",
+                source.name,
+                connection_string,
+                source.options
+            );
+
+            register_dynamodb_tables(
+                session_ctx,
+                &source.name,
+                connection_string,
+                source.options.as_ref(),
+            )
+            .await
+            .map_err(|e| {
+                tracing::error!(
+                    "DynamoDB registration failed for '{}': {:?}",
+                    source.name,
+                    e
+                );
                 ConfigError::DataSourceRegistrationFailed {
                     name: source.name.clone(),
                     error: format!("{:?}", e),
--- a/crates/skardi/src/sources/providers/dynamodb.rs
+++ b/crates/skardi/src/sources/providers/dynamodb.rs
@@ -134,20 +134,7 @@ impl DynamoTableProvider {
             );
         }
 
-        // Merge attributes across the sampled items; first observation of an
-        // attribute fixes its type. The SDK item maps iterate in arbitrary
-        // order, so sort by name afterwards to give a deterministic non-key
-        // column order across runs (keys are placed explicitly downstream).
-        let mut attrs: Vec<(String, DataType)> = Vec::new();
-        let mut seen: HashSet<String> = HashSet::new();
-        for item in items {
-            for (key, value) in item.iter() {
-                if seen.insert(key.clone()) {
-                    attrs.push((key.clone(), attribute_value_to_arrow_type(value)));
-                }
-            }
-        }
-        attrs.sort_by(|(a, _), (b, _)| a.cmp(b));
+        let attrs = merge_sampled_attributes(items);
 
         Ok(build_schema_fields(partition_key, sort_key, &attrs))
     }
@@ -156,6 +143,24 @@ impl DynamoTableProvider {
     fn plan_read(&self, pushable: &[Expr]) -> DFResult<DynamoRead> {
         classify_read(&self.partition_key, self.sort_key.as_deref(), pushable)
     }
+}
+
+/// Merge attributes across sampled items; the first observation of an
+/// attribute fixes its type. The SDK item maps iterate in arbitrary order, so
+/// the merged list is sorted by name to give a deterministic non-key column
+/// order across runs (keys are placed explicitly downstream).
+fn merge_sampled_attributes(items: &[HashMap<String, AttributeValue>]) -> Vec<(String, DataType)> {
+    let mut attrs: Vec<(String, DataType)> = Vec::new();
+    let mut seen: HashSet<String> = HashSet::new();
+    for item in items {
+        for (key, value) in item.iter() {
+            if seen.insert(key.clone()) {
+                attrs.push((key.clone(), attribute_value_to_arrow_type(value)));
+            }
+        }
+    }
+    attrs.sort_by(|(a, _), (b, _)| a.cmp(b));
+    attrs
 }
 
 /// Build an ordered field list: key attributes first (non-nullable, in key
@@ -1962,7 +1967,9 @@ mod tests {
     use super::*;
     use arrow::array::Array;
     use datafusion::common::ScalarValue;
+    use datafusion::logical_expr::dml::InsertOp;
     use datafusion::logical_expr::{BinaryExpr, col, lit};
+    use datafusion::physical_plan::empty::EmptyExec;
 
     fn n(s: &str) -> AttributeValue {
         AttributeValue::N(s.to_string())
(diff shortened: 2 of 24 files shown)


crates/skardi/src/sources/providers/dynamodb.rs
```diff
@@ -0,0 +1,2017 @@
...
+            schema: self.schema.clone(),
+        }))
+    }
+
+    async fn delete_from(
+        &self,
+        _state: &dyn Session,
+        filters: Vec<Expr>,
+    ) -> DFResult<Arc<dyn ExecutionPlan>> {
+        let pushable: Vec<Expr> = filters
```
abbccdda: **Critical — silent full-table DELETE/UPDATE.** `delete_from` (here) and `update` (L458) keep only the filters that pass `is_pushable_binary_filter` and silently discard the rest. DataFusion hands *all* WHERE conjuncts to this method, and `OR`, `LIKE`, `IN`, function calls, etc. are all non-pushable — so `DELETE FROM t WHERE a = 1 OR b = 2` produces `pushable = []`, `build_filter_expression` returns `None`, and `matching_keys(None)` scans and deletes **every item in the table**. Partially-pushable predicates over-delete too (`WHERE id = 'A' AND name LIKE 'x%'` deletes all `id = 'A'` rows). Unlike the SELECT path there is no Inexact re-filtering here — the dropped predicate is simply never applied.

Sibling providers guard this: mongo passes the full filter list to `exprs_to_mongo_filter`, which returns a `Plan` error on any unsupported expression (mongo/mod.rs:1049), and mysql unparses the whole WHERE clause or errors. Suggest erroring when any DML filter can't be fully converted (or evaluating the residual predicates client-side over the scanned items).

crates/server/src/config.rs
```diff
@@ -1266,6 +1270,41 @@ async fn register_data_source(
...
+            })?;
+
+            tracing::debug!(
+                "Endpoint for {}: {} (options: {:?})",
+                source.name,
+                connection_string,
+                source.options
+            );
+
+            register_dynamodb_tables(
```
abbccdda, line 1289: **`access_mode: read_only` is not enforced for DynamoDB.** `register_dynamodb_tables` takes no read/write flag, while the postgres/mysql/seekdb/sqlite arms all pass `source.access_mode.is_read_write()` (L1087/1129/1160/1196), and `DynamoTableProvider` implements `insert_into`/`update`/`delete_from` unconditionally. `validate_pipeline_sql` only covers YAML-declared pipeline SQL at config-load time, so there is no provider-level backstop.

The CLI has the same gap and no SQL validation layer at all: cli/main.rs:925 registers dynamodb with no gate, whereas the postgres/mysql arms hard-code `read_write = false` for ad-hoc SQL — so `DELETE FROM t` typed into the CLI mutates a live AWS table whose source is configured read-only (the default). Mongo/redis share this pre-existing pattern, but the blast radius here is production AWS. Fix is cheap: thread `is_read_write` into `register_dynamodb_tables` and gate the three DML methods.

crates/skardi/src/sources/providers/dynamodb.rs
```diff
@@ -0,0 +1,2017 @@
...
+        let future = async move {
+            let mut count: u64 = 0;
+            use futures::StreamExt;
+            while let Some(batch) = input_stream.next().await {
+                let batch = batch?;
+                let items = record_batch_to_items(&batch, &schema)
+                    .map_err(|e| DataFusionError::External(e.into()))?;
+                for item in items {
+                    client
+                        .put_item()
```
abbccdda: **INSERT silently destroys existing items on duplicate key.** This is a bare `put_item` with no `condition_expression("attribute_not_exists(#pk)")`, and `insert_into` ignores its `_insert_op` parameter (L410) — so `InsertOp::Append`, `Overwrite`, and `Replace` all behave as whole-item upsert. Inserting a row whose key exists replaces the entire item (all other attributes deleted) and reports a successful insert count. Every other writable provider surfaces a duplicate-key error (mongo `insert_many` E11000, SQL providers PK violation).

The README does document PutItem-upsert semantics, but the whole-item replacement (not attribute merge) plus the ignored `InsertOp` make this an easy silent-data-loss path. Suggest `attribute_not_exists` for `InsertOp::Append` and reserving upsert for `Overwrite`/`Replace`.

crates/skardi/src/sources/providers/dynamodb.rs
```diff
@@ -0,0 +1,2017 @@
...
+        filters: &[&Expr],
+    ) -> DFResult<Vec<TableProviderFilterPushDown>> {
+        Ok(filters
+            .iter()
+            .map(|expr| {
+                if is_pushable_binary_filter(expr) {
+                    // Inexact (not Exact) keeps the filter in the logical plan so
+                    // DataFusion's UPDATE/DELETE planner can still hand it to
+                    // delete_from/update — matching the MongoDB provider.
+                    TableProviderFilterPushDown::Inexact
```
abbccdda, line 256: **Inexact pushdown returns a *subset*, so rows silently vanish.** DynamoDB `FilterExpression` comparison is type-strict, but the read path is type-coercive: `av_to_f64`/`av_to_i64` parse `S`-typed values into numeric columns (L638/646). So for an item storing `price` as `S("50")` in a Float64-inferred column, `SELECT *` shows `50.0`, but `WHERE price < 100` pushes `#n0 < :v0` with `:v0 = N("100")` and DynamoDB excludes the item server-side — it never reaches DataFusion's re-filter, even though the local filter would match. The `Inexact` contract requires returning a **superset** of matching rows; mixed-type attributes are perfectly normal in schemaless DynamoDB, so this is silent wrong results.

Options: only push filters when type mismatch is impossible, push an OR over the type variants, or apply the predicate client-side after an unfiltered read.

crates/skardi/src/sources/providers/dynamodb.rs
```diff
@@ -0,0 +1,2017 @@
...
+        other => format!("{other:?}"),
+    }
+}
+
+fn av_to_i64(v: &AttributeValue) -> Option<i64> {
+    match v {
+        AttributeValue::N(n) => n
+            .parse::<i64>()
+            .ok()
+            .or_else(|| n.parse::<f64>().ok().map(|f| f as i64)),
```
abbccdda: **Silent numeric truncation.** The `parse::<f64>().map(|f| f as i64)` fallback truncates fractional `N` values (`N("99.99") → 99`) and saturates out-of-range ones; the Int32 arm additionally wraps via `as i32` (L587). The unit test at L1509 codifies this: `assert_eq!(av_to_i64(&n("7.9")), Some(7))`.

This is reachable in practice because schema inference samples one item and maps a whole-number `N` to Int64 — every other row's fractional value in that column is then corrupted. Compounding bug: `WHERE price = 99.99` returns zero rows — the pushed server-side filter matches `N("99.99")`, then DataFusion's Inexact re-filter compares the truncated `99` and drops it, making the row unreachable via its own stored value. Suggest inferring `N` as Float64 by default, and returning `None` (null) instead of truncating on lossy coercion.

crates/skardi/src/sources/providers/dynamodb.rs
```diff
@@ -0,0 +1,2017 @@
...
+    async fn infer_schema(
+        client: &Client,
+        table_name: &str,
+        partition_key: &str,
+        sort_key: Option<&str>,
+    ) -> Result<Schema> {
+        let sample = client
+            .scan()
+            .table_name(table_name)
+            .limit(1)
```
abbccdda: **Schema inferred from a single arbitrary item, with no override.** `scan().limit(1)` means any attribute absent from whichever item DynamoDB happens to return first simply doesn't exist as a column — and scan order is hash-of-key order, so the schema is nondeterministic across environments and as data changes. An empty table degrades to a keys-only schema, so a `read_write` source over a fresh table can't accept INSERTs of non-key columns at all.

`DynamoTableProvider::new` already accepts `schema: Option<SchemaRef>` (the integration tests use it), but `register_dynamodb_tables` hardcodes `None` (L1413) and no source option exposes it. Suggest plumbing a declared-columns option through, and falling back to sampling several items and merging their attribute sets (mongo consults the `$jsonSchema` validator before falling back to sampling).

crates/skardi/src/sources/providers/dynamodb.rs
```diff
@@ -0,0 +1,2017 @@
...
+/// A DynamoDB `FilterExpression` plus its attribute-name/value placeholder maps.
+struct DynamoFilter {
+    expression: String,
+    names: HashMap<String, String>,
+    values: HashMap<String, AttributeValue>,
+}
+
+/// Same predicate the MongoDB provider uses: a comparison between a bare column
+/// and a literal, with an operator DynamoDB can evaluate server-side.
+pub(crate) fn is_pushable_binary_filter(expr: &Expr) -> bool {
```
abbccdda: Two issues here. **(1)** The pushability check is shape-only (column vs literal), but `scalar_to_attribute_value` (L974) supports only Utf8/ints/floats/Boolean/Null and returns a `Plan` error otherwise — and that error propagates out of `scan()` via `build_filter_expression`, failing the *whole query* for a filter that was only `Inexact` and could have been evaluated locally. Not reachable from config registration today (inferred schemas only produce convertible types), but the explicit-schema constructor path hits it with e.g. a Timestamp column. Cheap fix: skip inconvertible predicates in `build_filter_expression` — safe since pushdown is Inexact.

**(2)** This function is a character-for-character copy of the already-`pub(crate)` `is_pushable_binary_filter` in mongo/mod.rs:517 (the comment even says so). Lift it to `providers/mod.rs` and share it, so the two providers can't silently diverge on pushdown behavior.

crates/skardi/src/sources/providers/dynamodb.rs
```diff
@@ -0,0 +1,2017 @@
...
+                    table = %self.table_name,
+                    filtered = filter.is_some(),
+                    "DynamoDB read via Scan (no key constraint)"
+                );
+                self.scan_items(filter.as_ref(), limit).await
+            }
+        }
+        .map_err(|e| DataFusionError::External(e.into()))?;
+
+        let batch = self
```
abbccdda: **Scan path buffers the whole result set and over-fetches.** Four compounding costs: (a) `scan_items`/`query_items` accumulate every page into a `Vec` and this builds one monolithic `RecordBatch` — unbounded memory, no output until the last page; a per-page `RecordBatch` stream (`futures::stream::unfold`) bounds memory at ~1MB. (b) No `ProjectionExpression` anywhere — `SELECT id FROM wide_table` transfers and converts all attributes, then the projection block (L377) throws most away; the projection is already available here. (c) The SQL LIMIT stops pagination but `.limit()` is never set on the Scan/Query request, so full 1MB pages are read (and RCU-billed) even for `LIMIT 1`. (d) Pages are deep-cloned via `out.items().iter().cloned()` when `out.items.take().unwrap_or_default()` would move them for free (`last_evaluated_key` is a separate field, so the move is safe).

crates/skardi/src/sources/providers/dynamodb.rs
```diff
@@ -0,0 +1,2017 @@
...
+                .get(sk)
+                .cloned()
+                .ok_or_else(|| anyhow::anyhow!("item missing sort key '{sk}'"))?;
+            key.insert(sk.clone(), sk_val);
+        }
+        Ok(key)
+    }
+
+    /// Scan keys of items matching `filter` (used to drive key-based DML).
+    async fn matching_keys(
```
abbccdda, line 475: **DML ignores the PR's own key-aware routing.** `matching_keys` always issues a full-table Scan — `DELETE FROM t WHERE pk = 'x'` pays RCUs for every item in the table even though `classify_read` (L914) already knows how to route that predicate to GetItem/Query. It also fetches all attributes when `key_of` only keeps the key columns — a `ProjectionExpression` on the key attributes would cut that. And the write side is one request per row: sequential `delete_item` per key (L1232) and `put_item` per row (L1121); `BatchWriteItem` (25/chunk, supports Put and Delete) would cut a 10k-row insert from 10k round-trips to 400. (Updates are legitimately per-item — `BatchWriteItem` doesn't support them.)

crates/skardi/src/sources/providers/dynamodb.rs
```diff
@@ -0,0 +1,2017 @@
...
+        };
+
+        Ok(Box::pin(RecordBatchStreamAdapter::new(
+            self.schema.clone(),
+            futures::stream::once(future),
+        )))
+    }
+}
+
+impl DynamoDmlOp {
```
abbccdda: **Cleanup: avoidable duplication.** (a) `clone_op`/`clone_filter` are exactly what `#[derive(Clone)]` on `DynamoDmlOp`/`DynamoFilter` generates — all fields are already `Clone` (the manual impls call `.clone()` on each) — and a future field added to `DynamoFilter` but missed here compiles fine while silently dropping data. (b) The ExclusiveStartKey/LastEvaluatedKey pagination loop is hand-rolled three times (`scan_items` L158, `query_items` L220, `matching_keys` L515), and the third copy already diverges (no limit handling); the SDK's `.into_paginator().items()` replaces all three. (c) `DynamoScanExec` stores `Arc<RwLock<Option<RecordBatch>>>` but nothing ever takes the `Option` — the "Batch already consumed" arm is unreachable dead code (copied from `MongoExecPlan`, which has the same dead arm); storing the `RecordBatch` directly (clone is cheap Arc bumps) removes the lock, the `Option`, and the impossible error path.

abbccdda (commented):
Automated deep review (7 finder angles, each candidate independently verified against the code). 8 confirmed findings + 1 plausible, ranked by severity below as inline comments.

**Most urgent:**
1. `delete_from`/`update` silently drop non-pushable WHERE predicates → `DELETE FROM t WHERE a = 1 OR b = 2` deletes **every row** in the table (dynamodb.rs:425).
2. `access_mode: read_only` (the default) is not enforced for DynamoDB — server and CLI both register the provider fully write-capable (config.rs:1289, cli/main.rs:925).
3. INSERT is an unconditional PutItem upsert that silently destroys the existing item on duplicate key, and `InsertOp` is ignored (dynamodb.rs:1123).

**Wrong-results bugs:** type-strict pushed filters violate the `Inexact` superset contract (rows silently vanish, dynamodb.rs:323); `av_to_i64` silently truncates fractional numbers (dynamodb.rs:637); schema is inferred from a single scanned item with no override (dynamodb.rs:104).

**Efficiency/cleanup:** unbounded buffering + no ProjectionExpression/request Limit on the scan path; DML always full-table Scans despite the PR's own key-aware routing; assorted duplication (3× hand-rolled pagination, manual Clone impls, verbatim copy of mongo's pushability predicate).

**Checked and found OK (not filed):** `DynamoInsertExec::properties()` reporting the input schema while emitting a count batch — the stream adapter consistently uses `count_schema()` and the CI round-trip passes; partial-write hazard for pipeline INSERTs — the PR correctly adds the executor.rs job-destination guard and the pipeline-path tradeoff pre-exists for mongo/redis/seekdb. One doc nit: `docs/jobs.md`'s rejected-destination list should now include DynamoDB (and InfluxDB).

crates/skardi/src/sources/providers/dynamodb.rs
abbccdda: Fixed in 90c6de6. `delete_from`/`update` no longer drop predicates: `plan_dml` now errors if *any* WHERE predicate isn't a convertible pushable comparison, so `DELETE ... WHERE a=1 OR b=2` is rejected instead of scanning+deleting the whole table (and partial-pushable predicates like `id='A' AND name LIKE 'x%'` are rejected too). New non-network test `delete_with_non_pushable_predicate_is_rejected` asserts the guard fires at plan time.

crates/server/src/config.rs
abbccdda: Fixed in 90c6de6. Threaded `read_write` through `register_dynamodb_tables` into `DynamoTableProvider`; `insert_into`/`update`/`delete_from` now return a plan error when the source is read-only. Server passes `source.access_mode.is_read_write()` (config.rs) and the CLI passes `source.is_read_write()` (previously hard-coded nothing) — so `DELETE` typed at the CLI against a default read-only source is now blocked.

crates/skardi/src/sources/providers/dynamodb.rs
abbccdda: Fixed in 90c6de6. Plain INSERT (Append) is now a conditional `PutItem` with `condition_expression("attribute_not_exists(#pk)")` — a duplicate key returns a clear error instead of silently replacing the whole item. `InsertOp` is honored: `Overwrite`/`Replace` opt into upsert (and batch via `BatchWriteItem`). README updated.

crates/skardi/src/sources/providers/dynamodb.rs
abbccdda: Fixed in 90c6de6. The read path is now type-strict: `av_to_i64`/`av_to_f64` no longer parse `S`-typed values into numeric columns (they return `None`/NULL). So a `price` stored as `S("50")` in a numeric column reads as NULL, and `NULL < 100` is false locally too — the server-side exclusion now matches what DataFusion's re-filter would do, restoring the Inexact superset contract. Covered by `number_coercions`.

crates/skardi/src/sources/providers/dynamodb.rs
abbccdda: Fixed in 90c6de6. `av_to_i64` only accepts an exact integer parse now — `N("7.9")` -> `None` (NULL) instead of truncating to 7, and the unit test was updated to assert that. Combined with inferring `N` as `Float64` by default (`attribute_value_to_arrow_type`), `WHERE price = 99.99` now round-trips correctly. README type table updated.

crates/skardi/src/sources/providers/dynamodb.rs
abbccdda: Fixed in 90c6de6. `infer_schema` now samples up to 50 items (`SCHEMA_SAMPLE_SIZE`) and merges their attribute sets in first-seen order (deterministic over the sample). Added a `columns` source option (`name:type,…`) plumbed through `register_dynamodb_tables` -> `DynamoTableProvider::new`, so an empty/read_write table can declare non-key columns explicitly. `parse_columns_option` has unit coverage; README + schema-notes updated.

crates/skardi/src/sources/providers/dynamodb.rs
abbccdda: Fixed in 90c6de6, both parts. (1) The scan path now filters on `is_convertible_pushdown` (shape + literal convertibility), so a pushable-shaped-but-inconvertible predicate (e.g. a Timestamp column via the explicit-schema path) is skipped rather than failing the whole query — safe since pushdown is Inexact. DML instead *errors* on such a predicate (it can't be dropped). (2) `is_pushable_binary_filter` is lifted to `providers/mod.rs` and shared by both mongo and dynamodb, so they can't diverge. New test `convertible_pushdown_excludes_inconvertible_literal`.

crates/skardi/src/sources/providers/dynamodb.rs
abbccdda: Fixed in 90c6de6. (a) Scan/Query now stream one page at a time via `stream_pages` (a `futures::stream::unfold` yielding a `RecordBatch` per page) — memory is bounded to ~1 page. (b) A `ProjectionExpression` (`build_projection_expression`, `#p` namespace merged with filter/key names) fetches only the projected columns. (c) The SQL `LIMIT` is now set on the request (`.limit()`) and caps per-page reads. (d) Page items are moved via `out.items` (not `.iter().cloned()`). `count(*)` uses `Scan` `Select=COUNT`.

crates/skardi/src/sources/providers/dynamodb.rs
abbccdda: Fixed in 90c6de6. `matching_keys` now takes a `DmlKeyPlan` from `classify_dml`: partition-key-pinned DML routes to a `Query` (with any residual non-key predicate as a server-side `FilterExpression`, since a Query filter can't reference keys), falling back to `Scan` only when the partition key isn't pinned or a sort predicate is inexpressible (e.g. `sk <> …`). Key resolution uses a key-only `ProjectionExpression`. DELETE now removes keys via `BatchWriteItem` (25/chunk with unprocessed-item retry); upsert INSERT batches puts the same way. UPDATE stays per-item as you noted. Routing covered by `classify_dml_*` tests.

crates/skardi/src/sources/providers/dynamodb.rs
abbccdda: Fixed in 90c6de6. (a) `clone_op`/`clone_filter` are gone — `DynamoDmlOp`/`DynamoFilter`/`DynamoRead`/`DmlKeyPlan` all `#[derive(Clone)]`. (b) The three hand-rolled pagination loops collapse into one `fetch_page` helper used by the scan stream and `matching_keys`. (c) `DynamoScanExec` no longer holds `Arc<RwLock<Option<RecordBatch>>>`; it streams pages on execute, so the lock, the `Option`, and the unreachable "Batch already consumed" arm are all gone.

abbccdda:
Addressed all 9 review findings + the doc nit in 90c6de6 (replies inline on each thread).

**Safety/correctness:** DML rejects non-pushable predicates (no more silent full-table DELETE); `access_mode` enforced provider-side (server + CLI); Append INSERT is a conditional PutItem (duplicate-key errors, not silent replace); type-strict numeric coercion + `N`→Float64 so Inexact pushdown stays a superset.

**Robustness:** multi-item schema sampling + explicit `columns` option.

**Efficiency:** per-page streaming scan with ProjectionExpression + request LIMIT; key-aware DML routing (Query vs Scan); BatchWriteItem for DELETE and upsert INSERT.

**Cleanup:** shared `is_pushable_binary_filter` (mongo+dynamodb), derive(Clone), single `fetch_page` paginator, removed the dead `RwLock<Option<RecordBatch>>` scan path.

Doc nit: `docs/jobs.md` now lists DynamoDB and InfluxDB as non-transactional destinations.

`cargo fmt`, `cargo check --workspace --all-targets`, and the skardi (26 dynamodb) + skardi-server unit suites are green. Live integration tests remain `#[ignore]`-gated for CI (dynamodb-local) — I couldn't run them locally (no Docker), but added non-network unit tests for the DML guard, routing, convertibility and `columns` parsing.

PR #142, opened 2026-06-25
feat(influxdb): add InfluxDB 3 provider via Flight SQL

abbccdda:
## Summary

Adds **InfluxDB 3** as a Skardi data source. InfluxDB 3 (Core/Enterprise) is itself built on Apache Arrow + DataFusion and exposes an Arrow **Flight SQL** endpoint, so rather than write a bespoke `TableProvider`, this reuses the generic Flight SQL provider already vendored via `datafusion-table-providers` (newly enabling its `flight` feature).

A thin `crates/skardi/src/sources/providers/influxdb.rs` module translates Skardi's friendly `options` map into the Flight SQL driver's option keys and registers the resulting `FlightTable` (which already implements `TableProvider`). The driver runs with `persistent_headers` so the `database` + bearer-token headers reach the `DoGet` data fetch, not just `GetFlightInfo`.

**Access is read-only** — Flight SQL serves `SELECT`s only; writes via line protocol are out of scope.

### Options
- `query` — full SQL backing the table, or
- `measurement` / `table` — shorthand → `SELECT * FROM "<name>"`
- `database` — InfluxDB 3 database; sent as the `database` gRPC header
- `token` — sent as `authorization: Bearer <token>`
- any `flight.sql.*` key forwarded verbatim (advanced/basic-auth)

### Wiring
`DataSourceType::Influxdb` variant + exhaustive match updates across `config.rs`, `pipeline_handlers.rs`, `gui.rs`, `jobs/executor.rs` (non-transactional destination), and the CLI string dispatch. Read-only, so deliberately **not** in `WRITABLE_SOURCE_TYPES` / `CATALOG_SUPPORTED_SOURCES`.

### Demo
`docs/influxdb/` — README with a Dockerised InfluxDB 3 Core Quick Start, `ctx_influxdb_demo.yaml`, five pipelines (scan / tag filter / threshold / aggregation / federated CSV join), and a sample CSV. Root README supported-sources table updated.

## Verification
- [x] Provider implementation + config/CLI/handler wiring
- [x] Unit tests (14) for option→Flight-SQL mapping + registration error paths
- [x] `cargo check --workspace --all-targets` clean; rustdoc clean under `-D warnings`; no new clippy warnings in touched files
- [x] Demo README + context/pipeline YAMLs
- [x] **Ran the demo end-to-end against a live InfluxDB 3 Core** — all 5 pipelines return correct results over Flight SQL (incl. the federated InfluxDB⨝CSV join). README response bodies captured from that run.
- [ ] `#[ignore]` integration test + CI service block for InfluxDB 3 (follow-up — gate on `INFLUXDB_URL`/`INFLUXDB_DATABASE`, run a query against a CI service)

## Ready for review
Draft only because the automated `#[ignore]` integration test + CI wiring remain. The feature itself is implemented and manually verified end-to-end.

### DIFF
--- a/crates/skardi/src/sources/providers/influxdb.rs
+++ b/crates/skardi/src/sources/providers/influxdb.rs
@@ -312,4 +312,100 @@ mod tests {
         .unwrap_err();
         assert!(err.to_string().contains("requires either a 'query'"));
     }
+
+    // ─── Integration tests (need a live InfluxDB 3 endpoint) ────────────
+    //
+    // Gated with `#[ignore]`; CI runs them via `cargo nextest -- --ignored`
+    // after starting an InfluxDB 3 Core container and seeding the `metrics`
+    // database (see .github/workflows/ci.yml and docs/influxdb/README.md).
+    // The endpoint and database are read from env so the same test runs
+    // locally and in CI.
+
+    fn influx_url() -> String {
+        std::env::var("INFLUXDB_URL").unwrap_or_else(|_| "http://127.0.0.1:8181".to_string())
+    }
+
+    fn influx_database() -> String {
+        std::env::var("INFLUXDB_DATABASE").unwrap_or_else(|_| "metrics".to_string())
+    }
+
+    async fn register_ci_measurement(ctx: &mut SessionContext, name: &str, measurement: &str) {
+        let options = opts(&[
+            ("database", influx_database().as_str()),
+            ("measurement", measurement),
+        ]);
+        register_influxdb_tables(ctx, name, &influx_url(), Some(&options))
+            .await
+            .unwrap_or_else(|e| panic!("register {name} failed: {e}"));
+    }
+
+    fn total_rows(batches: &[datafusion::arrow::record_batch::RecordBatch]) -> usize {
+        batches.iter().map(|b| b.num_rows()).sum()
+    }
+
+    #[tokio::test]
+    #[ignore]
+    async fn integration_register_measurement_and_scan() {
+        let mut ctx = SessionContext::new();
+        register_ci_measurement(&mut ctx, "cpu", "cpu").await;
+
+        let batches = ctx
+            .sql("SELECT host, usage_user FROM cpu")
+            .await
+            .expect("plan query")
+            .collect()
+            .await
+            .expect("collect");
+        assert!(
+            total_rows(&batches) >= 5,
+            "expected at least 5 seeded cpu rows, got {}",
+            total_rows(&batches)
+        );
+    }
+
+    #[tokio::test]
+    #[ignore]
+    async fn integration_aggregation_pushes_through_flight() {
+        let mut ctx = SessionContext::new();
+        register_ci_measurement(&mut ctx, "cpu", "cpu").await;
+
+        let batches = ctx
+            .sql("SELECT count(*) AS n FROM cpu WHERE host = 'host1'")
+            .await
+            .expect("plan query")
+            .collect()
+            .await
+            .expect("collect");
+        assert_eq!(total_rows(&batches), 1);
+    }
+
+    #[tokio::test]
+    #[ignore]
+    async fn integration_register_with_explicit_query_option() {
+        // Exercise the `query` option path (rather than `measurement`).
+        let mut ctx = SessionContext::new();
+        let options = opts(&[
+            ("database", influx_database().as_str()),
+            (
+                "query",
+                "SELECT host, usage_user FROM cpu WHERE usage_user > 50",
+            ),
+        ]);
+        register_influxdb_tables(&mut ctx, "hot_cpu", &influx_url(), Some(&options))
+            .await
+            .expect("register with query option");
+
+        let batches = ctx
+            .sql("SELECT * FROM hot_cpu")
+            .await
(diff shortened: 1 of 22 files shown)


docs/influxdb/pipelines/cpu_by_host.yaml
```diff
@@ -0,0 +1,9 @@
+kind: pipeline
+
+metadata:
+  name: "cpu_by_host"
+  version: "1.0"
+  description: "CPU samples for a single host (filter pushdown on a tag)"
```
abbccdda: This description says "filter pushdown on a tag", but no pushdown actually happens at the InfluxDB side. The vendored `FlightTable::scan` ignores its `filters` argument and doesn't implement `supports_filters_pushdown`, and the backing query is fixed at registration (`SELECT * FROM "cpu"`). So this pipeline pulls the **entire `cpu` measurement** over Flight and applies `WHERE host = {host}` locally in DataFusion.

Suggest correcting the wording (e.g. "filter on a tag (applied in Skardi)") and adding a README note that selective predicates/aggregations should be pushed into the `query` option to limit data transferred.

crates/skardi/src/sources/providers/influxdb.rs
```diff
@@ -0,0 +1,506 @@
...
+impl TableProvider for CountSafeFlightTable {
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+
+    fn schema(&self) -> SchemaRef {
+        self.inner.schema()
+    }
+
+    fn table_type(&self) -> TableType {
```
abbccdda, line 212: The wrapper forwards `schema`, `table_type`, and `scan`, but not `statistics()`. The inner `FlightTable` returns `Some(self.stats)`, so this wrapper falls back to the trait default `None` and the optimizer loses row-count estimates for InfluxDB tables (matters for join ordering in the federated case). Suggest delegating:

```rust
fn statistics(&self) -> Option<Statistics> {
    self.inner.statistics()
}
```

(Delegating `supports_filters_pushdown` too would be good forward-hygiene, though it's moot until the upstream provider pushes filters.)

abbccdda (commented):
Reviewed the InfluxDB 3 Flight SQL provider. Design is sound and the read-only constraint is enforced everywhere (excluded from `WRITABLE_SOURCE_TYPES`, rejected as a job destination, `token` not exposed via `get_data_sources`). Exhaustive match arms and thorough option-mapping unit tests are great.

Two actionable items inline below, plus notes:

- **No predicate/limit pushdown (perf):** the vendored `FlightTable::scan` ignores `_filters`/`_limit` and the backing query is fixed at registration, so every pipeline pulls the *full measurement* over Flight and filters locally in Skardi. Fine for the demo, surprising at scale. See the inline note on `cpu_by_host.yaml`; the `query` option is the real pushdown path and worth promoting in the README.
- **`statistics()` not delegated:** see inline note on `CountSafeFlightTable`.
- **Optional:** the wrapper's empty-projection `scan` logic is only covered by `#[ignore]` integration tests; a small offline test with a mock `TableProvider` would lock in the workaround.
- **Housekeeping:** the PR body lists the `#[ignore]` integration test + CI service block as an unchecked follow-up, but they're already in this diff — worth updating the checklist / marking ready.

docs/influxdb/pipelines/cpu_by_host.yaml
abbccdda: Fixed in 988bfaf. The pipeline description now reads "tag filter, applied in Skardi" so it no longer implies server-side pushdown. Also added a dedicated **Query Pushdown** section to the demo README explaining that pipeline-level predicates run locally in DataFusion and that the `query` option is the lever to push filters/projections/aggregations into InfluxDB's own engine. Dynamic per-request pushdown is tracked in #143.

crates/skardi/src/sources/providers/influxdb.rs
abbccdda: Fixed in 988bfaf. `CountSafeFlightTable` now delegates `statistics()` to the inner `FlightTable`, so row-count estimates flow through to the optimizer for join ordering again. Also delegated `supports_filters_pushdown()` as suggested for forward-hygiene — the wrapper is now transparent to the optimizer apart from the `count(*)` empty-projection interception.

BtXin:
For all the testing pipelines in `pipeline` directory, let's use parameterized pipeline instead.

abbccdda:
Done in c3f752a — all five demo pipelines now take request parameters instead of hardcoding filters / running a bare scan:

- `list_all_cpu` → `{limit}`
- `cpu_by_host` → `{host}` (already had it; now documented)
- `high_cpu` → `{threshold}` (already had it; now documented)
- `avg_usage_by_host` → scoped by `{region}`
- `federated_cpu_by_datacenter` → scoped by `{owner}`

Each has a `# Parameters:` doc block matching the convention in the other demos, and the README curls pass the params. The new filter examples only drop rows, so the captured aggregate values stay a truthful subset of the original live run.

Re-ran the full demo against a live InfluxDB 3 Core: all five return the expected results, `{limit}` is actually applied, and an empty body now returns `400 Missing required parameters: <name>` for each.

PR #145, opened 2026-06-30
feat(documents): liteparse-backed `documents` source connector

BtXin:
## What

Adds a reusable `type: documents` data source to skardi OSS: points at a directory / object-store prefix and turns files (PDF, `.docx/.xlsx/.pptx`, ODF, images) into queryable `(file, page)` rows via the pure-Rust **liteparse** crate. `SELECT * FROM documents` like any other source. Feature-gated behind `documents`.

Design spec & plan (in skardi-cloud): `docs/superpowers/specs/2026-06-29-documents-source-connector-design.md`, `docs/superpowers/plans/2026-06-30-documents-source-connector-plan.md`.

## Row schema (fixed)

`doc_id, path, page, markdown, tables_json, page_image_ref, image_refs, file_type` — one row per (file, page). Custom structured fields are not produced here; they come from `llm_extract` over the `markdown` column. Path/naming-derived columns are done in SQL.

## Notable decisions / deviations (please review)

- **liteparse** is a **git dependency pinned to rev `c8bda5b`** (optional, under the `documents` feature). crates.io was rate-limited at build time; git was the safe choice.
- **`tokio` bumped 1.51.1 → 1.52.3** (lockfile-only; within the workspace `^1.44.2` range) — liteparse needs ≥1.52.3.
- **Bundled Tesseract OCR dropped** (`default-features = false`) due to a native `lzma` link collision with DataFusion's `liblzma-sys` (Cargo forbids two crates linking the same native lib). **OCR now works only via liteparse's HTTP OCR engine** (`ocr_server_url`). `ocr=on` **hard-fails at registration** when no engine is configured; `ocr=auto` proceeds best-effort without OCR.
- **`--features documents` downloads a prebuilt PDFium native binary at build time** (via liteparse's `pdfium-sys`, no checksum). `PDFIUM_LIB_PATH`/`PDFIUM_INCLUDE_PATH` are the hermetic/offline escape hatches — documented in `docs/documents.md`.
- Real liteparse API differed from the spec sketch (async `parse`/`is_complex`; `ImageMode::Embed`); implementation matches the real crate.
- `render_page_images` is implemented via liteparse's PDFium `screenshot()` (populates `page_image_ref`, which the PrimeParticle multimodal-extraction path depends on).

## Tests

`cargo test -p skardi --lib sources::providers::documents --features documents` → **20 passed, 0 failed** (4 docx/xlsx/OCR/image tests skip when LibreOffice/Tesseract/ImageMagick are absent — they run on CI images that ship the converters). Server e2e (`config::tests::test_register_documents_source_via_context`) green; `cargo build` clean with and without the feature.

## Known follow-ups (TODOs in code)

- `parse_source` runs eagerly in `ExecutionPlan::execute` and blocks a worker for the whole directory — consider `spawn_blocking`/streaming for scale.
- `limit` parses everything then truncates (no pushdown) — v1 acceptable.
- Remote (`s3://`) image-crop upload deferred (local writes work).


BtXin:
Reviewed the current PR head and, with CI green, I’m skipping the environment-specific local test discrepancy I saw. A few remaining issues/questions:

1. **Object-store input is advertised but not implemented.**  
   The PR body and docs say `documents` can point at a “directory / object-store prefix”, and the example uses `path: s3://pp/inbound`. The implementation currently discovers files with `std::fs::read_dir` in `collect_files`, so an `s3://...` source will warn and return zero rows. Either wire this through the existing object-store layer, or narrow the PR/docs to local directories for v1.

2. **Missing/unreadable local roots silently become empty tables.**  
   `collect_files` logs `cannot read dir ...` and continues. For the initial root, a typo or permission issue should probably fail registration or scan rather than look like a valid empty corpus.

3. **README supported-source list is stale.**  
   This adds a new data source and `docs/documents.md`, but `README.md`’s “Supported Data Sources” table doesn’t list `documents`. The repo guidance calls out keeping that table in sync when adding features/sources.

Open question: should remote `image_store` refs be usable immediately for multimodal `llm_extract`? Right now remote stores record `s3://...` refs without uploading bytes, so consumers will only succeed if a later upload pass is guaranteed.

PR #151, opened 2026-07-11
Open-Connector integration proposal

BtXin:
Proposal to integrate open-connector

### DIFF
--- a/docs/superpowers/specs/2026-07-11-open-connector-integration-design.md
+++ b/docs/superpowers/specs/2026-07-11-open-connector-integration-design.md
@@ -1,0 +1,816 @@
+# Open Connector Integration Design
+
+**Status:** Approved design
+**Date:** 2026-07-11
+**Branch:** `codex/open-connector-proposal`
+
+## Summary
+
+Skardi will integrate with [Open Connector](https://github.com/oomol-lab/open-connector) as a separate authenticated SaaS gateway. Open Connector will continue to own provider credentials, OAuth flows, token refresh, action policies, and provider-specific HTTP execution. Skardi will add the relational layer: stable table definitions, JSON-to-Arrow conversion, pagination, safe filter and limit pushdown, optional caching, and DataFusion registration.
+
+The integration will expose three SQL interfaces:
+
+1. Persistent-style stable catalog tables created through `CREATE EXTERNAL TABLE` or equivalent context YAML bindings.
+2. `open_connector_query`, a built-in source-pack UDTF with a stable schema and known pagination behavior.
+3. `open_connector_scan`, a lower-level UDTF for explicitly allowlisted raw Open Connector read actions.
+
+Milestone one delivers the reusable foundation and validates it with GitHub, Jira, and Notion. All milestone-one behavior is read-only.
+
+## Motivation
+
+Skardi currently implements each database or storage source as a DataFusion `TableProvider`. Open Connector exposes more than one thousand SaaS providers through a uniform HTTP action API, including input and output JSON Schemas, authentication metadata, connection aliases, and execution envelopes.
+
+The two systems are complementary:
+
+- Open Connector is an authentication and action-execution gateway.
+- Skardi is a federated relational query engine.
+
+Open Connector actions are not automatically SQL tables. Their results may be scalar, nested, binary, mutating, asynchronous, incompletely pageable, or dependent on required resource identifiers. Skardi therefore needs a relational contract between an Open Connector action and a DataFusion table.
+
+## Research Findings
+
+The Open Connector catalog audited at commit `6d23b1341475df7c60984186882ff826beedd9b2` contained 1,045 providers and 10,104 actions. A mechanical scan found hundreds of read-like actions with array outputs, but action metadata does not standardize:
+
+- the JSON path containing relational rows;
+- pagination input and output fields;
+- primary keys;
+- SQL filter translations;
+- stable relational schemas;
+- whether an action represents a complete collection scan.
+
+Coverage for Skardi's prioritized SaaS sources is uneven:
+
+| Priority source | Open Connector coverage | Relational readiness |
+|---|---|---|
+| Google Workspace | Gmail, Calendar, Drive, Docs, Forms, Sheets, Slides, Tasks | Broad; Sheets and document content need special mapping |
+| Discord | OAuth and bot providers | Bot messages/channels are viable with explicit bindings |
+| Slack | Conversations, users, files, messages, threads | Metadata viable; message history lacks complete cursor handling |
+| Feishu | App Bot and Custom Bot | App Bot IM is viable but bot-visible only; Custom Bot is send-only |
+| HubSpot | CRM objects, contacts, companies, deals, properties, campaigns | Strong |
+| Notion | Data sources, databases, pages, blocks, users | Strong, with dynamic property schemas |
+| Jira | Projects, JQL issue search, issues, comments | Strong |
+| GitHub | Repositories, issues, pull requests, commits, workflows, releases | Very strong |
+| Salesforce | No current provider | Blocked pending upstream provider work |
+| Confluence | Search, spaces, pages | Viable but narrow |
+| Zendesk | Tickets, users, organizations | Strong |
+| Intercom | Contacts, conversations, admins | Strong |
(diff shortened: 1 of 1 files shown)


docs/superpowers/specs/2026-07-11-open-connector-integration-design.md
```diff
@@ -0,0 +1,816 @@
...
+- Embedding or forking the Open Connector runtime into Skardi.
+- Claiming full product-suite coverage when only particular APIs are supported.
+- Providing transaction or snapshot-isolation guarantees across SaaS APIs.
+- Returning partial results after a failed page.
+
+## Decisions
+
+The approved design choices are:
+
+- Use a separate Open Connector service over HTTP.
```
abbccdda, line 86: Are we introducing a hard dependency on open-connectors, so Skardi oss needs to keep up with latest connector version?

docs/superpowers/specs/2026-07-11-open-connector-integration-design.md
```diff
@@ -0,0 +1,816 @@
...
+
+## Future Extensions
+
+The design deliberately leaves room for:
+
+- persistent or distributed scan caches;
+- scheduled snapshot materialization into Parquet, Lance, or Iceberg;
+- automatic generation of candidate source-pack definitions for review;
+- additional typed pagination strategies;
+- write actions behind an explicit non-SQL action API;
```
abbccdda, line 846: I assume we don't support any SQL write operations on the saas source right?

docs/superpowers/specs/2026-07-11-open-connector-integration-design.md
```diff
@@ -0,0 +1,816 @@
...
+- the source-pack contribution guide.
+
+Marketing language must describe concrete supported resources. For example, use “Feishu bot-visible messages” rather than “full Feishu support,” and “Outlook Mail and Excel” rather than “complete Microsoft 365 support.”
+
+## Future Extensions
+
+The design deliberately leaves room for:
+
+- persistent or distributed scan caches;
+- scheduled snapshot materialization into Parquet, Lance, or Iceberg;
```
abbccdda, line 842: I feel to make Skardi oss version runnable in production, this is a must-have, we don't even worry about permission model yet (for single person app only)

docs/superpowers/specs/2026-07-11-open-connector-integration-design.md
BtXin: open-connector is a gateway service, it's not a lib dependency.

docs/superpowers/specs/2026-07-11-open-connector-integration-design.md
BtXin: We can already support this via skardi jobs.

docs/superpowers/specs/2026-07-11-open-connector-integration-design.md
```diff
@@ -0,0 +1,816 @@
...
+
+    DDL --> Factory
+    Factory --> Pack
+    Pack --> Binding
+    Binding --> Catalog
+```
+
+The user does not declare columns, row paths, pagination, or filter mappings for built-in definitions.
+
+DDL-created tables live for the DataFusion session. Reproducible server deployments use the equivalent context YAML binding.
```
abbccdda: **Blocking: this interface conflicts with the existing SQL validator, and the session-lifetime claim is wrong for Skardi's server model.**

1. `crates/skardi/src/sources/sql_validator.rs` rejects **all** `CREATE TABLE` / `CREATE SCHEMA` statements unconditionally ("DDL operations are not permitted on any data source"), so `CREATE EXTERNAL TABLE ... STORED AS OPEN_CONNECTOR` cannot execute on the server SQL path today. The design needs to specify the validator carve-out: exactly which DDL forms become allowed, and how that interacts with parameterized-query validation.

2. The server uses one shared long-lived `SessionContext` (`crates/server/src/config.rs`), so "DDL-created tables live for the DataFusion session" actually means *visible to every user until server restart*, not scoped to an interactive session. Please state who is authorized to run this DDL on a shared multi-user context, and what the visibility/cleanup semantics are.

docs/superpowers/specs/2026-07-11-open-connector-integration-design.md
```diff
@@ -0,0 +1,816 @@
...
+
+- gateway identity;
+- connection alias;
+- action ID and source-pack version;
+- resource and fixed action inputs;
+- translated filters;
+- upstream projection;
+- stable Arrow schema fingerprint.
+
+Milestone one cache invalidation is TTL-based. The cache is bounded by bytes and entries and uses least-recently-used eviction. Cache state is observable in traces and metrics.
```
abbccdda, line 553: **Cache key omits scan completeness — this can serve truncated results as complete.**

Per the execution rules, `LIMIT` stops pagination early, and the truncated batches are then stored under a key that does not include the limit. A later query with a higher limit (or no limit) hitting the same key would silently return partial data as a complete result — contradicting the "no partial results" non-goal.

Fix by either (a) caching only scans that ran to natural exhaustion, or (b) including the effective limit / a completeness marker in the cache key. Please state the chosen rule in this section.

docs/superpowers/specs/2026-07-11-open-connector-integration-design.md
```diff
@@ -0,0 +1,816 @@
...
+
+- Provider credentials remain in Open Connector.
+- Skardi loads only the Open Connector runtime token from an environment variable.
+- Logs never include tokens, provider credentials, authorization headers, or full sensitive inputs.
+- Stable tables can execute only the read actions hard-coded in their source packs.
+- YAML overrides cannot replace a stable table action with another action.
+- `open_connector_query` can use only built-in stable definitions.
+- `open_connector_scan` can use only action IDs explicitly allowlisted for the gateway.
+- Milestone one registers no SQL DML provider methods and exposes no mutating actions.
+- Open Connector action policies provide a second independent allowlist boundary.
```
abbccdda, line 580: **The read-only guarantee has no defined enforcement mechanism.**

Acceptance criterion 11 says mutating actions are rejected before any HTTP request, but nothing in the document defines *how* an action is classified as read vs. mutating. `raw_action_allowlist` is user-editable YAML, so the allowlist alone does not enforce read-only — an operator could allowlist a mutating action ID and every layer described here would pass it through.

Please name the source of truth for the classification (Open Connector action metadata flag, underlying HTTP method, or a Skardi-curated read-action list) and specify the behavior when that metadata is absent or ambiguous (presumably default-deny).

abbccdda (commented):
Second review pass on the proposal. The overall design holds up well (clear scoping, honest coverage table, verified claims against the codebase), but three issues below should be resolved in the doc before the Phase 0 implementation plan is written.

docs/superpowers/specs/2026-07-11-open-connector-integration-design.md
```diff
@@ -0,0 +1,816 @@
...
+- fixed row path;
+- stable Arrow schema and field mappings;
+- pagination strategy;
+- resource input requirements;
+- filter and optional projection translations;
+- primary-key metadata;
+- default safety bounds;
+- expected action-contract fingerprint.
+
+Source-pack definitions are maintained and versioned by Skardi. Users bind them to concrete resources but do not define their internal relational contracts.
```
bakey, line 181: Source-pack versioning should be concrete and pinnable.
The proposal mentions source-pack versions and action-contract fingerprints, but it does not define where versions appear. I’d recommend supporting explicit version pinning in YAML/DDL, e.g. github.issues@v1 or source_pack_version: 1, so a Skardi upgrade cannot silently change a table schema after restart.

docs/superpowers/specs/2026-07-11-open-connector-integration-design.md
BtXin: Let's by-default use read-only on the catalog providers, but if they user really want to write with open-connector, they will have the ability to do so, and it's their choice.

bakey (approved):

abbccdda (approved):
Overall looks solid. We may want to add a separate design as follow-up to discuss how we are going to support ETL based on top of open-connector and skardi jobs/ API to make the data materialized for future processing.

PR #152, opened 2026-07-13
feat(dynamodb): add catalog mode for whole-account table discovery

bakey:
## Summary

This PR adds `HierarchyLevel::Catalog` support to the DynamoDB provider, allowing users to register DynamoDB tables as a named DataFusion catalog instead of configuring one table at a time.

DynamoDB catalog mode now supports two registration shapes:

- Discover all accessible tables with `ListTables`.
- Register only an explicit comma-separated allow-list via `allowed_tables`, which avoids `ListTables` and bounds startup work.

Catalog registration is best-effort: if a single DynamoDB table cannot be described or sampled for schema inference, Skardi logs a warning, skips that table, and continues registering the remaining tables.

## Motivation

SQLite, PostgreSQL, MySQL, and SeekDB already support catalog mode: point Skardi at the database and it auto-discovers tables. DynamoDB was the odd one out, requiring a separate data-source entry per table. Catalog mode removes that friction for DynamoDB-backed workloads, while `allowed_tables` keeps large accounts or restricted IAM setups practical.

## What changed

- `crates/skardi/src/sources/providers/dynamodb.rs`
  - `register_dynamodb_tables` now accepts a `HierarchyLevel` and dispatches to either single-table registration (existing behavior) or catalog registration.
  - Catalog mode registers DynamoDB tables under `<catalog>.tables.<table_name>` because DynamoDB has no native schema/database layer.
  - Added paginated `ListTables` discovery for catalog mode.
  - Added `allowed_tables` parsing for catalog mode. When present, Skardi registers only those tables and skips `ListTables`.
  - Added best-effort catalog construction: per-table `DescribeTable` or schema inference failures now warn and skip only that table instead of failing the whole catalog.
  - Added bounded concurrent table introspection during catalog registration.

- `crates/server/src/config.rs`
  - Adds `DataSourceType::Dynamodb` to `CATALOG_SUPPORTED_SOURCES`.
  - Passes `source.hierarchy_level` to `register_dynamodb_tables`.
  - Rejects empty `allowed_tables` values such as `""` or `" , "`, so an empty allow-list cannot accidentally mean "load all tables".

- `crates/cli/src/main.rs`
  - Passes the parsed `hierarchy_level` through to DynamoDB registration, fixing the CLI compile break from the provider signature change.

- Documentation
  - Updates `docs/catalog.md` to list DynamoDB catalog support and show the fixed `tables` schema.
  - Updates `docs/dynamodb/README.md` with catalog mode, `allowed_tables`, and best-effort skip behavior.

## Usage example

Discover all accessible DynamoDB tables:

```yaml
kind: context
spec:
  data_sources:
    - name: myddb
      type: dynamodb
      connection_string: "http://localhost:8000"
      hierarchy_level: catalog
      access_mode: read_only
      options:
        region: us-east-1
```

```sql
SELECT * FROM myddb.tables.products WHERE product_id = 'PROD001';
```

Register only selected tables:

```yaml
kind: context
spec:
  data_sources:
    - name: myddb
      type: dynamodb
      connection_string: "http://localhost:8000"
      hierarchy_level: catalog
      options:
        region: us-east-1
        allowed_tables: "products,orders"
```

## Tests

- Updated existing `register_dynamodb_tables` test calls to pass `HierarchyLevel::Table`.
- Added table-mode regression coverage for missing options/table config.
- Added unit coverage for `allowed_tables` parsing, trimming, sorting, deduplication, and empty-value rejection.
- Added server config validation coverage for empty DynamoDB catalog `allowed_tables`.
- Kept DynamoDB Local catalog integration coverage as ignored tests (`catalog_mode_accepts_empty_options`, `catalog_mode_registers_tables_under_fixed_schema`).

Verified locally:

```bash
cargo fmt --all
cargo check -p skardi --all-targets
cargo check -p skardi-server --all-targets
cargo check -p skardi-cli --all-targets --no-default-features
cargo test -p skardi parse_allowed_tables
cargo test -p skardi-server validate_rejects_dynamodb_catalog_empty_allowed_tables
cargo test -p skardi missing_options_errors
cargo test -p skardi missing_table_option_errors
cargo test -p skardi table_mode_explicitly_requires_options
git diff --check
```

Note: full default-feature CLI check was not used for this verification because it triggers `llama-cpp-sys`/ccache writes outside the sandbox; the no-default-features CLI check verifies the DynamoDB registration path and signature synchronization.

## Checklist

- [x] Code follows the project's import and error-handling conventions
- [x] Unit tests added
- [x] Documentation updated
- [x] `cargo fmt --all` applied
- [x] `cargo check -p skardi --all-targets` passes
- [x] `cargo check -p skardi-server --all-targets` passes
- [x] `cargo check -p skardi-cli --all-targets --no-default-features` passes

### DIFF
--- a/crates/server/src/config.rs
+++ b/crates/server/src/config.rs
@@ -240,6 +240,11 @@ pub enum ConfigError {
     )]
     EmptyAllowedSchemas { name: String },
 
+    #[error(
+        "Data source '{name}' has an empty 'allowed_tables' option. Either omit it to load all DynamoDB tables, or provide a non-empty comma-separated list such as \"products,orders\"."
+    )]
+    EmptyAllowedTables { name: String },
+
     #[error("Data source '{name}' has a non-UTF8 path: {path:?}")]
     NonUtf8Path { name: String, path: PathBuf },
 }
@@ -784,6 +789,22 @@ fn validate_data_sources(data_sources: &[DataSource]) -> Result<()> {
                         name: source.name.clone(),
                     }
                     .into());
+                }
+            }
+
+            if source.source_type == DataSourceType::Dynamodb {
+                if let Some(value) = source
+                    .options
+                    .as_ref()
+                    .and_then(|o| o.get("allowed_tables"))
+                {
+                    let has_entry = value.split(',').any(|s| !s.trim().is_empty());
+                    if !has_entry {
+                        return Err(ConfigError::EmptyAllowedTables {
+                            name: source.name.clone(),
+                        }
+                        .into());
+                    }
                 }
             }
         }
@@ -2249,6 +2270,29 @@ options:
             matches!(
                 config_err,
                 ConfigError::MissingConnectionString { name } if name == "products"
+            ),
+            "got {config_err}"
+        );
+    }
+
+    #[test]
+    fn validate_rejects_dynamodb_catalog_empty_allowed_tables() {
+        let mut options = HashMap::new();
+        options.insert("allowed_tables".to_string(), " , , ".to_string());
+        let mut source = dynamodb_source(
+            "ddb",
+            Some("http://localhost:8000"),
+            Some(options),
+            AccessMode::ReadOnly,
+        );
+        source.hierarchy_level = HierarchyLevel::Catalog;
+
+        let err = validate_data_sources(&[source]).unwrap_err();
+        let config_err = err.downcast_ref::<ConfigError>().unwrap();
+        assert!(
+            matches!(
+                config_err,
+                ConfigError::EmptyAllowedTables { name } if name == "ddb"
             ),
             "got {config_err}"
         );
--- a/crates/server/src/config.rs
+++ b/crates/server/src/config.rs
@@ -708,6 +708,7 @@ const CATALOG_SUPPORTED_SOURCES: &[DataSourceType] = &[
     DataSourceType::Mysql,
     DataSourceType::Sqlite,
     DataSourceType::Seekdb,
+    DataSourceType::Dynamodb,
 ];
 
 /// Data source types that support read_write access mode
@@ -1292,6 +1293,7 @@ async fn register_data_source(
                 connection_string,
                 source.options.as_ref(),
                 source.access_mode.is_read_write(),
+                source.hierarchy_level,
             )
             .await
             .map_err(|e| {
--- a/crates/server/src/config.rs
+++ b/crates/server/src/config.rs
@@ -2299,6 +2299,45 @@ options:
     }
 
     #[test]
+    fn validate_accepts_dynamodb_catalog_allowed_tables() {
+        let mut options = HashMap::new();
+        options.insert("allowed_tables".to_string(), "products, orders".to_string());
+        let mut source = dynamodb_source(
+            "ddb",
+            Some("http://localhost:8000"),
+            Some(options),
+            AccessMode::ReadOnly,
+        );
+        source.hierarchy_level = HierarchyLevel::Catalog;
+
+        validate_data_sources(&[source]).expect("valid DynamoDB catalog allow-list");
+    }
+
+    #[test]
+    fn validate_rejects_dynamodb_catalog_table_option() {
+        let mut options = HashMap::new();
+        options.insert("table".to_string(), "products".to_string());
+        let mut source = dynamodb_source(
+            "ddb",
+            Some("http://localhost:8000"),
+            Some(options),
+            AccessMode::ReadOnly,
+        );
+        source.hierarchy_level = HierarchyLevel::Catalog;
+
+        let err = validate_data_sources(&[source]).unwrap_err();
+        let config_err = err.downcast_ref::<ConfigError>().unwrap();
+        assert!(
+            matches!(
+                config_err,
+                ConfigError::CatalogModeConflictingOptions { name, option }
+                    if name == "ddb" && option == "table"
+            ),
+            "got {config_err}"
+        );
+    }
+
+    #[test]
     fn unsupported_write_mode_error_lists_dynamodb() {
         let err = ConfigError::UnsupportedWriteMode {
             name: "events".to_string(),
--- a/crates/skardi/src/sources/providers/dynamodb.rs
+++ b/crates/skardi/src/sources/providers/dynamodb.rs
@@ -43,6 +43,13 @@ use futures::StreamExt;
 use futures::stream::{self, Stream};
 
 use super::is_pushable_binary_filter;
+use crate::sources::DataSourceType;
+use crate::sources::hierarchy::{HierarchyLevel, SourceLabel, build_catalog, retry_with_timeout};
+
+/// Schema name used for DynamoDB catalog mode. DynamoDB has no native schema/database
+/// layer; all tables live under a single endpoint/region. We expose them under a fixed
+/// schema so SQL references are consistently three-part: `catalog.tables.<table>`.
+const DYNAMODB_CATALOG_SCHEMA: &str = "tables";
 
 /// Maximum items sampled to infer a schema when no explicit `columns` option is
 /// given. Merging several items (rather than one) makes the inferred column set
@@ -51,6 +58,10 @@ const SCHEMA_SAMPLE_SIZE: i32 = 50;
 
 /// DynamoDB `BatchWriteItem` accepts at most 25 write requests per call.
 const BATCH_WRITE_CHUNK: usize = 25;
+
+/// Maximum tables returned per `ListTables` page. AWS allows up to 100; we use the
+/// default and follow `last_evaluated_table_name` when present.
+const LIST_TABLES_PAGE_SIZE: i32 = 100;
 
 /// A DynamoDB table exposed to DataFusion as a `TableProvider`.
 pub struct DynamoTableProvider {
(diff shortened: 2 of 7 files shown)


crates/skardi/src/sources/providers/dynamodb.rs
```diff
@@ -1867,10 +1924,280 @@ pub async fn register_dynamodb_tables(
...
+    mode_str: &str,
+) -> Result<()> {
+    tracing::info!(
+        catalog = %catalog_name,
+        endpoint = %connection_string,
+        read_write,
+        "Registering DynamoDB catalog"
+    );
+
+    let opts = options.cloned().unwrap_or_default();
```
BtXin, line 2074: [P2] Reject catalog conflicts at the provider boundary

`validate_data_sources` catches `table` in server configuration, but the CLI calls this public provider directly. In CLI/library use, `hierarchy_level: catalog` plus `options.table` silently ignores the requested table and falls through to `ListTables`, potentially registering every accessible table. Please reject `table`/`schema` here (and ideally validate catalog-only versus table-only options) so every caller gets the same scope validation.

crates/skardi/src/sources/providers/dynamodb.rs
```diff
@@ -1867,10 +1924,280 @@ pub async fn register_dynamodb_tables(
...
+        describe_keys(&client, table_name).await.with_context(|| {
+            format!("DynamoDB catalog '{catalog_name}': DescribeTable failed for '{table_name}'")
+        })?;
+
+    DynamoTableProvider::new(
+        (*client).clone(),
+        table_name,
+        &partition_key,
+        sort_key.as_deref(),
+        None,
```
BtXin, line 2212: [P2] Preserve key types for empty tables

This passes no declared schema after `DescribeTable` returns only the key names. If a table is empty and its key type is Number, schema inference has no sample and `build_schema_fields` defaults that key to `Utf8`. Catalog mode ignores `columns`, so predicates and writes can use `S` instead of `N` with no workaround. Please carry the key scalar types from `DescribeTable.attribute_definitions` into the schema and add an empty numeric-key table test.

crates/skardi/src/sources/providers/dynamodb.rs
```diff
@@ -1867,10 +1924,280 @@ pub async fn register_dynamodb_tables(
...
+    let table_names = match parse_allowed_tables(Some(&opts))? {
+        Some(allowed) => {
+            tracing::info!(
+                catalog = %catalog_name,
+                allowed_tables = ?allowed,
+                "Using DynamoDB catalog table allow-list"
+            );
+            allowed
+        }
+        None => retry_with_timeout(label, "ListTables", || async {
```
BtXin: [P2] Avoid timing out the entire pagination pass

The shared timeout is five seconds, but this wraps the complete multi-page `ListTables` traversal. If traversal takes longer, the attempt is cancelled and the next attempt restarts at page one, so a sufficiently large or slow catalog can never finish discovery. Please apply timeout/retry to each page request (preserving the same `exclusive_start_table_name`) or use an overall deadline sized for a complete traversal.

docs/dynamodb/README.md
```diff
@@ -274,6 +285,37 @@ curl -X POST http://localhost:8080/federated_join/execute \
 Writes (INSERT/UPDATE/DELETE) additionally require `access_mode: read_write` on
 the source; the default `read_only` rejects them at plan time.
 
+### Catalog mode
+
+Set `hierarchy_level: "catalog"` to expose multiple DynamoDB tables under one
```
BtXin, line 291: [P3] Synchronize the option and supported-source tables

Catalog mode changes `table` from always required to table-mode-only and adds `allowed_tables`, but the Connection Options table above still says `table` is unconditionally required and omits `allowed_tables`. The root README DynamoDB row should also mention table-or-catalog registration, per the repository documentation rules.

BtXin (commented):
Comprehensive review completed at `0784565`. The targeted catalog-registration, allow-list parser, and server-validation tests pass. I found three functional edge cases plus one documentation consistency gap; details are inline.

docs/catalog.md
```diff
@@ -44,3 +47,4 @@ See the docs directories for full working examples:
 - [docs/postgres/](postgres/) — `ctx_postgres_catalog_demo.yaml`
 - [docs/mysql/](mysql/) — `ctx_mysql_catalog_demo.yaml`
 - [docs/sqlite/](sqlite/) — `ctx_sqlite_catalog_demo.yaml`
+- [docs/dynamodb/](dynamodb/) — use `hierarchy_level: "catalog"` with optional `allowed_tables`
```
abbccdda, line 50: Wonder if we should also update https://github.com/SkardiLabs/skardi/#supported-data-sources to have a separate column for showing the data sources supporting catalog mode? @BtXin

crates/server/src/config.rs
```diff
@@ -1292,6 +1314,7 @@ async fn register_data_source(
                 connection_string,
                 source.options.as_ref(),
                 source.access_mode.is_read_write(),
+                source.hierarchy_level,
```
abbccdda, line 1317: Are we making all tables read-write by default when signing up in catalog mode? Could we have compound mode for read-only on some tables while others are read-write?

crates/skardi/src/sources/providers/dynamodb.rs
```diff
@@ -1867,10 +1924,280 @@ pub async fn register_dynamodb_tables(
...
+}
+
+/// Build a [`DynamoTableProvider`] for `table_name`.
+async fn build_dynamodb_table_provider(
+    client: Arc<Client>,
+    table_name: &str,
+    read_write: bool,
+    catalog_name: &str,
+) -> Result<DynamoTableProvider> {
+    let (partition_key, sort_key) =
```
abbccdda: do we want to retry_with_timeout here?

crates/skardi/src/sources/providers/dynamodb.rs
```diff
@@ -1867,10 +1924,280 @@ pub async fn register_dynamodb_tables(
...
+    })
+    .await
+    .with_context(|| format!("Failed to build DynamoDB client for catalog '{catalog_name}'"))?;
+
+    let table_names = match parse_allowed_tables(Some(&opts))? {
+        Some(allowed) => {
+            tracing::info!(
+                catalog = %catalog_name,
+                allowed_tables = ?allowed,
+                "Using DynamoDB catalog table allow-list"
```
abbccdda, line 2098: [P2] Best-effort skipping masks typos in an explicit allow-list

When `allowed_tables` is set, a misspelled table name (e.g. `"products,ordersx"`) hits `ResourceNotFoundException` in `DescribeTable` and is downgraded to a skip-with-warning, so the config mistake only surfaces at query time as a missing table. Best-effort makes sense for `ListTables` discovery across a large account, but when the user explicitly enumerated the tables, a hard error at startup (or at least a distinct, louder failure path) would catch the typo immediately. Consider failing catalog registration when a table from `allowed_tables` cannot be described.

crates/skardi/src/sources/providers/dynamodb.rs
```diff
@@ -1867,10 +1924,280 @@ pub async fn register_dynamodb_tables(
...
+    let prepared = stream::iter(build_futures)
+        .buffer_unordered(DYNAMODB_CATALOG_BUILD_CONCURRENCY)
+        .collect::<Vec<_>>()
+        .await;
+
+    register_prepared_dynamodb_catalog(session_ctx, catalog_name, prepared)
+}
+
+/// Register already-built DynamoDB catalog table providers, skipping failed builds.
+fn register_prepared_dynamodb_catalog(
```
abbccdda: [P2] Near-duplicate of `hierarchy::build_catalog` with divergent failure semantics

This reimplements the schema/table registration loop from `hierarchy::build_catalog` almost verbatim, and the two now differ in failure semantics: the shared helper is fail-fast (`try_collect`), while this one skips failed tables. Consider extending `build_catalog` with a best-effort option so the registration loop lives in one place — otherwise future fixes to schema/table registration have to land twice and the providers drift further apart.

crates/skardi/src/sources/providers/dynamodb.rs
```diff
@@ -1867,10 +1924,280 @@ pub async fn register_dynamodb_tables(
...
+) -> Result<(usize, usize)> {
+    let catalog_provider = Arc::new(MemoryCatalogProvider::new());
+    let mut registered = 0usize;
+    let mut skipped = 0usize;
+    let schema_name = DYNAMODB_CATALOG_SCHEMA.to_string();
+
+    for (table_name, provider_result) in prepared {
+        match provider_result {
+            Ok(provider) => {
+                if catalog_provider.schema(&schema_name).is_none() {
```
abbccdda: [P3] Empty or all-skipped catalogs end up without the `tables` schema

The schema is registered lazily on the first successful table, so a catalog where every table was skipped (or none were discovered) has no `tables` schema at all. `SELECT * FROM ddb.tables.x` then errors with "schema not found" instead of "table not found", which is confusing given the docs promise a fixed `tables` schema. Registering the `MemorySchemaProvider` unconditionally before the loop is also simpler: it removes this lazy `is_none()` check and the "not found after registration" error path below.

crates/skardi/src/sources/providers/dynamodb.rs
bakey: I have moved retry/timeout into the paginated loop so exclusive_start_table_name progress is preserved across page retries, and keep a bounded per-page retry policy.

docs/dynamodb/README.md
bakey: done

crates/skardi/src/sources/providers/dynamodb.rs
bakey: done

crates/skardi/src/sources/providers/dynamodb.rs
bakey: done

crates/skardi/src/sources/providers/dynamodb.rs
bakey: done

crates/skardi/src/sources/providers/dynamodb.rs
bakey: done

crates/server/src/config.rs
bakey: Today catalog mode intentionally inherits the data-source-level access_mode, so all registered DynamoDB tables are read-only or read-write together. I’d avoid adding per-table mixed access in this PR because it changes the source config model and DML safety surface. If we need it later, I’d prefer a separate design with explicit per-table overrides.

docs/catalog.md
bakey: Agree

BtXin (approved):

PR #154, opened 2026-07-15
feat(cli): add `skardi doctor` single-machine self-check

Userneima:
## What
Adds a `skardi doctor` subcommand: a quick single-machine self-check of the skardi binary. It checks the CLI version, that the query engine can build a `SessionContext` and run SQL, and reports which optional features were compiled in. A failed **core** check exits non-zero so scripts / CI can detect a broken install.

## Scope (v1)
Binary only — it deliberately does **not** read `ctx.yaml`, so it does not verify that configured data sources are reachable (planned for a later version via `--ctx`). Environment concerns outside the engine (Python deps, the `auto_knowledge_base` skill) are handled on the skills side, not here.

## Motivation
Give users a one-command "is my install healthy?" check, so a broken binary is caught up front instead of failing later with an opaque error.

## Tests & verification
- Includes unit tests (query-engine check, feature reporting, report-on-success / -failure).
- `cargo test -p skardi-cli --no-default-features --features candle` → 59 passed (incl. the doctor tests); `cargo fmt --check` clean; clippy clean for the new files.
- Note: the full default build (with `gguf`) was not run locally (no `cmake` on the dev machine); CI covers that path.

### DIFF
--- a/crates/cli/src/doctor.rs
+++ b/crates/cli/src/doctor.rs
@@ -1,0 +1,141 @@
+//! `skardi doctor` — quick self-check of the skardi binary.
+//!
+//! Scope (v1): this checks the BINARY itself — its version, and that the query
+//! engine can build a context and run SQL — plus which optional features were
+//! compiled in. It deliberately does NOT read your `ctx.yaml`, so it does not
+//! verify that your configured data sources are reachable or correctly set up.
+//! Verifying data sources is planned for a future version (via `--ctx`).
+//!
+//! Environment concerns outside the engine (Python deps, the
+//! `auto_knowledge_base` skill) are handled by the skills/setup side, not here.
+
+use anyhow::Result;
+
+/// Run the checks, print a report, and return whether all CORE checks passed.
+/// The caller decides the process exit code, so this stays testable and free of
+/// `process::exit`.
+pub async fn run() -> Result<bool> {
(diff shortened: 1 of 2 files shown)


abbccdda:
The build is broken

Userneima:
Closing this.

#170 moved the query engine and the compile-time model features out of the CLI, so both things `doctor` checks — building a session context to run `SELECT 1`, and reporting which of onnx/candle/gguf/remote-embed/chunking were compiled in — no longer live here. The merge conflict in `main.rs` is a symptom of that, not something worth resolving: the command would end up checking things that aren't there.

I also don't think a replacement is worth building right now. The quick start already doubles as the confirmation — install both binaries, start a server, run one query, see rows — and `skardi health` plus `skardi schema` already answer "is the server reachable" and "what is registered", which are the failure modes the two-binary split actually introduced. What a `doctor` command would add on top is thin, and we have no reports of anyone getting stuck there. Meanwhile the measured first-run cost is elsewhere: the server takes ~2.5 minutes to build from source while the release workflow already produces prebuilt CLI binaries for macOS arm64 and Linux x86_64/arm64.

If first-run diagnostics come back, it should be driven by real failure reports rather than a guess, and it would look different from this PR.

Sorry I left your "the build is broken" comment hanging for so long — no longer relevant here, but noted.

PR #157, opened 2026-07-15
feat(sources): add ClickHouse table provider

abbccdda:
## Summary

Adds **ClickHouse** as a read-only data source, backed by the `clickhouse` feature of `datafusion-table-providers`. Filters, projections, and `LIMIT`s are unparsed back to ClickHouse SQL and executed server-side, so scans stream only the rows a query actually needs — and ClickHouse tables federate with any other Skardi source in a single query.

## Details

- **Two registration modes**, mirroring the other SQL providers:
  - **Table mode** (default): one ClickHouse table registered under the source name, with an optional `database` option.
  - **Catalog mode**: every table across the server's non-system databases registered as `name.<database>.<table>`, filtered by an optional `allowed_schemas` allow-list.
- **Credentials** come from environment variables via `user_env` / `pass_env` options and are never embedded in the connection URL.
- **Read-only by design**: ClickHouse mutations (`ALTER TABLE ... UPDATE/DELETE`) are asynchronous background rewrites and a mid-stream INSERT failure leaves partial parts visible, which doesn't match the semantics Skardi's write path promises. `access_mode: read_write` and job destinations are rejected for ClickHouse sources.
- **CI**: a `clickhouse/clickhouse-server:24.8` service container with seeded fixture tables (including a NULL-column and an empty table for schema-inference coverage); integration tests are `#[ignore]`-gated and run via the existing `--ignored` pass.
- **Robust catalog registration** (review follow-up): best-effort per-table assembly (one broken view or Kafka-engine table no longer aborts startup), with stream-like engines and materialized-view `.inner*` tables filtered up front, and introspection batched into a single `system.tables` query.
- **Validation at the provider boundary** (review follow-up): unknown options, mode-mismatched options, empty `allowed_schemas`, URL-embedded credentials, query strings, and `access_mode: read_write` are all hard errors in `register_clickhouse_tables` itself, so the CLI and public API enforce the same contract as server config validation — and secrets can't reach logs or the data-sources API.
- **Docs**: quick-start guide, table- and catalog-mode demo contexts, and federated demo pipelines (ClickHouse × CSV join) under `docs/clickhouse/`, plus a README source-matrix row.

## Test plan

- [ ] Unit tests: option parsing, URL validation, and `DataSourceType` roundtrip (`cargo nextest run`)
- [ ] `#[ignore]`-gated integration tests against the CI ClickHouse service container: registration, scans, filter/limit pushdown, NULL handling, empty-table schema inference, catalog mode
- [ ] `submit_rejects_clickhouse_destination_as_non_transactional` covers the job-destination rejection

### DIFF
--- a/crates/server/src/config.rs
+++ b/crates/server/src/config.rs
@@ -830,10 +830,11 @@ fn validate_data_sources(data_sources: &[DataSource]) -> Result<()> {
         }
 
         // Catalog mode must not mix with per-table / per-schema options
+        // ("database" is ClickHouse's schema-analog spelling)
         if CATALOG_SUPPORTED_SOURCES.contains(&source.source_type)
             && source.hierarchy_level == HierarchyLevel::Catalog
         {
-            for conflicting in &["table", "schema"] {
+            for conflicting in &["table", "schema", "database"] {
                 if source
                     .options
                     .as_ref()
@@ -2685,6 +2686,72 @@ bindings:
                 config_err,
                 ConfigError::CatalogModeConflictingOptions { name, option }
                     if name == "ddb" && option == "table"
+            ),
+            "got {config_err}"
+        );
+    }
+
+    fn clickhouse_source(
+        name: &str,
+        options: Option<HashMap<String, String>>,
+        access_mode: AccessMode,
+    ) -> DataSource {
+        DataSource {
+            name: name.to_string(),
+            source_type: DataSourceType::Clickhouse,
+            path: PathBuf::new(),
+            connection_string: Some("http://localhost:8123".to_string()),
+            schema: None,
+            options,
+            hierarchy_level: HierarchyLevel::default(),
+            access_mode,
+            enable_cache: false,
+            description: None,
+        }
+    }
+
+    #[test]
+    fn validate_rejects_clickhouse_read_write() {
+        let mut options = HashMap::new();
+        options.insert("table".to_string(), "events".to_string());
+        let source = clickhouse_source("events", Some(options), AccessMode::ReadWrite);
+        let err = validate_data_sources(&[source]).unwrap_err();
+        let config_err = err.downcast_ref::<ConfigError>().unwrap();
+        assert!(
+            matches!(
+                config_err,
+                ConfigError::UnsupportedWriteMode { name, source_type }
+                    if name == "events" && *source_type == DataSourceType::Clickhouse
+            ),
+            "got {config_err}"
+        );
+    }
+
+    #[test]
+    fn validate_accepts_clickhouse_table_mode_with_database_option() {
+        let mut options = HashMap::new();
+        options.insert("table".to_string(), "events".to_string());
+        options.insert("database".to_string(), "analytics".to_string());
+        let source = clickhouse_source("events", Some(options), AccessMode::ReadOnly);
+        validate_data_sources(&[source]).expect("table mode accepts a database option");
+    }
+
+    #[test]
+    fn validate_rejects_clickhouse_catalog_database_option() {
+        // "database" is ClickHouse's schema-analog option; letting it through
+        // in catalog mode would silently change the pool's default database.
+        let mut options = HashMap::new();
+        options.insert("database".to_string(), "analytics".to_string());
+        let mut source = clickhouse_source("ch", Some(options), AccessMode::ReadOnly);
+        source.hierarchy_level = HierarchyLevel::Catalog;
+
+        let err = validate_data_sources(&[source]).unwrap_err();
+        let config_err = err.downcast_ref::<ConfigError>().unwrap();
+        assert!(
+            matches!(
+                config_err,
+                ConfigError::CatalogModeConflictingOptions { name, option }
+                    if name == "ch" && option == "database"
             ),
             "got {config_err}"
         );
--- a/crates/server/src/config.rs
+++ b/crates/server/src/config.rs
@@ -1518,6 +1518,7 @@ async fn register_data_source(
                 &source.name,
                 connection_string,
                 source.options.as_ref(),
+                source.access_mode.is_read_write(),
                 source.hierarchy_level,
             )
             .await
--- a/crates/server/src/config.rs
+++ b/crates/server/src/config.rs
@@ -2704,6 +2704,7 @@ bindings:
             connection_string: Some("http://localhost:8123".to_string()),
             schema: None,
             options,
+            open_connector: None,
             hierarchy_level: HierarchyLevel::default(),
             access_mode,
             enable_cache: false,
--- a/crates/server/src/config.rs
+++ b/crates/server/src/config.rs
@@ -5,6 +5,7 @@ use datafusion::prelude::*;
 use serde::{Deserialize, Serialize};
 use skardi::jobs::JobDefinition;
 use skardi::pipeline::pipeline::{Pipeline, StandardPipeline};
+use skardi::sources::providers::clickhouse::register_clickhouse_tables;
 use skardi::sources::providers::dynamodb::register_dynamodb_tables;
 use skardi::sources::providers::iceberg::register_iceberg_table;
 use skardi::sources::providers::influxdb::register_influxdb_tables;
@@ -743,6 +744,7 @@ const CATALOG_SUPPORTED_SOURCES: &[DataSourceType] = &[
     DataSourceType::Sqlite,
     DataSourceType::Seekdb,
     DataSourceType::Dynamodb,
+    DataSourceType::Clickhouse,
     // OpenConnector is catalog-only; its tables come from typed bindings, so
     // the same catalog-mode guards (no per-table `options`) must apply.
     DataSourceType::OpenConnector,
@@ -890,6 +892,7 @@ fn validate_data_sources(data_sources: &[DataSource]) -> Result<()> {
                 | DataSourceType::Redis
                 | DataSourceType::Seekdb
                 | DataSourceType::Influxdb
+                | DataSourceType::Clickhouse
                 | DataSourceType::OpenConnector
                 | DataSourceType::Dynamodb,
                 false,
@@ -1065,6 +1068,7 @@ async fn register_data_source(
             | DataSourceType::Redis
             | DataSourceType::Seekdb
             | DataSourceType::Influxdb
+            | DataSourceType::Clickhouse
             | DataSourceType::OpenConnector
             | DataSourceType::Dynamodb,
             _,
@@ -1486,6 +1490,39 @@ async fn register_data_source(
             .map_err(|e| {
                 tracing::error!(
                     "InfluxDB registration failed for '{}': {:?}",
+                    source.name,
+                    e
+                );
+                ConfigError::DataSourceRegistrationFailed {
+                    name: source.name.clone(),
+                    error: format!("{:?}", e),
+                }
+            })?;
+        }
+        DataSourceType::Clickhouse => {
+            tracing::info!(
+                "Registering ClickHouse table: {} (hierarchy_level: {:?})",
+                source.name,
+                source.hierarchy_level
+            );
+
+            let connection_string = source.connection_string.as_ref().ok_or_else(|| {
+                ConfigError::MissingConnectionString {
+                    name: source.name.clone(),
+                }
+            })?;
+
+            register_clickhouse_tables(
+                session_ctx,
+                &source.name,
+                connection_string,
+                source.options.as_ref(),
+                source.hierarchy_level,
+            )
+            .await
+            .map_err(|e| {
+                tracing::error!(
+                    "ClickHouse registration failed for '{}': {:?}",
                     source.name,
                     e
                 );
--- a/crates/skardi/src/sources/providers/clickhouse.rs
+++ b/crates/skardi/src/sources/providers/clickhouse.rs
@@ -1,0 +1,835 @@
+//! ClickHouse data source provider.
+//!
+//! ClickHouse is a columnar OLAP database with an HTTP query interface. This
+//! module drives the ClickHouse table provider shipped by
+//! `datafusion-table-providers`: filters, projections, and limits are unparsed
+//! back to ClickHouse SQL and executed server-side, so scans stream only the
+//! rows the query actually needs.
+//!
+//! Access is **read-only**. ClickHouse has no transactional UPDATE/DELETE —
+//! mutations (`ALTER TABLE ... UPDATE/DELETE`) are asynchronous background
+//! rewrites, and a mid-stream INSERT failure leaves partial parts visible.
+//! Neither matches the semantics Skardi's write path promises, so ClickHouse
+//! sources never participate in CRUD or job destinations.
+//!
+//! Two registration modes, mirroring the other SQL providers:
+//! - **Table mode** (default): one ClickHouse table registered under `name`.
+//! - **Catalog mode**: every table across the server's non-system databases is
+//!   registered under the `name` catalog as `name.<database>.<table>`.
+
+use crate::sources::DataSourceType;
+use crate::sources::hierarchy::{
+    HierarchyLevel, SourceLabel, build_catalog, parse_allowed_schemas, retry_with_timeout,
+};
+use anyhow::{Context, Result, anyhow};
+use async_trait::async_trait;
+use datafusion::arrow::datatypes::SchemaRef;
+use datafusion::catalog::Session;
+use datafusion::common::Statistics;
+use datafusion::datasource::TableProvider;
+use datafusion::logical_expr::{Expr, TableProviderFilterPushDown, TableType};
+use datafusion::physical_expr::PhysicalExpr;
+use datafusion::physical_plan::ExecutionPlan;
+use datafusion::physical_plan::projection::ProjectionExec;
+use datafusion::prelude::SessionContext;
+use datafusion::sql::TableReference;
+use datafusion_table_providers::clickhouse::ClickHouseTableFactory;
+use datafusion_table_providers::sql::db_connection_pool::clickhousepool::ClickHouseConnectionPool;
+use datafusion_table_providers::sql::db_connection_pool::dbconnection::AsyncDbConnection;
+use secrecy::SecretString;
+use std::any::Any;
+use std::collections::HashMap;
+use std::sync::Arc;
+
+// Skardi-side `options` keys (as written in a ctx YAML). Centralised so the
+// recognised vocabulary can't drift between the parser, the docs, and the
+// tests.
+
+/// ClickHouse table to register (required in table mode).
+const OPT_TABLE: &str = "table";
+/// ClickHouse database holding the table (table mode; optional — defaults to
+/// the server's default database, usually `default`).
+const OPT_DATABASE: &str = "database";
+/// Name of an environment variable holding the username. ClickHouse's
+/// out-of-the-box `default` user needs no credentials, so both are optional.
+const OPT_USER_ENV: &str = "user_env";
+/// Name of an environment variable holding the password.
+const OPT_PASS_ENV: &str = "pass_env";
+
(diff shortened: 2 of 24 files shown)


crates/skardi/src/sources/providers/clickhouse.rs
```diff
@@ -0,0 +1,835 @@
...
+    if schema_tables.is_empty() {
+        tracing::warn!(
+            "No tables found in ClickHouse catalog for source '{}'",
+            catalog_name
+        );
+    }
+
+    let table_count = schema_tables.len();
+
+    build_catalog(
```
abbccdda: ### 1. Catalog mode is fail-fast, and real ClickHouse servers will trip it — **reproduced live**

`register_clickhouse_catalog` uses the fail-fast `build_catalog`: the first table-provider failure aborts the whole catalog. Provider construction runs `SELECT * FROM db.table LIMIT 0` (upstream `get_schema`) against every row of `system.tables`, and upstream's `tables()` does not filter by engine. Real ClickHouse databases routinely contain:

- **Kafka / RabbitMQ / NATS engine tables** — direct SELECT errors ("direct select is not allowed") unless `stream_like_engine_allow_direct_select` is set. One Kafka table anywhere in an allowed database kills the entire catalog registration, and with it server startup. **Verified empirically**: adding one `ENGINE = Kafka` table to the demo `mydb` database makes `skardi-server --ctx ctx_clickhouse_catalog_demo.yaml` exit 1 with `Code: 620 ... Direct select is not allowed (QUERY_NOT_ALLOWED)`.
- **Broken views** — a VIEW whose `as_select` references a since-dropped table fails schema inference the same way.
- **Materialized-view inner tables** (`.inner_id.<uuid>` in Atomic databases) — these register successfully but are implementation-detail clutter with near-unqueryable names.

DynamoDB already solved exactly this with `build_catalog_best_effort` (`dynamodb.rs:2150`) — skip-and-warn per table. Recommend the same here, plus optionally filtering `.inner%` names and stream-like engines in `list_clickhouse_tables` (the engine column is right there in `system.tables`).

crates/skardi/src/sources/providers/clickhouse.rs
```diff
@@ -0,0 +1,835 @@
...
+        "http" | "https" => {}
+        other => {
+            return Err(anyhow!(
+                "ClickHouse connection string must use the http:// or https:// interface \
+                 (got '{other}://'); the native TCP protocol is not supported"
+            ));
+        }
+    }
+
+    if !parsed.username().is_empty() || parsed.password().is_some() {
```
abbccdda, line 484: ### 2. Embedded credentials: warn-and-continue leaks the secret and fails confusingly

`parse_connection_params` only *warns* when the URL contains `user:pass@`, then proceeds. I verified live that clickhouse-rs 0.14.3 does ignore URL credentials (a context with only embedded creds fails with `default: Authentication failed`), so the warning text is accurate — but the behavior still has two problems:

- The raw connection string — real password included — is logged at info level in both registration paths (`clickhouse.rs:136`, `:179`) and returned verbatim by the `get_data_sources` API (`pipeline_handlers.rs`, whose comment asserts "credentials are not stored in connection strings" — an assumption this provider enforces only by warning).
- The user who embedded working credentials gets a fail-late, misleading error (`default: Authentication failed` after 3 retries) instead of being told up front their URL format is unsupported.

Since embedded credentials never work anyway, make this a **hard error** instead of a warning — it closes the log/API leak and turns a confusing auth failure into an actionable config error.

crates/server/src/config.rs
```diff
@@ -714,6 +715,7 @@ const CATALOG_SUPPORTED_SOURCES: &[DataSourceType] = &[
     DataSourceType::Sqlite,
     DataSourceType::Seekdb,
     DataSourceType::Dynamodb,
+    DataSourceType::Clickhouse,
```
abbccdda, line 747: ### 3. Catalog mode rejects `table` but not `database`

Config validation rejects `table`/`schema` options in catalog mode (the conflict check at `config.rs:767`, just below this list), and the README states "`table` / `schema` are rejected when `hierarchy_level: catalog` is set" — but ClickHouse's schema-analog option is named **`database`**, which is not in the conflict list. A `database` option on a catalog-mode ClickHouse source slips through validation and silently changes the pool's default database. Add `"database"` to the conflict check (or make the conflict list per-source-type).

crates/skardi/src/sources/providers/clickhouse.rs
```diff
@@ -0,0 +1,835 @@
...
+        DataSourceType::Clickhouse,
+        HierarchyLevel::Catalog,
+        catalog_name,
+    );
+    let pool = build_pool(label, params).await.with_context(|| {
+        format!("Failed to create ClickHouse connection pool for catalog '{catalog_name}'")
+    })?;
+
+    let allowed_schemas = parse_allowed_schemas(options);
+    let schema_tables = retry_with_timeout(label, "system.tables introspection", || async {
```
abbccdda, line 274: ### 4. One 5-second timeout wraps the entire N+1 introspection loop

`list_clickhouse_tables` runs one `schemas()` query plus one `tables()` query *per database*, sequentially — all inside a single `retry_with_timeout` window of `CONNECT_TIMEOUT` = 5s. A server with many databases (or WAN latency) blows the budget, retries the whole loop three times, and fails startup. MySQL doesn't have this problem because its introspection is one `information_schema` query. Either batch into a single query (`SELECT database, name FROM system.tables WHERE database IN (...)`) or scope the timeout per database.

crates/skardi/src/sources/providers/clickhouse.rs
```diff
@@ -0,0 +1,835 @@
...
+        state: &dyn Session,
+        projection: Option<&Vec<usize>>,
+        filters: &[Expr],
+        limit: Option<usize>,
+    ) -> datafusion::common::Result<Arc<dyn ExecutionPlan>> {
+        match projection {
+            // Empty projection (count(*) / EXISTS): fetch one column so the
+            // unparsed ClickHouse SQL stays well-formed, then drop it again to
+            // honour the requested zero-column output.
+            Some(p) if p.is_empty() => {
```
abbccdda: ### 5. `count(*)` streams an entire column, and column 0 is an arbitrary choice

The enabled Cargo feature is `clickhouse` only, not `clickhouse-federation`, so aggregates are never pushed down: `SELECT count(*)` scans column 0 in full over HTTP and counts client-side. Parity with the InfluxDB wrapper, but ClickHouse tables are OLAP-sized — `count(*)` on a 10B-row table becomes a multi-GB transfer the server could answer from metadata instantly. Cheap improvements: pick the narrowest fixed-width column from `schema()` instead of index 0 (column 0 could be a wide String), and document the limitation in `docs/clickhouse/README.md`. Longer term, the federation feature fixes this class of issue wholesale.

abbccdda (commented):
## Review: ClickHouse table provider

Overall a solid, well-tested addition that follows the house patterns (env-based credentials, eager schema fetch, read-only enforced at config/executor/provider layers, the count(*) wrapper mirroring `influxdb.rs`). Findings are posted as inline comments, ordered by production impact (#1–#5). I verified the upstream claims against the vendored `datafusion-table-providers 0.10.1` source.

If prioritizing: #1 (best-effort catalog) before merge; #2/#3 are small changes with security/correctness payoff; #4/#5 can be fast-follows.

### Minor

- **Doc inaccuracy:** `clickhouse.rs:256` (and the README) say schema is inferred via `DESCRIBE TABLE`; upstream actually uses `SELECT * … LIMIT 0` through the ArrowStream format plus an engine lookup in `system.tables`. Worth correcting since the mechanism determines which failures occur (e.g. the Kafka-engine case in #1).
- **No timeout around per-table schema fetch:** `build_clickhouse_table_provider` isn't wrapped in `retry_with_timeout`, so an endpoint that hangs after listing succeeds stalls startup indefinitely. Matches mysql (inherited gap), so not a blocker — but catalog mode multiplies exposure by table count.
- **No config-validation tests:** the DynamoDB PR added `validate_accepts_dynamodb_read_write` / catalog-option tests in `config.rs`; ClickHouse has none — `read_write` rejection rides the generic `UnsupportedWriteMode` path untested for this type, and a test for #3 (`database` in catalog mode) would have caught that gap.
- **Type-coverage gap in fixtures:** CI tables only use UInt32/String/Float64/Bool/DateTime/Nullable(String). The types most likely to break ClickHouse→Arrow mapping — `LowCardinality(String)` (dictionary encoding can differ between the LIMIT-0 schema fetch and real data batches, and upstream derives the stream schema from the first batch), `DateTime64`, `Decimal`, `Enum8`, `UUID`, `Array` — are untested. One kitchen-sink types fixture would harden this a lot. A catalog-mode test including a VIEW would also cover the view branch in upstream `get_schema`.

### What checks out

- Read-only stance enforced consistently: config validation, `JobExecutor` destination rejection (with test), no write path in the provider.
- Option validation fires before any network call (tested offline-safe).
- Upstream behaviors the docs rely on verified against vendored 0.10.1 source: `schemas()` excludes `system`/`information_schema`, and nonexistent `allowed_schemas` entries contribute nothing rather than erroring.

### Verification (run locally against `clickhouse/clickhouse-server:24.8`)

- **Unit + integration tests**: `cargo test -p skardi clickhouse -- --include-ignored` → **20 passed, 0 failed** (12 unit + all 8 live integration tests: register/scan, filter pushdown, count(*) empty projection, NULL handling, empty-table schema inference, numeric aggregation, catalog mode, plus the executor destination-rejection test).
- **README demo, table mode**: all 5 pipelines (`query_user_by_id`, `list_all_users`, `products_by_category`, `user_order_summary`, `federated_stock_value`) return exactly the documented rows. One cosmetic nit: `federated_stock_value` returns `stock_value: 16999.390000000003` where the README shows `16999.39` — consider a `ROUND(..., 2)` in the pipeline or noting float formatting.
- **README demo, catalog mode**: both `catalog_demo` pipelines return the documented rows; catalog registered 4 tables.
- **Finding #1 repro**: `CREATE TABLE mydb.kafka_events (...) ENGINE = Kafka` → catalog-mode server exits 1 at startup with `Code: 620 Direct select is not allowed`.
- **Finding #2 repro**: context with `connection_string: http://skardi_user:skardi_pass@localhost:8123` and no `user_env`/`pass_env` → embedded creds ignored, 3 retries of `default: Authentication failed`, registration fails; warning + full connection string appear in server logs.

crates/skardi/src/sources/providers/clickhouse.rs
abbccdda: Fixed in ad2e81a. Catalog assembly now uses `build_catalog_best_effort` (same as DynamoDB), and `list_clickhouse_tables` filters stream-like engines (Kafka/RabbitMQ/NATS/FileLog) and `.inner*` materialized-view inner tables up front via the engine column. Verified live: with `mydb.kafka_events` (ENGINE = Kafka) and a broken view present, the catalog-demo server now starts, logs one skip line per excluded table, registers 4 tables (1 skipped), and both demo pipelines return the documented rows.

crates/skardi/src/sources/providers/clickhouse.rs
abbccdda: Fixed in ad2e81a. `parse_connection_params` now hard-errors on URL-embedded credentials, and both registration paths validate the connection string *before* the `tracing::info!` that logs it, so the secret never reaches the logs. Verified live: a ctx with `http://skardi_user:skardi_pass@localhost:8123` exits 1 immediately with the actionable message, and the password appears zero times in the log output.

crates/server/src/config.rs
abbccdda: Fixed in ad2e81a. `database` joined `table`/`schema` in the catalog-mode conflict list (it's only used by ClickHouse among catalog-supported types — InfluxDB uses it too but doesn't support catalog mode, so no cross-type fallout). Added config-validation tests: `validate_rejects_clickhouse_catalog_database_option`, `validate_rejects_clickhouse_read_write`, and `validate_accepts_clickhouse_table_mode_with_database_option`. README updated to list all three rejected options.

crates/skardi/src/sources/providers/clickhouse.rs
abbccdda: Fixed in ad2e81a — took the batching option: introspection is now a single `SELECT database, name, engine FROM system.tables WHERE database IN (...)` (or `NOT IN` the system list when no allow-list is set), so one 5s window covers one query regardless of database count. This required a direct dep on the `clickhouse` crate — same version the vendored provider already uses, so nothing new enters the tree. The batched query also supplies the engine column that finding #1's filtering needed. Separately, per-table schema inference is now wrapped in `retry_with_timeout` so a hanging endpoint can't stall startup indefinitely (the minor from the review summary).

crates/skardi/src/sources/providers/clickhouse.rs
abbccdda: Cheap improvements landed in ad2e81a: the empty-projection scan now picks the narrowest fixed-width column via `narrowest_column_index` (Boolean counted as 1 byte since Arrow bit-packs it; falls back to index 0 when every column is variable-width), with unit tests, and the no-aggregate-pushdown limitation is documented in the Query Pushdown section of `docs/clickhouse/README.md`. The `clickhouse-federation` feature remains a follow-up.

crates/skardi/src/sources/providers/clickhouse.rs
```diff
@@ -0,0 +1,966 @@
...
+        "http" | "https" => {}
+        other => {
+            return Err(anyhow!(
+                "ClickHouse connection string must use the http:// or https:// interface \
+                 (got '{other}://'); the native TCP protocol is not supported"
+            ));
+        }
+    }
+
+    if !parsed.username().is_empty() || parsed.password().is_some() {
```
bakey, line 484: [security] URL query parameters can carry credentials, bypassing the embedded-credential check                                                                                           
                                                                                                                                                                                            
   parse_connection_params rejects user:pass@host in the URL (clickhouse.rs:504), which is great — but ClickHouse's HTTP interface also accepts credentials as query parameters             
   (?user=u&password=p), and those pass validation today:                                                                                                                                   
                                                                                                                                                                                            
   ```                                                                                                                                                                                      
     http://host:8123/?user=admin&password=secret                                                                                                                                           
   ```                                                                                                                                                                                      
                                                                                                                                                                                            
   Since the connection string is logged at INFO on registration (clickhouse.rs:146, clickhouse.rs:191) and returned verbatim as url by the /data_source endpoint, a query-string password  
   would leak into both logs and the API — the exact exposure the embedded-credentials check was added to prevent (and the CI seed script itself uses this auth style, so users are likely  
   to reach for it).                                                                                                                                                                        
                                                                                                                                                                                            
   Could we reject any URL with a query string (or at least one containing user/password params) in parse_connection_params? Something like:                                                
                                                                                                                                                                                            
   ```rust                                                                                                                                                                                  
     if parsed.query().is_some() {                                                                                                                                                          
         return Err(anyhow!(                                                                                                                                                                
             "ClickHouse connection string must not contain query parameters \                                                                                                              
              (credentials in the query string would be logged and exposed by the \                                                                                                         
              data-sources API). Use the '{OPT_USER_ENV}' and '{OPT_PASS_ENV}' \                                                                                                            
              options instead."                                                                                                                                                             
         ));                                                                                                                                                                                
     }                                                                                                                                                                                      
   ```                                                                                                                                                                                      
                                                                                                                                                                                            
   A unit test alongside parse_connection_params_embedded_credentials_are_rejected would lock this in.

crates/skardi/src/sources/providers/clickhouse.rs
```diff
@@ -0,0 +1,966 @@
...
+/// the inner table, then strip it back to zero columns with a
+/// [`ProjectionExec`], which preserves the row count. All other projections
+/// delegate straight to the inner table.
+///
+/// The column still streams in full over HTTP (the enabled feature set does
+/// no aggregate pushdown — see `docs/clickhouse/README.md`), so we pick the
+/// narrowest fixed-width column rather than whatever sits at index 0, which
+/// could be an arbitrarily wide String.
+#[derive(Debug)]
+struct CountSafeClickHouseTable {
```
bakey: [maintainability] Duplicated count(*) workaround — share it?                                                                                                                             
                                                                                                                                                                                            
   CountSafeClickHouseTable is functionally identical to CountSafeFlightTable in influxdb.rs — both wrap the same upstream empty-projection bug. Consider extracting a shared CountSafeTable
    in providers/mod.rs (same dedup pattern as is_pushable_binary_filter) so the eventual upstream fix only needs to be unwound in one place. Non-blocking.

crates/skardi/src/sources/providers/clickhouse.rs
```diff
@@ -0,0 +1,966 @@
...
+///     "events",
+///     "http://localhost:8123",
+///     Some(&options),
+///     HierarchyLevel::Table,
+/// )
+/// .await?;
+/// # Ok(())
+/// # }
+/// ```
+pub async fn register_clickhouse_tables(
```
bakey, line 109: P1 Correctness/Security: ClickHouse configuration validation only runs in the server path, so the CLI and public provider API bypass it.
**validate_data_sources** rejects catalog-mode conflicts such as table / database and rejects access_mode: read_write for ClickHouse. However, the CLI calls register_clickhouse_tables directly, while the provider itself does not validate mode-specific options.

As a result, the CLI accepts configurations that the documentation says should be rejected, including access_mode: read_write and catalog configurations containing table-scoped options. An empty allowed_schemas value also bypasses server validation and becomes “all schemas” through parse_allowed_schemas. More generally, unknown options such as a misspelled pass_env are silently ignored, which can cause an unintended connection as ClickHouse’s default user.

Consider move ClickHouse-specific option and mode validation to the provider boundary, then reuse it from both the server and CLI paths. The validation should reject unknown options and enforce the same table-mode/catalog-mode contract everywhere.

crates/skardi/src/sources/providers/clickhouse.rs
abbccdda: Fixed in 394c83d. `parse_connection_params` now rejects **any** query string in the connection URL — credential-shaped or not — since the pool ignores query parameters entirely, so none of them can work and credential ones would leak through logs and the data-sources API. Unit test `parse_connection_params_query_string_is_rejected` covers both the `?user=&password=` case and a benign `?compress=1`.

crates/skardi/src/sources/providers/clickhouse.rs
abbccdda: Extracted in 394c83d: shared `CountSafeTable` (plus `narrowest_column_index`) now lives in `providers/mod.rs`, and both the ClickHouse and InfluxDB providers use it — as a bonus the InfluxDB path now also streams the narrowest fixed-width column instead of column 0. Unwinding the upstream empty-projection fix will be a one-place change.

crates/skardi/src/sources/providers/clickhouse.rs
abbccdda: Fixed in 394c83d + 4c34bc1. Validation now lives at the provider boundary (`validate_clickhouse_options`, called first thing in `register_clickhouse_tables`), so the CLI and public API get the same contract as server config validation: unknown option keys (e.g. a misspelled `pass_env`) are hard errors, `table`/`database` are rejected in catalog mode, `allowed_schemas` is rejected in table mode, and an `allowed_schemas` with no non-empty entry is rejected instead of falling through as "all schemas". The `access_mode: read_write` gap needed a signature change (4c34bc1): `register_clickhouse_tables` now takes `read_write` and hard-errors on it, following the OpenConnector/DynamoDB pattern, with both front-ends forwarding their access mode. Provider-level tests cover each rejection, and CLI-level tests (`register_clickhouse_source` module) verify the wiring.

bakey (approved):

PR #158, opened 2026-07-17
feat(server): add POST /query endpoint for ad-hoc SQL

BtXin:
## Summary

Adds a `POST /query` endpoint to skardi-server that executes a single ad-hoc SQL statement against the data sources registered from `ctx.yaml` — bringing the CLI's `skardi query --sql "..."` capability to the HTTP server.

Design spec and implementation plan: `docs/superpowers/specs/2026-07-17-query-endpoint-design.md`, `docs/superpowers/plans/2026-07-17-query-endpoint.md`.

## API

`POST /query` with `{ "sql": "...", "max_rows": 500 }` (`max_rows` optional, default 1000). Returns the same JSON envelope as pipeline execution, plus a `truncated` flag:

```json
{ "success": true, "data": [...], "rows": 500, "truncated": true, "execution_time_ms": 42, "timestamp": "..." }
```

## Statement policy

Enforced by the existing `sql_validator`, run per request against the current `ctx.yaml` access modes:

- **DDL** (CREATE/DROP/ALTER/TRUNCATE) and **COPY** — always rejected.
- **DML** (INSERT/UPDATE/DELETE) — allowed only against sources with `access_mode: read_write`; `read_only` (the default) is rejected.
- **SELECT / EXPLAIN-of-query / SHOW / DESCRIBE** — allowed.
- Multi-statement input, `EXPLAIN ANALYZE <write>`, `SET`, and `PREPARE`/`EXECUTE` are rejected — these otherwise smuggle writes/DDL past the access-mode gate via the shared `SessionContext`.

## Result cap

Default 1000 rows, overridable via `max_rows` (no server-side maximum). The cap is pushed into the DataFusion plan via a new `DataFusionEngine::execute_with_limit`; the endpoint fetches `max_rows + 1` to detect truncation, then slices. Overflow of `max_rows` is guarded (`saturating_add` + `i64::MAX` clamp for DataFusion's `i64` LIMIT literal).

## Other changes

- Extracted response/JSON helpers from the 1,600-line `pipeline_handlers.rs` into a shared `response.rs`; pipeline responses remain byte-identical.
- Extracted a shared `require_session` auth helper used by both the query and pipeline handlers.
- Hardened the SQL validator (COPY, single-statement, EXPLAIN recursion, SET, PREPARE/EXECUTE) — this also tightens pipeline config-load validation.

## Testing

- Validator unit tests: 26 passing (COPY, single-statement, EXPLAIN/SET/PREPARE bypass coverage).
- Engine unit tests: `execute_with_limit` truncation / full-result / empty-result.
- HTTP integration (`crates/server/tests/query_http.rs`): 15 passing — SELECT envelope, truncation boundaries, DDL/COPY/multi-statement/parse rejection, read-only vs read-write DML, `EXPLAIN ANALYZE`/`PREPARE` bypass rejection, `usize::MAX` overflow guard, execution error, and a 401 with the real better-auth layer.
- `cargo fmt --all` clean; `cargo test --workspace` green.

## Notes for reviewers

The validator is a denylist (`_ => Ok(())` catch-all with explicit rejections). Three same-class bypasses (EXPLAIN ANALYZE, SET, PREPARE/EXECUTE) were found and patched during review. A follow-up to convert it to an explicit read-only allowlist — to prevent a future DataFusion/sqlparser bump from reintroducing an executable statement type — is worth tracking separately.

### DIFF
--- a/crates/server/src/auth/bridge.rs
+++ b/crates/server/src/auth/bridge.rs
@@ -178,8 +178,15 @@ impl TableProvider for AuthSessionsTable {
 // Registration helper
 // ---------------------------------------------------------------------------
 
+/// Schema the auth virtual tables (`users`, `sessions`) register under.
+///
+/// Single source of truth for the reserved schema name: the ad-hoc `/query`
+/// policy denies exactly this schema (see `config::adhoc_policy_from_sources`),
+/// so the two must never drift apart.
+pub const AUTH_SCHEMA: &str = "auth";
+
 /// Register `auth.users` and `auth.sessions` virtual tables into the given
-/// DataFusion `SessionContext` under a dedicated `auth` schema.
+/// DataFusion `SessionContext` under the [`AUTH_SCHEMA`] schema.
 pub fn register_auth_tables(
     ctx: &mut SessionContext,
     auth: Arc<BetterAuth<DieselSqliteAdapter>>,
@@ -204,12 +211,14 @@ pub fn register_auth_tables(
         .catalog("datafusion")
         .ok_or_else(|| anyhow::anyhow!("Default catalog 'datafusion' not found"))?;
 
-    if catalog.schema("auth").is_some() {
-        return Err(anyhow::anyhow!("Auth schema 'auth' is already registered"));
+    if catalog.schema(AUTH_SCHEMA).is_some() {
+        return Err(anyhow::anyhow!(
+            "Auth schema '{AUTH_SCHEMA}' is already registered"
+        ));
     }
 
     catalog
-        .register_schema("auth", schema)
+        .register_schema(AUTH_SCHEMA, schema)
         .map_err(|e| anyhow::anyhow!("Failed to register auth schema: {}", e))?;
 
     tracing::info!("Registered DataFusion tables: auth.users, auth.sessions");
--- a/crates/server/src/auth/routes.rs
+++ b/crates/server/src/auth/routes.rs
@@ -394,6 +394,7 @@ mod tests {
             metrics: PipelineMetrics::new(),
             auth_layer: AuthLayer::None,
             jobs: None,
+            validator_config: Arc::new(crate::config::validator_config_from_sources(&[])),
         }
     }
 
@@ -448,6 +449,7 @@ mod tests {
             metrics: PipelineMetrics::new(),
             auth_layer: layer,
             jobs: None,
+            validator_config: Arc::new(crate::config::validator_config_from_sources(&[])),
         }
     }
 
--- a/crates/server/src/auth/routes.rs
+++ b/crates/server/src/auth/routes.rs
@@ -363,13 +363,12 @@ mod tests {
     fn make_no_auth_state() -> AppState {
         use crate::auth::layer::AuthLayer;
         use crate::config::{CliArgs, ServerConfig};
-        use crate::metrics::PipelineMetrics;
         use crate::semantics::SemanticsRegistry;
         use crate::server::AppState;
         use datafusion::prelude::SessionContext;
         use skardi::engine::datafusion::DataFusionEngine;
         use std::path::PathBuf;
-        use std::sync::{Arc, RwLock};
+        use std::sync::Arc;
 
         let config = ServerConfig {
             pipelines: Default::default(),
@@ -387,15 +386,7 @@ mod tests {
         };
         let session_ctx = Arc::new(SessionContext::new());
         let engine = Arc::new(DataFusionEngine::new_with_arc(session_ctx.clone()));
-        AppState {
-            config: Arc::new(RwLock::new(config)),
-            engine,
-            session_ctx,
-            metrics: PipelineMetrics::new(),
-            auth_layer: AuthLayer::None,
-            jobs: None,
-            validator_config: Arc::new(crate::config::validator_config_from_sources(&[])),
-        }
+        AppState::new(config, engine, session_ctx, AuthLayer::None, None)
     }
 
     #[tokio::test]
@@ -408,13 +399,12 @@ mod tests {
     async fn make_better_auth_state() -> AppState {
         use crate::auth::layer::AuthLayer;
         use crate::config::{CliArgs, ServerConfig};
-        use crate::metrics::PipelineMetrics;
         use crate::semantics::SemanticsRegistry;
         use crate::server::AppState;
         use datafusion::prelude::SessionContext;
         use skardi::engine::datafusion::DataFusionEngine;
         use std::path::PathBuf;
-        use std::sync::{Arc, RwLock};
+        use std::sync::Arc;
 
         unsafe {
             std::env::set_var("AUTH_SECRET", "test-secret-that-is-at-least-32-characters!");
@@ -442,15 +432,7 @@ mod tests {
         };
         let session_ctx = Arc::new(SessionContext::new());
         let engine = Arc::new(DataFusionEngine::new_with_arc(session_ctx.clone()));
-        AppState {
-            config: Arc::new(RwLock::new(config)),
-            engine,
-            session_ctx,
-            metrics: PipelineMetrics::new(),
-            auth_layer: layer,
-            jobs: None,
-            validator_config: Arc::new(crate::config::validator_config_from_sources(&[])),
-        }
+        AppState::new(config, engine, session_ctx, layer, None)
     }
 
     #[tokio::test]
--- a/crates/server/src/auth/routes.rs
+++ b/crates/server/src/auth/routes.rs
@@ -8,6 +8,7 @@
 use std::collections::HashMap;
 
 use axum::{
+    Json,
     body::Body,
     extract::State,
     http::{HeaderName, HeaderValue, Request, Response, StatusCode},
@@ -16,6 +17,7 @@ use axum::{
 use better_auth::{AuthRequest, HttpMethod, SessionOps};
 use cookie::Cookie;
 
+use crate::response::{ErrorResponse, create_error_response};
 use crate::server::AppState;
 
 async fn to_auth_request(req: Request<Body>) -> Result<AuthRequest, String> {
@@ -179,6 +181,27 @@ pub async fn verify_session(
             .body(Body::from(r#"{"error":"Invalid or expired session"}"#))
             .expect("static response always builds"))
     }
+}
+
+/// Enforce the session gate for JSON API handlers: on auth failure, convert
+/// the raw `verify_session` response into the shared `ErrorResponse`
+/// envelope handlers return.
+pub(crate) async fn require_session(
+    state: &AppState,
+    headers: &axum::http::HeaderMap,
+) -> Result<(), (StatusCode, Json<ErrorResponse>)> {
+    if let Err(unauth_response) = verify_session(state, headers).await {
+        let status = unauth_response.status();
+        let body_bytes = axum::body::to_bytes(unauth_response.into_body(), 512)
+            .await
+            .unwrap_or_default();
+        let msg = serde_json::from_slice::<serde_json::Value>(&body_bytes)
+            .ok()
+            .and_then(|v| v["error"].as_str().map(|s| s.to_string()))
+            .unwrap_or_else(|| "Authentication required".to_string());
+        return Err((status, create_error_response(&msg, "unauthorized", None)));
+    }
+    Ok(())
 }
 
 #[cfg(test)]
--- a/crates/server/src/query_handlers.rs
+++ b/crates/server/src/query_handlers.rs
@@ -100,7 +100,9 @@ pub async fn execute_query(
                 SqlValidationError::NotExactlyOneStatement { count } => {
                     Some(serde_json::json!({ "statement_count": count }))
                 }
-                SqlValidationError::CopyNotAllowed | SqlValidationError::ParseError(_) => None,
+                SqlValidationError::CopyNotAllowed
+                | SqlValidationError::StatementNotAllowed { .. }
+                | SqlValidationError::ParseError(_) => None,
             };
 
             return Err((
--- a/crates/server/src/query_handlers.rs
+++ b/crates/server/src/query_handlers.rs
@@ -16,7 +16,6 @@ use skardi::sources::sql_validator::{SqlValidationError, StatementKind, validate
 use std::time::Instant;
 
 use crate::auth::routes::require_session;
-use crate::config::validator_config_from_sources;
 use crate::response::{
     ErrorResponse, create_error_response, create_success_response, record_batch_to_json,
 };
@@ -69,17 +68,7 @@ pub async fn execute_query(
         None => DEFAULT_MAX_ROWS,
     };
 
-    // Build the validator config from the current data sources on every
-    // request so runtime config updates are respected.
-    let validator_config = {
-        let config = app_state
-            .config
-            .read()
-            .unwrap_or_else(|poisoned| poisoned.into_inner());
-        validator_config_from_sources(&config.data_sources)
-    };
-
-    let statement_kind = match validate_single_sql(&request.sql, &validator_config) {
+    let statement_kind = match validate_single_sql(&request.sql, &app_state.validator_config) {
         Ok(kind) => kind,
         Err(e) => {
             tracing::info!("Rejected ad-hoc query: {}", e);
@@ -100,8 +89,10 @@ pub async fn execute_query(
                 SqlValidationError::NotExactlyOneStatement { count } => {
                     Some(serde_json::json!({ "statement_count": count }))
                 }
-                SqlValidationError::CopyNotAllowed
-                | SqlValidationError::StatementNotAllowed { .. }
+                SqlValidationError::SchemaNotAllowed { schema, table } => {
+                    Some(serde_json::json!({ "schema": schema, "table": table }))
+                }
+                SqlValidationError::StatementNotAllowed { .. }
                 | SqlValidationError::ParseError(_) => None,
             };
 
@@ -111,6 +102,15 @@ pub async fn execute_query(
             ));
         }
     };
+
+    // Audit trail: every statement this endpoint hands to the engine is
+    // recorded before execution, so a crash mid-query still leaves a record.
+    tracing::info!(
+        sql = %request.sql,
+        max_rows,
+        kind = ?statement_kind,
+        "Executing ad-hoc query"
+    );
 
     // Queries get the row cap pushed into the plan (fetch cap + 1 so
     // truncation is detectable). Writes and other statements return small
--- a/crates/server/src/query_handlers.rs
+++ b/crates/server/src/query_handlers.rs
@@ -137,6 +137,9 @@ pub async fn execute_query(
     let record_batch = match result {
         Ok(batch) => batch,
         Err(e) => {
+            // Log the engine error server-side; do not echo it to the client.
+            // DataFusion errors can quote row/column values and internal
+            // schema, so the response stays generic.
             tracing::error!("Ad-hoc query execution failed: {}", e);
             tracing::debug!("Failed SQL query: {}", request.sql);
 
@@ -147,18 +150,12 @@ pub async fn execute_query(
                 "query_execution_error",
             );
 
-            let error_details = serde_json::json!({
-                "engine_error": e.to_string(),
-                "registered_tables": "Check server logs for data source registration status",
-                "suggestion": "Verify that data sources are properly registered and accessible"
-            });
-
             return Err((
                 StatusCode::INTERNAL_SERVER_ERROR,
                 create_error_response(
-                    &format!("SQL query execution failed: {}", e),
+                    "SQL query execution failed; see server logs for details",
                     "query_execution_error",
-                    Some(error_details),
+                    None,
                 ),
             ));
         }
@@ -174,7 +171,14 @@ pub async fn execute_query(
     let data = match record_batch_to_json(&record_batch) {
         Ok(json_data) => json_data,
         Err(e) => {
-            tracing::error!("Failed to convert results to JSON: {}", e);
+            // Keep the schema/error detail in the server log only; the client
+            // response must not echo internal schema back.
+            tracing::error!(
+                "Failed to convert results to JSON: {} (schema: {:?}, rows: {})",
+                e,
+                record_batch.schema(),
+                record_batch.num_rows()
+            );
 
             let elapsed_ms = start_time.elapsed().as_millis() as f64;
             app_state.metrics.record_error(
@@ -183,18 +187,12 @@ pub async fn execute_query(
                 "result_conversion_error",
             );
 
-            let error_details = serde_json::json!({
-                "conversion_error": e.to_string(),
-                "record_batch_schema": format!("{:?}", record_batch.schema()),
-                "record_batch_rows": record_batch.num_rows()
-            });
-
             return Err((
                 StatusCode::INTERNAL_SERVER_ERROR,
                 create_error_response(
-                    &format!("Failed to convert query results to JSON: {}", e),
+                    "Failed to convert query results to JSON; see server logs for details",
                     "result_conversion_error",
-                    Some(error_details),
+                    None,
                 ),
             ));
         }
--- a/crates/server/src/query_handlers.rs
+++ b/crates/server/src/query_handlers.rs
@@ -68,7 +68,7 @@ pub async fn execute_query(
         None => DEFAULT_MAX_ROWS,
     };
 
-    let statement_kind = match validate_single_sql(&request.sql, &app_state.validator_config) {
+    let statement_kind = match validate_single_sql(&request.sql, &app_state.adhoc_policy) {
         Ok(kind) => kind,
         Err(e) => {
             tracing::info!("Rejected ad-hoc query: {}", e);
--- a/crates/server/src/query_handlers.rs
+++ b/crates/server/src/query_handlers.rs
@@ -49,6 +49,13 @@ pub async fn execute_query(
 
     let max_rows = match request.max_rows {
         Some(0) => {
+            let elapsed_ms = start_time.elapsed().as_millis() as f64;
+            app_state.metrics.record_error(
+                QUERY_METRICS_LABEL,
+                elapsed_ms,
+                "parameter_validation_error",
+            );
+
             return Err((
                 StatusCode::BAD_REQUEST,
                 create_error_response(
@@ -108,9 +115,18 @@ pub async fn execute_query(
     // result batches (e.g. an insert count) and run uncapped.
     let result = match statement_kind {
         StatementKind::Query => {
+            // DataFusion's LIMIT plan stores `fetch` as an `i64` literal
+            // internally, so a `usize` fetch that doesn't fit in `i64`
+            // (reachable via a client-supplied `max_rows` near `usize::MAX`)
+            // round-trips to a negative literal and fails query planning.
+            // `saturating_add` avoids the arithmetic overflow from `+ 1`;
+            // the `min` additionally clamps to `i64::MAX`, a cap far beyond
+            // any result set this endpoint could ever materialize, so no
+            // real request is affected.
+            let fetch = max_rows.saturating_add(1).min(i64::MAX as usize);
             app_state
                 .engine
-                .execute_with_limit(&request.sql, max_rows + 1)
+                .execute_with_limit(&request.sql, fetch)
                 .await
         }
         StatementKind::Other => app_state.engine.execute(&request.sql).await,
--- a/crates/server/src/query_handlers.rs
+++ b/crates/server/src/query_handlers.rs
@@ -1,0 +1,205 @@
+//! Axum handler for the ad-hoc SQL endpoint.
+//!
+//! Endpoint mounted here:
+//!
+//! * `POST /query` — execute one SQL statement against the data sources
+//!   registered from the ctx file. DDL and COPY are always rejected; DML is
+//!   allowed only against sources configured with `access_mode: read_write`.
+//!   Query results are capped at `max_rows` (default 1000) and the response
+//!   carries a `truncated` flag.
+
+use axum::{Json, extract::State, http::StatusCode};
+use serde::Deserialize;
+use serde_json::Value;
+use skardi::engine::Engine;
+use skardi::sources::sql_validator::{SqlValidationError, StatementKind, validate_single_sql};
+use std::time::Instant;
+
+use crate::auth::routes::require_session;
+use crate::config::validator_config_from_sources;
+use crate::response::{
+    ErrorResponse, create_error_response, create_success_response, record_batch_to_json,
+};
+use crate::server::AppState;
+
+/// Default row cap applied when the request does not specify `max_rows`.
+const DEFAULT_MAX_ROWS: usize = 1000;
+
+/// Metrics label for ad-hoc queries (pipelines record under their own name).
+const QUERY_METRICS_LABEL: &str = "query";
+
+/// Request structure for ad-hoc query execution
+#[derive(Debug, Deserialize)]
+pub struct QueryRequest {
+    /// A single SQL statement to execute
+    pub sql: String,
+    /// Result row cap; defaults to [`DEFAULT_MAX_ROWS`]. Must be >= 1.
+    pub max_rows: Option<usize>,
+}
+
+/// Execute ad-hoc SQL endpoint - POST /query
+pub async fn execute_query(
+    State(app_state): State<AppState>,
+    headers: axum::http::HeaderMap,
+    Json(request): Json<QueryRequest>,
+) -> Result<Json<Value>, (StatusCode, Json<ErrorResponse>)> {
+    require_session(&app_state, &headers).await?;
+
+    let start_time = Instant::now();
+
+    let max_rows = match request.max_rows {
+        Some(0) => {
+            return Err((
+                StatusCode::BAD_REQUEST,
+                create_error_response(
+                    "max_rows must be a positive integer",
+                    "parameter_validation_error",
+                    None,
+                ),
+            ));
+        }
+        Some(n) => n,
+        None => DEFAULT_MAX_ROWS,
+    };
+
+    // Build the validator config from the current data sources on every
+    // request so runtime config updates are respected.
+    let validator_config = {
+        let config = app_state
+            .config
+            .read()
+            .unwrap_or_else(|poisoned| poisoned.into_inner());
+        validator_config_from_sources(&config.data_sources)
+    };
+
+    let statement_kind = match validate_single_sql(&request.sql, &validator_config) {
+        Ok(kind) => kind,
+        Err(e) => {
+            tracing::info!("Rejected ad-hoc query: {}", e);
+            tracing::debug!("Rejected SQL: {}", request.sql);
+
+            let elapsed_ms = start_time.elapsed().as_millis() as f64;
+            app_state
+                .metrics
+                .record_error(QUERY_METRICS_LABEL, elapsed_ms, "sql_validation_error");
+
+            let details = match &e {
+                SqlValidationError::DdlNotAllowed { operation } => {
+                    Some(serde_json::json!({ "operation": operation }))
+                }
+                SqlValidationError::WriteNotAllowed { operation, table } => {
+                    Some(serde_json::json!({ "operation": operation, "table": table }))
+                }
+                SqlValidationError::NotExactlyOneStatement { count } => {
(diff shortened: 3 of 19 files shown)


docs/superpowers/specs/2026-07-17-query-endpoint-design.md
```diff
@@ -0,0 +1,175 @@
...
+`execute_query` handler and its `QueryRequest` deserialization struct. Flow:
+
+1. `verify_session` auth check (identical to `execute_pipeline_by_name`).
+2. Build a `SqlValidatorConfig` from `state.config.read().data_sources` —
+   per request, so runtime config updates via the `RwLock` are respected.
+   This mirrors `validate_pipeline_sql` (`crates/server/src/config.rs`).
+3. `validate_sql` — rejects DDL/COPY/multi-statement, enforces per-table
+   access modes.
+4. Execute via `DataFusionEngine::execute_with_limit(sql, max_rows + 1)`.
+5. Slice to `max_rows` rows, set `truncated = fetched > max_rows`.
```
abbccdda, line 94: how do we expect client to react when they saw "truncated = true"? Are they having a way to paginate and get filtered rows on the next run?

crates/server/src/query_handlers.rs
```diff
@@ -0,0 +1,223 @@
...
+            // (reachable via a client-supplied `max_rows` near `usize::MAX`)
+            // round-trips to a negative literal and fails query planning.
+            // `saturating_add` avoids the arithmetic overflow from `+ 1`;
+            // the `min` additionally clamps to `i64::MAX`, a cap far beyond
+            // any result set this endpoint could ever materialize, so no
+            // real request is affected.
+            let fetch = max_rows.saturating_add(1).min(i64::MAX as usize);
+            app_state
+                .engine
+                .execute_with_limit(&request.sql, fetch)
```
abbccdda, line 131: **Security (high): any authenticated user can read every session token.** `/query` executes against the same `SessionContext` where `register_auth_tables` mounts `auth.sessions`, whose `token` column holds the live bearer credential that `get_session` authenticates with. The validator allows all reads unconditionally and `require_session` only checks that the caller is logged in — no row/tenant filter. So `SELECT token, user_id FROM auth.sessions` returns every active session, enabling account takeover.

Suggest restricting reads to registered data-source tables (a read allowlist), or excluding the `auth` schema from what `/query` can reach.

crates/skardi/src/sources/sql_validator.rs
```diff
@@ -152,6 +212,33 @@ fn validate_statement(
...
+        // above. Ad-hoc one-shot queries have no use for them — always blocked.
+        Statement::Prepare { .. } => Err(SqlValidationError::StatementNotAllowed {
+            operation: "PREPARE".to_string(),
+        }),
+        Statement::Execute { .. } => Err(SqlValidationError::StatementNotAllowed {
+            operation: "EXECUTE".to_string(),
+        }),
+
         // Read operations and others - always allowed
         _ => Ok(()),
```
abbccdda, line 385: **Statement policy is a default-allow denylist.** Ending in `_ => Ok(())` is why EXPLAIN/SET and PREPARE/EXECUTE each needed a follow-up patch commit, and holes remain. Against the vendored sqlparser 0.53 these still fall through today: `SetTimeZone`, `SetTransaction`, `Deallocate`, `Merge` (a DML write that also skips `check_write_access`), `Unload`, `StartTransaction`/`Commit`/`Rollback`, `Use`, `Pragma`, `Call`, `Grant`. Every sqlparser upgrade reopens this.

Invert to an allowlist: permit `Query`, access-checked `Insert`/`Update`/`Delete`, and `Explain`-of-an-allowed-statement; reject everything else (including unknown future variants).

crates/skardi/src/sources/sql_validator.rs
```diff
@@ -189,13 +276,13 @@ fn check_write_access(
...
     config: &SqlValidatorConfig,
 ) -> Result<(), SqlValidationError> {
-    if let Some(mode) = config.table_access_modes.get(table_name) {
-        if *mode == AccessMode::ReadOnly {
-            return Err(SqlValidationError::WriteNotAllowed {
-                operation: operation.to_string(),
-                table: table_name.to_string(),
-            });
-        }
+    if let Some(mode) = config.table_access_modes.get(table_name)
```
abbccdda: **`check_write_access` default-allows unknown tables, and matches only the bare table name.** Two related gaps:

1. It errors only when the table is present in `table_access_modes` and `ReadOnly`; anything absent falls through to `Ok(())`. `validator_config_from_sources` populates that map solely from `config.data_sources`, so `INSERT INTO auth.users` passes validation despite the documented "DML only on read_write sources" gate. (Not exploitable today — auth tables implement only `scan` and hierarchy children are built read-only — but the validation contract is wrong and only a second layer stops it.)
2. `extract_table_name` (line 247) keeps only `ObjectName.0.last()`, so `schema_a.orders` is checked against an unrelated flat source named `orders` — wrongly rejected in one direction, and a write to a read-only catalog table whose base name matches nothing passes entirely. Match on the fully-qualified name.

crates/skardi/src/sources/sql_validator.rs
```diff
@@ -64,6 +77,49 @@ pub fn validate_sql(sql: &str, config: &SqlValidatorConfig) -> Result<(), SqlVal
...
+///
+/// Applies the same rules as [`validate_sql`] (DDL and COPY always rejected,
+/// writes checked against per-table access modes) and additionally rejects
+/// input that parses to zero or more than one statement. Returns the
+/// statement's [`StatementKind`] on success.
+pub fn validate_single_sql(
+    sql: &str,
+    config: &SqlValidatorConfig,
+) -> Result<StatementKind, SqlValidationError> {
+    let preprocessed_sql = preprocess_parameters(sql);
```
abbccdda: **Validator preprocesses `{…}` but the engine executes the raw SQL.** `validate_single_sql` runs `preprocess_parameters` (pipeline-template brace substitution, no quote-awareness), while the handler executes the untransformed `request.sql`. A legitimate ad-hoc query such as `SELECT 'a{b' AS "c}d" FROM t` gets its brace span rewritten to `(NULL)`, fails to parse, and returns a spurious 400. A bypass isn't constructible (quote parity holds), but valid queries are wrongly rejected — skip parameter preprocessing on the ad-hoc path.

crates/skardi/src/sources/sql_validator.rs
```diff
@@ -152,6 +212,33 @@ fn validate_statement(
...
         }
 
+        // EXPLAIN wraps an inner statement that DataFusion may execute
+        // (EXPLAIN ANALYZE runs the plan). Validate the inner statement so
+        // EXPLAIN can never smuggle past DDL/COPY/write-access checks.
+        Statement::Explain { statement, .. } => validate_statement(statement, config),
+
+        // Session-mutating statements affect the process-wide SessionContext
+        // shared across all requests — always blocked.
+        Statement::SetVariable { .. } => Err(SqlValidationError::StatementNotAllowed {
```
abbccdda: **Breaking change to existing pipelines at startup.** These new denials live in the shared `validate_statement`, which the pipeline-load path also calls (`config.rs` `validate_pipeline_sql` → `validate_sql`). On `main` these statements were allowed. A deployment whose pipeline SQL uses `COPY … TO` (a legitimate DataFusion export) or `SET` now fails `load_server_config`, and the `?` there is fatal — the server refuses to start. Either scope the new denials to the ad-hoc `/query` path or flag this as a breaking change in the PR.

crates/server/src/server.rs
```diff
@@ -254,6 +255,7 @@ pub fn configure_routes(state: AppState) -> Router {
         .route("/pipelines", get(list_pipelines))
         .route("/pipeline/:name", get(get_pipelines_info))
         .route("/data_source", get(get_data_sources))
+        .route("/query", post(execute_query))
```
abbccdda, line 284: **Auth is opt-in per handler rather than middleware.** `require_session` is called manually inside each handler body, so protection is easy to forget — the jobs routes below (`/jobs/:name/run`, `/jobs/runs/:run_id/cancel`) trigger execution but never call it. Applying auth as a `route_layer` on the protected sub-router makes authentication the default and public routes the explicit exception.

crates/server/src/query_handlers.rs
```diff
@@ -0,0 +1,223 @@
...
+    };
+
+    // Build the validator config from the current data sources on every
+    // request so runtime config updates are respected.
+    let validator_config = {
+        let config = app_state
+            .config
+            .read()
+            .unwrap_or_else(|poisoned| poisoned.into_inner());
+        validator_config_from_sources(&config.data_sources)
```
abbccdda: **Per-request validator rebuild.** `validator_config_from_sources` allocates a fresh `HashMap` plus a lowercased `String` per data source on every request, under the config read lock. The justifying "so runtime config updates are respected" comment describes a writer that doesn't exist in the server crate. Build the `SqlValidatorConfig` once into `AppState` at startup.

crates/server/src/auth/routes.rs
```diff
@@ -181,6 +183,27 @@ pub async fn verify_session(
...
+/// Enforce the session gate for JSON API handlers: on auth failure, convert
+/// the raw `verify_session` response into the shared `ErrorResponse`
+/// envelope handlers return.
+pub(crate) async fn require_session(
+    state: &AppState,
+    headers: &axum::http::HeaderMap,
+) -> Result<(), (StatusCode, Json<ErrorResponse>)> {
+    if let Err(unauth_response) = verify_session(state, headers).await {
+        let status = unauth_response.status();
+        let body_bytes = axum::body::to_bytes(unauth_response.into_body(), 512)
```
abbccdda, line 195: **Auth error message round-trips through HTTP.** `require_session` serializes `verify_session`'s response, buffers the body (512-byte cap), and re-parses it as JSON just to recover a message the module already had as a `&str`. If the body shape/size ever changes, every JSON handler silently degrades to the generic "Authentication required". Have `verify_session` return a structured error (`StatusCode` + message) and build the HTTP response only at the edges.

crates/skardi/src/engine/datafusion.rs
```diff
@@ -124,28 +150,74 @@ impl Engine for DataFusionEngine {
...
+fn batches_to_single(schema: SchemaRef, batches: Vec<RecordBatch>) -> Result<RecordBatch> {
+    match batches.len() {
+        0 => Ok(RecordBatch::new_empty(schema)),
+        1 => Ok(batches
+            .into_iter()
+            .next()
+            .expect("len == 1 guarantees first element")),
+        _ => {
+            let batch_schema = batches[0].schema();
+            concat_batches(&batch_schema, &batches)
```
abbccdda, line 168: **Avoidable copies on the result hot path.** `batches_to_single` `concat_batches` deep-copies every column into one contiguous batch, only for the handler to `slice` it. Separately, `record_batch_to_json` (response.rs) encodes arrow → JSON bytes → re-parses into a `Value` tree → the envelope re-serializes it — three passes over client-controlled-size data. Returning `Vec<RecordBatch>` and feeding the slice straight to the JSON writer (which accepts multiple batches) avoids both.

abbccdda (commented):
Automated review of the POST /query endpoint (7 finder angles, each finding independently verified). Comments below, most severe first. The most important one is the `auth.sessions` token read — worth resolving before merge. CSRF and an uncapped-`Other`-path candidate were both investigated and refuted (JSON content-type + wildcard-CORS-without-credentials + SameSite=Lax block the former; DataFusion rejects `RETURNING` at plan time for the latter).

crates/server/src/query_handlers.rs
```diff
@@ -0,0 +1,223 @@
...
+#[derive(Debug, Deserialize)]
+pub struct QueryRequest {
+    /// A single SQL statement to execute
+    pub sql: String,
+    /// Result row cap; defaults to [`DEFAULT_MAX_ROWS`]. Must be >= 1.
+    pub max_rows: Option<usize>,
+}
+
+/// Execute ad-hoc SQL endpoint - POST /query
+pub async fn execute_query(
```
abbccdda, line 40: I feel we should incorporate a tracer here as a bearer minimum to make sure we remember what SQL queries the server has executed. Could start with local dump only or opentelemetry instrument if that's straightforward.

crates/server/src/query_handlers.rs
BtXin: Fixed in b6ed69b. The validator config now carries a `denied_schemas` set (`auth` registered by the server), and `validate_single_sql` walks every relation in the statement via sqlparser's `visit_relations` — FROM, JOINs, subqueries, CTEs, DML targets, EXPLAIN bodies, DESCRIBE targets, including 3-part `datafusion.auth.sessions` references — and rejects any whose qualifier names a denied schema. Scoped to the ad-hoc path only; pipeline SQL can still read auth tables. Covered by unit tests (8 bypass shapes) and HTTP tests.

crates/skardi/src/sources/sql_validator.rs
BtXin: Fixed in b6ed69b. `validate_single_sql` is now a strict allowlist: `Query`, access-checked `Insert`/`Update`/`Delete`, `Explain` of an allowed statement, and SHOW/DESCRIBE metadata reads. Everything else — including MERGE, GRANT, transactions, DEALLOCATE, SET TIME ZONE, and any statement type a future sqlparser bump introduces — falls into the reject arm. The listed bypasses are covered by tests.

crates/skardi/src/sources/sql_validator.rs
BtXin: Qualified-name matching fixed in b6ed69b: `extract_table_name` keeps the full name, and `check_write_access` looks up the full name first, then the first qualifier — so `schema_a.orders` no longer matches an unrelated flat `orders` source, and a write into a read-only source's schema (`mysrc.child`) is now rejected. The unknown-table default-allow is retained deliberately: the pipeline path depends on it (targets not listed in ctx.yaml), and on the ad-hoc path the auth schema is now denied outright while unregistered tables fail at execution. Flipping to default-deny would break existing pipeline deployments.

crates/skardi/src/sources/sql_validator.rs
BtXin: Fixed in b6ed69b — the ad-hoc path now parses the raw SQL with no brace preprocessing; `{param}` substitution stays on the pipeline-template path where it belongs. Regression tests cover the quote-crossing example at both the validator and HTTP layers (the literal survives execution intact).

crates/skardi/src/sources/sql_validator.rs
BtXin: Fixed by scoping in b6ed69b. `validate_sql` (the pipeline-load path) is reverted to its pre-PR policy: COPY, SET, and PREPARE are allowed again; DDL and DML access checks are unchanged. The one intentional addition kept on the shared path is EXPLAIN validating its inner statement, since EXPLAIN ANALYZE executes the plan and the DML access-mode contract already covers that. All new denials live only in the ad-hoc `validate_single_sql` path, so existing deployments keep booting.

crates/server/src/server.rs
BtXin: The jobs routes predate this PR (shipped in #98), so the gap wasn't introduced here — tracked the `route_layer` refactor as #160 rather than expanding this PR's scope. Agreed that auth-by-default is the right end state.

crates/server/src/query_handlers.rs
BtXin: Fixed in b6ed69b — the `SqlValidatorConfig` is built once in `setup_app_state` and stored in `AppState`. The justifying comment was wrong: the only `config.write()` in the crate is in test code.

crates/server/src/auth/routes.rs
BtXin: Deferred: folding this into #160, where auth moves to a `route_layer` and `verify_session`'s signature changes anyway — returning a structured (status, message) error there removes the serialize/re-parse round-trip in one place instead of patching it twice.

crates/skardi/src/engine/datafusion.rs
BtXin: Tracked as #161. Agreed on both halves (concat-then-slice, and the arrow→bytes→Value→re-serialize triple pass); it touches the shared response envelope used by pipeline responses, so it's split out rather than folded into this PR.

crates/server/src/query_handlers.rs
BtXin: Added in b6ed69b — every statement the endpoint hands to the engine is logged before execution (`tracing::info!` with the SQL text, max_rows, and statement kind), alongside the existing rejection and completion logs. Rejected statements were already logged. Since this rides the existing `tracing` pipeline, an OTel exporter picks it up with no extra wiring; a persistent audit table can build on this later if needed.

docs/superpowers/specs/2026-07-17-query-endpoint-design.md
BtXin: No cursor in v1: `truncated: true` means the client should refine the query — ad-hoc SQL already provides the paging primitives (ORDER BY + LIMIT/OFFSET, or keyset `WHERE id > last_seen`). Server-side pagination would require holding result state across stateless requests (cursor lifecycle, expiry, memory), which is a separate design if a client actually needs it.

abbccdda:
One clarification request on the design doc (`docs/superpowers/specs/2026-07-17-query-endpoint-design.md`): the **Auth** section says `/query` uses "the same session gate as pipeline execution (`verify_session`)", but the doc never explains where a session comes from — the endpoint only *validates* a token; creation happens in the pre-existing better-auth routes (`POST /api/auth/sign-in/*`), which issue the token/cookie that `verify_session` then checks against `auth.sessions` (expiry + active flag). Could you add a line making that explicit, so the doc stands alone?

Related: the doc also predates the review follow-ups in b6ed69b (the `auth`-schema read denial and the statement allowlist), so its statement-policy section no longer matches the implementation. Worth a small update while you're in there.

abbccdda:
## Second review round — commits `b6ed69b` (scoped policy / auth-schema denial / allowlist) + `b0d9169` (sqlparser 0.59)

Reviewed only the two commits landed since the last round (`1f8c0bb..HEAD`), focused on whether they actually close the prior findings. The auth-schema denial and the fail-closed statement allowlist are solid — `visit_relations` covers FROM/JOIN/subquery/CTE/INSERT-target/DESCRIBE/SHOW, quoted/`AUTH`/`datafusion.auth.sessions` variants are all caught, and `other => Err(StatementNotAllowed)` correctly fails closed for future sqlparser variants. But the access-mode rework introduced a real regression. Findings ranked most-severe first.

### 1. 🔴 Read-only enforcement bypassed by default-schema-qualifying a flat source — `crates/skardi/src/sources/sql_validator.rs:368`
`validator_config_from_sources` keys `table_access_modes` by the bare source name (`config.rs:890`), and flat sources register into DataFusion's default `datafusion.public` catalog/schema. `check_write_access` looks up the full joined name, then falls back to only the **first** dotted segment. So for a `read_only` source `users`:
- `INSERT INTO public.users …` → lookup `public.users` misses, fallback `get("public")` misses → **write allowed**
- `INSERT INTO datafusion.public.users …` → fallback `get("datafusion")` misses → **write allowed**

DataFusion resolves all three forms to the same read-only table, so the control is defeated (same for UPDATE/DELETE). The old `.last()`-segment logic caught this; the test `test_qualified_write_does_not_match_unrelated_flat_source` (line 759) actually locks in the behavior that enables the bypass. Fix: strip the default `datafusion`/`public` qualifiers before lookup, or check the last segment as well as the first.

### 2. 🟠 INSERT with a non-`TableName` target fails open — `sql_validator.rs:294`
The `Statement::Insert` arm matches `TableObject::TableName` and routes everything else to `_ => Ok(())`, while INSERT is on the strict `/query` allowlist. Any INSERT form that parses to a `TableObject::TableFunction` (e.g. `INSERT INTO FUNCTION …`) skips `check_write_access` entirely and reaches the engine unvetted. Low exploitability today, but it's a fail-open on the path whose whole point is to fail closed.

### 3. 🟡 `/query` validates against a startup snapshot of access modes — `server.rs:61`/`:224`
`validator_config` is now built once into `AppState`; the old per-request rebuild under `config.read()` is gone. `ServerConfig` sits behind an `RwLock` "for runtime updates", so if a source's `access_mode` is ever flipped to `read_only` at runtime, `/query` keeps serving the stale (writable) policy until restart. Latent today (no production writer mutates `data_sources`), but the snapshot silently couples correctness to that staying true.

### 4. 🟡 Auth denial is honored by only one of the two validator entry points — `sql_validator.rs` (`denied_schemas` field)
`denied_schemas` lives on the shared `SqlValidatorConfig` but is enforced only by `validate_single_sql`; `validate_sql` (pipeline path) ignores it. The trust boundary is encoded in *which function you call*, not in the type. A future endpoint author validating untrusted SQL who reaches for the obviously-named `validate_sql` gets neither the allowlist nor the auth denial — reopening both original vulns. Consider distinct trusted/untrusted policy types, or pass denied schemas explicitly at the `/query` boundary.

### 5. 🟡 Auth-schema denial is a syntactic denylist over a shared SessionContext — `config.rs:888`
Auth registers into the same `SessionContext` `/query` executes against, and the denial enumerates syntactic paths to the `auth` name. Any indirect handle that doesn't surface as an `auth`-named relation (a later view/semantic-model/federated alias over `auth.sessions`, or a table-reference form `visit_relations` doesn't descend into) leaks bearer tokens with zero validator change. A `/query`-scoped context (or a catalog that never exposes `auth`) would make it structurally unreachable rather than name-filtered. Worth at least a coverage test asserting `visit_relations` reaches table-function arguments.

### 6. 🟡 `sqlparser` is a parallel pin, not DataFusion's re-export — `Cargo.toml`
Unifying to a single workspace 0.59 is good, but it's manually kept in lockstep with DataFusion's internal `sqlparser`. The next DataFusion bump can desync them and reopen the exact validator/engine parse-divergence class this commit closed. Importing `datafusion::sql::sqlparser` would make them share one parser by construction.

### 7. Cleanup — `validator_config` derived state duplicated at ~9 sites
`Arc::new(validator_config_from_sources(&config.data_sources))` is copy-pasted across `server.rs:224`, `pipeline_handlers.rs:861`/`:954`, two `#[cfg(test)]` helpers in `auth/routes.rs`, and four integration-test builders. It's a pure function of `config.data_sources` that `AppState` already holds — an `AppState` constructor that derives it internally removes the boilerplate and the drift risk.

### 8. Cleanup — reserved `"auth"` schema name duplicated
`config.rs:888` denies the literal `"auth"` independently of where `auth/bridge.rs` registers it. Export a `pub const AUTH_SCHEMA` and use it at both sites so the security policy can't silently stop covering the tables it's meant to protect.

### 9. Cleanup — `check_denied_schemas` re-derives the table name inconsistently — `sql_validator.rs:~196`
It builds the error's `table` via `relation.to_string().to_lowercase()` (keeps quote chars) while `WriteNotAllowed` uses `extract_table_name` (strips quoting). Quoted references get reported as `"auth"."sessions"` in one error and `auth.sessions` in the other. Call `extract_table_name` here too.

### 10. Efficiency — `statement_keyword` renders the whole AST for one word — `sql_validator.rs:~169`
On every allowlist rejection `statement.to_string()` Display-renders the entire statement just to split off the first token. Map the enum variant to a `&'static str` keyword instead. (Same micro-cost note applies to the per-part `to_lowercase()` allocations in `check_denied_schemas`/`extract_table_name` on the success path — `eq_ignore_ascii_case` avoids them.)

---
Findings 1–3 are the ones I'd block on. #1 is a genuine security regression with a test pinning the wrong behavior.

crates/skardi/src/sources/sql_validator.rs
```diff
@@ -189,13 +362,20 @@ fn check_write_access(
...
-        if *mode == AccessMode::ReadOnly {
-            return Err(SqlValidationError::WriteNotAllowed {
-                operation: operation.to_string(),
-                table: table_name.to_string(),
-            });
-        }
+    // Look up the full qualified name first; for `source.table` references
+    // also honor the source's access mode, since hierarchical sources
+    // register their tables under the source name as schema.
+    let mode = config.table_access_modes.get(table_name).or_else(|| {
```
abbccdda: 🔴 **Read-only enforcement bypassed by default-schema-qualifying a flat source.**

`table_access_modes` is keyed by the bare source name (`config.rs:890`), and flat sources register into DataFusion’s default `datafusion.public` catalog/schema. This lookup checks the full joined name, then falls back to only the **first** dotted segment. For a `read_only` source `users`:

- `INSERT INTO public.users …` → `get("public.users")` miss, fallback `get("public")` miss → **write allowed**
- `INSERT INTO datafusion.public.users …` → fallback `get("datafusion")` miss → **write allowed**

DataFusion resolves all three forms (`users`, `public.users`, `datafusion.public.users`) to the same read-only table, so the write goes through — same for UPDATE/DELETE. The old `.last()`-segment logic caught this, and `test_qualified_write_does_not_match_unrelated_flat_source` (line 759) currently pins the behavior that enables the bypass.

Fix: strip the default `datafusion`/`public` qualifiers before lookup, or also check the **last** segment (the bare table name) in addition to the first.

crates/skardi/src/sources/sql_validator.rs
```diff
@@ -139,10 +286,13 @@ fn validate_statement(
...
-        Statement::Insert(insert) => {
-            let table_name = extract_table_name(&insert.table_name);
-            check_write_access("INSERT", &table_name, config)
-        }
+        Statement::Insert(insert) => match &insert.table {
+            TableObject::TableName(name) => {
+                check_write_access("INSERT", &extract_table_name(name), config)
+            }
+            // Table-function targets have no registered name to check.
+            _ => Ok(()),
```
abbccdda: 🟠 **INSERT with a non-`TableName` target fails open.**

This arm matches `TableObject::TableName` and routes everything else to `_ => Ok(())`, while `INSERT` is on the strict `/query` allowlist. Any INSERT form that parses to a `TableObject::TableFunction` (e.g. `INSERT INTO FUNCTION …`) skips `check_write_access` entirely and reaches the engine unvetted.

Low reachability today (DataFusion won't plan an `INSERT INTO FUNCTION` at a registered read-only table), but this is a fail-open on the one path whose whole purpose is to fail closed — the rest of `validate_statement_strict` rejects unknown shapes via `other => Err(StatementNotAllowed)`. Prefer rejecting the non-`TableName` target here (e.g. `WriteNotAllowed`/`StatementNotAllowed`) rather than allowing it.

crates/skardi/src/sources/sql_validator.rs
BtXin: Fixed in f4b9475 — confirmed the bypass end-to-end first: an HTTP test showed `INSERT INTO public.products` executing against the read-only source before the fix. `check_write_access` now strips the default `datafusion`/`public` qualifiers before lookup (rather than also checking the last segment, which would re-introduce the `schema_a.orders` false rejection), so `users`, `public.users`, and `datafusion.public.users` all resolve to the same access-mode entry, and `datafusion.<source>.<table>` still honors the source's mode. Covered at both the validator and HTTP layers; `test_qualified_write_does_not_match_unrelated_flat_source` stays valid since `schema_a` is not a default qualifier.

crates/skardi/src/sources/sql_validator.rs
BtXin: Fixed in f4b9475 — the non-`TableName` arm now returns `StatementNotAllowed` instead of `Ok(())`. Verified reachability while writing the test: `supports_insert_table_function` is ClickHouse-only in sqlparser 0.59, so GenericDialect can't produce the shape today — the test constructs it via `ClickHouseDialect` and asserts the rejection, so the path stays closed if a future parser bump or dialect change makes it reachable.

BtXin:
Second-round general review — resolutions for findings #3–#10 (findings #1–#2 were fixed in `f4b9475` with inline replies). Landed in `66818c9`; full workspace suite green, clippy clean on touched files.

**#3 — startup snapshot of access modes.** Kept the snapshot (it was round-1's explicit ask) and documented the invariant it rests on: no runtime writer mutates a source's `access_mode` — verified the only `config.write()` is test-only. The doc now lives on both `AppState.adhoc_policy` and `config::adhoc_policy_from_sources`, stating that any future access-mode writer must rebuild the policy or the snapshot goes stale.

**#4 — trust boundary in the type, not the function name.** Split the policy types. `SqlValidatorConfig` now holds only access modes (no trust-boundary state); the untrusted `/query` path validates against a new `AdhocSqlPolicy` (access modes + denied schemas) that `validate_single_sql` *requires*. A future caller reaching for the trusted `validate_sql` with a bare config can no longer get the untrusted input past the allowlist + schema denial by accident.

**#5 — syntactic denylist over a shared context.** Added coverage tests proving `visit_relations` descends into indirect relations (UNION, scalar/IN/EXISTS subqueries), so the guard is complete for any single ad-hoc statement. Documented the true residual — an operator-defined view or federated alias over `auth.sessions`, which ad-hoc SQL cannot create (DDL is rejected) and which only trusted operator config could introduce. The structural fix (a `/query`-scoped SessionContext) is filed as #164: it's deferred because `auth.*` is a real feature (`docs/auth/pipelines/active-users.yaml` reads it via the shared engine context), so isolating `/query` needs a second data-source registration pass that risks doubling DB connection pools — disproportionate to a trusted-only residual. Happy to pull #164 forward if you'd rather it block here.

**#6 — parallel sqlparser pin.** Removed the standalone `sqlparser` dependency entirely. The validator now parses via `datafusion::sql::sqlparser` (DataFusion enables the `visitor` feature, so `visit_relations` is available). Validator and engine now share one parser by construction; a DataFusion bump that moves sqlparser becomes a compile error here, not a silent parse divergence.

**#7 — duplicated derived state.** Added `AppState::new(config, engine, session_ctx, auth_layer, jobs)`, which derives the policy (and `metrics`) once. Removed the copy-pasted `Arc::new(validator_config_from_sources(..))` from ~9 call sites.

**#8 — duplicated `"auth"` literal.** Exported `pub const AUTH_SCHEMA` from `auth::bridge`; both the registration site and the deny site (`adhoc_policy_from_sources`) reference it.

**#9 — inconsistent table name in error.** `check_denied_schemas` now builds the reported table via `extract_table_name` (quote-stripped), matching `WriteNotAllowed`.

**#10 — `statement_keyword` rendered the whole AST.** Replaced with a direct variant → `&'static str` match; no more `Display` of the full statement on each rejection.

abbccdda (commented):
## Follow-up: hardening the `/query` security story

Did another security-focused pass on the branch after the last review round. **No exploitable HIGH/MEDIUM issue** — the validator is tight (type-split trusted/untrusted policies, strict statement allowlist that fails closed on unknown/future statement types, auth-schema denial via `visit_relations`, default-qualifier stripping, quoted-identifier normalization, shared parser with the engine). The remaining items are defense-in-depth, left as inline comments below by value.

Happy to take the DELETE and error-shape items as a fast-follow.

crates/server/src/auth/bridge.rs
```diff
@@ -178,8 +178,15 @@ impl TableProvider for AuthSessionsTable {
...
+///
+/// Single source of truth for the reserved schema name: the ad-hoc `/query`
+/// policy denies exactly this schema (see `config::adhoc_policy_from_sources`),
+/// so the two must never drift apart.
+pub const AUTH_SCHEMA: &str = "auth";
+
 /// Register `auth.users` and `auth.sessions` virtual tables into the given
-/// DataFusion `SessionContext` under a dedicated `auth` schema.
+/// DataFusion `SessionContext` under the [`AUTH_SCHEMA`] schema.
 pub fn register_auth_tables(
```
abbccdda, line 190: **Defense-in-depth (design-level): token exposure rests on a single string-based parser guard.**

`register_auth_tables` registers `auth.sessions` — live bearer tokens — into the **same** `SessionContext` the `/query` engine executes against. The only thing between ad-hoc SQL and those tokens is `check_denied_schemas`' `visit_relations` string match. That's one layer, and the code already documents a residual (operator-defined views / federated aliases); a future untagged `sqlparser` relation field would be another.

**Hardening:** give `/query` a catalog/`SessionContext` that never has `auth.*` registered, so a parser-level bypass has nothing to reach. Trade-off: pipelines legitimately JOIN `auth.*` and share the context today, so this means splitting the ad-hoc catalog from the pipeline catalog. Non-trivial, but it converts token safety from "the parser must be perfect" to "the tokens aren't in the catalog." Suggest a tracking issue rather than blocking this PR.

crates/skardi/src/sources/sql_validator.rs
```diff
@@ -152,22 +361,45 @@ fn validate_statement(
             check_write_access("DELETE", &table_name, config)
```
abbccdda, line 373: **Defense-in-depth: fail closed on multi-table DELETE.**

This access-checks only `delete.from`, ignoring `delete.tables` (`DELETE t1, t2 FROM …`) and `delete.using`. Not exploitable today — DataFusion rejects both forms at planning — but validation is *relying* on the engine to reject them. Reject non-empty `tables`/`using` in the strict path (or access-check every relation) so we don't depend on downstream engine behavior. Cheap.

crates/server/src/query_handlers.rs
```diff
@@ -0,0 +1,223 @@
...
+                QUERY_METRICS_LABEL,
+                elapsed_ms,
+                "query_execution_error",
+            );
+
+            let error_details = serde_json::json!({
+                "engine_error": e.to_string(),
+                "registered_tables": "Check server logs for data source registration status",
+                "suggestion": "Verify that data sources are properly registered and accessible"
+            });
```
abbccdda: **Defense-in-depth: don't return raw engine internals to the client.**

`engine_error: e.to_string()` is returned in the HTTP body. DataFusion errors can echo row/column values and internal schema. Caller is authenticated so risk is low, but prefer a generic client message with detail kept in the server-side log.

(The accepted-path audit log at `tracing::info!(sql = %request.sql, …)` above is deliberate — leave as-is, or hash if audit-log PII becomes a concern.)

crates/server/src/query_handlers.rs
```diff
@@ -0,0 +1,223 @@
...
+                QUERY_METRICS_LABEL,
+                elapsed_ms,
+                "result_conversion_error",
+            );
+
+            let error_details = serde_json::json!({
+                "conversion_error": e.to_string(),
+                "record_batch_schema": format!("{:?}", record_batch.schema()),
+                "record_batch_rows": record_batch.num_rows()
+            });
```
abbccdda: Same issue as the `engine_error` block: `record_batch_schema` echoes internal schema back to the client. Keep the detail in the server log and return a generic conversion-failed message.

crates/server/src/auth/bridge.rs
BtXin: Agreed this is the right end state, and it's tracked as #164 (filed after the last round with exactly this framing — split the ad-hoc catalog from the pipeline catalog so a parser-level bypass has nothing to reach). Not blocking here per your suggestion. In the meantime this round adds the indirect-relation coverage tests and the DELETE fail-closed below, so the single string-guard layer is at least well-pinned. The one caveat noted in #164: pipelines legitimately JOIN auth.* through the shared context (docs/auth/pipelines/active-users.yaml), so the split needs a second data-source registration pass without doubling DB connection pools.

crates/skardi/src/sources/sql_validator.rs
BtXin: Fixed in 133e1f4. The ad-hoc strict path now rejects a DELETE with a non-empty `tables` list or a `USING` clause outright (StatementNotAllowed), rather than access-checking only `delete.from` and leaning on the engine to reject the multi-relation forms. Verified parsing under GenericDialect first: `DELETE t1, t2 FROM …` doesn't even parse there, but `DELETE FROM x USING y` does — so `USING` was the reachable case, now failing closed at the validator with a test.

crates/server/src/query_handlers.rs
BtXin: Fixed in 133e1f4. The execution-error response is now a generic message with no `details`; the full DataFusion error stays in the server log (`tracing::error!`). Added an HTTP test asserting the offending identifier isn't echoed back and `details` is absent. Left the accepted-path audit log as-is per your note.

crates/server/src/query_handlers.rs
BtXin: Fixed in the same commit — the conversion-error path no longer returns `record_batch_schema`; the schema and row count are logged server-side only, and the client gets a generic message.

abbccdda (approved):
Nice work！Can't wait to see this go live

PR #159, opened 2026-07-17
feat(sources): Open Connector config foundation and gateway client 

bakey:
## Summary

> **Task tracker**: [2026-07-11-open-connector-integration-tasks.md](docs/superpowers/specs/2026-07-11-open-connector-integration-tasks.md) — the integration lands as a series of independently reviewable PRs; this file shows where this PR fits in the whole.

First two milestones of the [Open Connector integration design](docs/superpowers/specs/2026-07-11-open-connector-integration-design.md) (#151): the **typed-config foundation** and the **HTTP client + action registry** that every later piece (source packs, scan engine, external-table factory, UDTFs) builds on.

The design requires a typed `open_connector` configuration block on `DataSource` because nested gateway bindings, resources, and overrides do not fit the flat `options: HashMap<String, String>`. This PR lands that typed configuration, its validation, the gateway HTTP layer, and the registration entry point — with **no tables registered yet** (catalog registration arrives with the source-pack milestone).

## Milestone 1 — typed-config foundation

- **`DataSourceType::OpenConnector`** — new variant with an explicit serde rename so the public YAML spelling is `open_connector` (the `lowercase` rule would produce `openconnector`).
- **`open_connector/config.rs`** — `OpenConnectorConfig` / `OpenConnectorBinding` matching the design spec's YAML shape: `runtime_token_env`, request/scan timeouts, `max_pages`/`max_rows`, `cache_max_bytes`, `cache_ttl_seconds` (0 = live reads), `raw_action_allowlist`, and named `bindings` (`name`, `source_pack`, optional `source_pack_version` pin, `connection_alias`, `resource`, `tables`). The version pin keeps a Skardi upgrade from silently changing a bound table's schema after restart. Both structs use `deny_unknown_fields`, so a misspelled key (e.g. `source_pack_versions`) fails loudly instead of silently disabling the pin.
- **`open_connector/error.rs`** — `OpenConnectorError` with pre-network validation variants. `validate()` is pure, and runs in **both** the server's `validate_data_sources` and the provider's registration path — so CLI and server reject misconfiguration identically.
- **Server wiring** — `DataSource.open_connector` (required for `type: open_connector`, rejected for every other type), validation, registration dispatch, `/data_source` and dashboard type mapping. Read-only by construction: not in `WRITABLE_SOURCE_TYPES`; job destinations rejected as non-transactional.
- **CLI wiring** — same typed field and registration arm.

## Milestone 2 — HTTP client + action registry

- **`open_connector/client.rs`** — `OpenConnectorClient` against the gateway's `/v1` contract:
  - health checks, action discovery, action execution; endpoint paths centralized in private constants as the single reconciliation point with the upstream API.
  - runtime token read from the configured env var, sent as a Bearer header; gateway URLs must be `http(s)` without embedded credentials so secrets never reach logs; the `Debug` impl excludes the token.
  - bounded retries on 429 / transient 5xx / transport errors with exponential backoff + jitter, honoring `Retry-After` (capped).
  - bounded response decoding enforced on both declared `Content-Length` and the actual streamed bytes; per-request timeout from config.
  - connection-alias header on execute calls.
- **`open_connector/action_registry.rs`** — `ActionRegistry`:
  - discovers `raw_action_allowlist` actions at registration (deduplicated, concurrency-bounded); actions that are not locally executable are rejected, a **missing** executability flag is rejected as `ActionExecutabilityUnknown` (default-deny — `Option<bool>` end to end so "not declared" is never read as "executable"), and a partially loaded registry is never returned.
  - per-action **compatibility fingerprint**: canonicalized output schema (recursive key sort) hashed with FNV-1a — stable across processes without a crypto-hash dependency, used by later milestones to fail registration on incompatible upstream action changes.
- **Registration flow** — `register_open_connector_tables` now validates config, health-checks the gateway, loads the allowlist registry, then fails with `ExecutionNotImplemented` until source packs land.
- **`reqwest` becomes a hard dependency** of the `skardi` crate (the client is unconditional); `remote-embed` / `llm-extract` features now only gate their UDF code.
- **Test infrastructure** — a hand-rolled mock gateway (`testutil.rs`) over `tokio::net::TcpListener`, so no mock-HTTP crate is added. It records requests (method, path, headers, body) for header/auth assertions and scripts retry sequences (429 → `Retry-After` → 200, repeated 5xx, …).

## Test coverage

49 new `skardi` tests, including:

- config: design-spec YAML round-trip, defaults, every validation variant, unknown-field rejection (misspelled `source_pack_version`, plural-typo `raw_action_allowlist`, unknown binding keys, misspelled cache field);
- client: bearer-token header, retry on 5xx then success, `Retry-After` honored, terminal 4xx without retry, retry exhaustion on unreachable gateway, URL validation, missing token env, discovery parsing, malformed JSON, response-size bound, execute envelope + alias header + request body, missing executability flag preserved as `None`;
- registry: dedup, empty allowlist (zero network calls), non-executable action rejection, missing executability flag rejected as `ActionExecutabilityUnknown` (default-deny), error propagation, fingerprint stability across key order / sensitivity to schema changes;
- registration: every failure stage asserted to fire **before** any network call, health-before-discovery ordering, and the full healthy-gateway path ending at `ExecutionNotImplemented` with health + discovery calls verified.

Plus 7 `skardi-server` and 4 `skardi-cli` tests from milestone 1. `cargo check --workspace --all-targets`, `cargo clippy` (no new warnings), `cargo fmt --check`, and full `cargo test` for all three crates pass.

## Out of scope (later milestones)

Source-pack registry, JSON-to-Arrow conversion, pagination, filter/limit pushdown, cache, `open_connector_query` / `open_connector_scan` UDTFs. SQL DDL registration (`CREATE EXTERNAL TABLE ... STORED AS OPEN_CONNECTOR`) stays out per the approved design — stable tables register through context YAML only; DDL is a documented future extension. README's supported-sources table is intentionally **not** updated yet — the source is not queryable until the scan engine lands.


abbccdda (commented):
On the high level, could we use openspec skill to write a full tasks.md before starting the implementation, if one PR couldn't finish everything? It helps us/agent keep track of previous changes and review the full plan first along with the actual code changes

bakey:
> On the high level, could we use openspec skill to write a full tasks.md before starting the implementation, if one PR couldn't finish everything? It helps us/agent keep track of previous changes and review the full plan first along with the actual code changes

Sure. Let me do that

docs/superpowers/specs/2026-07-11-open-connector-integration-tasks.md
```diff
@@ -0,0 +1,115 @@
...
+- [ ] 4.5 Docs: `docs/open-connector.md`, ctx/UDTF examples, README supported-sources entry (first time the source is actually queryable)
+
+**Verification**: both UDTFs return the stable schema and values of their
+corresponding YAML-registered tables; federated join of the mock pack
+against a local CSV.
+
+## Milestone 5+ — Real source packs (one PR each, per design rollout)
+
+- [ ] 5.1 GitHub pack (API-key auth, page-number pagination): repositories, issues, issue comments, pull requests, reviews, commits, workflow runs, releases
+- [ ] 5.2 Jira pack (OAuth, cursor pagination, JQL-backed filters): projects, issues, comments
```
abbccdda: Change this to Slack

crates/cli/src/main.rs
```diff
@@ -915,6 +930,30 @@ async fn register_source(
                 .await
                 .with_context(|| format!("Failed to register InfluxDB '{}'", source.name))?;
         }
+        "open_connector" => {
+            let endpoint = source.connection_string.as_deref().ok_or_else(|| {
```
abbccdda, line 943: Since connection_string is always the same for most times, could we just add a default connection_string that overrides all open_connector sources at the server config level?

crates/skardi/src/sources/providers/open_connector/client.rs
```diff
@@ -0,0 +1,793 @@
...
+    /// Join a relative path onto the gateway base URL.
+    fn endpoint(&self, path: &str) -> String {
+        self.base_url
+            .join(path)
+            .unwrap_or_else(|_| self.base_url.clone())
+            .to_string()
+    }
+
+    /// Build the URL for one action, optionally with a path suffix (`execute`).
+    fn action_url(&self, action_id: &str, suffix: Option<&str>) -> String {
```
abbccdda, line 462: **[1/10] Dot-segment action IDs escape the `/v1/actions/` namespace.**

`ACTION_ID_SET` preserves `.`, so an action ID like `..` survives percent-encoding, and `Url::join` in `endpoint()` resolves the dot segment. Verified against the `url` crate: `join("v1/actions/..")` on a slash-terminated base resolves to `{base}/v1/`.

Config validation of `raw_action_allowlist` only rejects empty-after-trim entries, so `".."` is accepted; `discover_action("..")` then issues `GET {base}/v1/` instead of `/v1/actions/..`, and any 2xx JSON from the misrouted endpoint is admitted into the `ActionRegistry` under that action ID.

Suggested fix: remove `.` from `ACTION_ID_SET` (encode it), or reject dot-only path segments in allowlist/action-ID validation.

crates/skardi/src/sources/providers/open_connector/client.rs
```diff
@@ -0,0 +1,793 @@
...
+            input_schema: raw.input_schema,
+            output_schema: raw.output_schema,
+            locally_executable: raw.locally_executable,
+            connection_aliases: raw.connection_aliases,
+        })
+    }
+
+    /// `POST /v1/actions/{action_id}/execute` — execute one action and return
+    /// the response envelope's `output` value.
+    pub async fn execute(
```
abbccdda: **[2/10] `execute()` structurally bypasses the default-deny gates.**

The `raw_action_allowlist` and `locally_executable` default-deny checks live only in `ActionRegistry::load` (discovery path), but `execute()` is `pub` on a type re-exported crate-publicly (`open_connector/mod.rs` → `providers/mod.rs` → `lib.rs`) and accepts any `action_id` with no tie to the registry.

When the scan engine (or any external consumer) calls `client.execute("github.delete_repo", ...)`, nothing enforces the gating that abb543a added. Currently latent — the only production path bails with `ExecutionNotImplemented` first — but the bypass is structural.

Suggested fix: make `execute` `pub(crate)`, or mediate execution through the registry so only discovered/allowlisted actions are executable.

crates/server/src/config.rs
```diff
@@ -758,6 +782,34 @@ fn validate_data_sources(data_sources: &[DataSource]) -> Result<()> {
             .into());
         }
 
+        // Open Connector typed config: required for that type, rejected for
+        // every other type. `config.validate()` is pure (no network I/O), so
+        // misconfigurations surface here at config load, not at first query.
+        match (&source.source_type, &source.open_connector) {
```
abbccdda, line 796: **[3/10] `hierarchy_level` is not validated here, but registration hard-requires `Catalog`.**

The comment above promises misconfigurations "surface here at config load, not at first query" — but `HierarchyLevel` defaults to `Table`, and this block never checks it. `register_open_connector_tables` then fails with `CatalogHierarchyRequired`, wrapped in a Debug-formatted `DataSourceRegistrationFailed`, aborting the whole server startup.

So the *natural minimal config* (omitting `hierarchy_level: catalog`) passes validation and fails at boot. The CLI path (`crates/cli/src/main.rs:933-952`) has the same gap.

Suggested fix: add a `hierarchy_level == Catalog` check to this match arm (and its CLI counterpart), with a targeted `ConfigError`.

crates/skardi/src/sources/providers/open_connector/client.rs
```diff
@@ -0,0 +1,793 @@
...
+        terminal_error: impl Fn(StatusCode, String) -> OpenConnectorError,
+    ) -> Result<Response, OpenConnectorError> {
+        let mut last_reason = String::new();
+        for attempt in 1..=self.max_attempts {
+            let request = build().bearer_auth(self.token.expose_secret());
+            match request.send().await {
+                Ok(response) if response.status().is_success() => return Ok(response),
+                Ok(response) => {
+                    let status = response.status();
+                    if RETRYABLE_STATUSES.contains(&status.as_u16()) {
```
abbccdda: **[4/10] `POST /execute` is retried on 5xx/transport errors — duplicate execution of non-idempotent actions.**

`send_with_retry` applies `RETRYABLE_STATUSES` (429/500/502/503/504) and the transport-error branch uniformly to all requests, with no method-based distinction and no idempotency key. If the gateway performs an action but the response is lost (502 from a proxy, connection reset), the client re-POSTs up to `MAX_ATTEMPTS` (3) times, re-executing the action against the SaaS provider.

Harmless for the read-only milestone, but the hazard is baked into the shared retry path. Suggested fix: restrict automatic retry to idempotent requests (GET health/discover), or send an idempotency key with execute and document the gateway contract.

crates/server/src/config.rs
```diff
@@ -758,6 +782,34 @@ fn validate_data_sources(data_sources: &[DataSource]) -> Result<()> {
...
+                    source_type: source.source_type,
+                }
+                .into());
+            }
+            (_, None) => {}
+        }
+
         // Catalog mode must not mix with per-table / per-schema options
         if CATALOG_SUPPORTED_SOURCES.contains(&source.source_type)
             && source.hierarchy_level == HierarchyLevel::Catalog
```
abbccdda, line 832: **[5/10] `OpenConnector` is missing from `CATALOG_SUPPORTED_SOURCES` (defined ~line 735), so the catalog-mode guards below silently skip it.**

The checks rejecting `table`/`schema` options and empty `allowed_schemas` are gated on membership in `CATALOG_SUPPORTED_SOURCES`, which lists only Postgres/MySQL/SQLite/Seekdb/DynamoDB. An `open_connector` source with `hierarchy_level: catalog` and `options: {table: issues}` (or `allowed_schemas: ""`) passes validation silently and the option is ignored — the identical config on postgres/dynamodb fails loudly with `CatalogModeConflictingOptions`/`EmptyAllowedSchemas`.

Since the type is catalog-only, it should be in the list.

crates/skardi/src/sources/providers/open_connector/client.rs
```diff
@@ -0,0 +1,793 @@
...
+            OpenConnectorError::InvalidGatewayResponse {
+                operation: operation.clone(),
+                reason: format!("execution envelope is not valid JSON: {e}"),
+            }
+        })?;
+
+        // The contract wraps results in an `output` field; tolerate gateways
+        // that return the bare value.
+        Ok(match value {
+            Value::Object(ref map) => map.get("output").cloned().unwrap_or(value.clone()),
```
abbccdda: **[6/10] A 200 object without `output` is returned whole — error/async envelopes become action output.**

`map.get("output").cloned().unwrap_or(value.clone())` means a gateway returning HTTP 200 with `{"error": "rate limited"}` or `{"status": "pending", "job_id": ...}` has that envelope treated as the action result (pinned by `execute_without_output_field_returns_bare_body`). No output-schema validation exists downstream in this PR, so once the scan engine lands, contract violations flow into Arrow rows as silently wrong data instead of a targeted error.

Suggested fix: treat a missing `output` field on an object response as `InvalidGatewayResponse`, and drop the bare-body tolerance (or gate it explicitly per gateway capability).

Efficiency note on the same line: `unwrap_or` is eager, so the full envelope (up to 16 MiB) is cloned even when `output` is present — `match` on `Value::Object(mut map)` + `map.remove("output")` avoids both clones.

crates/skardi/src/sources/providers/open_connector/config.rs
```diff
@@ -0,0 +1,518 @@
...
+    pub bindings: Vec<OpenConnectorBinding>,
+}
+
+impl OpenConnectorConfig {
+    /// Validate the configuration, failing on the first problem found.
+    ///
+    /// Pure and allocation-light: no network I/O, so it runs identically in
+    /// the server's config validation and in the provider's registration
+    /// path (which also covers the CLI).
+    pub fn validate(&self) -> Result<(), OpenConnectorError> {
```
abbccdda, line 174: **[7/10] `request_timeout_seconds: 0` / `scan_timeout_seconds: 0` pass validation but break every request.**

`validate()` rejects zero `max_pages`/`max_rows` as `ZeroSafetyBound` but not the timeout fields, and `from_config` feeds `Duration::from_secs(0)` straight into reqwest's client timeout. Every request then times out instantly, is classified as a retryable transport error, burns 3 attempts with backoff, and surfaces as an opaque `RetriesExhausted` at registration — instead of the targeted config-load error the other bounds get.

Suggested fix: include both timeout fields in the `ZeroSafetyBound` check.

crates/skardi/src/sources/providers/open_connector/client.rs
```diff
@@ -0,0 +1,793 @@
...
+    /// exponential backoff with jitter; both are capped at [`MAX_RETRY_WAIT`].
+    async fn send_with_retry(
+        &self,
+        operation: &str,
+        build: impl Fn() -> RequestBuilder,
+        terminal_error: impl Fn(StatusCode, String) -> OpenConnectorError,
+    ) -> Result<Response, OpenConnectorError> {
+        let mut last_reason = String::new();
+        for attempt in 1..=self.max_attempts {
+            let request = build().bearer_auth(self.token.expose_secret());
```
abbccdda, line 486: **[8/10] A header-invalid token (e.g. trailing newline) is retried 3× and surfaces as an opaque transport error.**

`export OPEN_CONNECTOR_TOKEN="$(cat token.txt)"` with a trailing newline: `bearer_auth` stores a reqwest builder error, every `send()` fails into the transport-error retry branch (there is no `reqwest::Error::is_builder()` check), and after 3 backoff sleeps the caller gets `RetriesExhausted { reason: "builder error" }` — no hint the token itself is malformed. Neither `from_config` nor `validate()` trims or checks the token value.

Suggested fix: trim/validate the token for header-legal characters in `from_config` (fail fast with a targeted error), and/or treat `is_builder()` errors as terminal rather than retryable.

crates/cli/src/main.rs
```diff
@@ -915,6 +930,30 @@ async fn register_source(
                 .await
                 .with_context(|| format!("Failed to register InfluxDB '{}'", source.name))?;
         }
+        "open_connector" => {
```
abbccdda, line 933: **[9/10] CLI accepts `access_mode: read_write` for open_connector; the server rejects it.**

The server rejects `read_write` at config validation (`OpenConnector` absent from `WRITABLE_SOURCE_TYPES` → `UnsupportedWriteMode`), but this CLI arm never consults `is_read_write()` (only the sqlite/dynamodb arms do). The same ctx YAML is accepted by one front-end and rejected by the other — the two already disagree on the read-only invariant this milestone is built around.

This arm also re-implements the required-block/wrong-type checks as anyhow strings that the server encodes as typed `ConfigError` variants. Suggested fix: share one validation helper (natural home: the provider module, which already owns `config.validate()`), and add the read-only check here.

crates/skardi/src/sources/providers/open_connector/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+                if env == "SKARDI_TEST_OC_TOKEN_DEFINITELY_UNSET"
+        ));
+    }
+
+    #[tokio::test]
+    async fn register_fails_health_check_before_discovery() {
+        let gateway = MockGateway::start(|_| MockResponse::new(503, "{}")).await;
+
+        unsafe {
+            std::env::set_var(TOKEN_ENV, "test-token");
```
abbccdda: **[10/10] Flaky-test risk: two parallel tests race on the same process-wide env var.**

`register_fails_health_check_before_discovery` and `register_reaches_not_implemented_with_healthy_gateway` both `set_var`/`remove_var` the same `SKARDI_TEST_OC_REGISTER_TOKEN`, and cargo runs `#[tokio::test]`s on parallel threads. One test's `remove_var` landing between the other's `set_var` and the `env::var` read inside `OpenConnectorClient::from_config` (during the await window) yields an intermittent `MissingRuntimeToken` failure. No mutex/`serial_test`/per-test var-name isolation is present.

Suggested fix: use a distinct env-var name per test (like the other tests in this file already do with never-set names), or serialize the two tests.

abbccdda:
## Code review summary (10 inline comments posted)

Automated high-effort review (7 finder angles, per-finding verification). 10 confirmed findings posted inline, ranked most-severe first in each comment's header.

Additional verified findings that didn't make the inline cut:

- **Startup behavior**: declaring any `open_connector` source makes server/CLI startup perform real network I/O (health check + discovery, up to 3 retries each) and then *always* abort with `ExecutionNotImplemented` (`open_connector/mod.rs:126-139`). Presumably intentional milestone scaffolding, but there is no config-time warning on the shipped YAML surface — worth a doc note or validation-time rejection.
- **Reuse**: `retry_after` (`client.rs:492`) is the fourth copy of Retry-After parsing in the crate (`remote_embed/mod.rs:32`, `llm_extract/anthropic.rs:166`, `llm_extract/openai_compat.rs:236`), and `send_with_retry` is the third bespoke reqwest retry loop. The `OpenConnector => "open_connector"` string arm was added in three places (`gui.rs`, `pipeline_handlers.rs`) when `DataSourceType::as_str()` already exists.
- **Dependency**: reqwest was promoted from optional (feature-gated behind `remote-embed`/`llm-extract`) to an unconditional dependency of the skardi crate — every downstream build now compiles hyper/rustls even with no connector configured. Consider a cargo feature gate for the `open_connector` module.
- **Efficiency**: `execute()` deep-clones the input via `json!({"input": input})` and eagerly clones the full envelope via `unwrap_or(value.clone())`; terminal error paths buffer up to 16 MiB of body to keep 512 chars; the 16 MiB / 3-attempt bounds are hardcoded with no wiring from `OpenConnectorConfig` despite similar operator-tunable bounds living there.

crates/cli/src/main.rs
bakey: The design model already eliminates the repetition                                                                                                                                    
                                                                                                                                                                                            
   Per the approved design, one gateway = one data source = one catalog. Multiple resources attach to a single source through bindings:                                                     
                                                                                                                                                                                            
   ```yaml                                                                                                                                                                                  
     - name: saas                                                                                                                                                                           
       type: open_connector                                                                                                                                                                 
       connection_string: http://open-connector:3000   # written once                                                                                                                       
       open_connector:                                                                                                                                                                      
         bindings:                                                                                                                                                                          
           - { name: github_skardi, source_pack: github, ... }                                                                                                                              
           - { name: jira_company, source_pack: jira, ... }                                                                                                                                 
           - { name: notion_roadmap, source_pack: notion, ... }

crates/skardi/src/model/llm_extract/anthropic.rs
```diff
@@ -165,12 +165,7 @@ fn parse_entities(body: &str) -> anyhow::Result<Vec<serde_json::Value>> {
 
 /// Parse `Retry-After` as seconds, falling back to `DEFAULT_RETRY_WAIT`.
 fn parse_retry_after(resp: &reqwest::Response) -> Duration {
-    resp.headers()
-        .get(reqwest::header::RETRY_AFTER)
-        .and_then(|v| v.to_str().ok())
-        .and_then(|v| v.parse::<u64>().ok())
-        .map(Duration::from_secs)
-        .unwrap_or(DEFAULT_RETRY_WAIT)
+    crate::util::http::parse_retry_after(resp).unwrap_or(DEFAULT_RETRY_WAIT)
```
BtXin: Don't use full path for a lib, put import create::util ... on top

crates/skardi/src/model/llm_extract/openai_compat.rs
```diff
@@ -235,12 +235,7 @@ fn parse_entities(body: &str) -> anyhow::Result<Vec<serde_json::Value>> {
 
 /// Parse `Retry-After` as seconds, falling back to `DEFAULT_RETRY_WAIT`.
 fn parse_retry_after(resp: &reqwest::Response) -> Duration {
-    resp.headers()
-        .get(reqwest::header::RETRY_AFTER)
-        .and_then(|v| v.to_str().ok())
-        .and_then(|v| v.parse::<u64>().ok())
-        .map(Duration::from_secs)
-        .unwrap_or(DEFAULT_RETRY_WAIT)
+    crate::util::http::parse_retry_after(resp).unwrap_or(DEFAULT_RETRY_WAIT)
```
BtXin: Ditto

BtXin (approved):
Overall looks good, some nitpick.

PR #162, opened 2026-07-21
feat(sources): Open Connector source packs and scan engine

bakey:
## Summary

> **Task tracker**: [2026-07-11-open-connector-integration-tasks.md](docs/superpowers/specs/2026-07-11-open-connector-integration-tasks.md) — the integration lands as a series of independently reviewable PRs; this file shows where this PR fits in the whole.

First **three** milestones of the [Open Connector integration design](docs/superpowers/specs/2026-07-11-open-connector-integration-design.md) (#151): the **typed-config foundation**, the **HTTP client + action registry**, and the **source packs + scan engine**. A configured gateway now registers as a real, queryable catalog (`<gateway>.<binding>.<table>`) — proven end to end with the synthetic `mock` source pack.

The design requires a typed `open_connector` configuration block on `DataSource` because nested gateway bindings, resources, and overrides do not fit the flat `options: HashMap<String, String>`. This PR lands that typed configuration, its validation, the gateway HTTP layer, and the relational core (stable table definitions → Arrow RecordBatches).

## Milestone 3 — source packs + scan engine (NEW)

- **`source_pack.rs` + `packs/mock.rs`** — built-in stable table definitions (action ID, row path, fixed Arrow schema/field mappings, typed pagination strategy, allowlisted filters, required resources, expected fingerprint). Packs are versioned; bindings may pin `source_pack_version` and mismatches fail at bind time. The synthetic `mock` pack proves the abstraction without a real SaaS.
- **`row_path.rs`** — `$.a.b` object-key extraction; missing keys / non-array targets fail with page + segment context, never echoing values.
- **`json_to_arrow.rs`** — conversion against the fixed schema only: missing required keys and type mismatches fail with (column, page, row, expected, found *kind* — never the value); nullable fields become null; RFC 3339/epoch timestamps; `List<Utf8>`; opaque fields serialize as JSON; extra upstream fields ignored.
- **`pagination.rs`** — typed strategies (page-number with short/empty-page termination; cursor with repeated-cursor `PaginationLoop` detection). Offset/next-link/has-more slot into the same enum as packs need them.
- **`filters.rs`** — allowlisted `FilterMapping` translation: Exact pushes into action inputs (verified in gateway request bodies), everything else stays in DataFusion; literal-side normalization; `Inexact` reserved for conservative mappings.
- **`cache.rs`** — bounded TTL whole-scan cache: canonical keys (gateway, alias, action, pack version, resource, translated filters, projection, **LIMIT**, Arrow-schema fingerprint), LRU eviction by bytes + entries; `cache_ttl_seconds: 0` = live reads. Documented boundary: completed scans only — overlapping scans (self-join sides) are not deduplicated.
- **`exec.rs` + `table.rs`** — `OpenConnectorExec` streams pages sequentially (drop = cancel), converts per page, truncates and stops at LIMIT, and fails with `ScanBoundsExceeded`/`ScanTimeout` instead of returning incomplete results; `OpenConnectorTableProvider` is read-only (`TableType::Base`, no `insert_into`) and classifies pushdown with the same allowlist the scan uses.
- **Registration now builds the real catalog** — `ExecutionNotImplemented` is gone: bindings resolve to pack tables, discovery covers allowlist + pack actions, the fingerprint gate runs, and `<gateway>.<binding>.<table>` becomes queryable.
- **Shared util dedup** — `util::json::canonical_json` (fingerprints + cache keys), `util::hash::fnv1a_hex` (schema/contract fingerprints).

## Milestone 1 — typed-config foundation

- **`DataSourceType::OpenConnector`** — new variant with an explicit serde rename so the public YAML spelling is `open_connector` (the `lowercase` rule would produce `openconnector`).
- **`open_connector/config.rs`** — `OpenConnectorConfig` / `OpenConnectorBinding` matching the design spec's YAML shape: `runtime_token_env`, request/scan timeouts, `max_pages`/`max_rows`, `cache_max_bytes`, `cache_ttl_seconds` (0 = live reads), `raw_action_allowlist`, and named `bindings` (`name`, `source_pack`, optional `source_pack_version` pin, `connection_alias`, `resource`, `tables`). The version pin keeps a Skardi upgrade from silently changing a bound table's schema after restart. Both structs use `deny_unknown_fields`, so a misspelled key (e.g. `source_pack_versions`) fails loudly instead of silently disabling the pin.
- **`open_connector/error.rs`** — `OpenConnectorError` with pre-network validation variants. `validate()` is pure, and runs in **both** the server's `validate_data_sources` and the provider's registration path — so CLI and server reject misconfiguration identically. Server validation also rejects non-catalog hierarchy at config load (`OpenConnectorHierarchyRequired`), so a minimal config (which defaults to `table` hierarchy) fails cleanly instead of aborting boot with a wrapped provider error.
- **Server wiring** — `DataSource.open_connector` (required for `type: open_connector`, rejected for every other type), validation, registration dispatch, `/data_source` and dashboard type mapping. `OpenConnector` is in `CATALOG_SUPPORTED_SOURCES`, so the catalog-mode guards (no `table`/`schema` flat options, no empty `allowed_schemas`) fire for it exactly as for postgres/dynamodb — a stray flat option fails loudly instead of being silently ignored. Read-only is enforced at the single shared point: `register_open_connector_tables` rejects `read_write` (`ReadWriteNotSupported`), so server and CLI apply the invariant identically; the server additionally keeps its typed `UnsupportedWriteMode` at config validation, and job destinations are rejected as non-transactional.
- **CLI wiring** — same typed field and registration arm.

## Milestone 2 — HTTP client + action registry

- **`open_connector/client.rs`** — `OpenConnectorClient` against the gateway's `/v1` contract:
  - health checks, action discovery, action execution; endpoint paths centralized in private constants as the single reconciliation point with the upstream API; action IDs validated at the client boundary (bare `.`/`..` and `/` rejected as `InvalidActionId` before any request, so an ID can't escape `/v1/actions/` via `Url::join` dot-segment resolution), with the same check on `raw_action_allowlist` at config-validation time.
  - runtime token read from the configured env var, trimmed and validated at client construction (`InvalidRuntimeToken` for control characters / empty-after-trim — the `export TOKEN="$(cat token.txt)"` newline case fails fast with the real cause; unbuildable requests are terminal via `RequestBuildFailed`, never retried), sent as a Bearer header; gateway URLs must be `http(s)` without embedded credentials **and** without query/fragment (a `?token=…` query would leak into logs, `Debug` output, and the data-sources API); the `Debug` impl excludes the token.
  - bounded retries split by idempotency: GET health/discovery retry 429 / transient 5xx / transport errors with exponential backoff + jitter, honoring `Retry-After` (capped); POST execute retries **only** a pre-execution 429 — 5xx is terminal and transport failure raises `NonIdempotentAmbiguousFailure`, so a possibly-executed action is never re-sent.
  - bounded response decoding enforced on both declared `Content-Length` and the actual streamed bytes; per-request timeout from config; `max_response_bytes` / `max_attempts` are operator-tunable config fields wired through `from_config` (zero rejected as `ZeroSafetyBound`); terminal error paths read a 4 KiB snippet instead of buffering a worst-case error page for a 512-char message; execute serializes a borrowing envelope so the caller's input `Value` is not deep-cloned.
  - connection-alias header on execute calls; `execute()` is `pub(crate)`, so the default-deny gating (allowlist + `locally_executable` checks in `ActionRegistry`) is structurally un-bypassable from outside the crate — discovery and health stay public metadata, execution is mediated by the registry/UDTF layers.
- **`open_connector/action_registry.rs`** — `ActionRegistry`:
  - discovers `raw_action_allowlist` actions at registration (deduplicated, concurrency-bounded); actions that are not locally executable are rejected, a **missing** executability flag is rejected as `ActionExecutabilityUnknown` (default-deny — `Option<bool>` end to end so "not declared" is never read as "executable"), and a partially loaded registry is never returned.
  - per-action **compatibility fingerprint**: canonicalized output schema (recursive key sort) hashed with FNV-1a — stable across processes without a crypto-hash dependency, used by later milestones to fail registration on incompatible upstream action changes.
- **Registration flow** — `register_open_connector_tables` now validates config, health-checks the gateway, loads the allowlist registry, then fails with `ExecutionNotImplemented` until source packs land.
- **`reqwest` becomes a hard dependency** of the `skardi` crate (the client is unconditional); `remote-embed` / `llm-extract` features now only gate their UDF code. Retry-After parsing is deduplicated into a new shared `skardi::util::http::parse_retry_after` used by all four reqwest clients (open_connector, remote_embed, llm_extract anthropic/openai_compat) — retry *loops* intentionally stay bespoke since their policies differ; the `open_connector` type string now comes from the existing `DataSourceType::as_str()` instead of three hand-maintained matches (gui, pipeline_handlers).
- **Test infrastructure** — a hand-rolled mock gateway (`testutil.rs`) over `tokio::net::TcpListener`, so no mock-HTTP crate is added. It records requests (method, path, headers, body) for header/auth assertions and scripts retry sequences (429 → `Retry-After` → 200, repeated 5xx, …).

## Test coverage

110 `skardi` open_connector tests, including:

- config: design-spec YAML round-trip, defaults, every validation variant (including zero `max_pages`/`max_rows` **and** zero request/scan timeouts/bounds as `ZeroSafetyBound`), unknown-field rejection (misspelled `source_pack_version`, plural-typo `raw_action_allowlist`, unknown binding keys, misspelled cache field);
- client: bearer-token header, retry on 5xx then success, `Retry-After` honored, terminal 4xx without retry, retry exhaustion on unreachable gateway, URL validation (credentials/query/fragment rejected), missing/malformed token env (trim + `InvalidRuntimeToken`, terminal `RequestBuildFailed`), discovery parsing, malformed JSON, response-size bound, execute envelope (object without `output` rejected, bare array whole) + alias header + request body, execute retry policy (5xx terminal / 429 retried / transport error not retried), config-wired `max_attempts`/`max_response_bytes`;
- registry: dedup, empty allowlist (zero network calls), non-executable action rejection, missing executability flag rejected as `ActionExecutabilityUnknown` (default-deny), fingerprint stability across key order / sensitivity to schema changes;
- scan engine: row-path extraction errors, fixed-schema conversion (nulls, wrong types, timestamps, lists, opaque JSON, empty pages), pagination termination + repeated-cursor loop detection, filter translation (Exact/Unsupported, literal-side normalization), cache (TTL expiry, LRU eviction, canonical keys, LIMIT in key);
- catalog end-to-end against a mock gateway: multi-page scan through SQL, Exact filter pushdown verified in request bodies, LIMIT stopping at one live page, cache replay with zero new requests, self-join identical-key/concurrent-fetch boundary test;
- registration: every failure stage asserted to fire **before** any network call (hierarchy, read-only, missing config, invalid config, missing token, unknown pack, version-pin mismatch).

Plus local CLI integration against a Python stub gateway: full-scan pagination, `min_value` pushdown, LIMIT early stop, and a federated JOIN with a local CSV — all verified end to end. 7 `skardi-server` and 4 `skardi-cli` tests from milestone 1. `cargo check --workspace --all-targets`, `cargo clippy` (no new warnings), `cargo fmt --check`, and full `cargo test` for all three crates pass (110 open_connector / 607 skardi lib / 133 server / 61 cli).

## Out of scope (later milestones)

Source-pack registry, JSON-to-Arrow conversion, pagination, filter/limit pushdown, cache, `open_connector_query` / `open_connector_scan` UDTFs. SQL DDL registration (`CREATE EXTERNAL TABLE ... STORED AS OPEN_CONNECTOR`) stays out per the approved design — stable tables register through context YAML only; DDL is a documented future extension. README's supported-sources table is intentionally **not** updated yet — the source is not queryable until the scan engine lands.


crates/skardi/src/util/hash.rs
```diff
@@ -0,0 +1,18 @@
...
+/// checksum. BLAKE3 is already used for stable document IDs in this crate.
+///
+/// # Example
+/// ```
+/// use skardi::util::hash::blake3_hex;
+///
+/// assert_eq!(blake3_hex(b"abc"), blake3_hex(b"abc"));
+/// assert_ne!(blake3_hex(b"abc"), blake3_hex(b"abd"));
+/// ```
+pub fn blake3_hex(bytes: &[u8]) -> String {
```
BtXin: Is it really necessary to create a new file for this?

BtXin (commented):
Solid milestone overall — disciplined error handling (kinds, never values), drop-to-cancel streaming, fail-instead-of-partial-results, and genuinely strong test coverage. Inline comments below; the two I'd consider blocking before merge:

1. **🔴 `GtEq` → `min_value` pushed as `Exact` returns wrong results** for `WHERE value >= x` (boundary row silently dropped, and DataFusion never reapplies the filter). See comment on `packs/mock.rs`.
2. **🟡 Test flake risk**: two parallel tests share the `SKARDI_TEST_OC_REGISTER_CATALOG_CACHE` env var and one removes it while the other may still need it. See comment on `mod.rs`.

Smaller items (cache key hardcodes pack version, LIMIT-satisfied scans never populate the cache, nullable columns swallowing structural errors, cursor-path re-parse per page, error-variant reuse) are inline — fine as fast follow-ups if you'd rather ticket them.

One PR-description nit: the body still says fingerprints use "FNV-1a … without a crypto-hash dependency" and names `util::hash::fnv1a_hex`, but the code ships BLAKE3 / `blake3_hex` (the tasks doc is already correct). Worth updating so the description matches what merges.

crates/skardi/src/sources/providers/open_connector/packs/mock.rs
```diff
@@ -0,0 +1,71 @@
...
+    // Small page size so tests exercise multi-page scans with few rows.
+    pagination: PaginationStrategy::PageNumber {
+        page_param: "page",
+        per_page_param: "per_page",
+        per_page: 2,
+    },
+    required_resources: &["workspace"],
+    filters: &[FilterMapping {
+        column: "value",
+        operators: &[Operator::Gt, Operator::GtEq],
```
BtXin: 🔴 **Correctness: `GtEq` here produces wrong query results.**

Both `Gt` and `GtEq` map to the same `min_value` input, and `translate_filters` classifies any operator match as `Exact`. The gateway's `min_value` semantics are strictly greater-than (the mock handler filters `item.value > min`, and the tasks doc describes this mapping as `` `value >` → `min_value` ``). So `SELECT … WHERE value >= 3.0` pushes `min_value: 3`, the provider drops the `value == 3.0` row, and because the filter was `Exact`, DataFusion never reapplies it — the boundary row silently vanishes.

`Inexact` wouldn't rescue this either: reapplying a filter can't resurrect rows the provider already excluded. `GtEq` has to be `Unsupported` for a strict-`>` input, i.e. removed from this list.

**Suggestion:** drop `Operator::GtEq` here, and add a `>= ` end-to-end test — the existing test only exercises `value > 3.0`, which is exactly why this slipped through. More structurally: a `FilterMapping` listing multiple operators against one input field is a footgun, since two operators with different semantics can never both be exactly represented by the same `(input_field, literal)` pair. Consider one operator per mapping (or per-operator translation) before real packs land.

crates/skardi/src/sources/providers/open_connector/filters.rs
```diff
@@ -0,0 +1,235 @@
...
+    // literal is on the left (`5 < col` → `col > 5`).
+    let (column, operator, literal) = match (binary.left.as_ref(), binary.right.as_ref()) {
+        (Expr::Column(column), Expr::Literal(literal, _)) => (column, binary.op, literal),
+        (Expr::Literal(literal, _), Expr::Column(column)) => (column, binary.op.swap()?, literal),
+        _ => return None,
+    };
+
+    let mapping = mappings
+        .iter()
+        .find(|mapping| mapping.column == column.name)?;
```
BtXin: Nit (related to the `GtEq` comment on `packs/mock.rs`): this resolves the mapping by column only, first match wins. A future pack can't express a range via two mappings for the same column (`Gt → min_value`, `Lt → max_value`) — the second is unreachable, and its filter silently stays `Unsupported`. Matching on column **and** operator (`.find(|m| m.column == column.name && m.operators.contains(&operator))`) fixes that and composes with one-operator-per-mapping.

crates/skardi/src/sources/providers/open_connector/exec.rs
```diff
@@ -0,0 +1,405 @@
...
+                .fields()
+                .iter()
+                .map(|f| f.name().clone())
+                .collect(),
+        };
+        let cache_key = scan_cache_key(&ScanKeyParts {
+            gateway: &exec.gateway,
+            connection_alias: exec.connection_alias.as_deref(),
+            action_id: exec.table.action_id,
+            source_pack_version: 1,
```
BtXin: 🟡 **Cache key hardcodes `source_pack_version: 1`** instead of the bound pack's actual version. Today it's harmless (one built-in version per process, in-memory cache), but it silently defeats the documented purpose of this key component the day a pack ships v2 — and both the PR body and tasks doc advertise "pack version" as part of the canonical key.

The exec can't see the version because `SourcePackTable` doesn't carry it. Either thread the pack version through `OpenConnectorTableProvider::new` (registration has it), or add a `version` field to `SourcePackTable`.

crates/skardi/src/sources/providers/open_connector/exec.rs
```diff
@@ -0,0 +1,405 @@
...
+            seconds: self.scan_timeout.as_secs(),
+        }
+    }
+
+    /// Fetch and convert one page; returns None when the scan is complete.
+    async fn next_page(&mut self) -> Result<Option<RecordBatch>, OpenConnectorError> {
+        if let Some(batch) = self.replay.pop_front() {
+            return Ok(Some(batch));
+        }
+        if self.done || self.limit_remaining == Some(0) {
```
BtXin, line 320: 🟢 **LIMIT-satisfied scans never populate the cache.** `store_cache()` only runs when `pagination.advance()` reports exhaustion. When LIMIT stops the scan early, this branch returns `None` without storing (and DataFusion's limit operator may stop polling before the state even reaches here). So repeated identical `LIMIT n` queries always fetch live, even with caching enabled.

Since LIMIT is part of the cache key, the truncated batches are the *complete* result for that key and are safe to cache — consider storing eagerly at the point `limit_remaining` hits zero. At minimum, document this boundary next to the existing "overlapping scans" note in `cache.rs`.

crates/skardi/src/sources/providers/open_connector/exec.rs
```diff
@@ -0,0 +1,405 @@
...
+            return Err(OpenConnectorError::ScanBoundsExceeded {
+                table: self.table.id.to_string(),
+                bound: "max_pages",
+                limit: u64::from(self.max_pages),
+            });
+        }
+
+        // Assemble the action input: fixed resource inputs, Exact filters,
+        // then page parameters.
+        let mut input = self.resource.as_object().cloned().unwrap_or_default();
```
BtXin: Nit: a non-object `resource` silently becomes empty action input here. Registration always builds a `Value::Object`, so per the repo's error-handling convention this invariant reads better as `.expect("resource is built as Value::Object at registration")` (or an early typed error) rather than a silent `unwrap_or_default()`.

crates/skardi/src/sources/providers/open_connector/json_to_arrow.rs
```diff
@@ -0,0 +1,537 @@
...
+        field: &CompiledField,
+        rows: &[Value],
+        page: usize,
+    ) -> Result<ArrayRef, OpenConnectorError> {
+        let mapping = &field.mapping;
+        let mut cells: Vec<Option<&Value>> = Vec::with_capacity(rows.len());
+        for (row_index, row) in rows.iter().enumerate() {
+            match field.path.extract(row, page) {
+                Ok(value) => cells.push(Some(value)),
+                Err(_) if mapping.nullable => cells.push(None),
```
BtXin: 🟢 **Nullable columns swallow structural errors.** `Err(_)` converts *any* extraction failure to null — including traversing a non-object mid-path. If upstream changes `user` from an object to a string, a nullable `user.login` column quietly becomes all-null instead of failing, which contradicts the module's "upstream breaking changes fail conversion" contract.

Consider nulling only a missing key at the *final* segment and keeping mid-path shape mismatches as `ConversionFailed` (the row-path error already distinguishes the segment).

crates/skardi/src/sources/providers/open_connector/pagination.rs
```diff
@@ -0,0 +1,263 @@
...
+                if more {
+                    self.page += 1;
+                    self.next_token = Some(self.page.to_string());
+                }
+                Ok(more)
+            }
+            PaginationStrategy::Cursor {
+                next_cursor_path, ..
+            } => {
+                let path = RowPath::parse(next_cursor_path)?;
```
BtXin: 🟢 `next_cursor_path` is re-parsed on every page, and a malformed (static, pack-authored) path only surfaces mid-scan instead of at registration. Parse once — e.g. in `Pagination::new` (returning `Result`), or validate alongside `row_path` in `OpenConnectorTableProvider::new` so a pack bug fails at bind time like the other paths do.

crates/skardi/src/sources/providers/open_connector/mod.rs
```diff
@@ -345,59 +450,389 @@ bindings:
...
     }
+    /// Env var for the catalog tests (unique per test file section).
+    #[cfg(test)]
+    const TOKEN_ENV_CATALOG_BASIC: &str = "SKARDI_TEST_OC_REGISTER_CATALOG_BASIC";
+
+    #[cfg(test)]
+    const TOKEN_ENV_CATALOG_FILTER: &str = "SKARDI_TEST_OC_REGISTER_CATALOG_FILTER";
+
+    #[cfg(test)]
+    const TOKEN_ENV_CATALOG_CACHE: &str = "SKARDI_TEST_OC_REGISTER_CATALOG_CACHE";
```
BtXin, line 836: 🟡 **Test flake risk:** `cached_scan_replays_without_new_requests` and `self_join_scans_compute_identical_keys_but_fetch_live` both set and then **remove** this same env var. Cargo runs tests on parallel threads, so one test's `remove_var` can land between the other's `set_var` and its client construction, failing registration with a missing-token error. Every other test here got a unique var name for exactly this reason — give these two distinct names as well.

crates/skardi/src/sources/providers/open_connector/mod.rs
```diff
@@ -137,29 +152,119 @@ pub async fn register_open_connector_tables(
...
+                ),
+                config.max_pages,
+                config.max_rows,
+                scan_timeout,
+            )?;
+            schema_provider
+                .register_table(table_name.clone(), Arc::new(provider))
+                .map_err(|e| OpenConnectorError::ActionContractMismatch {
+                    table: table.id.to_string(),
+                    reason: format!("failed to register into catalog schema: {e}"),
```
BtXin: Nit: catalog/schema registration failures here (and below for `register_schema`) are wrapped in `OpenConnectorError::ActionContractMismatch`, which is misleading — nothing about the action contract mismatched. A dedicated variant (or `anyhow::Context` on the registration call) would keep the error taxonomy honest for operators grepping logs.

BtXin (approved):

PR #163, opened 2026-07-21
docs(spec): RSS feed support

gracexmatin:
## Body
 
This PR contributes the design spec for first-class RSS/Atom support (`docs/superpowers/specs/2026-07-20-rss-feed-support-design.md`). Design-only; implementation follows in three milestones (M1 provider core, M2 surfaces, M3 skill).
 
### What's designed
 
- **Many dialects, one representation.** RSS 0.9x/1.0/2.0, Atom 0.3/1.0, and JSON Feed all normalize to a single protocol-pinned Arrow schema via `feed-rs` (Decisions 10, 12).
- **Conformance checking of fetched data.** After every parse, the provider records what the feed *claimed* to be vs what it *parsed as*, plus spec-required-field violations and sanitation repairs, in queryable `feeds.dialect` / `dialect_declared` / `conformance_notes` columns — tolerance made visible, never silent (Decision 16).
- **Annotated schema mapping.** A normative dialect → unified-schema Field Mapping table ships in `docs/rss.md` and as semantics-overlay column descriptions, so agents see each column's per-dialect provenance (Decision 17).
- **One table-provider, many feeds.** A single source binds a subscription list and exposes two fixed tables — `feeds` (health) and `items` (live union, `feed` discriminator column); per-feed tables explicitly rejected (Decisions 1–2).
- **TTL cache + conditional-request validators.** Scan-time fetch through a per-feed cache: within TTL → serve cached, zero network; expired + stored ETag/Last-Modified → revalidate, 304 re-arms cache without reparsing; otherwise full fetch/parse. Only complete parses are ever cached (Decision 5).
Plus: partition-per-feed execution with per-feed fault isolation, `rss_scan(url)` UDTF, `html_to_markdown()` UDF, fixture-corpus compatibility ratchet, and the `auto_news_base` skill contract.


docs/superpowers/specs/2026-07-20-rss-feed-support-design.md
```diff
@@ -0,0 +1,361 @@
...
+
+- **No full-article scraping.** Feeds that carry only summaries are served as summaries. Fetching and extracting article HTML is crawler territory (anti-bot, rendering, extraction) and out of scope.
+- **No scheduler.** Refresh cadence belongs to the caller (agent, cron, future jobs extension). The TTL cache makes repeated polling cheap; nothing in the provider runs unbidden.
+- **No push (WebSub).** Recorded as a future extension; it is a cache-invalidation signal, not a different provider shape.
+- **No write path.** `INSERT`/`UPDATE`/`DELETE` against a feed is meaningless; the source registers strictly read-only (`WRITABLE_SOURCE_TYPES` untouched).
+- **No authenticated feeds.** Cookie-, token-, or basic-auth-protected feeds are out of scope for v1; if demand appears they belong behind the Open Connector gateway or a scoped follow-up.
+- **No history retention in the provider.** The live window is the contract; archiving is a pipeline composition (Decision 4).
+- **No gateway.** RSS is unauthenticated public HTTP with an open wire format; the Open Connector gateway adds a moving part and buys nothing.
+
+## Decisions
```
abbccdda: A general feedback: it is in general hard to review the high-level approach for a design if we put 10+ decisions upfront without knowing what the architecture would look like. Consider adding a first-class arch diagram and walkthrough of how we interact with rs feed first before jumping into the detailed decisions, and try to put up explicit sections to group the decisions into logical blocks for the ease of reviewing.

docs/superpowers/specs/2026-07-22-rss-feed-support-design-v2.md
```diff
@@ -0,0 +1,512 @@
...
+
+### SQL layer
+
+#### Table providers and execution plan
+
+`feeds` and `items` are fixed-`SchemaRef` `TableProvider`s over a shared execution plan that exposes one partition per subscription. A `feeds` scan reads cache/state for feeds within TTL and revalidates the rest — the cheap health check.
+
+#### `rss_scan` UDTF and `html_to_markdown()` UDF
+
+`rss_scan(url)` returns the `items` schema for a single unregistered feed through the same fetch/sanitize/parse path, default TTL 0. `html_to_markdown()` is a scalar UDF registered alongside the existing model UDFs, backed by a pure-Rust HTML→Markdown crate (`htmd`/`html2md` class, selected at implementation); it is useful for any HTML-bearing column, not just RSS.
```
BtXin: Let's only allow registered rss feed for now, so we don't allow rss_scan(). As a result we don't need html_to_markdown() either, for now.

docs/superpowers/specs/2026-07-22-rss-feed-support-design-v2.md
```diff
@@ -0,0 +1,493 @@
...
+| Scenario | Behavior |
+|---|---|
+| Feed down / DNS failure | Partition serves stale cached window (`stale-error`) or zero rows if never fetched; other partitions unaffected; `feeds` row records error; tracing warns |
+| Malformed XML | Strict parse → sanitation → retry; success traced with repairs recorded; failure sets `last_status = 'error'`, `last_error` names the parse stage |
+| Dialect misdeclaration | Parses normally; mismatch recorded in `dialect_declared` vs `dialect` and `conformance_notes` |
+| Feed omits `guid` | `link` used as guid; dedup collapses to link identity |
+| Response exceeds `max_response_bytes` | Fetch aborts with a targeted error status; never partial-parsed |
+| Slow feed | Per-request timeout isolates it; scan deadline bounds the whole query |
+| HTTP 304 | Cache re-armed without reparse; `last_status = 'revalidated'` |
+| HTTP 429 / transient 5xx | Bounded jittered retries honoring `Retry-After` within the scan deadline |
```
abbccdda: **Should exhausted retries be surfaced in-band rather than only via the `feeds` health table?**

The spec absorbs 429 / transient 5xx into bounded retries, then on exhaustion degrades the partition to stale/zero rows and records the reason in `feeds.last_error`. My worry: this only satisfies "no silent incompleteness" if the caller actually inspects `feeds`. A plain `SELECT ... FROM items` (or the hybrid-search path an agent drives) gets a row stream back with no in-band signal that a feed was dropped — so to that consumer the incompleteness *is* silent. Relying on a side-table health check pushes the burden onto every caller to remember to join/query `feeds`, which most agent queries won't.

I don't think the answer is failing the whole scan — that's the rejected all-or-nothing alternative, and it defeats per-feed fault isolation. But is there a middle option that keeps isolation while making exhaustion visible without a separate query? A few to weigh:

- An in-band marker on the affected rows / scan (e.g. a warning surfaced through the query result, not just tracing), so the caller learns of degradation from the same response.
- Distinguishing "never fetched / hard error" from "serving stale" more loudly at the point of consumption.

At minimum, worth stating explicitly *who* is expected to poll `feeds` and how often, since the whole "visible degradation" guarantee hinges on it. Curious whether surfacing exhausted-retry failures immediately (alongside the partial rows) was considered and rejected, and why.

docs/superpowers/specs/2026-07-22-rss-feed-support-design-v2.md
```diff
@@ -0,0 +1,493 @@
...
+          - url: https://this-week-in-rust.org/rss.xml
+        # or: opml: subscriptions.opml # mutually exclusive with feeds:
+        ttl_seconds: 900               # 0 = always live
+        max_concurrent: 6              # fetch parallelism and per-host politeness bound
+        request_timeout_seconds: 10
+        max_response_bytes: 5242880
+        user_agent: "skardi-rss/<version> (+https://github.com/SkardiLabs/skardi)"
+```
+
+Subscription management is a configuration action, not a SQL action: no SQL statement can add, alter, or remove a subscription, preserving the SQL validator's no-DDL invariant. Agents manage subscriptions by editing `ctx.yaml`/OPML and reloading — a flow the `auto_news_base` skill owns.
```
abbccdda, line 271: **SSRF: agent-authored subscriptions are server-fetched, with no URL policy.**

Subscriptions are server-side-fetched URLs, and this section has *agents editing `ctx.yaml`/OPML* to manage them. That makes feed URLs an attacker-influenceable input: a prompt-injected agent adds `http://169.254.169.254/latest/meta-data/` (or an internal service) as a feed, the server fetches it, and the body comes back as `items` rows — SSRF + exfiltration. The security discussion is currently silent on this.

Worth noting the codebase already has the pattern to reuse: `fetch_image_with_policy` in `crates/skardi/src/model/llm_extract/mod.rs` is default-deny and blocks link-local/private ranges (there is even a `169.254.169.254` test). Could the RSS fetcher adopt the same default-deny policy (block loopback/link-local/private, optional allowlist) and have the spec state it — especially since subscriptions are agent-authored?

docs/superpowers/specs/2026-07-22-rss-feed-support-design-v2.md
```diff
@@ -0,0 +1,493 @@
...
+
+### Engine layer
+
+#### Fetcher
+
+The fetcher owns bounded HTTP: per-request timeout, total scan deadline, response-size cap, bounded jittered retries honoring `Retry-After`, cancellation, and conditional GET with stored ETag / Last-Modified validators. Concurrency is capped by `max_concurrent`, which doubles as the per-host politeness bound. A self-identifying `User-Agent` is sent by default — feed servers routinely ban anonymous clients.
+
+#### Cache
+
+A per-feed, in-memory, bounded (bytes + entries, LRU) TTL cache behind a small trait so a persistent implementation can be swapped in later without touching the table providers. An entry stores parsed Arrow batches, ETag / Last-Modified validators, and fetch metadata. Only complete, successfully parsed feed windows are ever stored.
```
abbccdda, line 192: **In-memory cache is process-local — politeness and 304 revalidation degrade under restart/replicas.**

Two consequences of the in-memory cache (and in-memory ETag/Last-Modified validators) that the spec should state explicitly:

- **Politeness is per-process, not per-host-global.** `max_concurrent` bounds one server's concurrency; with N replicas (or a crash/restart loop) a feed host can see up to N× the load, so the "per-host politeness bound" claim should be scoped to a single process.
- **Validators are lost on restart**, so the first post-restart scan is a full `200` for every feed rather than a `304` — the conditional-GET savings vanish precisely during a fleet redeploy. Worth calling out that cross-restart savings depend on the (future) persistent cache.

docs/superpowers/specs/2026-07-22-rss-feed-support-design-v2.md
```diff
@@ -0,0 +1,493 @@
...
+
+#### Parser and conformance checker
+
+A strict `feed-rs` parse is attempted first; on failure a bounded, deterministic sanitation pass runs and the parse is retried once. After every successful parse, a conformance check records the declared dialect, the parsed dialect, and any deviations. Details in "Parsing, Sanitation, and Conformance".
+
+### SQL layer
+
+#### Table providers and execution plan
+
+`feeds` and `items` are fixed-`SchemaRef` `TableProvider`s over a shared execution plan that exposes one partition per subscription. A `feeds` scan reads cache/state for feeds within TTL and revalidates the rest — the cheap health check.
```
abbccdda: **Querying `feeds` is not actually free — it can trigger N conditional GETs, and full bodies.**

"A `feeds` scan reads cache/state for feeds within TTL and revalidates the rest — the cheap health check." Revalidation is a conditional GET (network), and on `200` it pulls the full body — so it's only cheap for in-TTL feeds. This also compounds the in-band-signal concern raised on the 429 row: the recommended way to detect a broken feed is to query `feeds`, but that query performs the very fetch that may fail. Could we clarify whether a `feeds` scan pulls full bodies or takes a lighter path, and document that health-check cost is proportional to the number of expired feeds?

docs/superpowers/specs/2026-07-22-rss-feed-support-design-v2.md
```diff
@@ -0,0 +1,493 @@
...
+        end
+    end
+    TP-->>DF: RecordBatch stream
+```
+
+Execution rules:
+
+- Each feed is one DataFusion partition: parallel fetching without a bespoke pool, streaming batches (fast feeds emit before slow ones), and a natural fault boundary.
+- `feed`/`feed_url` equality and `IN` predicates prune partitions before any fetch and are reported `Exact`; all other predicates remain in DataFusion — RSS has no query parameters, so nothing else can reach the wire.
+- `LIMIT` stops launching further partitions once satisfied; items have no global order (documented).
```
abbccdda: **`LIMIT` short-circuit interacts badly with the "visible health" guarantee.**

"`LIMIT` stops launching further partitions once satisfied" combined with "items have no global order" means `SELECT * FROM items LIMIT 20` fetches a nondeterministic subset of feeds — and the un-launched feeds' health is never refreshed, so their `feeds` rows stay stale with no signal. Two identical `LIMIT` queries can also touch different feeds and do different amounts of work. This is worth an explicit note, and it sits in tension with the "no silent incompleteness" principle: incompleteness introduced by LIMIT pruning is itself invisible.

docs/superpowers/specs/2026-07-22-rss-feed-support-design-v2.md
```diff
@@ -0,0 +1,493 @@
...
+
+Caching claims no cross-feed consistency: a multi-feed scan can observe different feeds at different freshness, visible per row via `feeds.last_fetch`.
+
+## Parsing, Sanitation, and Conformance
+
+`feed-rs` parses RSS 0.9x/1.0/2.0, Atom 0.3/1.0, and JSON Feed 1.0/1.1. The tolerance strategy:
+
+1. **Sanitation pre-pass, on failure only.** A strict parse is attempted first. On failure, a bounded, deterministic sanitation pass runs — encoding sniff (BOM / XML declaration / `encoding_rs`), re-encode to UTF-8, strip control characters, repair naked ampersands — and the parse is retried once. Both attempts and applied repairs are traced and recorded in `conformance_notes`.
+2. **Conformance check, after every successful parse.** Detect the declared dialect from the document itself (root element + version attribute; Content-Type as corroboration), compare with what `feed-rs` parsed, and verify the dialect's spec-required fields (e.g. RSS 2.0 channel `title`/`link`/`description`). Deviations populate `feeds.dialect_declared` / `dialect` / `conformance_notes`; they never reject a feed that parsed — the check converts silent tolerance into queryable evidence.
```
abbccdda: **Conformance record feasibility depends on feed-rs actually exposing the *parsed* dialect.**

The check compares the declared dialect "with what `feed-rs` parsed." feed-rs normalizes everything into a unified `Feed` model and does not obviously surface "I parsed this as RSS 2.0 vs Atom 1.0 vs …". If it doesn't, both `dialect` and `dialect_declared` have to be sniffed independently from the raw document, and `dialect` no longer means "what the parser did." Could we confirm feed-rs surfaces the parsed dialect/version, and adjust the column semantics if not? (feed-rs isn't a dependency yet, so this is unverified in-tree.)

docs/superpowers/specs/2026-07-22-rss-feed-support-design-v2.md
```diff
@@ -0,0 +1,493 @@
...
+
+Cache entries are per feed, not per scan, so partial hits refetch only expired feeds. The completeness invariant is adopted at feed granularity: **only a complete, successfully parsed feed window is ever cached** — a half-parsed feed is never served. Conditional requests still minimize transfer under `ttl_seconds: 0`.
+
+Caching claims no cross-feed consistency: a multi-feed scan can observe different feeds at different freshness, visible per row via `feeds.last_fetch`.
+
+## Parsing, Sanitation, and Conformance
+
+`feed-rs` parses RSS 0.9x/1.0/2.0, Atom 0.3/1.0, and JSON Feed 1.0/1.1. The tolerance strategy:
+
+1. **Sanitation pre-pass, on failure only.** A strict parse is attempted first. On failure, a bounded, deterministic sanitation pass runs — encoding sniff (BOM / XML declaration / `encoding_rs`), re-encode to UTF-8, strip control characters, repair naked ampersands — and the parse is retried once. Both attempts and applied repairs are traced and recorded in `conformance_notes`.
```
abbccdda: **Sanitation "repair naked ampersands" is a known corruption foot-gun; and "wire-faithful" needs pinning down.**

Text-level ampersand/entity repair on XML routinely corrupts CDATA and already-valid entities — it is a classic source of *new* breakage. The evidence-loop + fixtures help, but it would be worth committing to a contract test that sanitation is a strict superset: it never breaks a feed that would have parsed after the encoding-fix alone.

Separately, if sanitation rewrites bytes, "store content wire-faithful (HTML)" should be defined as "the HTML fragment feed-rs extracted," not "the pre-sanitation bytes" — otherwise the term is misleading.

docs/superpowers/specs/2026-07-22-rss-feed-support-design-v2.md
```diff
@@ -0,0 +1,493 @@
...
+
+The archive contract is two tables. `news_items` retains one wire-faithful row per entry — primary key `(feed, guid)`, plus `title`, `link`, `author`, `published`, and `content` (HTML) — and is the anti-join target for ingest. `news_chunks` holds `(feed, guid, chunk_idx, chunk_text, embedding, ingested_at)` with the fts5/vec0 mirrors attached. Ingest is two `INSERT` steps: new entries land wire-faithful in `news_items`, then are chunked and embedded from `news_items` into `news_chunks`. Search joins the two inside the archive, so results remain citable (title + link + published) after entries fall out of the live window, and history can always be re-chunked or re-embedded from retained content.
+
+### Skill lifecycle
+
+First assembly is the five-step flow under Rollout. Afterwards, every rendered artifact except the `rss:` block is subscription-agnostic — pipelines, DDL, aliases, and semantics reference `news.main.items`, never individual feeds — so the lifecycle splits cleanly:
```
abbccdda, line 224: **Cross-repo coupling: the archive contract lives in an external skill with no version pin.**

`news_items`/`news_chunks` and the pipelines live in `skardi-skills`, yet acceptance criteria #6/#8/#11 test them. If the engine's `items`/`feeds` schema ever shifts, the external skill's rendered pipelines break with nothing enforcing the contract. The sibling Open Connector design added source-pack versioning precisely to avoid silent schema drift. How are engine and skill kept in lockstep — cross-repo CI, or a declared `items` schema version the skill checks?

docs/superpowers/specs/2026-07-22-rss-feed-support-design-v2.md
```diff
@@ -0,0 +1,493 @@
...
+- Isolate faults per feed; make feed health queryable in SQL, never silent.
+- Record declared-versus-parsed dialect conformance queryably.
+- Publish the dialect → unified-schema mapping as documentation and semantics annotations.
+- Support federated joins between feed items and existing Skardi sources.
+- Let `auto_news_base` assemble a searchable, citable news base from a natural-language subscription list, keep it maintainable afterwards through configuration edits alone, and keep results citable after entries leave the live window.
+
+## Non-goals
+
+- Full-article scraping. Feeds that carry only summaries are served as summaries; crawling is a different product.
+- A scheduler. Refresh cadence belongs to the caller; the TTL cache makes polling cheap; nothing runs unbidden.
```
abbccdda, line 42: **"Continuously updating" motivation vs. no scheduler — name the dependency.**

The motivation sells "continuously updating external information," but with no scheduler, freshness depends entirely on an external cron invoking `skardi sync`. That's a reasonable scope cut, but the dependency on external orchestration should be named in the motivation/rollout so readers don't assume the provider self-refreshes.

docs/superpowers/specs/2026-07-22-rss-feed-support-design-v2.md
gracexmatin: Addressed in commit 6d01285:
**In-band marker**: `items` rows now carry `window_status` (`fresh` | `revalidated` | `stale-error`). A feed that exhausts retries but has a cached window serves it with rows stamped `stale-error` — a plain `SELECT … FROM items` sees the degradation in the result stream itself, and strict-freshness consumers can filter it out with one predicate.
**A feed that fails.** Its absence leaves no trace in the result for a marker to ride on.  Did prototype inventing a placeholder row per failed feed (entry columns NULL, status stamped) and rejected it: a fake row breaks the `guid`/`position` non-null contract, archive ingest would swallow it as a real entry, and any content filter (`published >= …`) silently drops an all-NULL row — the marker vanishes under exactly the filtered queries agents actually run. A warning attached to the query result was also weighed: DataFusion exposes no warning channel, so that would be a CLI/server response-envelope mechanism — recorded in Alternatives as deferred, not rejected. For these feeds the signal lives in `feeds` and in `sync`'s closing report (below).
**How polls `feeds`** :`sync` is the one operation that touches every feed, and its pipeline now ends with a health `SELECT` whose rows are the command's response: each degraded feed with its reason and as-of time, an empty report meaning all healthy, report-only (a degraded feed never fails the run). Exhausted retries therefore reach the agent in the output of the very command that hit them — no follow-up query. Ad-hoc checks stay cheap on top: `feeds` is now a pure observation surface — it never fetches, and failures re-arm the TTL (negative caching), so a health read never re-pokes a feed that just 429'd — and the anti-join absence check is specced in SQL Interfaces and taught via the semantics overlay.

abbccdda:
@gracexmatin Any update?

docs/superpowers/specs/2026-07-22-rss-feed-support-design-v2.md
gracexmatin: The SSRF concern is valid, and the spec now specifies it — a new **Security section** plus the Fetcher component. Pushed on add_RSS_Feed (`0b000e2`, `aed30bc`).
According to `fetch_image_with_policy`, the logic is as below: resolve the host, default-deny loopback/link-local/private/CGNAT/ULA, re-validate on each redirect hop, and connect to the validated IP (DNS-rebinding safe), kept local to the RSS fetcher. An allowlist for intentional internal feeds is deferred to Future Extensions.
Feed content——unsolved

docs/superpowers/specs/2026-07-22-rss-feed-support-design-v2.md
gracexmatin: Stated explicitly in c2a4f71:
**Politeness scope.** The Fetcher section now scopes the claim: `max_concurrent` doubles as the per-host politeness bound *within a single process* — each replica counts independently, so N replicas (or a crash/restart loop) can present a feed host with up to N× the bound. The config example's comment
carries the same qualifier so the two statements can't drift.
**Validators across restarts.** The Cache section now states that the cache — validators included — is process-lifetime state: the first post-restart scan presents no validators and re-fetches every feed as a full `200`, never a `304`, so the conditional-GET savings vanish exactly when a fleet redeploys. Cross-restart savings arrive with the persistent cache, which Future Extensions already tracks; the Cache section now points at that dependency explicitly.

docs/superpowers/specs/2026-07-22-rss-feed-support-design-v2.md
gracexmatin: The latest commit redesigned exactly this:

**① Read/write separation — `feeds` is a pure observation surface.** A `feeds` scan never fetches and never revalidates; it only reads state already held in-process. All network I/O is initiated by `items` scans, so a health check costs no conditional GETs and pulls no bodies, regardless of how many feeds have expired.

**② Negative caching — a failure re-arms the TTL just like a success.** A feed that just `429`'d is not re-contacted within its failure window, even by `items` scans, and `Retry-After` politeness extends across scans instead of resetting with each one.

docs/superpowers/specs/2026-07-22-rss-feed-support-design-v2.md
gracexmatin: The Execution rules expand the `LIMIT` line into a full note: because fetching and health refresh ride on the scan, pruning bounds the side effects as well as the reading — un-launched feeds are not fetched, and their `feeds` observations keep aging past the one-TTL bound that a full scan restores. The Failure Modes entry is updated accordingly, absence diagnosis now documents three causes rather than two (legitimately empty, dead, and not-scanned — pruned by the query's own bare `LIMIT`), and the bundled semantics overlay carries the bare-`LIMIT` caveat, so agents learn the sampling behavior from the schema alone.
"no silent incompleteness."governs failure concealment by the engine, whereas `LIMIT` pruning is truncation requested by the query itself. The incompleteness it introduces is also recoverable from the response: the result's `feed` column is the scan's coverage manifest, and any subscription absent from it is diagnosed in `feeds` as usual.
`LIMIT` is the caller's work bound, and honoring it is the point of the short-circuit; launching every partition
regardless would turn `LIMIT 20` into a full fetch across all subscriptions. The completeness-sensitive shapes are already immune: `ORDER BY … LIMIT` is Top-K and consumes every partition, and the prescribed completeness reads — the absence check and `sync`'s ingest — use no bare `LIMIT`. The nondeterministic sample is therefore confined to exploratory bare-`LIMIT` queries, where it matches standard SQL expectations — now with the side-effect consequence documented instead of implicit.

docs/superpowers/specs/2026-07-22-rss-feed-support-design-v2.md
gracexmatin: Confirmed against the feed-rs source: it does expose the parsed dialect — `Feed.feed_type: FeedType` (model.rs:38), variants `Atom, JSON, RSS0, RSS1,RSS2` (model.rs:224-230), and each sub-parser stamps its own variant (rss2 → `FeedType::RSS2`, etc.; verified one by one). But it's family-level only — no version numbers. Against the value set the spec promised, `rss-0.9x` / `rss-1.0` / `rss-2.0` / `json-feed-1.x` map 1:1 onto `RSS0` / `RSS1` / `RSS2` / `JSON`; the one thing it can't produce is `atom-0.3` vs `atom-1.0`. The XML dispatch keys on exactly root element + version attribute.

Adjusted in e500a2c: `dialect`'s value set drops to feed-rs's true granularity (`atom` with no version; the rest unchanged), keeping its semantics as "what the parser did" (a direct `feed_type` mapping); the conformance check's mismatch axis is rewritten to Content-Type vs parsed family (in-band declaration and parse outcome are the same evidence by construction, so they can't disagree); the Atom 0.3/1.0 distinction stays
with `dialect_declared`.

docs/superpowers/specs/2026-07-22-rss-feed-support-design-v2.md
gracexmatin: Addressed in d5fdda4. **The ampersand repair is now specified, staged, and contract-bound.**
*Behavioral specification.* The repair is defined as lexical, not textual: CDATA sections, comments, and processing instructions pass through untouched, and in the remaining character data only an `&` that cannot open a valid reference is rewritten — the five XML-predefined entities and numeric character references are left alone, while undefined HTML names (`&nbsp;`) become `&amp;`-escaped and so survive into the extracted HTML.
 *Staged retry.* Sanitation now runs as a ladder — re-encode to UTF-8, strip control characters, repair ampersands — applied cumulatively with a re-parse after each rung, stopping at the first success. Your strict-superset property therefore holds structurally rather than by assertion: a feed that would parse after the encoding fix alone never receives the ampersand repair at all, and `conformance_notes` records exactly the minimal repair set that was applied.
*A contract stronger than parse-success.* The committed test is that every rung is a byte-level no-op on well-formed input — CDATA with legal ampersands included — and that fixtures rescued by sanitation pin their expected extracted content. We strengthened it deliberately: the corruption you describe usually does not break the parse, it silently dirties content, so a parses-afterward check would miss precisely the failure it exists to catch. Acceptance criterion 16 and the unit / fixture-corpus test commitments pin all of this.

**"Wire-faithful" is now defined.** Faithful to the HTML fragment the parser extracted: XML transport encoding (entity references, CDATA wrapping) is removed at extraction; sanitation repairs, when any ran, happened to the pre-extraction XML document and are queryable in `conformance_notes`; after extraction nothing is altered. It is a fragment-level fidelity claim, not byte-identity with the transport stream — a distinction that was in fact needed even before sanitation, since extraction itself decodes entities and unwraps CDATA. The definition covers `items.content`/`summary` and the archived `news_items` rows alike.

docs/superpowers/specs/2026-07-22-rss-feed-support-design-v2.md
gracexmatin: Agreed on the naming asymmetry — the promise lived in Motivation while the mechanism only lived as a Non-goals bullet. Named now, with the dependency stated precisely: query-time freshness is self-serve (pull model — any `items` scan refetches TTL-expired feeds at read time), so what external orchestration actually supplies is unattended continuity: archive capture and health-observation freshness. Feeds expose bounded live windows, so an entry that scrolls out between syncs is never captured — that's the real cost of assuming self-refresh.
 Addressed in 8ff1fb1: the Motivation sentence now says continuity is caller-supplied (a scheduled `skardi sync` — cron or a recurring agent session; nothing self-refreshes); the Rollout M3 flow states the steady state (externally paced, cadence faster than the fastest feed's window roll); the rendered `auto_news_base` README commits to teaching the scheduling note; and Failure Modes gains a row for the missed-cadence case — a permanent capture gap with no in-band trace, diagnosed by aging `feeds.last_fetch`.

docs/superpowers/specs/2026-07-22-rss-feed-support-design-v2.md
gracexmatin: Addressed in 394cd34. Direct answer to the closing question: both halves — they protect different populations. An in-repo contract test protects co-development; a declared version protects artifacts already rendered into user contexts, which no CI can see.

**The declared version.** `feeds`/`items` are now a declared stable surface: evolution is additive, and an enumerated list of breaking changes (column removal/rename/retype, nullability tightening, enum-domain repurposing, a change to `(feed, guid)` identity or window semantics) bumps an integer `rss` surface version. Registration surfaces the active version in logs and table metadata — the Open Connector convention. The skill stamps `requires: rss/<version>` into each rendered pipeline's metadata, and the pipeline loader refuses a mismatch with an error naming both versions and the remedy (re-run `auto_news_base` to re-render). One deliberate divergence from the sibling design: Open Connector could leave pinning as a future extension because its source packs compile into the binary and cannot skew; our rendered artifacts live in user space, where skew is the default failure mode — so the handshake is scheduled for M3 with the skill, not deferred.

**The CI half, without cross-repo plumbing.** A canonical render of the skill's artifacts (archive DDL, both pipelines), provenance-stamped, is vendored into the engine's fixture corpus as a pinned representative consumer. Acceptance criteria 6 and 11 execute against it in engine CI — which also answers where those criteria actually run — so an engine change that breaks the rendered surface fails in the same PR that made it. The fixture refreshes when the contract changes, not when the skill does; the skill repository conversely tests its render against a pinned engine release. Criterion 8 (the full `auto_news_base` flow) remains the release gate exercising the real skill end to end, and new criterion 17 pins the version surfacing and the load-time refusal.

Standing cross-repo CI was considered and rejected: it exercises only HEAD×HEAD, protecting neither released pairings nor already-rendered artifacts, at the cost of the heaviest plumbing.

docs/superpowers/specs/2026-07-22-rss-feed-support-design-v2.md
```diff
@@ -0,0 +1,564 @@
+# RSS Feed Support Design
```
abbccdda, line 1: nit: this file should be v1, not v2

BtXin (approved):

PR #165, opened 2026-07-22
feat(sources): Open Connector UDTFs, security policy, and observability

bakey:
## Summary

This PR implements the fourth milestone of the [Open Connector integration design](https://github.com/SkardiLabs/skardi/blob/main/docs/superpowers/specs/2026-07-11-open-connector-integration-design.md), following #159 (milestones 1–2) and #162 (milestone 3): the interactive SQL surface (`open_connector_query` / `open_connector_scan` UDTFs), default-deny security policy enforcement, scan observability, and the first user-facing documentation — the point at which the source is advertised in the README.

**No SQL DDL**, per the approved design: stable tables register exclusively through context YAML ("registration is a configuration action, not a SQL action"). Both UDTFs are plain `SELECT` table functions, so the SQL validator's no-DDL invariant and shared-`SessionContext` semantics are untouched.

### 4.1 `open_connector_query` — built-in pack tables, ad hoc

- **`table_functions.rs` (new)** — `open_connector_query(gateway, 'pack.table', resource_json[, connection_alias])` runs a built-in source-pack table without a persistent YAML binding. It resolves the table against `SourcePackRegistry::builtins()` only — there is no way to name an action through this function, so it can execute nothing but the read actions hard-coded in Skardi's packs.

- **Identical scan path as YAML tables** — The function constructs the same `OpenConnectorTableProvider` a binding does: same stable Arrow schema, same `FilterMapping` allowlist and pushdown classification, same pagination strategy, same `max_pages`/`max_rows`/`scan_timeout` bounds, and the **same per-gateway `ScanCache` instance** — a UDTF invocation with the same resource/filters/projection/LIMIT replays a YAML table's cache entry (and vice versa) with zero new gateway requests, verified in tests.

- **Planning-time gates, no network I/O** — `GatewayHandle` captures each gateway's planning-time state (client, `Arc<ActionRegistry>`, cache, allowlist, bounds) when `register_open_connector_tables` succeeds. Planning re-runs the registration gates in memory: required resource inputs present, action discovered, action-contract fingerprint matches the pack's pin. An action that was never discovered (its table not bound, its action not allowlisted) is a targeted planning error telling the operator to bind the table or allowlist the action — never a hidden gateway call.

- **`OpenConnectorGateways`** — `Arc<RwLock<HashMap<String, Arc<GatewayHandle>>>>`, the shared map both UDTFs plan against, owned by the front-end exactly the way the KNN/FTS `DatasetRegistry` is. Handles are published only after every registration gate passes, so a failed registration never leaves a queryable handle behind.

### 4.2 `open_connector_scan` — allowlisted raw actions

- **`open_connector_scan(gateway, action_id, input_json, row_path[, connection_alias])`** invokes an explicitly allowlisted raw read action. Raw actions declare no pagination contract, so the scan is one request, one page (new `PaginationStrategy::SinglePage` — injects nothing, never advances); callers pass any paging inputs explicitly in the input JSON.

- **`raw_schema.rs` (new)** — Deterministic row type or planning error, derived purely from the discovered action output JSON Schema at the row path (in-memory; planning never contacts the gateway):
  - the row-path target must be an array of objects with declared properties, walked through nested `properties` declarations;
  - declared primitives (`string`, `integer`, `number`, `boolean`), including the `["T","null"]` union spelling, become typed nullable columns;
  - objects, arrays, wider unions, and undeclared types become opaque JSON-string columns rather than guesses;
  - columns are sorted by name so the derived schema doesn't depend on the gateway's property serialization order; dotted property names are rejected (they would silently address a nested key through the converter's path parser);
  - anything else fails with `RawRowTypeIndeterminate`, whose message recommends a built-in pack table or a source-pack contribution.

- **Deliberately narrow execution** — `RawScanProvider` uses the shared `OpenConnectorExec` with no cache (raw scans are always live: with no pagination contract there is no "complete result" to store) and no filter pushdown (no allowlist to translate against; provider-side filters go in the input JSON, SQL predicates stay in DataFusion). LIMIT and all safety bounds still apply.

### 4.3 Security policy enforcement — default-deny, pre-HTTP

- **`read_only` classification in discovery** — `DiscoveredAction`/`ActionMetadata` gain `read_only: Option<bool>` with the same `Option` discipline as `locally_executable`: a missing field stays missing and is never read as "yes".

- **The allowlist alone does not grant execution** — a raw scan requires allowlist membership **and** `read_only: Some(true)` in the discovered metadata. `Some(false)` fails as `RawActionMutating`; `None` fails as `RawActionReadOnlyUnknown`, naming the classification gap — two distinct errors so operators can tell an explicit refusal from a metadata gap. Both fire at planning time; tests assert zero gateway POSTs were recorded.

- **Deny-by-order** — the allowlist check runs before any metadata lookup, so an unallowlisted action is rejected without leaking whether it exists.

- **YAML cannot swap the relational contract** — new config tests pin that a binding declaring `action`, `row_path`, `pagination`, or `columns` fails loudly at parse time (`deny_unknown_fields`), rather than being accepted-and-ignored while the config claims a different action than the one that runs.

- **Structural guarantees unchanged** — `OpenConnectorClient::execute` stays `pub(crate)`, so allowlist/classification gating remains un-bypassable from outside the crate; pack tables continue to vouch read-only by Skardi review (they do not consult `read_only`, so gateways that don't emit the flag yet still serve stable tables — only raw scans refuse).

### 4.4 Observability

- **Scan completion event** (once per scan, in `exec.rs`): gateway, binding (or `<udtf>` for ad-hoc scans), table, action, cache hit/miss, pages fetched, rows returned, duration — emitted for live, cached, and empty scans alike.
- **Scan failure event** with the same identity fields, pages fetched, and the error display.
- Tokens, headers, request inputs, and response bodies are never logged; conversion errors already carry JSON *kinds*, not values. Client retry warns (operation, attempt, status) from milestone 2 are unchanged.
- To carry the binding name, `OpenConnectorTableProvider`/`OpenConnectorExec` thread an optional `binding` through from registration.

### 4.5 Docs — the source becomes advertised

- **`docs/open-connector.md` (new)** — configuration reference with every field and default, catalog naming, all three SQL interfaces with runnable examples, federated-join example, security model, caching/freshness semantics, bounds/retries/error behavior, compatibility & schema drift, observability. Status wording is honest per the design's marketing rule: foundation + synthetic mock pack today, provider packs one PR each behind the admission gate.
- **README supported-sources entry** — first added now that the source is actually queryable, linking the new guide.
- **Tasks spec** — milestone 4 items checked off with implementation notes; review-notes section updated to point at this PR.

## Enabling Engine Refactor

The raw UDTF derives its schema at planning time, so the scan engine could no longer require `&'static` pack definitions:

- **`exec.rs`** — `OpenConnectorExec` takes an owned `ScanTarget { table_id, action_id, pagination, source_pack_version }` instead of `&'static SourcePackTable`; `ScanTarget::from_pack_table` keeps the stable-table path a one-liner. Cache-key composition is unchanged (verified by the existing pack-version key test).
- **`json_to_arrow.rs`** — new owned `ColumnSpec` (+ `RowConverter::from_columns`); `RowConverter::new(&[FieldMapping])` keeps its signature and now delegates. A test pins that both constructors produce identical schemas and conversions.
- **`pagination.rs`** — `SinglePage` variant wired through `new`/`apply`/`advance`/`validate`.
- **`row_path.rs`** — `segments()` accessor for the schema walk.

No leaked statics, no per-query allocation tricks: UDTF invocations build ordinary owned providers.

## Front-end Wiring

- **`mod.rs`** — `register_open_connector_tables` gains `udtf_gateways: Option<&OpenConnectorGateways>`; the `ActionRegistry` is now `Arc`-shared between registration and the handle. Module docs updated; stale "execute() unused until the scan engine lands" comment removed.
- **Server** — `OptimizerRegistry` owns the `OpenConnectorGateways` map (accessor + `register_udfs` arm for `DataSourceType::OpenConnector`); the registration arm in `config.rs` passes it through. A new test proves the UDTFs are registered and fail with the targeted "gateway not registered" error rather than "function not found" when no handle exists.
- **CLI** — `new_session_context()` creates the map and registers both UDTFs alongside the existing KNN/FTS functions; `load_and_register_all`/`register_source` thread it to the `open_connector` arm.

## New Error Variants

`UdtfGatewayNotRegistered`, `ActionNotDiscovered` (names the fix: bind the table or allowlist the action), `RawActionNotAllowlisted`, `RawActionMutating`, `RawActionReadOnlyUnknown`, `RawRowTypeIndeterminate` — all surfaced as planning errors with the typed message as the user-facing diagnostic. Argument-shape problems (wrong arity, non-literal args, invalid JSON, non-object input) use `plan_err!` in the established `lance_knn` style.

## Test Coverage

147 `skardi` open_connector tests (was 125), all passing. New coverage:

- **UDTF/YAML parity** — `open_connector_query` returns the same Arrow schema and the same rendered values as `saas.ws.items`; pushes the same `min_value` Exact filter into gateway request bodies; sends the explicit connection alias header on every execute.
- **Shared cache** — a YAML-table scan's cache entry is replayed by an identical UDTF scan with zero new gateway requests.
- **Raw scan** — exactly one POST per scan; derived columns sorted and conservatively typed (`Int64` id, JSON-string `tags`); LIMIT honored; complex fields render as canonical JSON text.
- **Security, asserted via recorded gateway traffic** — unallowlisted (even though discovered via a binding), mutating (`read_only: false`), unclassified (flag absent), and schema-indeterminate actions are all rejected with their distinct messages **before any HTTP execute**; row-path syntax errors too.
- **Planning-time discovery rule** — a pack table whose action was never discovered fails with the bind-or-allowlist error and no gateway traffic.
- **Federated join** — the mock pack, through the UDTF, joined against a local CSV (tempfile + `register_csv`).
- **Argument validation** — arity, non-literal arguments, malformed JSON, malformed table IDs, unknown gateway/pack/table, missing required resource.
- **Engine refactor** — `SinglePage` injects nothing and never advances; owned-`ColumnSpec` converter equivalence; `raw_schema` unit tests for every indeterminate case (no output schema, unknown segment, non-array target, non-object items, propertyless rows, dotted property names, wide unions).
- **Registry/client** — `read_only` passthrough (absent stays absent; explicit `true` carried).
- **Config** — relational-contract override rejection (`action`/`row_path`/`pagination`/`columns`).

Plus 1 new server test (UDTF registration path). Full suites green: `skardi` 646 lib tests, server 134 lib + integration, CLI 61, doc tests; `cargo fmt` applied; `cargo clippy` clean in all touched files.

## Invariants Held (from the tasks-spec review notes)

No provider credentials in Skardi; read-only by construction (execute stays `pub(crate)`, no DML registered); pure validation shared by CLI and server; **no network I/O at query-planning time**; no `.unwrap()` in production paths.

## Out of Scope (Later Milestones)

Real provider packs — GitHub (5.1), Slack (5.2), Notion (5.3), and later waves — land one PR each through the source-pack admission gate. `CREATE EXTERNAL TABLE ... STORED AS OPEN_CONNECTOR` remains a documented future extension gated on a DDL authorization design. Registration-time read-only gating of allowlisted actions (as opposed to scan-planning-time) was deliberately not added, so gateways that don't emit `read_only` metadata yet can still register and serve stable pack tables.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -51,7 +51,10 @@ use skardi::sources::providers::{
     lance::register_lance_table,
     mongo::register_mongo_tables,
     mysql::register_mysql_tables,
-    open_connector::{OpenConnectorConfig, register_open_connector_tables},
+    open_connector::{
+        OpenConnectorConfig, OpenConnectorGateways, register_open_connector_tables,
+        register_open_connector_udtfs,
+    },
     sqlite::{
         register_sqlite_fts_udtf, register_sqlite_knn_udtf, register_sqlite_tables,
         register_vec_to_binary_udf,
@@ -478,9 +481,10 @@ impl UrlTableFactory for SkardiUrlTableFactory {
 }
 
 /// Create a new SessionContext with custom URL table support (built-in files + Lance)
-/// and the `lance_knn` / `pg_knn` UDTFs registered.
-fn new_session_context() -> (SessionContext, DatasetRegistry) {
+/// and the `lance_knn` / `pg_knn` / Open Connector UDTFs registered.
+fn new_session_context() -> (SessionContext, DatasetRegistry, OpenConnectorGateways) {
     let dataset_registry: DatasetRegistry = Arc::new(RwLock::new(HashMap::new()));
+    let open_connector_gateways = OpenConnectorGateways::default();
     let session_store = SessionStore::new();
     let factory = Arc::new(SkardiUrlTableFactory::new(
         session_store,
@@ -515,6 +519,9 @@ fn new_session_context() -> (SessionContext, DatasetRegistry) {
     register_sqlite_knn_udtf(&ctx, Arc::clone(&dataset_registry));
     register_sqlite_fts_udtf(&ctx, Arc::clone(&dataset_registry));
     register_vec_to_binary_udf(&mut ctx);
+    // Open Connector UDTFs plan against the gateway state that
+    // register_open_connector_tables fills in during ctx registration.
+    register_open_connector_udtfs(&ctx, Arc::clone(&open_connector_gateways));
 
     // Embedding UDFs (gated by feature flags, lazy model loading on first call).
     #[cfg(feature = "onnx")]
@@ -543,7 +550,7 @@ fn new_session_context() -> (SessionContext, DatasetRegistry) {
         registry.register_chunk_udf(&mut ctx);
     }
 
-    (ctx, dataset_registry)
+    (ctx, dataset_registry, open_connector_gateways)
 }
 
 /// Resolve a path string: if relative (and not remote), resolve against cwd.
@@ -732,13 +739,19 @@ async fn load_and_register_all(
     ctx_path: &Path,
     session_ctx: &mut SessionContext,
     dataset_registry: &DatasetRegistry,
+    open_connector_gateways: &OpenConnectorGateways,
 ) -> Result<LocalContextConfig> {
     let config = read_context_file(ctx_path)?;
 
     for source in &config.data_sources {
-        register_source(session_ctx, source, dataset_registry)
-            .await
-            .with_context(|| format!("Failed to register data source '{}'", source.name))?;
+        register_source(
+            session_ctx,
+            source,
+            dataset_registry,
+            open_connector_gateways,
+        )
+        .await
+        .with_context(|| format!("Failed to register data source '{}'", source.name))?;
     }
 
     Ok(config)
@@ -773,6 +786,7 @@ async fn register_source(
     session_ctx: &mut SessionContext,
     source: &LocalDataSource,
     dataset_registry: &DatasetRegistry,
+    open_connector_gateways: &OpenConnectorGateways,
 ) -> Result<()> {
     let source_type = source.source_type.to_lowercase();
 
@@ -956,6 +970,7 @@ async fn register_source(
                 source.open_connector.as_ref(),
                 source.is_read_write(),
                 source.hierarchy_level,
+                Some(open_connector_gateways),
             )
             .await
             .with_context(|| format!("Failed to register Open Connector '{}'", source.name))?;
@@ -1203,8 +1218,9 @@ async fn show_schema(
     table_filter: Option<&str>,
     out: &mut dyn Write,
 ) -> Result<()> {
-    let (mut session_ctx, dataset_registry) = new_session_context();
-    let config = load_and_register_all(ctx_path, &mut session_ctx, &dataset_registry).await?;
+    let (mut session_ctx, dataset_registry, oc_gateways) = new_session_context();
+    let config =
+        load_and_register_all(ctx_path, &mut session_ctx, &dataset_registry, &oc_gateways).await?;
 
     // Build the semantics registry from the same inputs the server uses:
     //   - the ctx-inline `description` field on each data source (fallback), and
@@ -1349,12 +1365,13 @@ fn source_name_for<'a>(
 /// data sources first. If no context file is found, run the query in a bare session with
 /// URL table support (allowing direct file/lance paths in SQL).
 async fn run_query(ctx_override: Option<PathBuf>, sql: &str) -> Result<()> {
-    let (mut session_ctx, dataset_registry) = new_session_context();
+    let (mut session_ctx, dataset_registry, oc_gateways) = new_session_context();
 
     // Try to load context file, but don't fail if not found when no explicit --ctx was given
     match resolve_ctx_path(ctx_override.as_deref()) {
         Some(ctx_path) if ctx_path.exists() => {
-            load_and_register_all(&ctx_path, &mut session_ctx, &dataset_registry).await?;
+            load_and_register_all(&ctx_path, &mut session_ctx, &dataset_registry, &oc_gateways)
+                .await?;
         }
         Some(ctx_path) if ctx_override.is_some() => {
             anyhow::bail!("Context file not found: {}", ctx_path.display());
@@ -1506,9 +1523,9 @@ async fn run_pipeline_with_params(
         .with_context(|| format!("Failed to render SQL for pipeline '{}'", pipeline_name))?;
 
     // 3. Build a SessionContext with ctx data sources registered, then execute.
-    let (mut session_ctx, dataset_registry) = new_session_context();
+    let (mut session_ctx, dataset_registry, oc_gateways) = new_session_context();
     if let Some(p) = &ctx_path_for_load {
-        load_and_register_all(p, &mut session_ctx, &dataset_registry).await?;
+        load_and_register_all(p, &mut session_ctx, &dataset_registry, &oc_gateways).await?;
     }
 
     auto_register_object_stores_from_sql(&session_ctx, &sql)?;
@@ -2860,10 +2877,15 @@ spec:
 
         #[tokio::test]
         async fn errors_without_connection_string() {
-            let (mut session_ctx, registry) = new_session_context();
-            let err = register_source(&mut session_ctx, &dynamodb_source(None), &registry)
-                .await
-                .unwrap_err();
+            let (mut session_ctx, registry, oc_gateways) = new_session_context();
+            let err = register_source(
+                &mut session_ctx,
+                &dynamodb_source(None),
+                &registry,
+                &oc_gateways,
+            )
+            .await
+            .unwrap_err();
             let msg = format!("{err:?}");
             assert!(
                 msg.contains("connection_string (endpoint URL) required"),
@@ -2873,9 +2895,9 @@ spec:
 
         #[tokio::test]
         async fn errors_without_options() {
-            let (mut session_ctx, registry) = new_session_context();
+            let (mut session_ctx, registry, oc_gateways) = new_session_context();
             let source = dynamodb_source(Some("http://localhost:8000"));
-            let err = register_source(&mut session_ctx, &source, &registry)
+            let err = register_source(&mut session_ctx, &source, &registry, &oc_gateways)
                 .await
                 .unwrap_err();
             let msg = format!("{err:?}");
@@ -2921,9 +2943,9 @@ bindings:
 
         #[tokio::test]
         async fn errors_without_connection_string() {
-            let (mut session_ctx, registry) = new_session_context();
+            let (mut session_ctx, registry, oc_gateways) = new_session_context();
             let source = open_connector_source(None, Some(VALID_CONFIG));
-            let err = register_source(&mut session_ctx, &source, &registry)
+            let err = register_source(&mut session_ctx, &source, &registry, &oc_gateways)
                 .await
                 .unwrap_err();
             let msg = format!("{err:?}");
@@ -2937,11 +2959,11 @@ bindings:
         async fn errors_with_table_hierarchy() {
             // hierarchy_level defaults to Table; the CLI must reject it with
             // a clear message, not the provider's wrapped error.
-            let (mut session_ctx, registry) = new_session_context();
+            let (mut session_ctx, registry, oc_gateways) = new_session_context();
             let mut source =
                 open_connector_source(Some("http://localhost:3000"), Some(VALID_CONFIG));
             source.hierarchy_level = HierarchyLevel::Table;
-            let err = register_source(&mut session_ctx, &source, &registry)
+            let err = register_source(&mut session_ctx, &source, &registry, &oc_gateways)
                 .await
                 .unwrap_err();
             let msg = format!("{err:?}");
@@ -2953,9 +2975,9 @@ bindings:
 
         #[tokio::test]
         async fn errors_without_typed_config() {
-            let (mut session_ctx, registry) = new_session_context();
+            let (mut session_ctx, registry, oc_gateways) = new_session_context();
             let source = open_connector_source(Some("http://localhost:3000"), None);
-            let err = register_source(&mut session_ctx, &source, &registry)
+            let err = register_source(&mut session_ctx, &source, &registry, &oc_gateways)
                 .await
                 .unwrap_err();
             let msg = format!("{err:?}");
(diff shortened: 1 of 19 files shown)


docs/open-connector.md
```diff
@@ -0,0 +1,259 @@
+# Open Connector Integration
```
abbccdda, line 1: Could we add a README that shows a live local setup to have UDTF up and running using github as a source? Similar to how we do demo README for db sources like postgres/mysql.

bakey: How about adding this after GitHub lands as a data source in the next PR?

crates/skardi/src/sources/providers/open_connector/json_to_arrow.rs
```diff
@@ -183,34 +212,34 @@ impl RowConverter {
...
                 &cells,
-                mapping,
+                spec,
                 |v| v.as_f64(),
                 fail,
             )?))),
             FieldType::Utf8 => Ok(Arc::new(StringArray::from(collect_cells(
                 &cells,
-                mapping,
+                spec,
```
abbccdda, line 244: **🔴 Correctness: JSON `null` becomes the string `"null"`, not SQL NULL** — see the `FieldType::Json` arm just below (lines 246–252):

```rust
FieldType::Json => {
    let values: Vec<Option<String>> = cells
        .iter()
        .map(|cell| cell.map(|v| v.to_string()))   // Some(Value::Null) -> Some("null")
        .collect();
    Ok(Arc::new(StringArray::from(values)))
}
```

Unlike every other arm (which routes through `collect_cells` and maps a present JSON `null` → Arrow null for nullable columns), this arm ignores `spec.nullable` and stringifies. A present-but-null cell is `Some(&Value::Null)` (because `RowPath::extract` returns `Ok(&Value::Null)` when the key exists), so `.to_string()` yields the 4-char literal `"null"`. Only a genuinely-absent key (`None`) becomes null.

This is newly hot: `open_connector_scan` types every object/array/wide-union field as `FieldType::Json` (`raw_schema::column_type`), and nullable nested fields are ubiquitous in real SaaS payloads (`assignee: null`, `user: null`). Result: the column holds `"null"` strings, so `WHERE assignee IS NULL` returns nothing and `assignee = 'null'` matches instead. No test covers the JSON-null case.

Fix: map `Value::Null` → `None` in this arm (and honor `spec.nullable` for consistency with the other types).

crates/skardi/src/sources/providers/open_connector/exec.rs
```diff
@@ -349,13 +441,14 @@ impl ScanState {
...
             self.client.execute(
-                self.table.action_id,
+                &self.target.action_id,
                 &Value::Object(input),
                 self.connection_alias.as_deref(),
             ),
         )
         .await
         .map_err(|_| self.timeout_error())??;
+        self.pages_fetched += 1;
```
abbccdda, line 456: **🟡 Minor: `pages_fetched` over-counts on timeout.** This increment runs *before* the post-fetch deadline check on line 452 that returns `ScanTimeout`. A page that arrives exactly at the deadline is counted in the failure event even though its rows are never emitted. Observability-only off-by-one — consider incrementing after the deadline check.

crates/skardi/src/sources/providers/open_connector/table_functions.rs
```diff
@@ -0,0 +1,973 @@
...
+            return Err(plan_error(OpenConnectorError::RawActionNotAllowlisted {
+                gateway,
+                action_id,
+            }));
+        }
+        // The allowlist alone does not grant execution: the discovered
+        // metadata must classify the action as a non-mutating read. Both
+        // rejections fire at planning time, before any HTTP request.
+        let meta = discovered_action(&handle, &gateway, &action_id)?;
+        match meta.read_only() {
```
abbccdda, line 268: **🟡 Note: `read_only`/allowlist are validated only at registration.** The raw-scan gate reads metadata captured at startup and planning never re-discovers (by design — no network at plan time). If an upstream action flips read→mutating after registration, `open_connector_scan` keeps executing it until re-registration. Mitigated by Open Connector's own action policy being a second enforcement boundary, but the staleness window is worth a doc line.

crates/skardi/src/sources/providers/open_connector/exec.rs
```diff
@@ -190,7 +220,26 @@ impl ExecutionPlan for OpenConnectorExec {
...
+                    // row data (conversion and row-path failures report JSON
+                    // *kinds* only).
+                    tracing::warn!(
+                        gateway = %state.gateway,
+                        binding = state.binding.as_deref().unwrap_or("<udtf>"),
+                        table = %state.target.table_id,
+                        action = %state.target.action_id,
+                        pages_fetched = state.pages_fetched,
+                        error = %e,
+                        "Open Connector scan failed"
```
abbccdda, line 239: **🟡 Note: gateway error body flows into this event.** The typed error's Display can carry a bounded (≤512-char) snippet of the gateway's *error* response (`client.rs` `terminal_reason`), which lands in this `error = %e` field and in the SQL error. Skardi never puts the token there, but a gateway that reflects request context into 4xx/5xx bodies could echo something sensitive. Bounded and error-path-only, and already documented as an accepted tradeoff — flagging for awareness.

crates/cli/src/main.rs
```diff
@@ -773,6 +786,7 @@ async fn register_source(
     session_ctx: &mut SessionContext,
     source: &LocalDataSource,
     dataset_registry: &DatasetRegistry,
+    open_connector_gateways: &OpenConnectorGateways,
```
abbccdda, line 790: one naming concern: why do we need the word gateway here? It's additional context burden to confuse with our cloud offering

abbccdda: Agreed — and the collision with our cloud "gateway" offering is a good reason to drop it entirely rather than keep it in prose.

Two things the word is doing today, worth separating:

1. **The remote process** (oomol-lab/open-connector): "gateway" only meant "a mediating front for many providers" — the API-gateway sense. "Open Connector service" (or just "Open Connector") says the same thing without the loaded term.
2. **The Skardi-side handle** — the same word names the `type: open_connector` data source: `OpenConnectorGateways`, the first UDTF arg, the `gateway = %name` log field, and several error strings. This is the real burden — one word for both the remote service and our local named handle, now also clashing with the cloud product.

Suggestion: call the remote thing the **Open Connector service** in docs, and rename the Skardi-side concept to what it actually is — a named **connection** (or `endpoint`). So `open_connector_query('<connection>', ...)`, `OpenConnectorConnections`, `connection = %name`. That kills all three collisions at once.

It's a mechanical rename across the UDTF arg name, the map type, error/​log text, and docs — happy to scope it as a follow-up if you want to keep this PR focused, or fold it in here.

crates/skardi/src/sources/providers/open_connector/exec.rs
bakey: pages_fetched counts gateway requests actually made (rate-limit budget spent), not pages emitted — a failed scan emits nothing, so the field can only mean fetches there. Moving the increment would under-report real gateway traffic in timeout diagnostics and stay inconsistent with the extraction/conversion failure paths below it. Added a comment pinning the intent.

crates/cli/src/main.rs
bakey: I would like to do it as a follow-up task.

crates/skardi/src/sources/providers/open_connector/raw_schema.rs
```diff
@@ -0,0 +1,321 @@
...
+            nullable,
+        });
+    }
+    columns.sort_by(|a, b| a.name.cmp(&b.name));
+    Ok(columns)
+}
+
+/// The declared `type` of a schema node, collapsing the `["T", "null"]`
+/// union spelling to `T`. `None` for undeclared or wider unions.
+fn schema_type(node: &Value) -> Option<&str> {
```
abbccdda, line 104: Minor: `filter_map(Value::as_str)` silently drops non-string members of a `type` array, so a malformed schema like `{"type": ["string", 123]}` collapses to a typed `Utf8` column instead of the conservative JSON fallback the rest of this module uses for anything not cleanly a declared primitive. Very low severity (requires an invalid-JSON-Schema gateway), but if you want the fallback to be total, treating any non-string element as "wider union -> Json" would keep it consistent with the documented philosophy on line 12-13.

crates/skardi/src/sources/providers/open_connector/table_functions.rs
```diff
@@ -0,0 +1,1042 @@
...
+    action_id: &str,
+) -> DFResult<ActionMetadata> {
+    handle.actions.get(action_id).cloned().ok_or_else(|| {
+        plan_error(OpenConnectorError::ActionNotDiscovered {
+            gateway: gateway.to_string(),
+            action_id: action_id.to_string(),
+        })
+    })
+}
+
```
abbccdda, line 419: Cleanup (non-blocking): this `Expr -> String` literal extractor is re-implemented per provider (sqlite / lance / mongo table-function modules have the same shape). A shared helper would keep the error wording and non-literal/null handling consistent across all UDTFs.

abbccdda (commented):
A few non-blocking observations from a correctness + cleanup pass. Security gating, UDTF arg parsing, and the shared-cache path all check out; these are minor.

crates/skardi/src/sources/providers/open_connector/table_functions.rs
```diff
@@ -0,0 +1,1042 @@
...
+            .map(|expr| literal_string("open_connector_query", expr, "connection_alias"))
+            .transpose()?;
+
+        let handle = lookup_gateway(&self.gateways, &gateway)?;
+
+        // `pack.table` resolves against the built-in pack registry only —
+        // there is no way to name an action here, so the function can execute
+        // nothing but the read actions hard-coded in Skardi's packs.
+        let Some((pack_name, table_name)) = table_id.split_once('.') else {
+            return plan_err!(
```
abbccdda, line 160: Cleanup (non-blocking): this `serde_json::from_str` + `is_object()` validation is duplicated for `open_connector_scan`'s `input_json` (around line 274). A small `parse_json_object(fn_name, raw, arg_name)` helper would remove the copy and keep the two error messages in lockstep.

BtXin (approved):

abbccdda (approved):
Approve with minor comments

PR #168, opened 2026-07-23
feat(sources): GitHub source pack for Open Connector (milestone 5.1)

bakey:
The first real provider pack: repositories, issues, issue comments, pull requests, reviews, commits, workflow runs, and releases as stable SQL tables (packs/github.rs, page-number pagination at GitHub's 100-row maximum), queryable through YAML bindings, open_connector_query, and federated joins.

Engine additions the pack required, each sanctioned by the design spec:

- Per-mapping filter Fidelity (filters.rs): Exact mappings stay fully provider-side; Inexact mappings narrow the fetch and DataFusion reapplies the predicate. issues.updated_at >= maps to GitHub's `since` as Inexact (documented "updated at or after" — a guaranteed superset); commits' strictly-after `since` is deliberately NOT mapped, because a boundary row the provider drops is unrecoverable — the mock pack's `>=` rule applied for real. Timestamp filter literals render as RFC 3339 UTC strings at every DataFusion granularity.
- Utf8ListFromObjectKey (json_to_arrow.rs): the design's `$.labels[*].name` / `$.assignees[*].login` flattening — an array of objects becomes List<Utf8> of one declared key.
- A JSON-null *parent* on a nested path is absence, not a structural failure: nullable leaves under `commit.author: null` / `issue.user: null` become SQL NULL (required ones still fail with a targeted error). Extends the previous PR's JSON-null-to-SQL-NULL rule to nested paths.
- SourcePackTable::fixed_inputs: issues/pull_requests pin `state=all` so SELECT * reads the complete collection (GitHub defaults to open items only); a pushed state predicate overrides the pin, keeping SELECT * and WHERE state='closed' consistent.

Relational caveats encoded rather than papered over: GitHub's issues endpoint returns pull requests, so the stable schema exposes the pull_request marker as nullable opaque JSON (IS NULL selects pure issues). Nullability is conservative — identity fields only. Fingerprint pins stay None like the mock pack until validated against a live gateway's discovered contracts; the redacted per-table fixtures under packs/fixtures/github/ are the build-time conversion contract.

Tests: 180 open_connector total (was 158 on main). Fixture contract suites for all 8 tables (null-bearing, null-parent, empty-list, nested, extra-field rows; empty pages keep the stable schema), bind-time validation of every contract, and end to end through the mock gateway: a 150-row two-page scan carrying state=all and the resource inputs on every request, Exact state override, Inexact since narrowing plus local re-filter (gateway ignores since entirely — the harshest legal Inexact provider — and the boundary row is kept), pull_request IS NULL / IS NOT NULL, LIMIT stopping after one page, and open_connector_query parity with the YAML-bound table.

Docs: docs/open-connector-github.md (per-table filter/limit behavior, authorization/visibility incl. the issues-returns-PRs caveat, rate limits, freshness, compatibility), open-connector.md status, README row, milestone 5.1 ticked in the tasks spec.


crates/skardi/src/sources/providers/open_connector/packs/github.rs
```diff
@@ -0,0 +1,1357 @@
...
+//!   deliberately **not** mapped, exactly like the mock pack's `>=` note.
+//! - **`issues` includes pull requests**, because GitHub's issues endpoint
+//!   does. The stable schema exposes the `pull_request` marker as an opaque
+//!   nullable JSON column: `WHERE pull_request IS NULL` selects pure
+//!   issues.
+//! - **Nullability is conservative**: only identity fields (`id`, `number`,
+//!   `sha`, `tag_name`, …) are non-null. GitHub nulls out whole objects
+//!   (`commit.author: null`, `issue.user: null`); nullable columns under
+//!   them become SQL NULL per the converter's null-parent rule.
+//! - **No fingerprint pins yet** (`expected_fingerprint: None`), same as
```
abbccdda, line 34: **On `expected_fingerprint: None` and what it means for schema-drift handling.** Flagging the operational consequence for reviewers, since this pack is the first to be exercised end-to-end.

With no fingerprint pinned, there is **no registration/startup contract check** for these tables — so a breaking upstream change (a removed required field, or a mapped field whose type changes) is caught only at **query/scan time**, as a terminal `ConversionFailed`, for whoever happens to run the query. Additive changes and drift in unmapped/nullable fields are still absorbed silently, as intended.

Worth knowing about the failure mode today, because recovery options are thin:
- **Narrowing the `SELECT` does not dodge a broken column** — `RowConverter::convert()` builds every declared column and projection is applied *after* (`exec.rs` convert → `batch.project`). So `SELECT number` still fails if `labels` drifted.
- Bindings **cannot override** column types/nullability (`deny_unknown_fields` rejects a `columns` override), so there is no client-side schema fix.
- The only escape hatch is `open_connector_scan`, and only if the action is allowlisted + `read_only` — but it is `SinglePage` and returns opaque JSON, so it is a diagnostic workaround, not a drop-in. The real fix is a pack version bump.

Not blocking — the `None` choice is reasonable pre-live-validation and is already documented here. The ask: once validated against a live gateway, **pin the fingerprints** so incompatible drift fails fast at startup/registration rather than surfacing as a surprise mid-query error for a client. Might be worth a one-line pointer to that follow-up in this comment block.

abbccdda:
## Two non-blocking design-conformance deviations (foundation follow-ups)

Surfaced while reviewing this PR, but both live in **pre-existing foundation code** (`exec.rs` caching, from #165) — not in #168's diff, so they can't be inline-anchored here. Recording them so they're not lost; happy to move to a tracking issue if that's the better home.

**1. `exec.rs:520-522` — caches LIMIT-stopped scans, which the design doc (line 551) forbids.**
The design says: *"A scan that stops early — because a LIMIT was satisfied ... — is never cached."* The code stores the truncated result when `limit_remaining == Some(0)`. It's **safe today** because `limit` is part of the cache key (`exec.rs:319`, documented at `:323-324`), so a truncated entry only ever serves a byte-identical query — the design's *intent* (never serve truncated data as complete) still holds. But the literal rule is now violated. **Recommend:** confirm this is intentional and update the design doc, since a future edit that drops `limit` from the key would silently reintroduce the exact bug the original rule was written to prevent.

**2. Cache key omits `fixed_inputs`, which design line 548 lists as a key component.**
Currently safe: `fixed_inputs` are functionally determined by `(action_id, source_pack_version)`, both already in the key. **Recommend:** a one-line code comment at the `ScanKeyParts` construction noting *why* it's safe to omit, so the invariant isn't accidentally broken when a future pack varies fixed inputs per version (or makes them binding-configurable).

crates/skardi/src/sources/providers/open_connector/packs/github.rs
```diff
@@ -0,0 +1,1357 @@
...
+    // table reads as the complete collection (a pushed `state` predicate
+    // overrides the pin).
+    fixed_inputs: &[("state", FixedValue::Str("all"))],
+    filters: &[
+        // `state` takes exactly the SQL literal (open/closed/all).
+        FilterMapping {
+            column: "state",
+            operator: Operator::Eq,
+            input_field: "state",
+            fidelity: Fidelity::Exact,
```
abbccdda: **`state` as `Exact` is only faithful inside GitHub's enum domain (`open`/`closed`/`all`).** The mapping pushes any string literal as GitHub's `state` param with no value check, and `Exact` tells DataFusion *not* to re-apply the predicate. So for an out-of-domain literal — e.g. `WHERE state = 'merged'` on PRs, a natural thing to try — correctness hinges entirely on provider behavior:

- provider **422s** on the invalid enum → scan fails terminally (acceptable);
- provider **silently ignores** it and returns its default (`open`) → rows come back with `state='open'`, DataFusion does not re-filter (Exact), and you get open rows for a `state='merged'` query — **silently wrong** (should be empty).

GitHub validates `state` and 422s today, so this is low-severity *for GitHub* — but the `Exact` claim is leaning on provider validation the contract shouldn't assume.

**Recommend classifying `state` as `Inexact`.** The `state` column is in every output row, so DataFusion re-applying `state = X` is essentially free, gives identical results for valid values, and is immune to any out-of-domain mishandling regardless of provider. Same applies to `pull_requests.state` (L419-424).

crates/skardi/src/sources/providers/open_connector/packs/github.rs
```diff
@@ -0,0 +1,1357 @@
...
+            field_type: FieldType::Utf8,
+            nullable: true,
+        },
+    ],
+    pagination: GITHUB_PAGINATION,
+    required_resources: &["owner", "repo"],
+    fixed_inputs: &[],
+    // GitHub's `status` query parameter matches status OR conclusion values
+    // interchangeably — not a faithful translation of either column.
+    filters: &[],
```
abbccdda, line 665: **Optimization left on the table (non-blocking).** The `status`-not-mapped rationale here is correct and well-reasoned. But two other `workflow_runs` filters *are* faithful GitHub params and could be pushed later: `event` (Exact — exact string match) and `head_branch` → the `branch` query param (Exact). Not needed for this PR; noting so it's captured as a follow-up rather than lost. Everything unmapped still works via local DataFusion filtering — this is purely about reducing fetched rows.

crates/skardi/src/sources/providers/open_connector/packs/github.rs
```diff
@@ -0,0 +1,1357 @@
...
+            name: "html_url",
+            path: "html_url",
+            field_type: FieldType::Utf8,
+            nullable: true,
+        },
+    ],
+    pagination: GITHUB_PAGINATION,
+    required_resources: &["owner", "repo"],
+    // Same open-by-default listing as issues; pin the complete collection.
+    fixed_inputs: &[("state", FixedValue::Str("all"))],
```
abbccdda, line 441: **Test gap: end-to-end coverage is `issues`-only.** `pull_requests` is the *only other* table carrying the `state=all` pin + an Exact `state` filter, but its pin/override path never runs through the real scan engine — only `issues` exercises `fixed_inputs` override end-to-end. Shared code, so low risk, but this is a second declared instance of the headline behavior with zero direct test. A `pull_requests` variant of `state_predicate_overrides_the_fixed_input_exactly` would close it.

abbccdda (commented):
Test-gap notes from a thorough review. All non-blocking — the pack logic is sound and well-covered; these are the paths where a declared behavior has no direct test. Posting inline so each is anchored to the relevant table/definition.

crates/skardi/src/sources/providers/open_connector/packs/github.rs
```diff
@@ -0,0 +1,1357 @@
...
+        },
+        FieldMapping {
+            name: "html_url",
+            path: "html_url",
+            field_type: FieldType::Utf8,
+            nullable: true,
+        },
+    ],
+    pagination: GITHUB_PAGINATION,
+    required_resources: &["owner", "repo", "pull_number"],
```
abbccdda: **Test gap: numeric-resource forwarding is tested only for `issue_comments` (`issue_number`).** `reviews` takes `pull_number`, the same numeric-resource path, but there's no assertion it reaches the gateway as JSON `42` rather than `"42"`. `numeric_yaml_resource_values_reach_the_gateway_as_numbers` covers `issue_comments` only — one analogous assertion for `pull_number` would cover the second consumer.

crates/skardi/src/sources/providers/open_connector/packs/github.rs
```diff
@@ -0,0 +1,1357 @@
...
+            name: "html_url",
+            path: "html_url",
+            field_type: FieldType::Utf8,
+            nullable: true,
+        },
+    ],
+    pagination: GITHUB_PAGINATION,
+    required_resources: &["owner", "repo"],
+    fixed_inputs: &[],
+    // NOTE: GitHub's commits `since`/`until` are documented as commits
```
abbccdda, line 566: **Test gap: the deliberately-unmapped commits time filter has no regression test.** The PR argues hard (here, the module docs, and the guide) that `commits` must *not* map `since`. A test asserting a `committed_at >= …` predicate on `commits` produces **no** `since` in the request body (and re-filters locally) would guard against exactly the future accidental mapping this decision protects against. Same reasoning applies to `workflow_runs.status`.

crates/skardi/src/sources/providers/open_connector/packs/github.rs
```diff
@@ -0,0 +1,1357 @@
...
+    filters: &[],
+    expected_fingerprint: None,
+};
+
+/// Workflow runs (GitHub Actions) of one repository. `conclusion` is JSON
+/// null while a run is in progress.
+static WORKFLOW_RUNS: SourcePackTable = SourcePackTable {
+    id: "github.workflow_runs",
+    action_id: "github.list_workflow_runs",
+    row_path: "$.workflow_runs",
```
abbccdda, line 580: **Test gap: the wrapped-object row shape isn't covered end-to-end.** `workflow_runs` nests rows under `$.workflow_runs` alongside a sibling `total_count`; the sole e2e scan uses the unwrapped `issues` shape. The fixture *conversion* test covers extraction, but no end-to-end scan verifies pagination/termination against a `total_count`-bearing envelope. Minor, but it's the one structurally-different response shape in the pack.

crates/skardi/src/sources/providers/open_connector/packs/github.rs
```diff
@@ -0,0 +1,1357 @@
...
+        },
+        FieldMapping {
+            name: "html_url",
+            path: "html_url",
+            field_type: FieldType::Utf8,
+            nullable: true,
+        },
+    ],
+    pagination: GITHUB_PAGINATION,
+    required_resources: &["owner", "repo"],
```
abbccdda, line 261: **Test gap: `required_resources` enforcement isn't pinned for a GitHub binding.** A binding missing `owner`/`repo` should raise `MissingResourceInput`; the null-resource rejection is tested, but the missing-key path for a GitHub table isn't asserted in this PR (may ride the foundation's generic check). A one-line negative test would make the pack's resource contract self-documenting.

crates/skardi/src/sources/providers/open_connector/packs/github.rs
```diff
@@ -0,0 +1,1357 @@
...
+                    n,
+                    if n % 2 == 1 { "open" } else { "closed" },
+                    "2026-01-01T00:00:00Z",
+                )
+            })
+            .collect()
+    }
+
+    #[tokio::test]
+    async fn full_scan_paginates_the_complete_collection() {
```
abbccdda, line 1169: **Test gap: empty-page termination isn't exercised end-to-end through a GitHub table.** This test terminates on a *short* page (50 < 100). The realistic large-repo path — a full 100-row page followed by an empty page, which GitHub requires at exact 100-row boundaries and which forces the extra request — is only covered by the `pagination.rs` unit test, not through a pack scan. A fixture of exactly 100 (or 200) rows would drive the full-then-empty path here.

powerfulZ:
From the skardi-cloud RBAC side — our Syncer (Week 3) consumes the same OC GitHub provider. Our path reads **ACL** actions (`list_repository_collaborators`, `get_repository_permission_for_user`), orthogonal to your 8 content tables, so no overlap — nice clean split. One concrete finding worth surfacing before merge:

**Pagination sends snake_case keys that the live OC gateway rejects.** `pagination.rs` (`PageNumber`) inserts `per_page`/`page` as the OC action-input keys, but OC's action-input schemas are camelCase and strict. Confirmed just now against a live gateway using your own action:

```
POST /v1/actions/github.list_repository_issues
  {"input":{"owner":"...","repo":"...","per_page":5,"page":1}}  -> HTTP 400
    "Action input does not match the action schema" (additionalProperties: "per_page")
  {"input":{"owner":"...","repo":"...","perPage":5,"page":1}}   -> HTTP 200
```

The action's own `agent.md` contract lists `perPage` / `page`. The fixture tests pass because they feed local JSON and never round-trip the gateway, so CI stays green while a live OC scan of any paginated table 400s on page 1. Fix looks like `GITHUB_PAGINATION` using `perPage` (and confirming `page` is right — it is, per the contract).

(Context: we hit the exact same camelCase-strict behavior probing OC for the ACL path on 2026-07-23 — flagging so both paths stay consistent.)

abbccdda:
> [@powerfulZ's finding above](https://github.com/SkardiLabs/skardi/pull/168#issuecomment-5068434976): pagination sends snake_case `per_page`/`page`, which the camelCase-strict live OC gateway rejects with HTTP 400.

Confirmed against the code, and flagging one **extension** beyond pagination that your ACL path wouldn't have surfaced: if OC's action-input schema is uniformly camelCase-strict, the same mismatch hits the **resource keys**, not just the page params.

- `github.issue_comments` requires `issue_number` (`packs/github.rs:336`)
- `github.reviews` requires `pull_number` (`packs/github.rs:498`)

Both ship snake_case and are actively pinned that way by `numeric_yaml_resource_values_reach_the_gateway_as_numbers`, which asserts the body contains `"issue_number":42` / `"pull_number":7`. That test passes only because it round-trips `MockGateway`, never the real gateway — same reason the pagination case stays green in CI. So a live scan of those two tables would 400 on the resource key even before pagination.

Single-word keys (`owner`, `repo`, `state`, `since`) are unaffected, which is why the symptom looks pagination-specific. Net: the fix is likely `per_page`→`perPage` **and** `issue_number`→`issueNumber` / `pull_number`→`pullNumber` (plus updating the two tests). If you can share the `agent.md` input contract for the 8 `github.*` actions, that would settle the full key list definitively.

abbccdda (commented):
No obvious gap, just echo @powerfulZ 's point for a fix.

docs/superpowers/specs/2026-07-11-open-connector-integration-tasks.md
```diff
@@ -124,7 +124,38 @@ event carrying the scan identity and error.
...
+      `updated_at >=` → `since` pushes **Inexact** and DataFusion re-applies it — verified
+      against a gateway that ignores `since` entirely; commits' strictly-after `since` is
+      deliberately NOT mapped since a dropped boundary row is unrecoverable), RFC 3339
+      rendering of timestamp filter literals, `Utf8ListFromObjectKey` for the design's
+      `$.labels[*].name` / `$.assignees[*].login` flattening, a JSON-null *parent* on a
+      nested path is absence → SQL NULL for nullable columns (GitHub `commit.author:
+      null` / `issue.user: null`), and `SourcePackTable::fixed_inputs` pinning `state=all`
+      on issues/pull_requests so `SELECT *` reads the complete collection while a pushed
+      `state` predicate overrides the pin (GitHub defaults to open-only). `issues` exposes
+      the `pull_request` marker as nullable opaque JSON (`IS NULL` → pure issues, matching
```
abbccdda: **Stale after the live-gateway fixes.** Commit f6d4519c made `issues` pure issues (the OC action filters out the PRs GitHub's raw endpoint mixes in) and **removed** the `pull_request` marker column — `github.rs` now carries a negative-space guard test (`issues_declares_no_pull_request_marker`). This 5.1 entry still describes the marker as nullable opaque JSON and lists a `pull_request IS NULL` end-to-end test (line ~150) that no longer exists. The "22 pack tests" count is also understated now (~202 open_connector total). Please reconcile this entry with the current code.

crates/skardi/src/sources/providers/open_connector/packs/fixtures/github/issues.json
```diff
@@ -0,0 +1,50 @@
+{
```
abbccdda, line 1: **Admission-gate: no schema-mismatch fixture.** The design's Source-pack Admission Gate requires "empty, null-bearing, nested, **and schema-mismatch** fixtures" per pack. These github fixtures cover empty / null-bearing / null-parent / nested / extra-field, but none exercises a type mismatch (e.g. a string where an Int is declared) at the pack contract level — that path is only covered by engine-level `json_to_arrow` tests. Consider a mismatch-bearing github fixture so the per-pack contract test asserts the targeted (column, page, row, expected, found-kind) error, satisfying the gate to the letter.

docs/open-connector/stub_gateway.py
```diff
@@ -0,0 +1,216 @@
...
+        if self.path.startswith("/v1/actions/"):
+            action_id = self.path.removeprefix("/v1/actions/")
+            if action_id in ACTIONS:
+                _, _, schema = ACTIONS[action_id]
+                return self._ok({
+                    "inputSchema": {},
+                    "outputSchema": schema,
+                    # The real gateway publishes no readOnly classification
+                    # (yet); the stub does, so the raw-scan demo can run.
+                    "execution": {"locallyExecutable": True, "readOnly": True},
```
abbccdda, line 193: **The required raw-action example is not live-exercisable.** The stub hardcodes `readOnly: True`, so `open_connector_scan` (the gate's required raw-action example, via `raw_issue_scan.yaml`) passes here — but per f6d4519c the real gateway publishes **no** read/write classification, so default-deny **refuses** raw scans live. The stable-table and built-in-UDTF examples degrade gracefully to the credential wall; this one can only run against the stub. Worth calling out explicitly next to the demo (not just in the module docs) so a reader doesn't expect the raw-scan pipeline to work against a real Open Connector gateway yet.

abbccdda:
**Review note — item 2/4 (PR description is stale).** This finding is about the PR body itself, so it can't be an inline comment.

The description still reflects the pre-live-fix state and contradicts the current branch:
- Describes `issues` exposing the `pull_request` marker as nullable opaque JSON and the `pull_request IS NULL / IS NOT NULL` tests — **removed** by f6d4519c ("issues is pure issues").
- Keeps the "GitHub's issues endpoint returns pull requests" caveat and the `state=all` pin on **pull_requests** framing tied to it.
- States "180 open_connector total (was 158 on main)" — now ~202 after f6d4519c (200) and b3456dba (202).

Per the repo's PR-hygiene rule, please refresh the description to match the current code (pure-issues, no marker column, updated test count).

abbccdda (approved):

powerfulZ:
lgtm

PR #170, opened 2026-07-23
feat(cli): reframe skardi CLI as thin HTTP client for skardi-server

BtXin:
## Summary

Rewrites `crates/cli` from an embedded-DataFusion query tool into a pure HTTP client for skardi-server. Every command now sends one request to the server and renders the JSON response — the local query engine, the alias system, and all cargo features (`candle`/`gguf`/`onnx`/`remote-embed`/`chunking`/`embedding`/`rag`) are removed. The CLI no longer depends on the `skardi` engine crate at all (deps shrink to clap + reqwest + serde + tokio), so CLI builds no longer compile the engine.

### New command surface

| Command | Endpoint |
|---|---|
| `skardi query -e/-f [--max-rows] [--table]` | `POST /query` |
| `skardi run <name> [-d JSON\|@FILE\|-] [-p NAME=VALUE ...]` | `POST /{name}/execute` (replaces aliases) |
| `skardi pipeline list\|show <name>` | `GET /pipelines`, `GET /pipeline/{name}` |
| `skardi schema` | `GET /data_source` |
| `skardi health [name]` | `GET /health[/{name}]` |
| `skardi job run\|status\|list\|cancel\|show` | jobs API (ported onto shared client) |

- Connection resolution: `--server`/`--token` flags > `SKARDI_SERVER_URL`/`SKARDI_API_TOKEN` env > `~/.skardi/config.yaml` (`kind: client` manifest) > `http://127.0.0.1:8080`; token sent as `Authorization: Bearer`.
- Output: pretty-printed JSON `data` array on stdout by default (pipes into `jq`), `--table` for aligned ASCII tables, truncation notice on stderr.
- `-p` values are JSON-first typed (`-p user_id=1` sends a number); `-d` takes inline JSON, `@file`, or stdin, with `-p` overriding individual keys.
- Exit codes: 0 success, 1 request rejected / usage error, 2 server unreachable.
- Docs: `docs/cli.md` fully rewritten with a migration table; README CLI sections updated; `docs/jobs.md` param examples updated to JSON-first syntax.

### Breaking changes (intentional, per design spec)

- `skardi alias` removed — server-side pipelines invoked by name replace aliases.
- `skardi query --ctx/--semantics/--schema` and local pipeline-YAML execution removed — run a skardi-server instead.
- `NAME:TYPE=VALUE` param syntax replaced by JSON-first values.
- All docs migrated: per-source docs (`docs/lance/`, `docs/sqlite/`, …), `docs/semantics.md`, `docs/agent_data_plane.md`, and the `demo/llm_wiki`/`demo/rag` READMEs now describe the server-based workflow, with every example verified against the real CLI/server flags and pipeline params.

Design spec: `docs/superpowers/specs/2026-07-23-cli-reframe-design.md`. Plan: `docs/superpowers/plans/2026-07-23-cli-reframe.md`.

## Test plan

- [x] 57 unit/integration tests (config precedence, param typing/merging, table rendering, ApiClient error mapping via wiremock, per-command request-shape assertions)
- [x] 2 non-ignored e2e tests for the exit-code contract (usage error → 1, `--help` → 0)
- [x] 2 `#[ignore]` e2e smoke tests for use against a live server (`cargo test -p skardi-cli --test e2e_smoke -- --ignored`)
- [x] `cargo clippy -p skardi-cli --all-targets` clean; `cargo build --workspace` green
- [x] Manual exit-code checks: bogus flag → 1, `--help` → 0, unreachable server → 2


abbccdda (commented):
Focused review of the CLI reframe (scoped to `crates/cli`, ignoring the legacy migration path).

**Verdict: no serious security vulnerability introduced** — the change shrinks the attack surface (no local credential/driver handling), TLS uses default cert verification, and reqwest strips `Authorization` on cross-origin redirects. Three real-but-minor gaps noted inline. Only the path-encoding one (finding 1) is worth fixing before merge; the other two are hardening.

crates/cli/src/commands/run.rs
```diff
@@ -0,0 +1,139 @@
...
+/// failures keeps working).
+pub async fn run(
+    client: &ApiClient,
+    name: &str,
+    data: Option<&str>,
+    param_flags: &[String],
+    table: bool,
+) -> Result<()> {
+    let body = build_body(data, param_flags)?;
+    let path = format!("/{name}/execute");
```
abbccdda: **Path segments are never URL-encoded (correctness gap).** `name` is interpolated straight into the request path, and the same pattern repeats in `pipeline.rs:27` (`/pipeline/{name}`), `health.rs:13` (`/health/{name}`), and `jobs.rs:65,75,88`. A name containing `/`, `?`, `#`, `%`, or a space produces a malformed or mis-routed request (e.g. `skardi run "a/b"` hits `/a/b/execute`). Not a privilege-escalation issue (self-targeted at the user's own server) and unlikely for identifier-style names, but a genuine bug. Suggest percent-encoding path segments (`percent-encoding` crate) consistently. This is the one finding I'd fix before merge.

crates/cli/src/commands/jobs.rs
```diff
@@ -0,0 +1,283 @@
...
+            println!("{}", serde_json::to_string_pretty(&response)?);
+        }
+    }
+
+    Ok(())
+}
+
+/// Minimal url-encoding: only escape the characters most likely to appear in
+/// a job name (`/`, `&`, `?`). Good enough for a CLI.
+fn urlencode(s: &str) -> String {
```
abbccdda: **Inconsistent + incomplete encoding.** This hand-rolled `urlencode` is applied to the `job` *query* param (line 82) but the job/run_id *path* segments (lines 65, 75, 88) are not encoded at all — so path and query are handled differently. It also only escapes `& ? /`, missing `%`, `#`, and space. If finding 1 is addressed with a proper percent-encoder, this can be deleted in favor of it.

crates/cli/src/client.rs
```diff
@@ -0,0 +1,360 @@
...
+        request = self.with_auth(request);
+
+        self.send(request, url).await
+    }
+
+    /// Attach `Authorization: Bearer <token>` when a token is configured;
+    /// otherwise leave the request untouched (no auth header at all).
+    fn with_auth(&self, request: RequestBuilder) -> RequestBuilder {
+        match &self.token {
+            Some(token) => request.bearer_auth(token),
```
abbccdda, line 167: **Bearer token sent over cleartext `http://` with no warning (hardening).** The token is attached regardless of URL scheme. The default is loopback (fine), but nothing warns if a user configures a non-loopback `http://` server with a token — the token then travels in cleartext. Suggest emitting a stderr warning when a token is set and the resolved scheme is `http` on a non-loopback host.

crates/cli/src/client.rs
```diff
@@ -0,0 +1,360 @@
...
+    /// responses to `ApiError`. `url` is the request URL, kept for
+    /// `ApiError::Connect` messages.
+    async fn send(&self, request: RequestBuilder, url: String) -> Result<Value, ApiError> {
+        let response = request.send().await.map_err(|err| ApiError::Connect {
+            url: url.clone(),
+            message: err.to_string(),
+        })?;
+
+        let status = response.status();
+        let body_text = response.text().await.map_err(|err| ApiError::Connect {
```
abbccdda: **Unbounded response body buffering (robustness).** `response.text().await` reads the entire body into memory with no cap, so a very large (or malicious) response can OOM the CLI. The server's default `max_rows` mitigates the normal case, but there's no client-side ceiling. Low severity — a `Content-Length` / read cap would be defense-in-depth.

crates/cli/src/commands/run.rs
BtXin: Fixed in 5a6eadd. Added a shared `encode_component` helper in client.rs (percent-encoding crate, encodes everything outside RFC 3986 unreserved) and applied it at all seven interpolation sites: run.rs, pipeline.rs, health.rs, and the three jobs.rs paths. A wiremock test now pins the behavior: `skardi run "a/b c"` posts to `/a%2Fb%20c/execute` as a single encoded segment.

crates/cli/src/commands/jobs.rs
BtXin: Deleted in 5a6eadd — the `job` query param now goes through the same `encode_component` used for path segments, so path and query are encoded consistently (including `%`, `#`, and space) and the run_id path segments are encoded too.

crates/cli/src/client.rs
BtXin: Implemented in 5a6eadd: `ApiClient::new` now emits a stderr warning when a token is configured and the resolved URL is plain `http://` to a non-loopback host (loopback = localhost, 127.0.0.0/8, ::1 — covered by unit tests, including the `[::1]:port` and userinfo forms).

crates/cli/src/client.rs
BtXin: Implemented in 5a6eadd, slightly stronger than the Content-Length suggestion: the body is read via `Response::chunk()` with a 256 MiB ceiling, so chunked responses that carry no Content-Length are capped too. Over-cap responses fail with a clear client-side error (tested both sides of the limit with a shrunken cap).

crates/cli/src/config.rs
```diff
@@ -0,0 +1,248 @@
...
+    let (file_server, file_token) = match file_config {
+        Some(file) => (file.server, file.token),
+        None => (None, None),
+    };
+
+    let server = flag_server
+        .or(env_server)
+        .or(file_server)
+        .unwrap_or_else(|| DEFAULT_SERVER_URL.to_string());
+    let token = flag_token.or(env_token).or(file_token);
```
abbccdda: **Empty-string value defeats precedence fallback (medium, non-blocking).**

`std::env::var(...).ok()` returns `Some("")` for an exported-but-empty var, and `.or(...)` treats `Some("")` as set. So `SKARDI_SERVER_URL=` (common when a wrapper script exports vars unconditionally) yields `server = ""` instead of falling through to the config file / default — every request URL then becomes just the path and fails confusingly. Same for an empty `--token`/`SKARDI_API_TOKEN`, which produces an empty `Authorization: Bearer ` header.

Suggest treating empty strings as unset, e.g. `.filter(|s| !s.is_empty())` on each source before the `.or()` chain.

crates/cli/src/output.rs
```diff
@@ -0,0 +1,199 @@
...
+        .enumerate()
+        .map(|(i, col)| {
+            data_rows
+                .iter()
+                .map(|row| row[i].len())
+                .max()
+                .unwrap_or(0)
+                .max(col.len())
+        })
+        .collect();
```
abbccdda, line 95: **`--table` misaligns on multibyte cell values (low, cosmetic).**

Column widths use byte length (`row[i].len()` / `col.len()`), but `format!("{cell:<width$}")` pads by Unicode-scalar count. Any non-ASCII cell (`"São Paulo"`, CJK, cyrillic) shifts the `" | "` separators out of alignment. JSON mode is unaffected. If you want it correct, compute widths with `.chars().count()` to match the padding unit.

abbccdda (commented):
Code review vs the CLI-reframe design doc, focused on security/correctness. No blockers — request paths and request/response envelope shapes all match the server API, user-controlled names/ids are percent-encoded, the token only ever travels as a Bearer header, and the response-size cap is in place. Two non-blocking robustness findings inline; #1 (config empty-string) is the one I'd fix before merge.

crates/cli/src/config.rs
BtXin: Fixed in 5c502c3: a `non_empty` filter now normalizes every level of the precedence chain (flag, env, file) — empty or whitespace-only values fall through, so `SKARDI_SERVER_URL=` resolves to the file value/default and an empty token never produces an empty `Authorization: Bearer` header. Both scenarios are unit-tested.

crates/cli/src/output.rs
BtXin: Fixed in 5c502c3: widths are now computed with `.chars().count()` for both cells and headers, matching the unit `format!("{:<width$}")` pads by. Added a test with a 10-byte/9-char cell ("São Paulo") pinning the exact aligned output. (Char count still isn't terminal display width for double-wide CJK glyphs — noted in a code comment as the known remaining limitation.)

crates/cli/src/config.rs
```diff
@@ -0,0 +1,289 @@
...
+        .or(non_empty(env_token))
+        .or(non_empty(file_token));
+
+    ClientConfig { server, token }
+}
+
+/// Treat an empty (or whitespace-only) string as unset for precedence
+/// purposes.
+fn non_empty(value: Option<String>) -> Option<String> {
+    value.filter(|s| !s.trim().is_empty())
```
abbccdda: **Nit (non-blocking): `non_empty` trims for the check but passes the untrimmed value through.**

`value.filter(|s| !s.trim().is_empty())` correctly treats a whitespace-only value as unset, but a *padded* value survives with its surrounding whitespace intact — e.g. `SKARDI_SERVER_URL=" http://x "` resolves to `" http://x "`, which then produces a malformed request URL. Unlikely input, so purely optional; `value.map(|s| s.trim().to_string()).filter(|s| !s.is_empty())` would also normalize the value it keeps.

crates/cli/src/client.rs
```diff
@@ -0,0 +1,525 @@
...
+    let host_port = authority.rsplit('@').next().unwrap_or(authority);
+    let host = if let Some(v6) = host_port.strip_prefix('[') {
+        v6.split(']').next().unwrap_or("")
+    } else {
+        host_port.split(':').next().unwrap_or("")
+    };
+
+    !(host.eq_ignore_ascii_case("localhost")
+        || host == "::1"
+        || host.parse::<Ipv4Addr>().is_ok_and(|ip| ip.is_loopback()))
```
abbccdda: **Nit (non-blocking, cosmetic): a couple of loopback forms slip past the cleartext check.**

The check recognizes `localhost`, `::1`, and `127.0.0.0/8`, but not IPv4-mapped IPv6 loopback (`http://[::ffff:127.0.0.1]`) or `0.0.0.0`, so a token sent to those over `http://` triggers the "cleartext to a non-loopback host" warning even though the traffic never leaves the machine. Effect is limited to a spurious stderr warning — no functional impact — so this is optional hardening, not a bug.

abbccdda (approved):

crates/cli/src/config.rs
BtXin: Fixed in 299dadd: `non_empty` now maps through `trim()` before the emptiness filter, so a padded value is kept but normalized — `SKARDI_SERVER_URL=" http://x "` resolves to `http://x`. Covered by a new test with padded server and token values.

crates/cli/src/client.rs
BtXin: Fixed in 299dadd: the host check is now an `is_local_host` predicate that parses both address families — it accepts `localhost`, `127.0.0.0/8`, `::1`, IPv4-mapped loopback (`[::ffff:127.x.x.x]`), and the unspecified addresses (`0.0.0.0` / `[::]`). New assertions cover the three forms you listed.

Issue #167, opened 2026-07-23
documents connector: add object-store (S3) input + predicate/prefix pushdown for scans


abbccdda:
## Context

The `documents` data source connector (`crates/skardi/src/sources/providers/documents/`) currently reads a **local directory** and re-parses the entire directory into memory on **every** `SELECT` (parsing is lazy at query time, with no caching, no streaming, and no pushdown). See the `TODO(scale)` / `TODO(pushdown)` comments in `table.rs`.

Two limitations to track here.

### 1. Add S3 / object-store input support

- Today the read path is local-filesystem only: `collect_files()` uses `std::fs::read_dir` and `parse_file()` passes a local path string to `LiteParse::parse()`.
- The `mod.rs` doc comment already advertises "directory / object-store prefix" as the intent, and `image_store` (output side) accepts `s3://` URIs but `write_image_crop()` is a **no-op** for remote stores — so the aspiration exists but nothing is implemented.
- Goal: let a source `path` be an object-store prefix (e.g. `s3://bucket/prefix/`) so cloud customers can point at their own bucket instead of having to land files on a mounted volume (PVC/hostPath) inside our deployment.
- Scope to consider: listing objects under a prefix, fetching bytes for parse, and finishing the deferred `image_store` upload path for remote stores.

### 2. Handle the full-directory scan case (predicate / prefix pushdown)

- `TableProvider::scan()` currently ignores its `_filters: &[Expr]` argument and does not implement `supports_filters_pushdown`, so DataFusion applies any `WHERE` above the scan as a separate `FilterExec`. A query like `WHERE path LIKE 'batch-a/%'` still parses the **whole** directory; even `LIMIT 1` parses everything and then truncates.
- Goal: implement `supports_filters_pushdown` for `path`-prefix predicates and push the prefix into `collect_files()` (prune the walk, ideally only `read_dir` the matching subtree). Also consider `limit` pushdown and per-file streaming (`RecordBatchStream` yielding as each file parses) so large sources don't block a tokio worker while materializing one big batch.

### Notes / references

- `crates/skardi/src/sources/providers/documents/table.rs` — `scan()` (ignored filters), `fetch_partition()` (eager parse, `TODO(scale)`/`TODO(pushdown)`)
- `crates/skardi/src/sources/providers/documents/parse.rs` — `collect_files()`, `parse_file()`, `write_image_crop()` (remote no-op)
[...]

PR #171, opened 2026-07-23
feat(documents): S3 / object-store support for source path and image_store

abbccdda:
## What

Adds S3 / object-store support to the `documents` connector (issue [#167](https://github.com/SkardiLabs/skardi/issues/167) question 1) — **design + implementation + docs** in this PR.

The source `path` can be an `s3://` prefix and image crops / page renders can be written to an `s3://` `image_store`, in any local/s3 scheme combination. When *both* are `s3://` they must be the **same bucket** (a single store / region / credential set can't serve two; cross-bucket is a follow-up).

## Implementation

- **`documents/blob.rs`** (new): `Loc` (local vs `s3://` with bucket/key) and `BlobStore` (`Local` via `std::fs`, `Remote` via `object_store`) with `list`/`get`/`put`, prefix normalization, and crop `Content-Type` inference. `BlobStore::resolve` builds the S3 client lazily so it lives on the parse thread's own runtime (no reqwest cross-runtime pool hazard).
- **`parse.rs`**: all I/O routed through `BlobStore`; inputs fed to liteparse as `PdfInput::Bytes` (`parse_input`/`screenshot_input`/`is_complex`); `list_docs` filters by glob and **excludes the `image_store` prefix** (self-ingestion guard); a non-empty listing where every object fails is now a hard error, not a silent empty result.
- **`remote_storage.rs`**: `validate_documents_configuration` (same-bucket enforcement, `aws_*` credential rejection when *either* side is `s3://`, and **rejection of an `image_store` that equals or is an ancestor of the source `path`** — local or s3 — since that would exclude every source doc via the self-ingestion guard and yield an empty scan) and `preflight_documents_s3` (bounded read check that polls only the **first** list-stream item so a large prefix isn't fully inventoried at startup + put/delete write preflight). The `config.rs` Documents arm runs both before registration.
- `object_store` added to the `skardi` crate under the `documents` feature.

## Docs

`docs/documents.md` predated this change ("`path` is local-filesystem only in v1", "s3 `image_store` records refs but bytes are not uploaded") — both passages rewritten, plus a new **"S3 / object store"** section covering the env-only credential contract, same-bucket constraint, registration preflight, prefix scoping, and self-ingestion guard. README source table updated to match. A doc comment on `require_s3_region_and_creds` records that `AWS_PROFILE` satisfies the startup check only for parity with `setup_object_store` — `object_store` never reads `~/.aws/` profiles, so profile-only setups fail loudly at the preflight call; the docs point at `aws configure export-credentials` as the working path.

## Tests

`blob` (local + `InMemory` remote round-trips, prefix scoping, content-type), `parse` (`list_docs` glob + self-ingestion exclusion, `loc_is_under`, bytes routing via the existing fixtures), `remote_storage` (same-bucket, credential-gap, image_store equal/ancestor-of-path rejection for local + s3). All green: **42 documents + 13 remote_storage**.

Coverage-review follow-up closed three untested paths:
- `parse.rs` — regression test for the wholesale-failure guard (non-empty listing where every file fails to parse → hard error, not a silent empty corpus).
- `remote_storage.rs` — extracted `write_probe_key` and `require_s3_region_and_creds` as pure helpers (behavior-preserving) so they're unit-testable without live S3 or racy env mutation; added tests for the probe-key scoping, `s3_key_prefix`, `parse_bucket`, and all four region/credential branches.

Full-suite run with `--features documents`: `skardi` 750 unit + 31 doctests, `skardi-server` 151 unit + 48 integration — 0 failures. (`skardi-cli` excluded: its default `gguf` feature has a pre-existing `llama-cpp-sys` bindgen failure on the test machine, unrelated to this branch.)

## Live e2e (real S3, us-east-1)

Ran `skardi-server --features documents` against a dedicated bucket with a nested corpus (`corpus/two_pages.pdf`, `corpus/sub/nested.pdf`, a non-matching `.txt`, and a `corpus-sibling/decoy.pdf`), `path: s3://…/corpus`, `image_store: s3://…/corpus/crops`, `image_mode: embedded`, `render_page_images: true`:

- `POST /query` returned exactly 4 rows (2 files × 2 pages) with correct markdown; the sibling-prefix decoy and the `.txt` were excluded (prefix scoping + globs against real S3 listing).
- Page renders were genuinely uploaded: 4 valid PNGs under `corpus/crops/` with `ContentType: image/png`; the `.skardi-write-probe` object was cleaned up.
- **Self-ingestion guard live**: rescanning with the crops now inside the source prefix (and matching `*.png`) still returned 4 rows.
- Negative startup cases all fail with actionable errors: cross-bucket `image_store` (same-bucket message), nonexistent bucket (read-connectivity 404), `aws_secret_access_key` smuggled into `options` with a *local* path + s3 `image_store` (the gap-closer), missing `AWS_REGION`/credentials in env, and an `image_store` on a read-only public bucket (write preflight 403 on the probe key).

**Follow-ups:** a live HTTP-mock (localstack/`httpmock`) test for the reqwest cross-runtime path; cross-bucket (cross-region/account) support.

## Review context

Design spec at `docs/superpowers/specs/2026-07-23-documents-s3-object-store-support-design.md`; the review round's inline comments are resolved.

### DIFF
--- a/crates/server/src/remote_storage.rs
+++ b/crates/server/src/remote_storage.rs
@@ -249,7 +249,10 @@ fn build_bucket_store(
         .build()
         .map_err(|e| ConfigError::S3ObjectStoreRegistrationFailed {
             name: source_name.to_string(),
-            error: format!("Failed to build S3 object store for bucket '{}': {}", bucket, e),
+            error: format!(
+                "Failed to build S3 object store for bucket '{}': {}",
+                bucket, e
+            ),
         })?;
     Ok((Arc::new(store), region))
 }
@@ -676,10 +679,7 @@ path: "s3://corpus-bucket/in/"
                 &src,
             )
             .unwrap_err();
-        assert!(
-            err.to_string().contains("same bucket"),
-            "unexpected: {err}"
-        );
+        assert!(err.to_string().contains("same bucket"), "unexpected: {err}");
     }
 
     #[test]
@@ -698,14 +698,11 @@ options:
 "#,
         );
         let err = s3
-            .validate_documents_configuration(
-                "/data/corpus",
-                Some("s3://crops-bucket/out/"),
-                &src,
-            )
+            .validate_documents_configuration("/data/corpus", Some("s3://crops-bucket/out/"), &src)
             .unwrap_err();
         assert!(
-            err.to_string().contains("must not be stored in configuration files"),
+            err.to_string()
+                .contains("must not be stored in configuration files"),
             "unexpected: {err}"
         );
     }
@@ -725,7 +722,10 @@ options:
         let err = s3
             .validate_documents_configuration("s3://corpus-bucket/in/", None, &src)
             .unwrap_err();
-        assert!(err.to_string().contains("AWS configuration"), "unexpected: {err}");
+        assert!(
+            err.to_string().contains("AWS configuration"),
+            "unexpected: {err}"
+        );
     }
 
     #[test]
--- a/crates/server/src/remote_storage.rs
+++ b/crates/server/src/remote_storage.rs
@@ -91,6 +91,28 @@ impl S3Storage {
         image_store: Option<&str>,
         source: &DataSource,
     ) -> Result<()> {
+        // Reject an `image_store` that equals or is an ancestor of the source
+        // `path`. `list_docs` excludes everything under `image_store` as a
+        // self-ingestion guard, so such overlap silently drops every source
+        // document and the scan returns an empty-but-successful result. Applies
+        // to any backend (local or s3://); `image_store` nested *inside* the
+        // source is still allowed — that's the guard's intended use.
+        if let Some(img) = image_store {
+            if loc_is_under_or_equal(path, img) {
+                return Err(ConfigError::DataSourceRegistrationFailed {
+                    name: source.name.clone(),
+                    error: format!(
+                        "documents: `image_store` ('{img}') equals or is an ancestor of the \
+                         source `path` ('{path}'). Every source document lives under `image_store` \
+                         and would be excluded from the scan (self-ingestion guard), yielding an \
+                         empty result. Point `image_store` at a location nested inside or disjoint \
+                         from `path`.",
+                    ),
+                }
+                .into());
+            }
+        }
+
         let path_is_s3 = path.starts_with("s3://");
         let image_is_s3 = image_store.is_some_and(|s| s.starts_with("s3://"));
 
@@ -134,7 +156,9 @@ impl S3Storage {
     ///
     /// - **Read connectivity** — a *prefix-aware* `list` of the source `path`
     ///   (the object `head` used for CSV/Parquet returns `NotFound` for a
-    ///   prefix). An empty-but-reachable prefix is OK; auth/network errors fail.
+    ///   prefix). Only the stream's first item is polled — a full top-level
+    ///   inventory of a large prefix is never drained at startup — so an
+    ///   empty-but-reachable prefix is OK while auth/network errors still surface.
     /// - **Write preflight** — put+delete a probe object under `image_store` so a
     ///   missing `s3:PutObject` fails loudly here rather than silently dropping
     ///   crops mid-scan.
@@ -144,6 +168,7 @@ impl S3Storage {
         image_store: Option<&str>,
         source_name: &str,
     ) -> Result<()> {
+        use futures::StreamExt;
         use object_store::PutPayload;
         use object_store::path::Path as ObjectPath;
 
@@ -152,8 +177,11 @@ impl S3Storage {
             let (store, region) = build_bucket_store(&bucket, source_name)?;
             let prefix = s3_key_prefix(path);
             let op = (!prefix.is_empty()).then(|| ObjectPath::from(prefix.as_str()));
-            store.list_with_delimiter(op.as_ref()).await.map_err(|e| {
-                ConfigError::S3ObjectStoreRegistrationFailed {
+            // Poll only the first stream item: this proves auth/network/prefix
+            // reachability without draining a full inventory of a large prefix.
+            // `None` (empty but reachable) is a pass; a first-item error fails.
+            if let Some(res) = store.list(op.as_ref()).next().await {
+                res.map_err(|e| ConfigError::S3ObjectStoreRegistrationFailed {
                     name: source_name.to_string(),
                     error: format!(
                         "documents: S3 read connectivity check failed for '{}' (region '{}'): {}\n\
@@ -161,8 +189,8 @@ impl S3Storage {
                          credentials/region are configured via the environment.",
                         path, region, e
                     ),
-                }
-            })?;
+                })?;
+            }
         }
 
         if let Some(img) = image_store.filter(|s| s.starts_with("s3://")) {
@@ -214,6 +242,17 @@ fn write_probe_key(base: &str) -> String {
     } else {
         format!("{}/.skardi-write-probe", base.trim_end_matches('/'))
     }
+}
+
+/// Whether `inner` is the same location as, or nested under, `outer`
+/// (path-segment aware, trailing slashes ignored). Used at registration to
+/// detect a documents `image_store` that would swallow the source `path`.
+/// Cross-backend pairs (local vs `s3://`) never match, mirroring `list_docs`'s
+/// `loc_is_under`.
+fn loc_is_under_or_equal(inner: &str, outer: &str) -> bool {
+    let inner = inner.trim_end_matches('/');
+    let outer = outer.trim_end_matches('/');
+    inner == outer || inner.starts_with(&format!("{outer}/"))
 }
 
 /// The key/prefix portion of an `s3://bucket/key` URI (no leading `/`).
@@ -767,6 +806,63 @@ path: "s3://corpus-bucket/in/"
             )
             .unwrap_err();
         assert!(err.to_string().contains("same bucket"), "unexpected: {err}");
+    }
+
+    #[test]
+    fn documents_rejects_image_store_equal_or_ancestor_of_path() {
+        let s3 = S3Storage::new();
+        let src = documents_source(
+            r#"
+name: "docs"
+type: "documents"
+path: "/data/corpus"
+"#,
+        );
+
+        // Local: image_store == path → rejected (all docs excluded → empty scan).
+        let err = s3
+            .validate_documents_configuration("/data/corpus", Some("/data/corpus"), &src)
+            .unwrap_err();
+        assert!(err.to_string().contains("ancestor"), "unexpected: {err}");
+        // Local: image_store is an ancestor of path → rejected.
+        let err = s3
+            .validate_documents_configuration("/data/corpus", Some("/data"), &src)
+            .unwrap_err();
+        assert!(err.to_string().contains("ancestor"), "unexpected: {err}");
+        // Local: image_store nested inside path → allowed (self-ingestion guard).
+        assert!(
+            s3.validate_documents_configuration("/data/corpus", Some("/data/corpus/crops"), &src)
+                .is_ok()
+        );
+        // A sibling that shares a name prefix but not a path segment is disjoint.
+        assert!(
+            s3.validate_documents_configuration("/data/corpus", Some("/data/corpus-2"), &src)
+                .is_ok()
+        );
+
+        // S3: image_store == path → rejected.
+        let err = s3
+            .validate_documents_configuration(
+                "s3://bucket/corpus/",
+                Some("s3://bucket/corpus/"),
+                &src,
+            )
+            .unwrap_err();
+        assert!(err.to_string().contains("ancestor"), "unexpected: {err}");
+        // S3: image_store is the bucket root (ancestor of every key) → rejected.
+        let err = s3
+            .validate_documents_configuration("s3://bucket/corpus/", Some("s3://bucket"), &src)
+            .unwrap_err();
+        assert!(err.to_string().contains("ancestor"), "unexpected: {err}");
+        // S3: image_store nested inside path → allowed.
+        assert!(
+            s3.validate_documents_configuration(
+                "s3://bucket/corpus/",
+                Some("s3://bucket/corpus/crops/"),
+                &src
+            )
+            .is_ok()
+        );
     }
 
     #[test]
--- a/crates/server/src/remote_storage.rs
+++ b/crates/server/src/remote_storage.rs
@@ -169,12 +169,7 @@ impl S3Storage {
             let bucket = parse_bucket(img, source_name)?;
             let (store, region) = build_bucket_store(&bucket, source_name)?;
             let base = s3_key_prefix(img);
-            let probe_key = if base.is_empty() {
-                ".skardi-write-probe".to_string()
-            } else {
-                format!("{}/.skardi-write-probe", base.trim_end_matches('/'))
-            };
-            let probe = ObjectPath::from(probe_key);
+            let probe = ObjectPath::from(write_probe_key(&base));
             store
                 .put(&probe, PutPayload::from_static(b"skardi-write-probe"))
                 .await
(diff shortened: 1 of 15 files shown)


docs/superpowers/specs/2026-07-23-documents-s3-object-store-support-design.md
```diff
@@ -0,0 +1,277 @@
...
+        include_globs: "*.pdf, *.docx"
+        image_mode: "embedded"             # off | placeholder | embedded
+        image_store: "s3://my-bucket/crops/"  # local dir OR s3:// prefix
+        # NO aws_* keys — rejected by S3Storage::validate_configuration.
+        # Credentials/region come from env / IAM role.
+```
+
+The `path` and `image_store` values are **independent** — either may be local or
+`s3://`, and they may live in **different buckets**. All 2×2 combinations are
+valid.
```
abbccdda: **Region/credential model can't serve two S3 buckets (blocking as written).** `setup_object_store` derives one region from `AWS_REGION`/`AWS_DEFAULT_REGION` and uses one credential set (`remote_storage.rs:135`). If `path` and `image_store` are in different regions/accounts, one gets a 301 PermanentRedirect / AccessDenied.

**Decision:** when BOTH `path` and `image_store` are `s3://`, require them to be in the **same bucket**; reject at registration otherwise. Cross-bucket (and thus cross-region) is a follow-up. This section's "they may live in different buckets / all 2×2 combinations are valid" needs to be narrowed accordingly.

docs/superpowers/specs/2026-07-23-documents-s3-object-store-support-design.md
```diff
@@ -0,0 +1,277 @@
...
+
+### Registration flow (`config.rs`, `DataSourceType::Documents` arm)
+
+Before registering the `DocumentsTable`:
+
+1. Collect the S3 buckets referenced by `path` and (if set) `options.image_store`.
+2. For each source whose `path` is `s3://`, call
+   `S3Storage::validate_configuration(source)` so the credential-rejection rule
+   applies to `documents` exactly as it does to CSV/Parquet.
```
abbccdda: **Credential-rejection gap.** `validate_configuration` bails unless `source.path` starts with `s3://` (`remote_storage.rs:58`) and only inspects `options`. So a **local `path` + `s3://` image_store** source never runs the check — `aws_*` keys smuggled into `options` are silently accepted. Run the rejection whenever *either* `path` or `image_store` is `s3://`.

docs/superpowers/specs/2026-07-23-documents-s3-object-store-support-design.md
```diff
@@ -0,0 +1,277 @@
...
+/// All local-vs-S3 I/O for the documents connector lives here.
+enum BlobStore {
+    Local,
+    Remote(Arc<dyn object_store::ObjectStore>),
+}
+
+impl BlobStore {
+    /// List every object/file under `prefix` (honoring `recursive`), returning
+    /// (loc, rel_key) pairs where rel_key is relative to the prefix.
+    async fn list(&self, prefix: &Loc, recursive: bool) -> Result<Vec<(Loc, String)>>;
```
abbccdda, line 162: **Prefix normalization.** S3 `list` is string-prefix, so `s3://b/corpus` also matches `corpus-2/…`. Normalize the prefix to a trailing `/` before listing, and define `rel_key` stripping precisely so `doc_id = blake3(rel_path)` and `path` stay consistent with the local walk (which has real directory boundaries).

docs/superpowers/specs/2026-07-23-documents-s3-object-store-support-design.md
```diff
@@ -0,0 +1,277 @@
...
+    current `read_dir` walk; for `Remote` it is `ObjectStore::list(Some(prefix))`
+    (recursive) or a delimiter-scoped list (non-recursive).
+  - `parse_file()` signature changes from `(abs_path, rel_path, opts)` to
+    `(bytes, rel_path, file_type, opts, write_store, image_loc_base)`. It calls
+    `parser.parse_input(PdfInput::Bytes(bytes))` and, for page renders,
+    `parser.screenshot_input(PdfInput::Bytes(bytes), …)`.
+  - `write_image_crop()` is replaced by `write_store.put(loc, bytes)`. The
+    `s3://` no-op branch is **deleted**; both local and S3 now perform a real
+    write. `file_type` is derived from the `rel_path` extension (unchanged
+    `file_type_for`) rather than a filesystem `Path`.
```
abbccdda, line 202: **Local routing changes — the "byte-for-byte identical" claim (§9) is inaccurate.** Making `parse_file` always take `bytes` routes local files through `PdfInput::Bytes` (content-sniff, `conversion.rs:115-125`) instead of `PdfInput::Path` (extension-based, `conversion.rs:99-113`). For a correctly-named corpus these agree, but a mislabeled file (`report.pdf` that's really DOCX) or non-default text globs now behave differently than today. Either keep the `Local` backend on the `Path` API, or drop the byte-for-byte claim and add a regression test.

Also: the OCR-Auto probe `is_complex` (`parse.rs:312`) still uses `PdfInput::Path` — it must switch to `Bytes` too (the API accepts it; the spec only names `parse_input`/`screenshot_input`).

docs/superpowers/specs/2026-07-23-documents-s3-object-store-support-design.md
```diff
@@ -0,0 +1,277 @@
...
+
+### Sync ↔ async bridge (unchanged strategy)
+
+`parse_source` already runs its blocking loop on a **dedicated OS thread that
+owns a current-thread tokio runtime** (because DataFusion calls the scan from a
+tokio worker and liteparse is async). The same runtime `block_on`s the
+`BlobStore` async calls (`list` / `get` / `put`). The `Arc<dyn ObjectStore>`
+handles are `Send + Sync`, so they are moved into that thread. `parse_source`
+gains the resolved `BlobStore`s as parameters; it stays a synchronous call from
+DataFusion's perspective.
```
abbccdda, line 218: **Cross-runtime `object_store` client usage (blocking — and the proposed test won't catch it).** The store is built at config time on the server's multi-threaded runtime, then moved into `parse_source`'s dedicated thread with a *fresh current-thread runtime* (`parse.rs:538-554`) to `block_on` `list`/`get`/`put`. `object_store::aws` wraps a `reqwest` client whose connection pool is bound to the reactor it was created on — using it from another runtime can hang ("dispatch task is gone"). The §9 `InMemory` test is synchronous and won't exercise this. Build the S3 client lazily inside the parse thread, or add a test against a real HTTP-backed mock (localstack/httpmock).

docs/superpowers/specs/2026-07-23-documents-s3-object-store-support-design.md
```diff
@@ -0,0 +1,277 @@
...
+### Write (`image_store`, only when `image_mode=embedded` or `render_page_images`)
+
+1. `execute()` resolves `options.image_store` to `(write_store, base_loc)`.
+2. `image_ref_uri` / `page_image_uri` build the per-image URI exactly as today
+   (`{image_store}/{relpath_underscored}_{id}.png`). This URI is recorded in the
+   `image_refs` / `page_image_ref` output columns.
+3. `write_store.put(loc, bytes)` writes the crop. For `Local` this is the current
+   `create_dir_all` + `std::fs::write`; for S3 it is `ObjectStore::put`.
+4. On a per-image write failure the ref is dropped and a warning logged — the
+   existing local behavior, now applied uniformly to S3.
```
abbccdda: **Self-ingestion loop.** Default globs include image extensions (`*.png, *.jpg …`, `parse.rs:90-98`). If `image_store` overlaps the source prefix (e.g. `path=s3://b/corpus/`, `image_store=s3://b/corpus/crops/`), written crops get re-listed and re-parsed on the next scan — now easy to hit since same-bucket read+write is the headline capability. Exclude the `image_store` prefix from listing (or require it to be disjoint from `path`).

**Minor:** `ObjectStore::put` won't set `Content-Type: image/png`; consumers reading crops by HTTP content-type (vs the `.png` extension) get `application/octet-stream`.

abbccdda (commented):
Reviewed the design against the actual code (`parse.rs`, `table.rs`, `remote_storage.rs`, `config.rs`) and the `liteparse` 2.4.0 source. The core is solid — `PdfInput::Bytes` exists and liteparse content-sniffs non-PDF bytes (`guess_extension_from_data`), so DOCX/XLSX/PPTX work without a filename hint, and reusing the `object_store` + env-only-creds convention is right. Inline comments below cover three potentially-blocking items and several edge cases. Follow-up commit incorporating the fixes is on the way.

docs/superpowers/specs/2026-07-23-documents-s3-object-store-support-design.md
```diff
@@ -0,0 +1,277 @@
...
+
+## 7. Reliability & error handling
+
+- **Prefix-aware connectivity check.** The existing `test_connectivity` does
+  `head(object_path)`, which is correct for a single object but returns
+  `NotFound` for a **prefix** (`s3://bucket/corpus/`). The documents registration
+  uses a `list(prefix)` with a small bound (e.g. take the first item) to verify
+  credentials + bucket reachability without requiring the prefix itself to be a
+  key. An empty-but-reachable prefix is **not** a startup error (it is a valid,
+  currently-empty corpus); auth/network/permission failures **are**.
```
abbccdda, line 280: **Write-permission failures are silent.** This connectivity check only does `list` (read). A missing `s3:PutObject` on the `image_store` bucket surfaces only mid-scan as a per-image `warn` + dropped ref (`parse.rs:394-401`) — the query *succeeds* with silently missing crops. Add a write preflight (put+delete a probe key) when `image_store` is `s3://`.

docs/superpowers/specs/2026-07-23-documents-s3-object-store-support-design.md
```diff
@@ -0,0 +1,277 @@
...
+- **Missing/unreadable source.** Local semantics are unchanged (a missing root
+  fails the scan loudly rather than looking like an empty corpus). For S3, a
+  `list` that fails on auth/permission/network errors fails the scan with a clear
+  message; a successful list returning zero objects yields zero rows.
+- **Per-file fault isolation (unchanged).** A `get` or parse failure for one
+  object is logged (`tracing::warn`) and skipped; remaining objects still produce
+  rows. This already exists for local parse errors and now also covers per-object
+  fetch failures.
```
abbccdda: **Wholesale mid-scan fetch failure returns partial data with no error.** Per-object `get` failures are logged and skipped (matching local). But S3 creds expiring *after* the initial `list` succeeds means every `get` fails → zero/partial rows with only warnings, no hard error. Local almost never hits this; S3 will. Consider failing the scan when all (or a high fraction of) objects fail.

crates/skardi/src/sources/providers/documents/parse.rs
```diff
@@ -225,46 +228,60 @@ fn glob_match(pattern: &str, name: &str) -> bool {
...
+                return false;
             }
-        };
-        for entry in entries.flatten() {
-            let path = entry.path();
-            if path.is_dir() {
-                if opts.recursive {
-                    stack.push(path);
+            if let Some(base) = image_store {
+                if loc_is_under(loc, base) {
```
bakey, line 252: When image_store equals the source prefix, or is an ancestor of it, every source entry satisfies loc_is_under and is excluded before parsing. For example, path=s3://bucket/corpus/ with image_store=s3://bucket/corpus/ returns an empty successful result despite corpus/a.pdf existing; the same happens locally. Reject equal/ancestor source-image_store relationships at registration (while still allowing image_store nested inside the source), and add local and S3 regression cases.

crates/server/src/remote_storage.rs
```diff
@@ -39,6 +39,252 @@ impl S3Storage {
...
+    ) -> Result<()> {
+        use object_store::PutPayload;
+        use object_store::path::Path as ObjectPath;
+
+        if path.starts_with("s3://") {
+            let bucket = parse_bucket(path, source_name)?;
+            let (store, region) = build_bucket_store(&bucket, source_name)?;
+            let prefix = s3_key_prefix(path);
+            let op = (!prefix.is_empty()).then(|| ObjectPath::from(prefix.as_str()));
+            store.list_with_delimiter(op.as_ref()).await.map_err(|e| {
```
bakey: This is a registration-time connectivity probe, but list_with_delimiter drains and accumulates every page under the prefix before returning. On a large documents prefix, startup performs a full top-level inventory and can incur substantial S3 request volume and memory use. Probe with a bounded list stream read (for example, list(prefix).next().await) so empty prefixes still succeed while auth/network errors remain visible.

docs/superpowers/specs/2026-07-23-documents-s3-object-store-support-design.md
abbccdda: Done — `validate_documents_configuration` now requires `path` and `image_store` to be in the **same bucket** when both are `s3://`, rejecting cross-bucket at registration (`remote_storage.rs:137-150`). §4 of the design doc was narrowed accordingly (the "2×2 combinations are valid" language is gone); cross-bucket/cross-region is a tracked follow-up. Covered by `documents_same_bucket_ok_different_bucket_rejected`.

docs/superpowers/specs/2026-07-23-documents-s3-object-store-support-design.md
abbccdda: Fixed — credential rejection now fires whenever *either* `path` or `image_store` is `s3://`, not only for an `s3://` `path`. `validate_documents_configuration` calls `reject_credential_options` as soon as any side is S3 (`remote_storage.rs:116-124`), so a local-`path` + `s3://`-`image_store` source that smuggles `aws_*` keys into `options` is now rejected. Covered by `documents_rejects_credentials_when_only_image_store_is_s3`.

docs/superpowers/specs/2026-07-23-documents-s3-object-store-support-design.md
abbccdda: Done — `normalize_prefix` (`blob.rs:54`) forces the S3 prefix to end in exactly one `/` before listing, so `s3://b/corpus` no longer matches `corpus-2/…`. `rel_key` is stripped of the normalized prefix in `push_remote_entry`, and the local walk strips `root`, so both backends yield identical `/`-separated rel keys and `doc_id = blake3(rel_path)`/`path` stay consistent. Covered by `remote_backend_prefix_is_directory_scoped`.

docs/superpowers/specs/2026-07-23-documents-s3-object-store-support-design.md
abbccdda: Both parts addressed. Every input now goes through `PdfInput::Bytes` uniformly — there is no separate `Local`-on-`Path` route — so all backends share one content-sniff path. The OCR-Auto `is_complex` probe was switched to `Bytes` as well (`parse.rs:330`), matching `parse_input` (`:344`) and `screenshot_input` (`:358`). The inaccurate "byte-for-byte identical" claim was dropped from §9; the doc now states routing is content-sniff for all inputs.

docs/superpowers/specs/2026-07-23-documents-s3-object-store-support-design.md
abbccdda: Fixed the blocking issue — the S3 `object_store` client is now built lazily *inside the parse thread* via `BlobStore::resolve`, called from `parse_source_blocking` (`parse.rs:557-568`), so the reqwest connection pool is created on the same current-thread runtime that `block_on`s `list`/`get`/`put`. The config-time store from `setup_object_store` is only used for DataFusion CSV/Parquet scans, not the documents parse. Agreed the `InMemory` tests only cover backend logic and not the reactor binding — the fix is structural (client construction moved onto the polling thread).

docs/superpowers/specs/2026-07-23-documents-s3-object-store-support-design.md
abbccdda: Both handled. Self-ingestion: `list_docs` excludes everything under the `image_store` location (`loc_is_under`, `parse.rs:251-255`) so crops written into the source prefix are never re-listed; registration additionally rejects an `image_store` equal to / ancestor of `path`. Content-Type: `put` now uses `put_opts` with `Attribute::ContentType` inferred from the key extension (`content_type_for_key`, `blob.rs:143-146`), so `.png` crops are stored as `image/png` rather than `application/octet-stream`.

docs/superpowers/specs/2026-07-23-documents-s3-object-store-support-design.md
abbccdda: Done — `preflight_documents_s3` now runs a write preflight when `image_store` is `s3://`: it puts a `.skardi-write-probe` object under the prefix and deletes it (best-effort) (`remote_storage.rs:196-214`). A missing `s3:PutObject` fails loudly at registration instead of surfacing only as per-image `warn`s and silently-dropped crops mid-scan.

docs/superpowers/specs/2026-07-23-documents-s3-object-store-support-design.md
abbccdda: Added a wholesale-failure guard: when the listing is non-empty but *every* object fails to fetch/parse (`total > 0 && ok_files == 0`), the scan now hard-errors rather than returning a silent empty/partial result — this is exactly the creds-expire-after-list case (`parse.rs:604-613`). An empty listing (`total == 0`) still returns a valid empty corpus. Covered by `all_files_failing_to_parse_is_a_hard_error`.

crates/skardi/src/sources/providers/documents/parse.rs
abbccdda: Fixed — `validate_documents_configuration` now rejects at registration any `image_store` that equals or is an ancestor of the source `path` (`remote_storage.rs:100-114`, via `loc_is_under_or_equal`), for both local and S3, while still allowing `image_store` nested *inside* the source (the guards intended use). Regression cases in `documents_rejects_image_store_equal_or_ancestor_of_path` cover local (equal / ancestor / nested-ok / sibling-with-shared-name-prefix-ok) and S3 (equal / bucket-root-ancestor / nested-ok).

crates/server/src/remote_storage.rs
abbccdda: Fixed — the read probe no longer drains the prefix. It now polls only the first item of the list stream (`store.list(prefix).next().await`, `remote_storage.rs:183`) instead of `list_with_delimiter`, so an empty-but-reachable prefix still passes while auth/network errors still surface — without a full top-level inventory or the associated request/memory cost on a large documents prefix.

bakey (approved):

BtXin:
Let's do a live e2e test in azure object storage and gcp object storage too, since we already have azure and gcp credits (we can ignore gcp if we don't already have gcp credits)

abbccdda:
Good call — filed as #178 to run the same live e2e checklist against Azure Blob Storage (we have credits) and GCS (only if we already have GCP credits, per your note). Since it's provider validation of the existing `object_store` path rather than blocking work for this PR, I'd keep it as a separate follow-up rather than expanding #171's scope.

PR #172, opened 2026-07-24
feat(sources): Slack source pack (milestone 5.2) + built-in packs as embedded YAML

bakey:
## Summary

Milestone 5.2 of the [Open Connector integration](docs/superpowers/specs/2026-07-11-open-connector-integration-tasks.md): the **Slack source pack** — conversations (channels), users, and files. Where the GitHub pack (#168) validated page-number pagination, this pack validates **cursor pagination**: Slack's `cursor` / `response_metadata.next_cursor` contract, with complete termination on both of Slack's end-of-collection spellings and loop detection against non-advancing gateways.

> **Stacked on #168** (`feature/open-connector-github-pack`): depends on its `Fidelity::Inexact`, `fixed_inputs`/`FixedValue`, the fixtures convention, and the list/null conversion rules. Merge after #168; this PR's own diff is the last commit (`37b5780`).

## Tables (`packs/slack.rs`)

| Table | Action | Pagination | Notes |
|---|---|---|---|
| `conversations` | `slack.list_conversations` | cursor, `limit` 200 | `types=public_channel,private_channel` pinned — the `state=all` move: Slack defaults to public-only, the pin makes the table read as everything the bot can see. IMs/MPIMs deliberately excluded (message-shaped, not channels). |
| `users` | `slack.list_users` | cursor, `limit` 200 | Includes bots and deleted members (`deleted = true`); `email` is scope-gated (`users:read.email`) and NULL without it. |
| `files` | `slack.list_files` | classic `page`/`count` 100 | That endpoint never adopted cursors. `user_id = X` → `userId` pushed **Inexact** per the string-push rule; `created >= X` deliberately **not** mapped — the live `slack.list_files` contract declares **no time input at all**, and its strict schema would 400 one, so `created` predicates run in DataFusion. (Per-mapping value rendering exists: `ValueFormat::Verbatim`/`Rfc3339`/`EpochSeconds`, loader `format:`, tested in `filters.rs` — retained for future packs whose contracts do take epoch-seconds inputs.) |

**No message/thread tables**, per the design's Slack caveat: Open Connector lacks complete message-cursor handling, and an incomplete message table would violate the admission gate's complete-pagination rule. Stated in the module docs and the guide; `open_connector_scan` remains the ad-hoc path for allowlisted read actions.

No required resources on any table — a pack first; the binding's `resource:` block can be omitted entirely.

## Engine addition

**`FieldType::TimestampSecondsUtc`** — Slack's `created`/`updated` are epoch **seconds**; reading them through the existing millis variant would silently produce January-1970 dates. Same Arrow type (`Timestamp(ms, UTC)`), checked ×1000, strictly integers (fractional/string values fail with their kind). Unit-tested incl. null/absent handling.

## Cursor semantics pinned end to end

- Multi-page scan: no `cursor` on page 1, the gateway's token afterwards; `limit` hint and the `types` pin ride every request.
- Termination on **both** spellings: a final page with `next_cursor: ""`, and a final page with no `response_metadata` at all.
- A gateway that repeats its cursor fails as a targeted `PaginationLoop` after exactly one repeat — never an unbounded scan.
- `LIMIT` stops cursor pagination mid-sequence with no extra request.

## Verification

32 new tests (212 open_connector total; full skardi 732 / server 137 / CLI 63 green):

- **Fixture contract tests** per table (`packs/fixtures/slack/`): nested `topic.value` / `profile.*`, Slack's empty-string convention (empty string ≠ NULL), scope-gated `email` NULL, deleted users, empty channel lists (empty ≠ NULL), epoch-seconds columns, absent keys → NULL, empty pages preserving schemas, and the bind-time completeness sweep over all three tables.
- **End-to-end via the mock gateway**: the cursor semantics above, empty workspace, `files.user_id` pushed and re-applied against a stub that ignores the parameter (the Inexact contract's hostile-provider case), multi-table binding with zero required resources, and UDTF parity (`open_connector_query('saas','slack.users','{}')` returns the bound table byte-for-byte).

## Same conventions as #168

Conservative nullability (only `id` non-null), deliberately-unmapped filters guarded with rationale, no fingerprint pins yet (same live-validation follow-up as GitHub, referenced from the module docs), docs: `docs/open-connector-slack.md` + guide/README/tasks-spec updates with visibility/scope boundaries stated per the design's marketing rule ("Slack workspace metadata", not full Slack access).

## Pack format: built-in packs are now embedded YAML assets

All three built-in packs — `mock`, `github`, and this PR's `slack` — are now **declarative YAML assets** (`packs/mock.yaml`, `packs/github.yaml`, `packs/slack.yaml`) instead of Rust statics. This adopts the design doc's illustrative pack format (the design explicitly left YAML-vs-Rust as an implementation choice) and lays the groundwork for its deferred second tier (user-authored packs loaded from a directory).

- **Explicit boundary: this is a format change, not dynamic loading.** Adding a built-in pack still requires a small Rust accessor module, a `mod` declaration, a registry entry, and a rebuild. Directory-loaded user-authored packs remain the design's deliberately deferred second tier (kept out so the format can stabilize while Skardi-internal); this refactor standardizes the format that tier will consume. The loader's module docs state the same boundary.
- **Test placement.** The per-pack contract/e2e suites stay colocated in the provider modules (`packs/github.rs` / `packs/slack.rs` are now module docs + accessor + tests, with declarations gone to YAML) — matching the repo-wide colocated-`#[cfg(test)]` convention. Happy to split them into separate test files if maintainers prefer.
- **Contract boundary unchanged.** Assets are `include_str!`-compiled into the binary, parsed once at first registry access by the new `packs/loader.rs`, and leaked into the same `&'static` shapes the scan engine has always borrowed. The engine, bindings, UDTFs, fingerprint gate, and every scan-path type are untouched; packs remain versioned, fingerprint-gated, and never user-editable configuration.
- **Strict loader.** `deny_unknown_fields` end to end (a typo'd `total_page_path` fails parsing instead of silently disabling the authoritative total), `key` is bound to `utf8_list_from_object_key`, table keys are bare names with ids derived as `<pack>.<table>`, table order is BTreeMap-deterministic, and filter `format` defaults to `verbatim` (the safe spelling).
- **Full validation in the loader, no panics.** The loader cross-validates each document before converting it (duplicate columns, filters referencing undeclared columns, duplicate mappings, required/optional resource overlap, fixed-input collisions with resources or pagination inputs) alongside the structural checks (row/error paths, converter, pagination). A malformed asset surfaces as a targeted `SourcePackAssetInvalid` at registration / UDTF setup — `pack()`, `SourcePackRegistry::builtins()`, and `register_open_connector_udtfs` are fallible — never a panic, so an AI-generated pack that slips past review becomes a startup diagnostic. A table-driven test pins every rejection message, and `builtin_assets_parse_and_validate` keeps shipped assets valid.
- **Mechanical, test-pinned migration.** The YAML was generated from the previous statics; the existing suite (fingerprint sync tests, per-table contract fixtures, every pagination/filter/drift e2e) passes unchanged. Pack modules keep their module docs and tests; declaration-site rationale moved into YAML comments. Pack docs now point at the `.yaml` definitions.

Counts after the refactor: **238 open_connector tests** (`cargo test -p skardi --lib sources::providers::open_connector`), 794 full-lib.


crates/skardi/src/sources/providers/open_connector/packs/slack.rs
```diff
@@ -0,0 +1,1343 @@
...
+    fixed_inputs: &[],
+    filters: &[
+        // Inexact per the string-push rule: user IDs are arbitrary strings,
+        // and an Exact claim would lean on the provider rejecting unknown
+        // ones. The OC input key is userId (camelCase, strict schema).
+        //
+        // No time filter is mapped: the OC list_files contract declares no
+        // ts_from/ts_to inputs, and its strict schema would 400 a request
+        // carrying one. Time predicates run in DataFusion.
+        FilterMapping {
```
abbccdda: Nit: `value_format: ValueFormat::Rfc3339` is semantically meaningless on a string column — it's a no-op only because non-timestamp scalars render identically under every `ValueFormat` (`scalar_to_json`). It works, but it leans on that invariant, and the tasks doc says `ValueFormat` is only retained "for future packs." A `ValueFormat::Verbatim`/`None` variant would document intent and be robust if someone later reworks the string-rendering path.

crates/skardi/src/sources/providers/open_connector/packs/slack.rs
```diff
@@ -0,0 +1,1343 @@
...
+    required_resources: &[],
+    optional_resources: &[],
+    // Slack lists public channels only by default; pin both channel kinds so
+    // the table reads as the complete collection the bot can see. IMs/MPIMs
+    // are message-shaped and deliberately out of a channels table. The
+    // action schema takes an array of enum strings.
+    fixed_inputs: &[(
+        "types",
+        FixedValue::StrList(&["public_channel", "private_channel"]),
+    )],
```
abbccdda: Design-alignment note (knowingly deferred, same as GitHub): with `expected_fingerprint: None` on every Slack table, the design's compatibility gate and acceptance criterion #12 ("incompatible OC action changes fail *registration* with a targeted error") aren't in force here. A breaking OC change — `nextCursor` renamed, or `types` switching from array to string — is only caught at *scan time*, not at registration. Fine for a pre-live-validation pack, but the "fail loudly at registration" guarantee genuinely isn't active for Slack yet.

docs/open-connector-slack.md
```diff
@@ -0,0 +1,139 @@
...
+— not full Slack access.
+
+## No message or thread tables (yet)
+
+Per the integration design's Slack caveat, complete message-history cursor
+handling is not yet available through Open Connector, and an incomplete
+message table would violate the source-pack admission gate's
+complete-pagination requirement. Message and thread tables land in a later
+pack version once upstream support exists; until then, allowlisted read
+actions remain reachable ad hoc through
```
abbccdda, line 141: Doc gap: the default safety bounds impose a hard ceiling worth stating here. With defaults, `files` caps at `max_pages(100) × count(100)` = 10k rows and the cursor tables at `200 × 100` = 20k rows *before the scan errors* with `ScanBoundsExceeded` — correct per the design's fail-don't-truncate rule, but a large workspace hits it as a failure, not a silent truncation. A sentence under this section (or a link to the general guide's bounds) would set expectations.

abbccdda (commented):
Reviewed the Slack pack plus the shared infrastructure it drives (filters, pagination, JSON→Arrow, exec engine, config, HTTP client, per-table resource scoping) against the design doc and tasks plan.

**Verdict: no correctness or security bugs found — this is a solid PR.** Validated the Slack-risky paths specifically: epoch-seconds timestamp scaling with distinct overflow diagnosis, cursor termination on both end spellings + `PaginationLoop` detection, `paging.pages` beating the short-page heuristic, Inexact `user_id` filter + cache-key interaction (no poisoning), and per-table resource scoping withholding `channelId` from `conversations`/`users`. Security surface is clean: token in `SecretString`/env-only, URL rejects creds+query+fragment, action IDs traversal-guarded, response bodies size-bounded, logs/errors carry no secrets or row data.

Inline comments below cover two minor nits and a design-alignment note. Two things that don't map to this PR's diff, recorded here instead:

**1. `connection_alias` header-legality (pre-existing code in `client.rs::execute`).** The alias (user-supplied via the UDTF's 4th arg or binding YAML) isn't pre-validated for control chars the way the runtime token is in `from_config`. A bad alias surfaces as an opaque `RequestBuildFailed` at scan time rather than a targeted config error. Not a vulnerability — reqwest rejects CRLF, so no header injection — just a worse error. Optional follow-up.

**2. Roadmap deviation (non-blocking, worth a line in the PR description).** The design doc puts GitHub/Jira/Notion in Milestone 1 and Slack in *Phase 2* (after Google/Discord/Feishu). This branch ships GitHub + Slack and defers Jira/Notion (tasks doc reframes them as 5.3/5.4), using Slack to validate cursor pagination, which means we should update the design spec in the first place to be consistent.

BtXin:
There might be some misalignment here, I have 2 main concerns.
1. Do we need to have a dedicated rust file per source like github or slack? Is it possible to only utilize config files to load each source pack? If we need to have a dedicated rust file per source, it would be extremely hard to dynamically dispatch each source, other wise we have to build the entire thing into the same binary.
2. Can we use yaml for source-pack configs? yaml's contract is stricter than json and more expressive in my personal opinion.

bakey:
> There might be some misalignment here, I have 2 main concerns.
> 
> 1. Do we need to have a dedicated rust file per source like github or slack? Is it possible to only utilize config files to load each source pack? If we need to have a dedicated rust file per source, it would be extremely hard to dynamically dispatch each source, other wise we have to build the entire thing into the same binary.

In design doc, use rust file or config file is an implementation choice. There are some pros for rust file. But I am OK with just config files for scalability. How about keep some critical pack as built-in pack that with rust file and some other external source packs as config files? We can load a new source pack from config files after user authorized.
> 2. Can we use yaml for source-pack configs? yaml's contract is stricter than json and more expressive in my personal opinion.

I am OK with that. Json is more readable for human and  error-resistant. But if we maintain the config by Agent, yaml is fine

bakey:
**Pros of Rust-file source packs:**

- Compile-time contract checking. Pack declarations use strongly typed enums (FieldType::TimestampSecondsUtc, PaginationStrategy::Cursor { … }, Fidelity::Inexact, ValueFormat::Verbatim, FixedValue::StrList). A typo'd variant, missing field, or type mismatch is a build error — with YAML it degrades to startup-time deserialization errors plus a hand-maintained schema-validation layer.
- Exhaustiveness protection during evolution. Adding ValueFormat::Verbatim forced every match format arm to be completed; adding total_pages_path / error_path made the compiler point at every declaration site needing initialization. YAML keys are stringly typed — drift between the engine's serde model and existing pack files can only be caught by tests, at runtime.
- Zero-cost, infallible registry. The engine borrows statically end to end (&'static str row paths, &'static [(&'static str, FixedValue)] fixed inputs). Registry construction is const data — no parsing, no failure path, the strongest form of the design's "no file discovery or network I/O at startup". YAML means owned Strings or load-time Box::leak, rippling through core type signatures.
- Tests reference declarations directly. Contract tests assert against the statics themselves (convert_fixture(&ISSUES, …), ISSUES.fields.iter().all(|f| f.name != "pull_request"), iterating SLACK_PACK.tables). No "parse YAML → look up registry" indirection, and assertion failures point at source lines.
- Free refactoring and navigation tooling. rust-analyzer renames and find-references work across engine ↔ pack declarations; YAML values are opaque strings to the IDE.
- Same review pipeline as the engine. The design defines packs as "Skardi-reviewed relational contracts" — Rust files ride the existing fmt/clippy/build/review toolchain with no separate format versioning, schema evolution policy, or custom linting.
- First-class documentation carrier. Module docs (//!) hold each design decision's rationale ("issues is pure issues", "no time filter is pushed"), and reviewers treat them as verifiable claims. YAML comments have no equivalent tooling status (no rustdoc, no docs.rs).
- An escape hatch for logic. Packs are pure data today, but if a provider ever needs computed defaults or derived inputs, Rust absorbs it without inventing YAML DSL extensions.
 
**Cons of Rust-file source packs:**

- Adding a pack requires writing and compiling Rust. Authors must know the language and the crate layout; every pack change is a recompile. A declarative format would make packs reviewable/generatable as plain data.
- Data and boilerplate interleave. A table definition is ~100 lines of struct-literal syntax (field mapping blocks, nested braces) for what is conceptually a compact table; the design's YAML sketch expresses the same contract in roughly a third of the space, easier to diff and eyeball in review.
- Higher barrier for AI/codegen workflows. A generator (e.g. the /source-pack skill) must emit syntactically valid Rust matching crate conventions, rather than emitting data validated by a schema.
- No runtime introspection or hot iteration. Trying a contract tweak means recompiling the binary; a parsed format could support a dev-mode reload or a standalone validator CLI during pack development.
- Cross-language reuse is closed off. A YAML pack format could be consumed by other tooling (docs generators, catalog UIs, external validators) without linking the Rust crate; Rust statics are only visible to the binary itself.

bakey:
Verdict: Refactor to onboard Yaml config file for each source pack

BtXin:
If contracts are test related, let's move it to test directory.

crates/skardi/src/sources/providers/open_connector/packs/github.yaml
```diff
@@ -0,0 +1,206 @@
...
+      - { name: created_at, path: created_at, type: timestamp_ms_utc, nullable: true }
+      - { name: updated_at, path: updated_at, type: timestamp_ms_utc, nullable: true }
+      - { name: pushed_at, path: pushed_at, type: timestamp_ms_utc, nullable: true }
+      - { name: html_url, path: html_url, type: utf8, nullable: true }
+
+  issues:
+    action: github.list_repository_issues
+    fingerprint: 45fbef8622d90702e536b9b2e0ade6bcc7da61f7cb654d62a90b2e36df13c64b
+    row_path: "$.issues"
+    pagination:
```
BtXin, line 40: [P1] Do not use short-page termination for this filtered action. In Open Connector v1.3.1, `list_repository_issues` fetches exactly one GitHub issues page and then applies `issues.filter(issue => issue.pull_request == null)`. Skardi stops page-number pagination whenever the resulting row count is below 100, so any non-final provider page containing a pull request silently hides every later issue. The gateway needs to expose an unfiltered continuation/total signal, or this pack needs an action/strategy whose returned page is not post-filtered. Upstream implementation: https://github.com/oomol-lab/open-connector/blob/v1.3.1/src/providers/github/runtime-issue.ts#L275-L293

crates/skardi/src/sources/providers/open_connector/packs/fixtures/github/contracts/list_repository_issues.json
```diff
@@ -0,0 +1,129 @@
...
+            }
+          },
+          "pull_request": {
+            "type": "object",
+            "properties": {},
+            "additionalProperties": true,
+            "description": "A GitHub API object."
+          }
+        },
+        "additionalProperties": true
```
BtXin, line 124: [P1] This fingerprinted schema permits additional properties but does not declare `created_at`, `updated_at`, or `closed_at`, even though `github.yaml` exposes those columns and pushes `updated_at` through `since`. Removing or changing any of them leaves the fingerprint unchanged, so registration succeeds and drift appears later as NULLs, empty filtered results, or conversion failures. The same gap affects other GitHub tables and most `slack.files` columns. Please make the captured contract cover every mapped path/type, or add a pack-specific canonical contract and a test proving every stable column is fingerprint-covered.

crates/skardi/src/sources/providers/open_connector/packs/loader.rs
```diff
@@ -0,0 +1,650 @@
...
+        PaginationStrategy::Cursor {
+            cursor_param,
+            page_size_param,
+            ..
+        } => std::iter::once(cursor_param)
+            .chain(page_size_param)
+            .collect(),
+        PaginationStrategy::SinglePage => Vec::new(),
+    };
+    for (key, _) in table.fixed_inputs {
```
BtXin, line 263: [P2] Validate the entire request-input namespace, not only fixed inputs. A filter input can currently equal `page`, `perPage`, a cursor key, or a declared resource; pagination is applied after filters in `exec.rs`, so it can overwrite an Exact pushed predicate that DataFusion will not reapply, returning incorrect rows. Also reject duplicate pagination parameter names and zero page sizes. Please add targeted loader cases for these collisions/invariants.

crates/skardi/src/sources/providers/open_connector/packs/loader.rs
```diff
@@ -0,0 +1,650 @@
...
+    Str(String),
+    StrList(Vec<String>),
+}
+
+impl FixedValueDoc {
+    fn into_fixed(self) -> FixedValue {
+        match self {
+            Self::Bool(v) => FixedValue::Bool(v),
+            Self::Int(v) => FixedValue::Int(v),
+            Self::Float(v) => FixedValue::Float(v),
```
BtXin, line 496: [P2] Reject non-finite fixed floats here. YAML accepts `.nan`, `.inf`, and `-.inf`; this conversion preserves them as `FixedValue::Float`, while `serde_json::Value::from(f64)` later converts them silently to JSON `null`. That bypasses the loader's promised startup diagnostic and sends a different value at query time. Check `is_finite()` during parsing and cover all three spellings.

crates/skardi/src/sources/providers/open_connector/packs/slack.rs
```diff
@@ -0,0 +1,1076 @@
...
+            "saas",
+            &gateway.url,
+            Some(&config),
+            false,
+            HierarchyLevel::Catalog,
+            Some(&gateways),
+        )
+        .await
+        .expect("gateway registration succeeds");
+        register_open_connector_udtfs(&ctx, gateways);
```
BtXin: [P3] `register_open_connector_udtfs` now returns a `Result`, but this setup helper ignores it and emits an `unused_must_use` warning. Handle it with `expect`/`?` here and in the equivalent GitHub and `table_functions.rs` helpers so setup failures are explicit and the test build remains warning-free.

docs/open-connector.md
```diff
@@ -253,19 +255,39 @@ calls (health, discovery) retry `429`/transient `5xx` with capped
...
-schema). An Open Connector upgrade that changes an action incompatibly
-fails registration with a targeted error instead of silently changing a
-table's schema; additive upstream fields are ignored. Bindings may pin
-`source_pack_version` so a Skardi upgrade cannot silently change a bound
-table's schema either.
+Each pack table pins the full relational contract and, once captured from
+a live gateway, an expected action-contract fingerprint (a canonicalized
+BLAKE3 hash of the discovered output schema; the slack pack is pinned
+against v1.3.1 with the captured schemas committed next to the pack, the
+github pack's pins are a pending follow-up). At registration a pinned
```
BtXin: [P3] This is stale in the current head: `github.yaml` now pins fingerprints for all eight GitHub tables. Please update this sentence so the compatibility guide does not describe the completed pinning work as pending.

BtXin (changes requested):
Requesting changes for two correctness issues in the effective source-pack contracts and two validation gaps in the new YAML loader:

1. GitHub issue scans can silently truncate because Open Connector filters pull requests after fetching a provider page, while Skardi treats the shorter filtered page as end-of-collection.
2. Several pinned output schemas leave stable pack columns under `additionalProperties`, so those fields can drift without changing the fingerprint.
3. The loader does not reject all collisions in the action-input namespace.
4. YAML non-finite floats pass validation and later become JSON `null`.

Smaller cleanup is noted inline for ignored fallible registration results and stale compatibility documentation.

Architecture note: moving declarations to embedded YAML does not yet provide dynamic pack loading. Every new pack still needs a Rust module, `mod.rs` and registry wiring, and a rebuild. If runtime/directory-loaded packs are intentionally deferred, the PR description should keep that boundary explicit; otherwise this does not yet resolve the dynamic-dispatch concern. The contract suites also remain in the large provider modules rather than separate test modules/files.

crates/skardi/src/sources/providers/open_connector/filters.rs
```diff
@@ -155,16 +181,29 @@ fn scalar_to_json(literal: &ScalarValue) -> Option<Value> {
...
-        ScalarValue::TimestampNanosecond(Some(v), _) => {
-            Some(rfc3339(DateTime::from_timestamp_nanos(*v)))
-        }
+        // Verbatim arms return None on purpose: the mapping declared no
+        // timestamp spelling, so the predicate stays local rather than
+        // being pushed in a guessed format.
+        ScalarValue::TimestampSecond(Some(v), _) => match format {
+            ValueFormat::Verbatim => None,
+            ValueFormat::Rfc3339 => DateTime::from_timestamp(*v, 0).map(rfc3339),
+            ValueFormat::EpochSeconds => Some(Value::from(*v)),
```
abbccdda, line 190: This ships the per-mapping epoch-seconds rendering (`ValueFormat::EpochSeconds`, exposed as `format: epoch_seconds` by the loader) that the PR description still lists as *deferred* — "the filter engine renders timestamp literals as RFC 3339 only — recorded as a candidate for per-mapping value rendering." That note is now stale: the rendering exists and is tested here.

The actual reason `files.created` isn't pushed is that `slack.list_files` declares no time input at all (correctly stated in the slack.rs module doc and the slack.yaml comment). Worth updating the PR description to cite that reason so it doesn't read as blocked on missing value rendering.

crates/skardi/src/sources/providers/open_connector/packs/loader.rs
abbccdda: +1. The same fix should also cover two collisions this check misses today: a filter `input` equal to a `fixed_inputs` key, and two filters sharing one `input_field`. `validate_table` only iterates `table.fixed_inputs` (loader.rs:181-193), so a filter input is never checked against fixed inputs or against another filter's input — both hit the same "ambiguous request value" hazard as the pagination/resource cases you flagged. Worth folding into the same namespace check and test matrix.

bakey: Commit 1 — 3672fb1 fix(sources): validate the whole request-input namespace in the loader

The request-input namespace is shared by four writers — resources, fixed inputs, translated filter inputs, and pagination parameters — and exec.rs injects pagination last. A filter input named page, perPage, or cursor would therefore be silently overwritten after translation; if that filter was classified Exact, DataFusion never re-applies the predicate, and the scan returns wrong rows with a green status. validate_table now rejects, at asset-load time:

a filter input colliding with a pagination parameter — the error message names the cause ("pagination input, which is applied last and would overwrite the pushed predicate");
a filter input colliding with a declared resource (the pushed predicate would clobber the binding's resource value);
a pagination declaration reusing one name for two parameters (page_input == page_size_input, or a cursor strategy whose page-size input equals the cursor input);
zero page sizes (always for page_number; for cursor when a page-size input is declared).
Five new table-driven loader cases pin each rejection's targeted message.

Commit 2 — 9ace4e9 fix(sources): reject two filter mappings sharing one input field

Follow-up covering the two collisions the first pass missed — with one accepted and one deliberately rebutted:

Accepted: two filters sharing one input_field are now rejected at load. Notably this was not a wrong-rows hazard at runtime — the scan-time claimed_inputs guard already keeps the later predicate local, pinned by duplicate_action_input_keeps_later_predicate_local. But statically, which predicate gets pushed depends on the query's predicate order, and that ambiguity in a pack declaration is an authoring mistake — worth failing at load, especially for generated packs.
Rebutted: a filter input equal to a fixed_inputs key stays legal on purpose. That collision is the override mechanism: a pushed predicate replacing the complete-collection pin (the GitHub state=all pattern, exercised end-to-end by pull_requests_state_pin_and_override_run_end_to_end). Rejecting it would outlaw the filter pushdown both real packs rely on.
The validation comment in validate_table now records both decisions — why the fixed-input overlap is deliberately legal and why the shared-input case is rejected despite being runtime-safe — so future reviewers (and skill-generated packs) don't have to re-derive them.

Verification: six table-driven rejection cases in total across both commits; 797 skardi lib tests green; all three built-in pack assets pass the tightened validation unchanged.

crates/skardi/src/sources/providers/open_connector/packs/fixtures/github/contracts/list_repository_issues.json
bakey: Valid finding, and it points at an inherent boundary of the fingerprint mechanism rather than a fixable capture bug. Of the two remedies proposed, the first — widening the captured contract to cover every mapped path — is mechanically impossible on the pack side: the pin is compared against the live-discovered schema at registration, so hand-editing the captured file would immediately fail registration with ActionContractMismatch. We therefore implemented the second: a pack-specific coverage check proving exactly which stable columns are (and are not) fingerprint-covered.

1. Coverage computation + per-table pinning. A new testutil::fingerprint_uncovered_columns helper walks each mapped column's dotted path through the captured contract's row-item schema. Both packs gain a fingerprint_coverage_gap_is_pinned test asserting each table's exact uncovered-column set as a literal:

github: issues pins precisely the columns the review named (created_at, updated_at, closed_at); repositories pins 8 uncovered columns, pull_requests 6, commits 4, workflow_runs 2, issue_comments/reviews 1 each — and releases pins the empty set.
slack: conversations and users pin empty sets — the normalizing executors declare every emitted field, and these two empty pins double as proof the walker itself works. files pins 8 uncovered columns (raw Slack passthrough plus aliases), as the review anticipated.
Any future change — upstream declaring more fields, or a mapping change on our side — now fails a test and forces a conscious decision. If upstream ever completes its declarations, the pinned sets shrink and the test surfaces the good news too.

2. Documented boundary and consequences. The integration guide's compatibility section now states: the gate's coverage is exactly the upstream declaration; mapped columns riding additionalProperties passthrough sit outside the fingerprint, and drift there surfaces at scan time under the conversion rules instead — a shape change fails loudly (ConversionFailed with column/page/row identity), while a removed nullable field reads as NULL. Each pack pins its uncovered set in a test, so the gap is a reviewed fact rather than an implicit one.

3. On the updated_at → since pushdown specifically: that mapping is Inexact, so DataFusion unconditionally re-applies the predicate locally — even if upstream silently changed since semantics, the worst case is a superset being trimmed back, never wrong rows.

Truly closing the gap requires upstream schema declarations to become complete (an Open Connector change); when that lands, re-capturing and re-pinning shrinks the pinned sets automatically.

Verification: 799 skardi lib tests green; the two empty-set pins (slack conversations/users, github releases) prove the mechanism end to end.

crates/skardi/src/sources/providers/open_connector/packs/github.yaml
bakey: The upstream OC hardcoded the post filter for this action list_repository_issues, which leads to we could not conduct a perfect and elegant solution to fetch exactoly records. 

**Option 1** — Defer the issues table until upstream provides a termination signal.
Remove issues from the github pack and document the absence, exactly the way Slack message/thread tables are deferred under the admission gate's complete-pagination requirement ("complete terminating pagination, or the table doesn't ship").
Pros: strictly honest — no configuration of Skardi can silently return partial data; the admission gate keeps its meaning.
Cons: removes the pack's flagship table; sizable churn (issues is the primary e2e vehicle in the pack tests — roughly a dozen tests re-anchor onto pull_requests, whose executor is raw passthrough and unaffected); docs/demo updates; arguably a pack-version bump if we consider the merged 5.1 contract released.

**Option 2** — Fix it upstream in Open Connector, keep the table (recommended).

Contribute a small PR to oomol-lab/open-connector restoring the lost bit, in one of two shapes:
have the executor report the raw page length alongside the filtered rows (e.g. pageInfo: { fetched: <raw count> } in the output, declared in outputSchema), or
stop post-filtering and return the raw mixed page with the pull_request marker intact, leaving exclusion to callers.

BtXin: Let's open the PR to open-connector, and merge this PR for now.

BtXin (approved):

PR #177, opened 2026-07-31
feat(sources): Notion source pack (milestone 5.3)

bakey:
## Summary

Milestone 5.3: the **Notion source pack** — `users`, `pages`, `data_sources`, `block_children` as stable SQL tables over Open Connector's `notion.*` read actions. First pack authored end-to-end as an embedded YAML asset (`packs/notion.yaml`) under the validating loader, and the first produced by the `/source-pack` skill's full workflow (live reconciliation → design → implementation → self-review).

### Wire contract (live-reconciled)

Raw passthrough of the Notion API: rows verbatim under `$.results` (no normalization, **no post-pagination filtering** — verified in the executor source), Notion's native cursor envelope (`$.next_cursor`, null terminates; `has_more` redundant and deliberately unused), camelCase strict inputs (`startCursor`/`pageSize`/`blockId`).

### Design decisions (module doc carries the rationale)

- **`pages`/`data_sources` are two pinned views of `notion.search`**: required `query` pinned to `""` + per-table object `filter` pins — Notion's spelling for "the complete visible listing" (the `state=all` move). The object pin required this milestone's one engine extension: **`FixedValue::Json`** (loader-parsed & leaked, non-finite-float guard extended to nested values, backward-compatible).
- **No filter pushdown anywhere** — search's free-text relevance `query` maps to no SQL predicate faithfully; the other actions declare no filter inputs. A wire guard test pins exactly-the-declared-inputs.
- **No `rows` table yet** — dynamic property maps stay opaque JSON pending the design's binding-time schema freeze for `query_data_source`; `block_children` excludes the type-keyed payload; `users` excludes `person.email` (capability-gated, privacy).
- **Fingerprints pinned from live capture**; `pages`/`data_sources` share `notion.search`'s pin. The coverage-gap pin records that search declares an **empty** results-item schema — both search tables' columns ride `additionalProperties` passthrough, outside the fingerprint gate, drift surfacing at scan time.

### Live verification

All four tables register against a live gateway (pins match discovery); every scan reaches the credential wall (403 `authorization_failed`, never `invalid_input`), and the gateway's own run log shows the exact pinned wire inputs, e.g. `{"filter":{"property":"object","value":"page"},"query":"","pageSize":100}`.

### Verification

**260 open_connector tests (14 new) / 817 lib total** (`cargo test -p skardi --lib sources::providers::open_connector`): four fixture contract suites + the schema-mismatch fixture, fingerprint sync + coverage pins, drift-refusal e2e, and per-declaration e2e — two-page users cursor with row identity, per-table search filter pins, `blockId` forwarding + pre-HTTP required-resource enforcement, LIMIT early-stop, UDTF parity. clippy baseline unchanged; `cargo doc` clean. Docs: `docs/open-connector-notion.md`, guide status, spec 5.3 ticked with the counting command.

### DIFF
--- a/crates/skardi/src/sources/providers/open_connector/packs/fixtures/notion/users.json
+++ b/crates/skardi/src/sources/providers/open_connector/packs/fixtures/notion/users.json
@@ -1,0 +1,17 @@
+{
+  "object": "list",
+  "results": [
+    {
+      "object": "user", "id": "u-0001", "name": "Ada Lovelace",
+      "type": "person", "avatar_url": "https://img.example/ada.png",
+      "person": { "email": "redacted@example.test" }
+    },
+    {
+      "object": "user", "id": "u-0002", "name": null,
+      "type": "bot", "avatar_url": null,
+      "bot": { "owner": { "type": "workspace" } }
+    },
+    { "object": "user", "id": "u-0003", "type": "person" }
+  ],
+  "next_cursor": null, "has_more": false
+}
--- a/crates/skardi/src/sources/providers/open_connector/packs/fixtures/notion/users.json
+++ b/crates/skardi/src/sources/providers/open_connector/packs/fixtures/notion/users.json
@@ -2,16 +2,46 @@
   "object": "list",
   "results": [
     {
-      "object": "user", "id": "u-0001", "name": "Ada Lovelace",
-      "type": "person", "avatar_url": "https://img.example/ada.png",
-      "person": { "email": "redacted@example.test" }
+      "object": "user",
+      "id": "00000000-0000-4000-8000-000000000001",
+      "name": "Redacted Name",
+      "avatar_url": null,
+      "type": "person",
+      "person": {
+        "email": "redacted@example.test",
+        "email_verified": true
+      }
     },
     {
-      "object": "user", "id": "u-0002", "name": null,
-      "type": "bot", "avatar_url": null,
-      "bot": { "owner": { "type": "workspace" } }
+      "object": "user",
+      "id": "00000000-0000-4000-8000-000000000002",
+      "name": "Redacted Name",
+      "avatar_url": "https://img.example/redacted.png",
+      "type": "bot",
+      "bot": {}
     },
-    { "object": "user", "id": "u-0003", "type": "person" }
+    {
+      "object": "user",
+      "id": "00000000-0000-4000-8000-000000000003",
+      "name": "Redacted Name",
+      "avatar_url": null,
+      "type": "bot",
+      "bot": {
+        "owner": {
+          "type": "workspace",
+          "workspace": true
+        },
+        "workspace_name": "Redacted Workspace",
+        "workspace_id": "00000000-0000-4000-8000-000000000004",
+        "workspace_limits": {
+          "max_file_upload_size_in_bytes": 5242880
+        }
+      }
+    }
   ],
-  "next_cursor": null, "has_more": false
+  "next_cursor": null,
+  "has_more": false,
+  "type": "user",
+  "user": {},
+  "request_id": "00000000-0000-4000-8000-000000000005"
 }
--- a/crates/skardi/src/sources/providers/open_connector/packs/loader.rs
+++ b/crates/skardi/src/sources/providers/open_connector/packs/loader.rs
@@ -107,11 +107,15 @@ fn convert_table(
         // spelling for them — FixedValue::to_json would silently send null
         // at query time, bypassing the startup diagnostic this loader
         // promises. Finite-only, enforced here.
-        if let FixedValueDoc::Float(v) = &value
-            && !v.is_finite()
-        {
+        let non_finite = match &value {
+            FixedValueDoc::Float(v) => (!v.is_finite()).then(|| v.to_string()),
+            FixedValueDoc::Json(v) => first_non_finite(v),
+            _ => None,
+        };
+        if let Some(shown) = non_finite {
             return Err(format!(
-                "{id}: fixed input '{key}' is {v}, which has no JSON spelling;                  pin a finite number"
+                "{id}: fixed input '{key}' contains {shown}, which has no JSON \
+                 spelling; pin finite numbers only"
             ));
         }
         fixed_inputs.push((leak_str(key), value.into_fixed()));
@@ -333,6 +337,21 @@ fn convert_filter(doc: FilterDoc) -> FilterMapping {
     }
 }
 
+/// First non-finite float inside a JSON value, if any. serde_json parses
+/// no non-finite literals itself, but serde_yaml's `.nan`/`.inf` arrive
+/// through the untagged `Json` variant as f64s that `Number::from_f64`
+/// silently turns into null on write — the same trap as bare floats.
+fn first_non_finite(value: &serde_json::Value) -> Option<String> {
+    match value {
+        serde_json::Value::Number(n) => {
+            n.as_f64().filter(|f| !f.is_finite()).map(|f| f.to_string())
+        }
+        serde_json::Value::Array(items) => items.iter().find_map(first_non_finite),
+        serde_json::Value::Object(map) => map.values().find_map(first_non_finite),
+        _ => None,
+    }
+}
+
 fn leak_str(s: String) -> &'static str {
     Box::leak(s.into_boxed_str())
 }
@@ -490,6 +509,10 @@ enum FixedValueDoc {
     Float(f64),
     Str(String),
     StrList(Vec<String>),
+    /// Object-shaped inputs (e.g. Notion's search `filter`). Captured as a
+    /// JSON value so nesting is unrestricted; converted with the same
+    /// finite-floats guard as the scalar variants.
+    Json(serde_json::Value),
 }
 
 impl FixedValueDoc {
@@ -500,6 +523,7 @@ impl FixedValueDoc {
             Self::Float(v) => FixedValue::Float(v),
             Self::Str(v) => FixedValue::Str(leak_str(v)),
             Self::StrList(v) => FixedValue::StrList(leak_str_slice(v)),
+            Self::Json(v) => FixedValue::Json(Box::leak(Box::new(v))),
         }
     }
 }
@@ -560,6 +584,7 @@ mod tests {
             ("mock.yaml", include_str!("mock.yaml")),
             ("github.yaml", include_str!("github.yaml")),
             ("slack.yaml", include_str!("slack.yaml")),
+            ("notion.yaml", include_str!("notion.yaml")),
         ] {
             // parse_pack performs the full structural + cross-field
             // validation pass itself; parsing IS the gate.
--- a/crates/skardi/src/sources/providers/open_connector/packs/loader.rs
+++ b/crates/skardi/src/sources/providers/open_connector/packs/loader.rs
@@ -106,19 +106,32 @@ fn convert_table(
         // YAML happily parses `.nan` / `.inf` / `-.inf`, but JSON has no
         // spelling for them — FixedValue::to_json would silently send null
         // at query time, bypassing the startup diagnostic this loader
-        // promises. Finite-only, enforced here.
-        let non_finite = match &value {
-            FixedValueDoc::Float(v) => (!v.is_finite()).then(|| v.to_string()),
-            FixedValueDoc::Json(v) => first_non_finite(v),
-            _ => None,
+        // promises. Finite-only, enforced here. The nested case CANNOT be
+        // checked after the fact on a `serde_json::Value`: serde_json's
+        // f64 visitor maps a non-finite to `Value::Null` during
+        // deserialization (`Number::from_f64(...).map_or(Value::Null, …)`),
+        // so the loss has already happened by then — which is why the
+        // `Json` variant captures `serde_yaml::Value` and converts here,
+        // where the non-finite is still observable.
+        let fixed = match value {
+            FixedValueDoc::Float(v) if !v.is_finite() => {
+                return Err(format!(
+                    "{id}: fixed input '{key}' contains {v}, which has no JSON \
+                     spelling; pin finite numbers only"
+                ));
+            }
+            FixedValueDoc::Bool(v) => FixedValue::Bool(v),
+            FixedValueDoc::Int(v) => FixedValue::Int(v),
+            FixedValueDoc::Float(v) => FixedValue::Float(v),
+            FixedValueDoc::Str(v) => FixedValue::Str(leak_str(v)),
+            FixedValueDoc::StrList(v) => FixedValue::StrList(leak_str_slice(v)),
+            FixedValueDoc::Json(v) => {
+                let json = yaml_to_json(v)
+                    .map_err(|reason| format!("{id}: fixed input '{key}' {reason}"))?;
+                FixedValue::Json(Box::leak(Box::new(json)))
+            }
         };
-        if let Some(shown) = non_finite {
-            return Err(format!(
-                "{id}: fixed input '{key}' contains {shown}, which has no JSON \
-                 spelling; pin finite numbers only"
-            ));
-        }
-        fixed_inputs.push((leak_str(key), value.into_fixed()));
+        fixed_inputs.push((leak_str(key), fixed));
     }
     let table = SourcePackTable {
         id,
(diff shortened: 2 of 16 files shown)


crates/skardi/src/sources/providers/open_connector/packs/notion.yaml
```diff
@@ -0,0 +1,111 @@
...
+      strategy: cursor
+      cursor_input: startCursor
+      next_cursor_path: "$.next_cursor"
+      page_size_input: pageSize
+      page_size: 100
+    fixed_inputs:
+      query: ""
+      filter:
+        property: object
+        value: data_source
```
abbccdda, line 83: **Version-coupled and unverified against real data.** `filter.value: data_source` only exists in Notion API version 2025-09-03+; older versions spell this `database`. The Notion-Version header lives in the OC gateway, is invisible to Skardi, and is unpinned here — if the gateway is ever on an older version this table silently returns nothing (or the gateway 400s the filter). The captured contract declares `const: "data_source"`, so the reconciled gateway *is* on the new version, but nothing guards drift and no real scan ever confirmed rows come back. Please confirm with a live scan (see testing plan) and, ideally, pin/record the expected Notion-Version.

crates/skardi/src/sources/providers/open_connector/packs/notion.yaml
```diff
@@ -0,0 +1,111 @@
...
+    columns:
+      - { name: id, path: id, type: utf8, nullable: false }
+      - { name: created_time, path: created_time, type: timestamp_ms_utc, nullable: true }
+      - { name: last_edited_time, path: last_edited_time, type: timestamp_ms_utc, nullable: true }
+      - { name: title, path: title, type: utf8_list_from_object_key, key: plain_text, nullable: true }
+      - { name: archived, path: archived, type: boolean, nullable: true }
+      - { name: url, path: url, type: utf8, nullable: true }
+      - { name: parent, path: parent, type: json, nullable: true }
+      - { name: properties, path: properties, type: json, nullable: true }
```
abbccdda, line 93: **Column set doesn't match the captured contract — a tell the fixture is hand-crafted, not observed.** In `fixtures/notion/contracts/search.json` the `data_source` branch declares `id, title, properties, parent, url, in_trash` — but **not** `created_time`/`last_edited_time`/`archived`, which this table maps; and this table does **not** map `in_trash`, which the contract *does* declare. Because search columns ride `additionalProperties` (outside the fingerprint gate), this mismatch raises no registration error: if real data sources don't carry `created_time`/`last_edited_time`, these columns are silently always-NULL. Please reconcile against a real `notion.search` data_source response and align the column set (add `in_trash`? drop the unbacked timestamps?).

crates/skardi/src/sources/providers/open_connector/packs/notion.rs
```diff
@@ -0,0 +1,729 @@
...
+//!   objects): capability-gated on the integration and privacy-sensitive
+//!   — same call the Slack pack made.
+//! - **Nullability is conservative**: only `id` is non-null. Notion nulls
+//!   or omits most metadata fields depending on object age and API
+//!   version; both spellings become SQL NULL, pinned by fixtures.
+//! - **Fingerprints are pinned** from a live gateway capture
+//!   (`fixtures/notion/contracts/`); `pages` and `data_sources` share
+//!   `notion.search`'s contract and therefore its pin. The declared
+//!   search item schema is EMPTY (`results.items` declares no
+//!   properties), so every mapped column of the two search tables rides
```
abbccdda: **Weakest point of the safety story — worth flagging explicitly.** Because `notion.search` results items are an `anyOf` (page | data_source), the fingerprint walker doesn't descend the branches, so *every* column of `pages` and `data_sources` sits outside the drift gate. That's honest and pinned by `fingerprint_coverage_gap_is_pinned`, but it means the two highest-value tables get **no** registration-time drift protection — a gateway shape change surfaces only as a scan-time conversion error to the end user, not at config load. Given these are also the version-coupled tables, this raises the bar for real-data verification. Consider a follow-up that fingerprints the `anyOf` branches, and in the meantime call this residual risk out in the user-facing doc.

crates/skardi/src/sources/providers/open_connector/packs/fixtures/notion/users.json
```diff
@@ -0,0 +1,17 @@
+{
```
abbccdda, line 1: These fixtures are synthetic/redacted, not captured from a real workspace — which is fine as a conversion contract, but it means the wire **data** shape (not just the declared schema) is unverified. Per the testing plan in the review summary, please re-derive these from a real live scan so they reflect actual Notion response shapes (field presence/absence, timestamp spellings, `next_cursor` envelope, `title` rich-text arrays).

abbccdda (commented):
🚫 **Blocking / requesting changes — do not merge until there is a solid e2e verification against a real Notion workspace.**

Review focused on correctness vs. the 5.3 spec, and on the verification gap.

**Strong parts:** matches the pack admission gate (terminating cursor pagination, deterministic schema, read-only, null/empty/nested + schema-mismatch fixtures, docs); the one engine extension `FixedValue::Json` is minimal, backward-compatible, and correctly extends the non-finite-float guard to nested values; `person.email` privacy exclusion is right. Tests pass locally (14 new, 0 failed).

**Why this blocks:** the PR/commit/spec all state scans only *"reach the credential wall (403 authorization_failed)"* — i.e. the gateway was reached and the generated **input** shapes passed its strict schema, but **zero Notion rows were ever returned**. Every fixture is hand-authored/redacted (`Ada Lovelace`, `redacted@example.test`), not captured from a real workspace. So every assumption about actual **response data** is unproven against real bytes. Two of the four tables (`pages`, `data_sources`, both via `notion.search`) are exactly the ones that (a) carry a Notion-API-version coupling and (b) sit entirely outside the fingerprint drift gate — see inline comments. A pack whose data-shape contract has never seen real data isn't ready to land.

**Requested testing plan before merge — verify against a real Notion source:**
1. Stand up an Open Connector gateway with a real Notion integration token; share at least one page and one data source with the integration so results are non-empty. **A personal, free-tier Notion account is completely fine for this** — the free plan supports internal integrations and the read actions this pack uses; no paid workspace or org approval required.
2. Run one live scan per table (`users`, `pages`, `data_sources`, `block_children`) and confirm each returns rows (not just a 403). Capture the raw responses.
3. Re-derive the fixtures under `fixtures/notion/` from those captured responses (as the 5.1/5.2 "reconciled live" contracts were), so the conversion contract reflects real shapes — especially `data_source` objects' actual field set and the `title` rich-text array.
4. Record the gateway's Notion-Version used, and confirm `filter.value: data_source` actually returns data sources under it (vs. the pre-2025-09-03 `database` spelling).
5. Note the observed row counts / a redacted sample in the PR description as evidence, so the "live-verified" claim covers data and not only the credential wall.

Once steps 1–5 are done (or the scope is explicitly narrowed with a documented caveat and downgraded "live-verified" wording), I'll re-review and clear the block.

abbccdda (changes requested):
🚫 Requesting changes — blocking merge until this pack has a solid end-to-end verification against a real Notion workspace (a personal free-tier Notion account is fine).

crates/skardi/src/sources/providers/open_connector/packs/notion.yaml
bakey: Confirmed against a real workspace. The gateway pins `Notion-Version: 2026-03-11` in code (`open-connector src/providers/notion/executors.ts`, `notionCoreVersion`) — ≥ 2025-09-03, so the `data_source` spelling is the correct one for this gateway, and a live scan through skardi-server returned the workspace's real data source (1 row) with this exact filter pin. f86160f records the pinned version in `docs/open-connector-notion.md` and the module docs. On drift: a gateway upgrade that changes the action's declared output schema is refused at registration by the fingerprint gate; a Notion-Version bump that only changes wire rows would surface at scan time under the conversion rules (input-schema fingerprinting is an open design question tracked separately).

crates/skardi/src/sources/providers/open_connector/packs/notion.yaml
bakey: You were right that the column set was wrong, and live rows settled it in both directions: real data_source objects DO carry `created_time`/`last_edited_time` (the contract under-declares them — kept), DO carry `in_trash` (now mapped) plus `public_url` (now mapped), and carry NO archived flag (dropped). The same live pass also caught that pages spell the flag `is_archived` — the declared `archived` never appears on the wire, so the pages column was renamed and blocks' `archived` dropped too. Fixed in f86160f; the fixture tests now pin the live shapes and a real scan returns non-NULL values for every mapped flag.

crates/skardi/src/sources/providers/open_connector/packs/notion.rs
bakey: Accurate — the search items are `anyOf` (page | data_source) and the coverage walker does not descend branches, so both search tables sit outside the registration-time gate (pinned honestly by `fingerprint_coverage_gap_is_pinned`, wording corrected in f86160f). Agreed this is the right follow-up rather than in-PR work: teaching the walker to union/intersect `anyOf` branch properties would put the two highest-value tables under the gate. Spun off as a tracked follow-up task.

crates/skardi/src/sources/providers/open_connector/packs/fixtures/notion/users.json
bakey: Done in f86160f: all four fixtures are now redacted LIVE captures from a real workspace through the gateway — real envelope keys (`request_id`, `type` discriminators), real field presence/absence per row, real timestamp spellings (`.000Z`), real rich-text title arrays, and the actual `app.notion.com/p/...` URL shape. Redaction replaces UUIDs with synthetic ones and all names/titles/URLs/emails with placeholders, structure verbatim. `users_type_mismatch.json` stays deliberately synthetic (it is the admission-gate mismatch case). The fixture-derived expectations (e.g. pages carrying `is_archived`, no archived flag on blocks) all changed as a result, which is exactly the drift you predicted.

bakey:
## Real-workspace e2e verification (f86160f)

Ran the requested end-to-end verification against a **real Notion workspace** (free-tier internal integration, token configured in a local Open Connector gateway; Skardi never sees the credential): registration passed the fingerprint gate against live discovery, and **all four tables scanned real rows through skardi-server**:

| Table | Live result |
|---|---|
| `users` | 3 rows (1 person + 2 bots), mixed null/value `avatar_url` |
| `pages` | 4 rows via the pinned `{object: page}` filter; `is_archived`/`in_trash` extract real booleans; timestamps parse |
| `data_sources` | 1 row via the pinned `{object: data_source}` filter on Notion-Version 2026-03-11; `title` plucked from the rich-text array |
| `block_children` | 4 rows (`heading_1`/`to_do`/`paragraph`) with the bound `blockId` forwarded |

What the live pass changed (all in f86160f):
- **pages**: `archived` → `is_archived` — the live wire never carries the contract-declared spelling; the old column would have been always-NULL.
- **data_sources**: `+in_trash`, `+public_url`, `-archived`; `created_time`/`last_edited_time` confirmed real (contract under-declares them).
- **block_children**: `-archived` (blocks carry `in_trash` only).
- **fixtures**: all four re-derived from redacted live captures.
- **docs**: gateway's pinned `Notion-Version: 2026-03-11` recorded; coverage-gap wording corrected (search items are an `anyOf` the walker does not descend).

262 open_connector tests / 848 crate-wide, all green. The `anyOf`-branch fingerprint coverage suggestion is spun off as a follow-up.

crates/skardi/src/sources/providers/open_connector/packs/loader.rs
```diff
@@ -333,6 +337,21 @@ fn convert_filter(doc: FilterDoc) -> FilterMapping {
...
+fn first_non_finite(value: &serde_json::Value) -> Option<String> {
+    match value {
+        serde_json::Value::Number(n) => {
+            n.as_f64().filter(|f| !f.is_finite()).map(|f| f.to_string())
+        }
+        serde_json::Value::Array(items) => items.iter().find_map(first_non_finite),
+        serde_json::Value::Object(map) => map.values().find_map(first_non_finite),
+        _ => None,
+    }
+}
```
abbccdda: **Correctness: this guard is dead code — the nested non-finite trap it documents is still open.**

By the time `first_non_finite` runs, a nested `.nan`/`.inf` has already been lost. When the untagged `FixedValueDoc::Json(serde_json::Value)` variant deserializes, serde buffers the YAML into `Content::F64(NaN)` and serde_json's `Value` visitor converts it before this function ever sees it — in serde_json 1.0.150 (what Cargo.lock pins, `arbitrary_precision` off):

```rust
fn visit_f64<E>(self, value: f64) -> Result<Value, E> {
    Ok(Number::from_f64(value).map_or(Value::Null, Value::Number))
}
```

So `filter: {threshold: .nan}` arrives here as `{"threshold": null}`, and a non-finite `Value::Number` cannot exist for this function to find — it can never return `Some`. Two consequences:

1. A nested non-finite silently pins JSON `null` — exactly the "silently send null at query time" failure the comment at the call site promises to prevent. (Top-level `.nan` is still caught, but by the `Float` variant, not this.)
2. The loader tests only cover top-level `.nan`/`.inf`/`-.inf`; a nested-`.nan` test would expose this immediately.

No impact on the Notion pack itself (its pins are strings only) and packs are review-gated embedded assets, so severity is low — but the PR description's "non-finite-float guard extended to nested values" doesn't hold. Suggested fix: deserialize the `Json` variant as `serde_yaml::Value` and convert to `serde_json::Value` manually, rejecting non-finite floats in the conversion — then add the nested `.nan` case to `semantic_invariants_are_rejected_with_targeted_errors`. Alternatively, if you'd rather defer, delete `first_non_finite`, correct the doc comment to say the nested case is unguarded, and drop the claim from the PR body.

abbccdda (commented):
Reviewed the e2e reconciliation (f86160f) and the full implementation. The test suite and fixtures look solid — redaction verified (synthetic v4 UUIDs, example-domain emails/URLs, no tokens), column renames consistent across yaml/tests/coverage pins, and the mock-gateway e2e covers pagination, filter pins, resource enforcement, LIMIT early-stop, UDTF parity, and drift refusal. Security-wise this is clean: no credential exposure, counters-only logging, `person.email` unmapped, and SQL users can't override the pinned wire inputs.

One real correctness finding on the loader guard, plus a minor test nit — inline below.

crates/skardi/src/sources/providers/open_connector/packs/notion.rs
```diff
@@ -0,0 +1,761 @@
...
+        let batches = collect(&ctx, "SELECT id FROM saas.ws.users ORDER BY id").await;
+        assert_eq!(ids_of(&batches), vec!["u-1", "u-2", "u-3"]);
+
+        let inputs = execute_inputs(&gateway);
+        assert_eq!(inputs.len(), 2, "two cursor pages");
+        assert!(inputs[0].get("startCursor").is_none(), "{}", inputs[0]);
+        assert_eq!(inputs[1]["startCursor"], "cur-2");
+        for input in &inputs {
+            assert_eq!(input["pageSize"], 100, "page-size hint: {input}");
+        }
```
abbccdda, line 533: **Minor:** the exactly-the-declared-inputs guard ("nothing else on the wire") only exists in the search test — here the users scan asserts `startCursor`/`pageSize` presence but not the absence of extras. Since this test's doc comment sells itself as pinning users' wire declarations, consider asserting the exact key set per request (`["pageSize"]` on page 1, `["pageSize", "startCursor"]` on page 2), same as `search_tables_pin_their_object_filters_on_the_wire` does. Not blocking.

abbccdda (approved):
Approving with nit comments

crates/skardi/src/sources/providers/open_connector/packs/loader.rs
bakey: Confirmed and fixed in 24c9612 — your serde analysis was exactly right, and this was the checklist's "a gate whose failure arm no test exercises is dead code" item hiding in plain sight. `FixedValueDoc::Json` now captures `serde_yaml::Value` (where the non-finite is still observable) and converts through a fallible `yaml_to_json` that rejects everything JSON cannot spell: non-finite floats at any depth, non-string mapping keys, and YAML tags (that last one turns out to be rejected even earlier, by serde's untagged buffering — pinned with its actual error message and kept in the converter as defense in depth). `first_non_finite` is deleted rather than repaired, and `into_fixed` folds into the single `convert_table` match so no unreachable arm survives the refactor. Regression tests: nested `.nan` in an object, `.inf` inside a nested list, a numeric mapping key, and the pass side — nested finite floats/bools/nulls asserted faithful on `to_json` output. You're also right about the blast radius (Notion's pins are strings; assets are review-gated), but the PR description's claim is now true instead of aspirational.

crates/skardi/src/sources/providers/open_connector/packs/notion.rs
bakey: Fixed in 24c9612: the users cursor e2e now asserts the exact key set per request — `["pageSize"]` on page 1, `["pageSize", "startCursor"]` on page 2 — matching the nothing-extra-on-the-wire guard the search test applies, so users' wire declaration is pinned in full, absence included.

PR #175, opened 2026-07-28
feat(sources): add RSS feed support

gracexmatin:
Implements first-class RSS/Atom support (`type: rss`): one configured subscription list registers as a read-only catalog exposing `<name>.main.feeds` (per-subscription fetch health) and `<name>.main.items` (live union of current entries), fetched at scan time through a per-feed TTL cache with HTTP conditional requests, partition-per-feed execution, and in-band `window_status` freshness stamps.

Follow-up to the merged design PR #163. Implementation lands task-by-task on this branch, following the committed plan (`docs/superpowers/plans/2026-07-27-rss-feed-provider-m1-m2.md`, 20 TDD tasks — M1 provider core + M2 docs). Each task is committed only after a two-stage review (spec compliance + code quality).

## In this PR so far

**Design spec v2 refinements since #163** — `docs/superpowers/specs/2026-07-22-rss-feed-support-design.md` is the normative spec: Markdown item storage (converted once at extraction), default-deny egress (SSRF) policy, sanitation ladder + dialect conformance contract, LIMIT-pruning semantics, versioned engine↔skill surface.

**Implementation plan** — 20 tasks with exact schemas, interfaces, and test lists; M3 (auto_news_base skill, pipeline statement-sequences, version handshake) deferred to a follow-up plan.

**Task 1 — provider foundation** (`crates/skardi/src/sources/providers/rss/`):
- Typed `RssConfig` / `FeedSubscription` with spec defaults (ttl 900s, concurrency 6, request timeout 10s, scan deadline 60s, 5 MiB response cap, self-identifying UA) and a zero-I/O `validate()` — mutually-exclusive `feeds`/`opml`, URL scheme checks, effective-name uniqueness, non-zero bounds, unknown-field rejection
- `RssError` taxonomy, `RSS_SURFACE_VERSION = 1`, `rss` Cargo feature (config compiles unconditionally so feature-off builds still parse ctx.yaml and fail with a clear error)
- 10 unit tests, every negative test asserting the failure reason

## Task progress (1/20)

- [x] 1. `rss` feature, module skeleton, typed `RssConfig`
- [x] 2–6. OPML resolution · egress policy (SSRF guard) · mock feed server + bounded fetcher · sanitation ladder · HTML→Markdown converter
- [x] 7–11. dialect conformance + parse driver · field extraction · Arrow schemas · TTL cache · engine state machine
- [x] 12–16. partition-per-feed exec · table providers + pushdown · registration · server wiring · CLI wiring
[...]

PR #180, opened 2026-08-02
feat(sources): RSS provider 1/4 — config, OPML, egress seam (AllowAll default), bounded fetcher

gracexmatin:
Phase 1 of 4 stacked PRs splitting #175 (plan Tasks 1-4). Merge order: this one first.

## Scope
- `rss` Cargo feature; typed `RssConfig` with zero-I/O `validate()`. Subscription-config validation surfaces a **nameless `InvalidConfig`** — the server attaches the source name when it wraps registration errors, so nothing is re-stamped here (same shape as `register_open_connector_tables`).
- OPML subscription resolution converging with inline `feeds:` on one `ResolvedSubscription` type, nameable in featureless builds
- **Egress-policy seam — OSS ships no SSRF guard.** An `EgressPolicy` trait with `AllowAll` as the only OSS implementation: the fetcher reaches any address the host can route to (including link-local `169.254.169.254` and RFC 1918). Constraining fetch egress is delegated to the operator's infrastructure or an injected policy — OSS deliberately does not sandbox destinations (see the design spec's Security section and the Cloud egress design). `PolicyDns` (resolver-layer, DNS-rebinding-safe) and the per-hop IP-literal re-check remain in place so an *injected* policy holds on the initial URL, every redirect hop, and hostnames — both are no-ops under `AllowAll`.
- **Bounded fetcher**: conditional GET (ETag/Last-Modified, first hop only), manual per-hop redirect loop with the egress policy consulted on every hop (reqwest's connector bypasses custom resolvers for IP-literal hosts), decoded-stream size cap, `Retry-After` clamped to 10s, unconditional 304 treated as a protocol error
- `MockFeedServer` test harness

`egress`/`fetch` carry `#![allow(dead_code)]` until phase 3 wires them into the engine.

## Verification
`cargo test -p skardi --features rss` passes at this phase (848 tests, includes non-rss crate tests); the committed tree compiles standalone.


crates/skardi/src/sources/providers/rss/fetch.rs
```diff
@@ -0,0 +1,1025 @@
...
+    max_response_bytes: u64,
+}
+
+impl FeedFetcher {
+    /// Build the fetcher's one shared client. `policy` is consulted in two
+    /// ways: wrapped in a [`PolicyDns`] as the client's DNS resolver (so
+    /// every hostname connection is checked structurally, including
+    /// pooled-connection reuse), and held directly for the IP-literal
+    /// checks [`FeedFetcher::check_hop_target`] runs before the initial
+    /// request and before every redirect hop.
```
BtXin, line 240: [P1] Disable automatic proxy discovery before relying on PolicyDns

Reqwest automatically honors system proxy variables unless `no_proxy()` is set. In that mode the target hostname is handed to the proxy, so this resolver never sees or validates the destination address. I reproduced this with the existing `injected_policy_refuses_hostname_via_resolver` test: it unexpectedly fetched the denied URL with proxy variables present and passed after they were removed. This defeats the injected egress policy and DNS pinning for agent-authored URLs. Please disable automatic proxies here, or make proxy use explicit and enforce the same destination policy around it, with a regression test that sets a proxy.

crates/skardi/src/sources/providers/rss/opml.rs
```diff
@@ -0,0 +1,425 @@
...
+///
+/// OPML lets subscriptions nest inside grouping outlines — an untyped,
+/// folder-like `<outline text="Tech">` wrapping its feeds as children.
+/// quick-xml's event stream is flat regardless of nesting depth, so no
+/// recursion is needed: every `outline` start/empty tag is visited once, in
+/// document order, and only those carrying an `xmlUrl` become a
+/// subscription. A grouping outline with no `xmlUrl` of its own — like the
+/// "Tech" folder above — is silently skipped; it is structure, not a feed.
+fn read_opml(name: &str, path: &Path) -> Result<Vec<(String, Option<String>)>, RssError> {
+    let content = fs::read_to_string(path).map_err(|e| RssError::OpmlUnreadable {
```
BtXin: [P2] Bound OPML reads and keep them off async workers

This reads an arbitrary configured path to EOF with no byte or entry cap, and `register_rss_tables*` calls it synchronously from an async registration function. A large OPML file can block a Tokio worker and consume unbounded memory; a non-terminating special file such as `/dev/zero` never reaches parsing at all. Please use bounded async I/O or `spawn_blocking`, reject files above an explicit OPML size/entry limit, and cover the limit in tests.

BtXin (changes requested):
Requesting changes on the fetch/config foundation. The injected egress boundary is bypassable when system proxy variables are present, and OPML registration performs an unbounded blocking read. The proxy issue reproduced locally: the RSS suite fetched a hostname that the test policy denied when proxy variables were set, while the same test passed with those variables removed.

crates/skardi/src/sources/providers/rss/fetch.rs
gracexmatin: Fixed: `FeedFetcher::new` now builds its client with `no_proxy()`, so system proxy variables are ignored and every hostname connection goes through `PolicyDns` unconditionally. The module doc gained a "Proxies are disabled" section with the reasoning. I went with disabling outright; if a deployment need for proxies ever shows up, explicit proxy support can be a separate proposal.

Regression test: `proxy_env_vars_do_not_bypass_the_egress_policy` sets `HTTP_PROXY`/`http_proxy` around fetcher construction (the moment reqwest reads them — verified against the reqwest 0.12.28 source: `auto_sys_proxy` re-reads the environment on every `ClientBuilder::build`), fetches a `localhost` URL under a policy denying `127.0.0.1`/`::1`, and asserts the `Egress` refusal with zero requests reaching the mock server. `NO_PROXY` is deliberately pinned to `127.0.0.1` (not `localhost`) so a future regression fails this test without contaminating concurrently running tests that talk to loopback by IP; prior env values are restored afterwards. Verified by mutation: removing `no_proxy()` turns the test red with the transport error from the proxy address.

crates/skardi/src/sources/providers/rss/opml.rs
gracexmatin: The bound is in this PR: `MAX_OPML_BYTES` (1 MiB), enforced *while* reading — `File::open` + `take(MAX_OPML_BYTES + 1)`, never a read to EOF — so an oversized file and a non-terminating `/dev/zero` both stop at the cap and surface an `OpmlUnreadable` naming the limit. I went with a byte cap rather than an entry cap because only a byte cap acts at the read stage. Tests pin both sides of the boundary: one byte over is rejected, exactly at the cap parses.

For the async half, the plan is a `spawn_blocking` wrap, landing in the phase-4 PR — `register_rss_tables`, the async code that calls `resolve_subscriptions`, lives there. `resolve_subscriptions`'s doc now states the contract explicitly: bounded blocking I/O; async callers wrap it in `spawn_blocking`.

BtXin (changes requested):
The earlier proxy-policy bypass is fixed, and the OPML read is now byte-capped. I found four remaining issues in this round:

- **[P2] OPML reads can still block forever.** `File::take(MAX_OPML_BYTES + 1)` bounds bytes but not time. A FIFO or device that produces fewer than the limit while remaining open blocks registration indefinitely; wrapping this in `spawn_blocking` only moves the blocked read. Reject non-regular files or add an enforceable timeout/cancellation boundary.

- **[P2] Mid-body failures bypass retries.** `attempt_hop` returns `self.read_body(resp).await?` directly, while `read_body` maps streamed-body timeouts and transport failures into `FetchError`. Consequently only failures before response headers are retried. Route body `Timeout`/`Transport` errors back through the attempt loop while keeping `TooLarge` terminal.

- **[P2] The proxy regression test mutates process-wide environment state concurrently.** `HTTP_PROXY` and related variables remain changed across an `.await`, while the Rust test harness can run other tests and threads. This violates the safety assumptions behind `set_var` and can make unrelated HTTP tests flaky. Isolate this check in a subprocess whose environment is established before execution.

- **[P3] Inline type paths violate this repository's import rule.** For example, `url::Url::parse` should use an imported `Url`; the same cleanup applies to inline `SystemTime` and `Error` type paths in `fetch.rs`.

Local verification: the targeted RSS suite passes (48/48), formatting is clean, and `cargo check -p skardi --no-default-features` succeeds.

gracexmatin:
All four addressed:

**OPML liveness**: `read_opml` now stats before open and rejects non-regular files outright; the open+read runs on a helper thread, bounded externally by `OPML_READ_TIMEOUT` (5s). On timeout the helper thread is abandoned rather than killed — at most one per failed registration, holding at most the `take()`-bounded 1 MiB — in exchange for a registration path that can no longer hang.

**Body-phase retries**: `read_body`'s `Timeout`/`Transport` now re-enter the same attempt loop; a retry reissues the full `GET` and discards the partial body (no `Range` resume: feed bodies are small enough that resumption buys nothing, and a body spliced from two server states is worse than none). `TooLarge` stays terminal — a policy verdict on the response's size, not a transient fault.

**Proxy test isolation**: the test is now subprocess-isolated — the parent relaunches the test binary itself via `current_exe()` (`--exact --ignored` targeting an `#[ignore]`d child test), with the proxy variables injected at spawn via `Command::env`; the child skips rather than fails when the proxy variables are absent (accommodating the CI integration step that runs every `#[ignore]`d test wholesale).

**Import cleanup**: replaced per the repo convention — `SystemTime`/`UNIX_EPOCH`, `std::error::Error as StdError` (the alias `cli/client.rs` already uses, avoiding the clash with thiserror's derive macro), and `url::Url` in `config.rs`; the test modules' `Command`/`Instant` swept as well.

BtXin (commented):
Sweep 2 (head 649b226). Both prior threads are **confirmed fixed**: `.no_proxy()` is in `FeedFetcher::new` with a subprocess-isolated regression test that can't vacuously pass, and OPML reads are capped at 1 MiB via `take()` with a regular-file check and a 5s wall-clock bound. Nice work on both.

Remaining items, besides the inline comments:

**Stack hygiene (blocking mechanics):** this branch is 4 commits behind `main` (including the `rust-toolchain.toml` pin from #189), and the four fix commits here (`c3a8326`…`649b226`) are **not** in phases 2–4 — those branches are still based on the pre-fix phase 1, and their green CI predates the fixes. Please rebase this onto `main` and the rest of the stack onto this.

**[P3] Seam test gaps:**
- The `Host::Ipv6` arm of `check_hop_target` is never exercised — no bracketed-IPv6-literal test on either the initial URL or a redirect hop (directly relevant to the canonicalization comment inline).
- The redirect→hostname→`PolicyDns` composition is untested: redirect denial is only tested with an IP-literal target, resolver denial only with an initial-URL hostname. The composed path is the DNS-rebinding-via-redirect case the design doc calls out.
- The `sent_validator` vs `validators.is_some()` distinction (`Some(Validators { etag: None, last_modified: None })` answered with 304) is called out in a comment but untested.

**[P3] Dangling forward references:** `error.rs` doc comments and the test `max_error_chars_is_the_number_the_docs_publish` cite `docs/rss.md:974` / `docs/rss/semantics.yaml:85`, and `egress.rs` cites `2026-08-03-rss-cloud-egress-design.md` — none of these files exist at this merge point (they land in later stack PRs). If this PR merges alone the comments dangle and the test's claim is false. Either soften the wording to "published in the RSS docs (later in this stack)" or move the doc-pinning test to phase 4.

crates/skardi/src/sources/providers/rss/fetch.rs
```diff
@@ -0,0 +1,1209 @@
...
+    /// resolver) validates it structurally when reqwest actually connects.
+    fn check_hop_target(&self, url: &Url) -> Result<(), FetchError> {
+        if url.scheme() != "http" && url.scheme() != "https" {
+            return Err(FetchError::InvalidUrl {
+                reason: format!("scheme '{}' is not http or https", url.scheme()),
+            });
+        }
+        let ip = match url.host() {
+            Some(Host::Ipv4(v4)) => Some(IpAddr::V4(v4)),
+            Some(Host::Ipv6(v6)) => Some(IpAddr::V6(v6)),
```
BtXin, line 337: [P2] Canonicalize IPv4-mapped IPv6 before handing IPs to the policy

`Host::Ipv6(v6)` is passed to `check_ip` raw here, and `check_addrs` (egress.rs:94-106) never canonicalizes either — there is no `to_canonical()` anywhere in the module. A feed URL or redirect `Location` of `http://[::ffff:10.0.0.1]/` reaches an injected policy as `IpAddr::V6`, and the natural policy shape (`V4 => deny_if_private(v4), V6 => deny_if_v6_reserved(v6)` — note Rust's `Ipv6Addr` range checks do NOT unmap) sees no reserved v6 range and allows it, while the OS dual-stack connect delivers the TCP connection to IPv4 `10.0.0.1`. Same hole on the resolver path via an attacker-controlled AAAA record returning `::ffff:a.b.c.d`. This PR's own test `DenyList(Vec<IpAddr>)` holding `"10.0.0.1".parse()` is bypassed exactly this way.

This is not the documented "AllowAll allows SSRF" stance — it's the seam failing to hold for injected policies, which is the contract this phase claims ("an injected policy holds on the initial URL, every redirect hop, and hostnames"). Fix is one line at each seam call point: `ip.to_canonical()` before `check_ip` here and in `check_addrs`. Please also add a bracketed-mapped-literal regression test on both the initial URL and a redirect hop.

crates/skardi/src/sources/providers/rss/fetch.rs
```diff
@@ -0,0 +1,1209 @@
...
+        Err(last_err.unwrap_or(FetchError::Transport {
+            reason: "exhausted retry attempts without a recorded error".to_string(),
+        }))
+    }
+
+    /// Capture the validator/content-type headers, then stream the body,
+    /// enforcing the byte cap on the *decoded* stream — see the module doc.
+    async fn read_body(&self, resp: Response) -> Result<FetchOutcome, FetchError> {
+        let http_status = resp.status().as_u16();
+        let etag = header_string(&resp, ETAG);
```
BtXin, line 482: [P3] Redirected feeds permanently lose conditional GET

`read_body` captures `etag`/`last_modified` from whichever hop terminated the chain, but the caller will cache them against the *original* URL. For the ubiquitous http→https 301 feed: outcome carries `etag(B)`, cached under A; the next fetch sends `If-None-Match: etag(B)` to A on hop 0, A answers 301 again, hop 1 to B sends no validators (by design) — full body every scan, forever. This also contradicts the module doc's rationale that validators are "against the resource the caller actually cached" — they're B's, keyed by A.

Safe (a mismatched etag can only produce a full 200), but the engine PR should either store the final URL alongside the validators or explicitly document that redirected feeds never 304.

crates/skardi/src/sources/providers/rss/fetch.rs
```diff
@@ -0,0 +1,1209 @@
...
+            http,
+            policy,
+            request_timeout,
+            max_response_bytes,
+        })
+    }
+
+    /// Fetch `url`, sending `validators` as conditional-GET headers on the
+    /// first hop only. See the module doc for the redirect and retry rules.
+    pub async fn fetch(
```
BtXin, line 272: [P3] No overall per-fetch deadline

The hop loop gives each hop a fresh attempt budget: 6 hops (initial + `MAX_REDIRECT_HOPS`) × (3 attempts × `request_timeout` 10s + 2 waits × `MAX_RETRY_WAIT` 10s) ≈ 300s worst case for one `fetch()` — 5× the default `scan_timeout_seconds: 60`, which is config-only in this PR and enforced by nothing yet. Fine if (and only if) the phase-3 scan deadline wraps `fetch()`; flagging so that isn't missed when the engine lands.

crates/skardi/src/sources/providers/rss/opml.rs
```diff
@@ -0,0 +1,652 @@
...
+/// Errors are returned as bare reason strings: only the caller knows the
+/// source `name` that [`RssError::OpmlUnreadable`] wants.
+fn read_to_string_bounded(path: &Path, timeout: Duration) -> Result<String, String> {
+    let (tx, rx) = mpsc::channel();
+    let path: PathBuf = path.to_path_buf();
+    thread::spawn(move || {
+        let result = (|| {
+            let file = File::open(&path).map_err(|e| e.to_string())?;
+            let mut content = String::new();
+            file.take(MAX_OPML_BYTES + 1)
```
BtXin, line 270: [P3] Over-cap OPML cut mid-UTF-8-sequence reports the wrong error

`take(MAX_OPML_BYTES + 1)` can slice a multibyte character, so `read_to_string` fails with `InvalidData` and the operator sees `OpmlUnreadable { reason: "stream did not contain valid UTF-8..." }` instead of ever reaching the size-limit message — a misleading diagnostic that sends them hunting an encoding bug that isn't there. The at-cap/over-cap tests only use ASCII padding, which is why this isn't caught. Read into `Vec<u8>`, do the length check first, then `String::from_utf8`.

crates/skardi/src/sources/providers/rss/opml.rs
```diff
@@ -0,0 +1,652 @@
...
+                // entity resolution only, while 0.41 routes through
+                // `normalize_xml10_attribute_value`, which also performs XML
+                // attribute-value normalization — `is_xml10_normalization_char`
+                // (quick-xml-0.41.0/src/escape.rs) fires on `\t`, `\r`, `\n` and
+                // `&`, so a literal tab or newline inside an OPML attribute
+                // value now collapses to a space (`a\tb` → `a b`, pinned in this
+                // module's tests). That is the spec-correct reading of an
+                // attribute value, and it only ever touches whitespace inside a
+                // feed's display name or URL, so it is kept deliberately.
+                attr.decoded_and_normalized_value(quick_xml::XmlVersion::Implicit1_0, decoder)
```
BtXin: [P3] Convention: import `XmlVersion` at the top

`quick_xml::XmlVersion` is used as a full inline path while the file already `use`-imports `quick_xml::Reader` and `quick_xml::events::Event` at the top. Project rule: crate types are imported via `use`, never full paths in function bodies.

crates/skardi/src/sources/providers/rss/fetch.rs
gracexmatin: The phase-3 fix is decided, taking the first option (store the final URL): cache the final landing URL alongside the validators, send the conditional request straight to it so the etag and the address line up and 304 works again — paired with periodically re-probing the original URL to catch a redirect target that has drifted (301 is nominally permanent but drifts or gets misused in practice), the re-probe interval to be settled when the engine lands. Tracked so it won't be missed in phase 3.

crates/skardi/src/sources/providers/rss/fetch.rs
gracexmatin: Roger that. Already comment in phase3 PR as a reminder.

crates/skardi/src/sources/providers/rss/opml.rs
gracexmatin: Already fixed.

crates/skardi/src/sources/providers/rss/opml.rs
gracexmatin: Already fixed.

crates/skardi/src/sources/providers/rss/fetch.rs
gracexmatin: Fixed. Both seam call points (`check_hop_target` and `check_addrs`) now call `ip.to_canonical()` before `check_ip`, unmapping an IPv4-mapped v6 to V4 before it reaches the policy — `EgressDenied` reports the canonical form too, so the error lines up with the operator's rule, while `check_addrs` still returns the original addrs to connect to (canonicalization gates the check only, not what actually gets connected).

Three regression tests: a bracketed mapped literal on the initial URL, the same on a redirect hop, and a `check_addrs` unit test using a `DenyV4` policy that matches only the raw V4 (so it fires only once canonicalized).

BtXin (commented):
Sweep 3 (head 607828e). All round-2 findings on this PR are **confirmed fixed**, verified line-by-line:

- **P2 canonicalization**: `to_canonical()` now gates `check_ip` at both seam points (`check_hop_target`, `check_addrs`), the canonical address is reported in `EgressDenied`, and `check_addrs` correctly still returns the *original* addrs for the connect. Regression coverage is exactly right: mapped literal on the initial URL, on a redirect hop, and the `DenyV4`-based `check_addrs` unit test that can only pass through canonicalization (the AAAA path).
- **OPML mid-UTF-8 diagnostic**: `read_bytes_bounded` + size-check-before-decode; the multibyte-pad test math checks out (2-byte `é` tiled to cap+2, cut at cap+1 lands mid-character).
- **Seam test gaps**: the empty-`Validators`-answered-with-304 case and the redirect→hostname→`PolicyDns` composition are both pinned, with request-count assertions.
- **Dangling doc refs** softened to "published later in this stack"; **`XmlVersion`** imported at top. Branch also rebased onto main. CI green on 607828e.

Two non-blocking nits, take or leave:
- The `EgressPolicy` trait doc could state that implementations always receive canonical (unmapped) addresses — so a policy author knows they don't also need to handle `::ffff:a.b.c.d` themselves. Right now that guarantee is only documented at the two call sites.
- `injected_policy_refuses_hostname_redirect_target_via_resolver` depends on `localhost` resolving dual-stack (a set containing `::1`) — acknowledged in its comment, but on a runner with IPv6 disabled or no `::1 localhost` hosts entry the test fails spuriously. A guard that skips (with a log) when `lookup("localhost")` lacks `::1` would make it portable.

This PR is good to merge from my side once the stack coordination is sorted. Reminder: phases 2–4 are still based on the pre-fix history and haven't moved since sweep 2 — their open findings (incl. the P1s) all still stand.

crates/skardi/Cargo.toml
```diff
@@ -77,6 +104,18 @@ lance-index = { workspace = true }
 mongodb = "3.1"
 mysql_async = "0.36"
 percent-encoding = "2.3"
+# rss source connector (optional, behind the `rss` feature): parses OPML
+# subscription lists — the only hand-rolled XML in the provider besides
+# `feed-rs` (used for feed bodies themselves), whose own XML backend is also
+# quick-xml. `Cargo.lock` carries three quick-xml versions regardless of what
+# this line pins: `reqsign` requires exactly 0.37, `opendal` requires exactly
+# 0.38, and `feed-rs` 2.4 requires 0.41 (with its `encoding` feature) — three
```
abbccdda: This rationale reads as present fact, but `feed-rs` isn't a dependency yet in this PR (it lands in a later phase) — it's absent from both `Cargo.toml` and `Cargo.lock` at this commit. The `quick-xml 0.41` entry in the lock is pulled in by *this* line's own direct dep, not by `feed-rs`, and it resolved **without** the `encoding` feature (the lock shows `dependencies = [memchr]` only). So "`feed-rs` 2.4 requires 0.41 (with its `encoding` feature)" and "three deps this crate does not control" don't match the tree here. Suggest phrasing it as forthcoming (or landing the pin together with `feed-rs`), so a reader verifying against the current lock doesn't hit a contradiction.

crates/skardi/src/sources/providers/rss/config.rs
```diff
@@ -0,0 +1,468 @@
...
+    #[serde(default)]
+    pub opml: Option<PathBuf>,
+
+    /// Per-feed cache TTL, in seconds. `0` means always-live (every scan
+    /// re-fetches every feed) — the only bound where zero is legal.
+    #[serde(default = "default_ttl_seconds")]
+    pub ttl_seconds: u64,
+
+    /// Maximum number of feeds fetched concurrently: both the fetch
+    /// parallelism for THIS source's feeds. It is neither a per-host nor a
```
abbccdda: Two things here:

1. **Grammar** — "both the fetch parallelism for THIS source's feeds" has no second clause; the `both` dangles and the sentence reads as broken.
2. **Contradicts the design spec** — the spec says `max_concurrent` "doubles as the per-host politeness bound" (§Fetcher, and the YAML block comment "fetch parallelism and per-host politeness bound"), whereas this says "neither a per-host nor a per-process bound." The code comment is the accurate one (there's no per-host accounting anywhere), but the spec still advertises a politeness guarantee the implementation won't deliver — N feeds on one host ⇒ N concurrent hits. Please reconcile: correct the spec, or track a real per-host bound for the engine phase.

crates/skardi/src/sources/providers/rss/config.rs
```diff
@@ -0,0 +1,468 @@
...
+        if self.request_timeout_seconds == 0 {
+            return Err(invalid_config("request_timeout_seconds must be at least 1"));
+        }
+        if self.scan_timeout_seconds == 0 {
+            return Err(invalid_config("scan_timeout_seconds must be at least 1"));
+        }
+        if self.max_response_bytes == 0 {
+            return Err(invalid_config("max_response_bytes must be at least 1"));
+        }
+        if self.user_agent.trim().is_empty() {
```
abbccdda, line 219: `validate()` is the zero-I/O path the server's `validate_data_sources()` calls, but it only checks the UA is non-empty. A `user_agent` containing a control character (e.g. an embedded newline) passes config-load validation and only fails later at fetcher construction (`RssError::HttpClientBuild`) — a late error raised where no source name is in scope. `error.rs` documents this deliberately, but validating the UA as a `reqwest::header::HeaderValue` here would fail early, at config load, with the source name attributable. Minor.

crates/skardi/src/sources/providers/rss/error.rs
```diff
@@ -0,0 +1,148 @@
+//! Error taxonomy for the RSS provider, plus the one length bound every
+//! feed-influenced diagnostic string in this provider is held to.
+
+use thiserror::Error;
+
+/// Length cap, in characters, on any feed-influenced diagnostic string this
+/// provider stores or logs.
+///
+/// Three call sites share it, and they share it because they are bounded by the
```
abbccdda: The `docs/rss.md` / `semantics.yaml` references were nicely fixed to future tense, but this "Three call sites share it" claim (and the `engine.rs` / `cache.rs` / `conformance.rs` / `parse.rs` names just below) is still present tense — none of those files exist in this PR yet, so there are currently **zero** call sites and the constant is exercised only by its own test. Suggest marking these as forthcoming too, to match the docs wording.

crates/skardi/src/sources/providers/rss/fetch.rs
```diff
@@ -0,0 +1,1341 @@
...
```
abbccdda, line 272: Reliability flag for the engine phase (not a defect in this PR): `fetch()` has no total-time bound of its own. Worst case, a single call spans up to `MAX_REDIRECT_HOPS + 1` hops × `MAX_ATTEMPTS` attempts, each with a request timeout plus a retry wait capped at `MAX_RETRY_WAIT` — on the order of minutes for one hostile feed that redirect-chains retryable statuses. `scan_timeout_seconds` is configured but unused here. When the engine lands, please ensure it enforces a per-feed / total-scan deadline so a single feed can't park a scan far past intent.

abbccdda (commented):
Re-reviewed against the current head (607828ed) after the round-2/3 force-push. The mapped-IPv6 egress canonicalization and the future-tense fixes for the `docs/rss.md` references from my earlier pass are already in — nice. No blocking correctness or security issues: the fetcher (unconditional-304 mapping, per-hop validator suppression, guard-before-follow redirects, decoded-body size cap, `MAX_RETRY_WAIT` clamp) and the egress seam (`PolicyDns` fail-closed resolution + per-hop IP-literal recheck) both hold up, and the crate compiles with and without the `rss` feature. The comments below are maintainability / spec-alignment, plus one reliability flag to carry into the engine phase. Thorough, well-tested groundwork.

crates/skardi/Cargo.toml
gracexmatin: Kept the 0.41 pin but rewrote the comment as forthcoming. It now states the lock's current reality accurately — two uncontrolled versions (0.37 from `reqsign`, 0.38 from `opendal`/`object_store`) plus 0.41 from this line's own direct dependency — and frames `feed-rs` as "added later in this stack", the same softening the rest of the provider uses for forward references. The pin's rationale is preserved: pinning to 0.41 now means feed-rs's arrival adds no fourth quick-xml version, since our direct usage already rides the version it will pull.

crates/skardi/src/sources/providers/rss/error.rs
gracexmatin: Already shift to future tense for forward references

crates/skardi/src/sources/providers/rss/fetch.rs
gracexmatin: Already comment on phase3 PR https://github.com/SkardiLabs/skardi/pull/182#discussion_r3727644543

crates/skardi/src/sources/providers/rss/config.rs
gracexmatin: Grammar fixed. The design spec is corrected where per-host is decided unnecessary.

crates/skardi/src/sources/providers/rss/config.rs
gracexmatin: Added in head `531ba64`

crates/skardi/src/sources/providers/rss/fetch.rs
```diff
@@ -0,0 +1,1361 @@
...
+                            status: status.as_u16(),
+                        };
+                        if attempt + 1 >= MAX_ATTEMPTS {
+                            return Err(err);
+                        }
+                        tokio::time::sleep(retry_wait(&resp, attempt)).await;
+                        last_err = Some(err);
+                        continue;
+                    }
+                    if status.is_success() {
```
BtXin: [P2] `is_success()` admits the whole 2xx range — a spurious 206/204 becomes a "complete" feed with a sticky validator

A CDN or misconfigured origin that answers the (Range-less) GET with `206 Partial Content` streams a truncated body through `read_body` cleanly, and the outcome carries the response's ETag. The engine then caches the truncated document *with the strong validator*, so the next scan sends `If-None-Match`, gets 304, and the truncated view sticks until the server's validator changes. `204 No Content` similarly yields an empty "successful" body. Tightening to `status == StatusCode::OK` (or excluding 204/205/206) turns the mismatch into a `FetchError::Status` and a normal degrade instead. Note `testutil.rs` already carries a 206 reason phrase, so this is one small test away.

crates/skardi/src/sources/providers/rss/fetch.rs
```diff
@@ -0,0 +1,1361 @@
...
+    }
+}
+
+fn is_retryable_status(status: StatusCode) -> bool {
+    RETRYABLE_STATUSES.contains(&status.as_u16())
+}
+
+/// `max(Retry-After, backoff)`, capped at [`MAX_RETRY_WAIT`] — see the
+/// module doc's Retries section.
+fn retry_wait(resp: &Response, attempt: u32) -> Duration {
```
BtXin, line 603: [P3] The HTTP-date form of `Retry-After` is silently dropped

`util::http::parse_retry_after` parses delta-seconds only (its doc says "none of the APIs we integrate with emit it" — true for the first-party gateway it was written for, not for arbitrary third-party feed hosts; CDN-fronted origins commonly emit the date form). A `429 Retry-After: <http-date>` therefore falls to `backoff(attempt)` ≈ 250–750ms, and the fetcher re-hits a host that asked for a multi-minute pause three times within a second — while the module doc and the reconciled spec both claim host politeness rests on honoring `Retry-After`. Worth teaching the shared helper the date form (`httpdate` is already in the tree via reqwest's deps) or noting the limitation where the spec makes the politeness claim. Only the numeric form is tested today.

crates/skardi/Cargo.toml
```diff
@@ -29,6 +29,33 @@ chunking = ["dep:text-splitter"]
...
+# does not weaken its response-size bound — the bound is enforced again
+# while streaming the body regardless — but it does mean its
+# `content_length()` early reject simply does not fire for a compressed
+# response: reqwest strips both `Content-Encoding` and `Content-Length` from
+# the header set when it installs a gzip decoder, and `content_length()`
+# itself reads the body's size hint rather than the header, so it returns
+# `None` — not an under-reported wire size — the moment decoding is applied.
+rss = [
+    "dep:quick-xml",
+    "reqwest/gzip",
```
BtXin, line 65: [P2, one-line hardening requested] `reqwest/gzip` gives `llm_extract`'s image fetch a decompression amplifier

The blast-radius comment audits open_connector and the CLI, but the same feature unification also flips auto-gzip on for `llm_extract`'s ad-hoc image client (`model/llm_extract/mod.rs`, the `LLM_EXTRACT_IMAGE_FETCH=1` path), which reads with an unbounded `resp.bytes()`. Today that read is bounded only by what a 30s timeout can move over the wire; with gzip compiled in, an attacker-controlled `image_ref` can serve a small compressed body that inflates ~1000:1 in memory. Release artifacts don't enable `rss` (so today this needs a custom build *plus* the env opt-in), but phase 4 makes rss-enabled server builds normal. Since that client wants no content negotiation anyway, please add `.no_gzip()` to its builder (same defensive pattern object_store already uses) — either in this PR as part of the blast-radius handling or as an immediate follow-up, and extend the audit comment to name it.

crates/skardi/src/sources/providers/rss/config.rs
```diff
@@ -0,0 +1,506 @@
...
+    for (url, sub_name) in raw {
+        let parsed = Url::parse(&url)
+            .map_err(|e| invalid_config(format!("invalid subscription URL '{url}': {e}")))?;
+        if parsed.scheme() != "http" && parsed.scheme() != "https" {
+            return Err(invalid_config(format!(
+                "subscription URL '{url}' must use http or https"
+            )));
+        }
+
+        let effective_name = sub_name.unwrap_or_else(|| url.clone());
```
BtXin: [P3] Empty-string names defeat the URL fallback and produce an ungreppable duplicate error

`sub_name.unwrap_or_else(|| url.clone())` treats only `None` as "no name", but OPML 2.0 requires the `text` attribute, so title-less exporters emit `text=""` — which arrives as `Some("")`, becomes the effective feed name, and with two such outlines the whole source is rejected with `duplicate subscription name ''` (nothing the operator can grep their OPML for). `name: " "` sails through entirely. The `user_agent` check just above uses `.trim().is_empty()` — applying the same normalization here (empty/whitespace → `None` before the fallback) closes it.

BtXin (approved):
Sweep 4 (head cf00e35). Every finding from rounds 1–3 is verified fixed, including the follow-up commits (canonical-address trait contract, dual-stack test guard, `user_agent` header-value validation at config load, `max_concurrent` spec reconciliation, forthcoming-doc rewording). A fresh multi-angle pass (correctness, removed-behavior, cross-file, conventions, efficiency, reuse, altitude) found **no severe issues** — approving. The inline comments and notes below are follow-up material, none blocking.

Non-blocking notes beyond the inline comments:

- **Pooled-connection nuance in the seam docs.** `FeedFetcher::new` claims PolicyDns checks "every hostname connection … including pooled-connection reuse", but a pooled connection performs no resolution, so `check_ip` is not re-consulted on reuse. This is *not* a bypass — a reused connection only ever talks to the previously-approved address — but the guarantee is temporal: a policy revocation doesn't sever warm connections for up to the idle window (~90s default). Suggest rewording the doc to "policies must treat approval as valid for the life of a connection" (and optionally exposing `pool_idle_timeout` for Cloud). The test harness can't observe this because the mock sends `connection: close` on every response.
- **`EgressDenied.host` on the IP-literal path** is set to the canonical IP, so a denied `http://[::ffff:10.0.0.1]/` produces "host '10.0.0.1' resolves to private address 10.0.0.1" — a tautology that names neither the configured literal nor the subscription. Passing the URL host as written (like `check_addrs` does with the DNS name) would make `feeds.last_error` greppable.
- **Non-ASCII `Location` is a hard `InvalidUrl` failure** — a server emitting a raw-UTF-8 redirect path permanently kills the feed while browsers percent-encode and follow. Percent-encoding the non-ASCII octets before `Url::join` would recover the hop.
- **No total per-fetch deadline** (hop × attempt × wait products reach ~300s+ against a 60s `scan_timeout`) — carried note: phase 3's scan deadline and parse fuse must be the enforcement point; flagged there.
- **`.no_proxy()` operational note:** in a corporate egress-restricted network RSS is now the only provider that can't traverse the mandated proxy — total, undiagnosable feed outage while other sources work. Fine as the security default; consider a config knob later (honor proxy only when no policy is injected, or document loudly).
- **CI never runs doctests** (nextest doesn't support them), so the `RssConfig` doc example's assertions can silently rot; `cargo test --doc -p skardi --all-features` would be a cheap CI line. Related: `resolve_subscriptions`' intra-doc links to `pub(crate)` items warn under an all-features doc build (`private_intra_doc_links`) — the docs job is featureless today so it can't see them, but docs.rs builds all-features.
- **Duplication digest** (deliberate stack-phase scoping accepted, listing for a cleanup pass): `MockFeedServer` forks ~130 lines of `open_connector/testutil.rs`'s HTTP/1.1 mock; `backoff`/`MAX_RETRY_WAIT`/`RETRYABLE_STATUSES` are third copies of open_connector's (with already-diverged jitter); `read_body`'s metering loop mirrors `read_body_bounded`; `truncate` is the crate's third string-bounding helper. `util::http`'s own module doc claims this territory — a consolidation pass after M1 would pay for itself. Also: the file-wide `#![allow(dead_code)]` in fetch.rs/egress.rs could narrow to the root items so genuinely-orphaned helpers still warn.

gracexmatin:
Follow-up on the `.no_proxy()` operational note: implemented the suggested conditional in `e0bb5f3`, on the reasoning that the bypass `no_proxy()` defends against only exists when there is a policy to bypass.

`FeedFetcher::new` now takes `Option<Arc<dyn EgressPolicy>>`, and the switch is the **fact of injection**, never the policy's runtime behavior:

- **`None` (the OSS default)** — no destination filtering exists, so there is nothing a proxy could bypass: system proxy variables are honored, and feeds keep working on networks that mandate a proxy (the corporate case in the note).
- **`Some(policy)`** — `no_proxy()` applies as before, so the policy always sees the real destination. The module doc now states the operational consequence loudly: on a proxy-mandating network a policy-bearing fetcher fails every feed with connect timeouts while other providers work, by design; an operator in that position enforces destination policy at the proxy and injects none here.

A new subprocess test pair (`no_policy_fetcher_honors_proxy_env_vars`) pins the new half of the conditional — with no policy injected, the request goes to the proxy and the target is never contacted directly — mirroring the existing `proxy_env_vars_do_not_bypass_the_egress_policy`. Note for the stack: the signature change surfaces at phase 3's rebase, where the engine decides which side it constructs.

crates/skardi/src/sources/providers/rss/fetch.rs
gracexmatin: Fixed in `615136c`, taking the strict form: the success check is now `status == StatusCode::OK` rather than `is_success()`. A bare GET (no `Range`, no `A-IM`) has exactly one conformant success answer, so every other 2xx is a protocol mismatch and falls through to `FetchError::Status` — a visible per-feed degrade that re-fetches next scan, and correctly non-retried within the hop (the 2xx is the server's answer, not a transient fault). The module doc gains an "Only 200 is a success" section spelling out the sticky-validator failure mode you described.

crates/skardi/src/sources/providers/rss/fetch.rs
gracexmatin: Fixed in `18c52fc`, taking the first option: the shared helper now parses the date form. `httpdate` was already in the tree transitively (reqwest → hyper), so it's promoted to a direct dependency and tried after the integer parse fails; the date converts to a duration from now, with a past date yielding `Duration::ZERO` — "you may retry immediately" is what such a header means. The helper's stale rationale ("none of the APIs we integrate with emit it") is gone with it.

Callers keep their own fallback and cap unchanged, so for the rss fetcher this is bounded politeness rather than an uncapped sleep: `MAX_RETRY_WAIT` still clamps any wait at 10s — a host asking for a multi-minute pause gets ~10s gaps across the hop's attempts instead of sub-second re-hits, not the full named wait (an uncapped `Retry-After` stays a one-header DoS, per the existing design). New end-to-end test drives a 429 whose `Retry-After` is an http-date minted at now+2s and asserts the wait was honored.

crates/skardi/Cargo.toml
gracexmatin: The image-fetch builder now pins `no_gzip()` — that client wants raw image bytes and no content negotiation, so this simply restores its pre-`rss` behavior in every build: reqwest provides `no_gzip()` independent of the Cargo feature, so the line compiles (as a no-op) even when gzip isn't unified in. The Cargo.toml audit comment now names this client as the third in-crate one reached by the feature, and — unlike OC — the one with no streaming bound to fall back on, which is exactly why it gets the hard opt-out rather than a note.

crates/skardi/src/sources/providers/rss/config.rs
gracexmatin: Fixed in `3d88be1`. `finalize()` now applies the same blank-is-absent normalization the `user_agent` check uses: an empty or whitespace-only name is filtered to `None` before the fallback, so it falls back to the URL — non-empty, and greppable in any collision message. One deliberate boundary: non-blank names are used exactly as written, no trimming — `"a "` and `"a"` remain two distinct names, so existing configs keep their dedup semantics. Tests pin the two-blank-OPML-outlines case (resolves to distinct URLs instead of the ungreppable `duplicate subscription name ''`), the whitespace-only fallback, and the no-trim boundary.

PR #181, opened 2026-08-02
feat(sources): RSS provider 2/4 — sanitation ladder, HTML→Markdown, parsing, schemas, TTL cache

gracexmatin:
Phase 2 of 4 stacked PRs splitting #175 (plan Tasks 5-10). Base: `rss-m1-phase1` (#180); retarget to `main` after #180 merges.

## Scope
- **Sanitation ladder**: encoding repair incl. a lying non-UTF-8 declaration over valid UTF-8 bytes, control-char stripping, internal-DTD refusal run case-insensitively on both raw and sanitized bytes (closes the restore-after-guard evasions)
- **HTML→Markdown** via htmd behind a lexical `MAX_HTML_DEPTH = 100` pre-scan: htmd's DOM walk recurses unboundedly, and a ~7 KB feed with 600 nested divs aborts the process on a 2 MiB Tokio stack; above the ceiling content degrades to tag-stripped text. The scanner survived eight distinct bypass attempts during review (stray close tags, `<!-->`, foreign-content self-closing tags, …) and is mutation-verified.
- **Dialect conformance + parse driver**: RSS 0.9x/1.0/2.0, Atom, JSON Feed; field extraction per the spec's Field Mapping table (MIME subtype compared case-insensitively)
- **Arrow schemas**: items (17 cols) / feeds (15 cols), `skardi.rss.surface_version` metadata, builders routed through the `FeedStatus` domain type
- **Per-feed TTL cache**: LRU-evicted windows, health observations structurally never evicted, HTTP validators bundled with the window they validate, bounded observation map
- `docs/rss/semantics.yaml` ships here because a schema test pins the overlay against both schemas

## Verification
148 rss tests (cumulative); featureless check clean; `cargo doc -D warnings` clean both configs; fmt clean.

### DIFF
--- a/crates/skardi/src/sources/providers/rss/cache.rs
+++ b/crates/skardi/src/sources/providers/rss/cache.rs
@@ -1,0 +1,1038 @@
+//! Per-feed TTL cache: the health record and parsed window a scan reads
+//! between fetches.
+//!
+//! Each feed key holds two things that age independently:
+//!
+//! - an [`FeedObservation`] — the feed's health (status, last error, dialect,
+//!   item count, ...) — which persists across every call that doesn't remove
+//!   its whole entry: [`FeedCache::record_success`] replaces it wholesale,
+//!   [`FeedCache::record_failure`] and [`FeedCache::record_not_modified`]
+//!   mutate select fields of it in place, and only [`MemoryFeedCache`]'s
+//!   last-resort `max_observations` backstop (see below) ever discards one
+//!   outright; and
+//! - an optional [`CachedWindow`] — the parsed `items` batch plus the
+//!   conditional-GET validators that produced it — which the byte/entry
+//!   budget can evict independently of the observation.
+//!
+//! ## TTL re-arms on every attempt
+//!
+//! [`FeedCache::record_success`], [`FeedCache::record_not_modified`], and
+//! [`FeedCache::record_failure`] all take an `armed_until: Instant` the
+//! caller computes and push the feed's next-attempt time to exactly that
+//! instant, regardless of which of the three ran. This is what makes a dead
+//! feed a bounded cost rather than a per-scan one: a failure is retried at
+//! most once per [`failure_fuse`] window instead of on every scan, and a
+//! server's `Retry-After` (folded into the caller's `armed_until`) survives
+//! across scans instead of resetting the moment the attempt that read it
+//! finishes.
+//!
+//! ## Only a complete window is ever stored
+//!
+//! The only method that takes a window is [`FeedCache::record_success`] —
+//! there is no partial-window setter, so a half-parsed feed has no path into
+//! the cache.
+//!
+//! ## Eviction keeps the observation
+//!
+//! When the byte or entry budget forces a window out, [`MemoryFeedCache`]
+//! drops the [`CachedWindow`] (batch and validators together — they live in
+//! the same `Option`) but leaves the feed's [`FeedObservation`] in place, so
+//! a `feeds` scan — specified to be a pure state read that never fetches —
+//! never loses a feed's health just because its window happened to be
+//! evicted.
+//!
+//! A separate, much larger `max_observations` bound exists purely as a
+//! last-resort backstop: `feed` is a fully generic `&str`, nothing in this
+//! module restricts it to a known subscription list, so the map of
+//! observations alone (unlike the windowed subset above) has no bound
+//! without one. See [`MemoryFeedCache`]'s doc for why removing a whole entry
+//! there does not conflict with the guarantee just described.
+//!
+//! ## A `304` can outlive the window it would have confirmed
+//!
+//! Eviction is driven by *other* feeds' `record_success` calls, so it can land
+//! between the moment a caller reads a feed's validators out of
+//! [`FeedCache::snapshot`] and the moment the resulting `304` comes back. The
+//! conditional request was well-formed and the answer is truthful — the
+//! caller's copy *is* still current upstream — but the copy itself is gone, so
+//! there is nothing left to serve. This is a reachable state under memory
+//! pressure, not a caller bug: [`FeedCache::record_not_modified`] handles it
+//! by re-arming the timer like any other attempt and recording
+//! [`FeedStatus::Error`] with [`WINDOW_EVICTED_ON_REVALIDATION`] as the
+//! reason, so the zero rows that follow have an explanation an operator can
+//! read. See that method's doc for why `Error` rather than `Revalidated`.
+//!
+//! ## A single window larger than the whole byte budget
+//!
+//! [`MemoryFeedCache::record_success`] measures the incoming window with
+//! `RecordBatch::get_array_memory_size()` before inserting anything. If that
+//! alone exceeds `max_bytes`, the window can never fit no matter what else is
+//! evicted, so it is never stored — not even transiently — while the
+//! observation is still recorded. Any window the feed previously held is
+//! dropped too: `record_success` always replaces wholesale, so an
+//! oversized new window does not leave a stale one behind under the new
+//! observation's identity.
+//!
+//! ## No in-flight coalescing
+//!
+//! Two scans that both find the same feed's TTL expired can both fetch it —
+//! there is no request coalescing here, matching `open_connector`'s cache.
+//! That is a documented future extension, not this module's job.
+
+use std::collections::{HashMap, VecDeque};
+use std::sync::Mutex;
+use std::time::{Duration, Instant};
+
+use arrow::record_batch::RecordBatch;
+
+use super::error::{MAX_ERROR_CHARS, truncate};
+
+/// A feed's freshness state, and the single source of the exact strings the
+/// `feeds.last_status` column serves.
+///
+/// Callers must go through [`FeedStatus::as_str`] rather than writing the
+/// literal — `"never"`, `"fresh"`, `"revalidated"`, `"stale-error"`,
+/// `"error"` — themselves. Within this module, `as_str` is the only place
+/// that spells those strings for `feeds.last_status`, so a typo like
+/// `stale_error` fails to compile here instead of silently breaking that
+/// column's contract.
+///
+/// The same holds for `items.window_status`: `schema.rs`'s two sites both go
+/// through [`FeedStatus::window_status_str`] — `build_items_batch` stamps a
+/// freshly built window from [`FeedStatus::Fresh`], and `with_window_status`
+/// takes a [`FeedStatus`] rather than a `&str` — so neither column's domain
+/// can be broken by a typo at a call site.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum FeedStatus {
+    /// No fetch attempt has ever completed for this feed.
+    #[default]
+    Never,
+    /// The last attempt fetched a fresh body and parsed it successfully.
+    Fresh,
+    /// The last attempt got a conditional `304`; the cached window is still
+    /// current.
+    Revalidated,
+    /// The last attempt failed, but a window from an earlier success is
+    /// still cached and can be served stale.
+    StaleError,
+    /// The last attempt failed and no window is cached to fall back on.
+    Error,
+}
+
+impl FeedStatus {
+    /// The exact string this status serializes to as `feeds.last_status`.
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            FeedStatus::Never => "never",
+            FeedStatus::Fresh => "fresh",
+            FeedStatus::Revalidated => "revalidated",
+            FeedStatus::StaleError => "stale-error",
+            FeedStatus::Error => "error",
+        }
+    }
+
+    /// The `items.window_status` label a snapshot with this status should
+    /// serve, or `None` for the two statuses that serve zero rows (nothing
+    /// has ever been fetched, or the last attempt failed with nothing
+    /// cached to fall back on).
+    pub fn window_status_str(&self) -> Option<&'static str> {
+        match self {
+            FeedStatus::Fresh => Some("fresh"),
+            FeedStatus::Revalidated => Some("revalidated"),
+            FeedStatus::StaleError => Some("stale-error"),
+            FeedStatus::Never | FeedStatus::Error => None,
+        }
+    }
+}
+
+/// One feed's health, independent of whether its window is currently cached.
+///
+/// `Default` is the state of a feed that has never been attempted: every
+/// field `None` and [`FeedStatus::Never`].
+#[derive(Debug, Clone, Default)]
+pub struct FeedObservation {
+    pub last_fetch_ms: Option<i64>,
+    pub last_status: FeedStatus,
+    pub http_status: Option<u16>,
+    pub last_error: Option<String>,
+    pub dialect: Option<String>,
+    pub dialect_declared: Option<String>,
+    pub conformance_notes: Option<String>,
+    pub title: Option<String>,
+    pub site_url: Option<String>,
+    pub description: Option<String>,
+    pub item_count: Option<u64>,
+}
+
+/// A feed's cached window: the parsed `items` batch plus the conditional-GET
+/// validators that produced it. The two travel together — a batch is never
+/// kept without the validators that would let a future fetch revalidate it,
+/// and eviction drops both at once by dropping this whole struct.
+#[derive(Debug, Clone)]
+pub struct CachedWindow {
+    pub batch: RecordBatch,
+    pub etag: Option<String>,
+    pub last_modified: Option<String>,
+}
+
+/// A point-in-time read of one feed's cached state.
+#[derive(Debug, Clone)]
+pub struct FeedSnapshot {
+    pub observation: FeedObservation,
+    pub window: Option<CachedWindow>,
+    pub within_ttl: bool,
+}
+
+/// Per-feed TTL cache. A sync trait with a `Mutex` behind each implementation
+/// (see [`MemoryFeedCache`]) rather than an async one, so a later persistent
+/// implementation can swap in without touching callers — every observation of
+/// this cache's state flows through these four methods.
+pub trait FeedCache: Send + Sync {
+    /// Read `feed`'s current state as of `now`. Unknown feeds report
+    /// [`FeedStatus::Never`], no window, and `within_ttl: false`.
+    fn snapshot(&self, feed: &str, now: Instant) -> FeedSnapshot;
+
+    /// Record a successful fetch-and-parse: replace both the window and the
+    /// observation wholesale, and arm the feed's next-attempt time to
+    /// `armed_until`.
+    fn record_success(
+        &self,
+        feed: &str,
+        window: CachedWindow,
+        observation: FeedObservation,
+        armed_until: Instant,
+    );
+
+    /// Record a conditional `304`: the existing window is still current, so
+    /// it is left untouched, but the observation's fetch metadata and status
+    /// move to [`FeedStatus::Revalidated`], and the next-attempt time re-arms
+    /// to `armed_until`.
+    ///
+    /// If the window was evicted while this attempt was in flight — see the
+    /// module doc — the fetch metadata and the re-arm still apply (an attempt
+    /// happened, and skipping the re-arm would leave the feed expired and
+    /// refetching on every scan), but the status becomes
+    /// [`FeedStatus::Error`] with [`WINDOW_EVICTED_ON_REVALIDATION`] as
+    /// `last_error` rather than [`FeedStatus::Revalidated`].
+    ///
+    /// `Error` is the honest answer there even though the feed itself is
+    /// healthy. It is the one status whose contract is "no window is cached to
+    /// fall back on", which is exactly the situation; it is also the status
+    /// whose [`FeedStatus::window_status_str`] is `None`, so *from the next scan
+    /// on* it agrees with the zero rows that scan will serve. `Revalidated`
+    /// would instead claim a current window exists indefinitely — leaving
+    /// `feeds` reporting a healthy feed with a non-zero `item_count` while
+    /// `items` returns nothing, and no column anywhere explaining why.
+    ///
+    /// The serve that *records* this state is the one exception: the engine
+    /// already read a window out of its pre-permit snapshot and stamps those
+    /// rows `revalidated` from a literal, so for that one racing scan
+    /// `items.window_status = 'revalidated'` sits beside
+    /// `feeds.last_status = 'error'`. That is the same no-coalescing
+    /// consequence the engine's `304` branch documents — the rows it serves are
+    /// a real window the `304` really did vouch for — and it lasts exactly one
+    /// scan.
+    ///
+    /// Recovery does not depend on this choice: with no window there are no
+    /// validators to send, so the next attempt after the timer expires is an
+    /// unconditional `GET` that rebuilds the window whatever status was
+    /// recorded here.
+    fn record_not_modified(
+        &self,
+        feed: &str,
+        http_status: u16,
+        last_fetch_ms: i64,
+        armed_until: Instant,
+    );
+
+    /// Record a failed attempt: the observation's status becomes
+    /// [`FeedStatus::StaleError`] if a window is currently cached (serve
+    /// stale) or [`FeedStatus::Error`] if not, `last_error` is recorded, and
+    /// the next-attempt time re-arms to `armed_until` regardless — this is
+    /// the negative-caching half of the TTL contract.
+    ///
+    /// `dialect_declared` carries a sniff that succeeded even though the
+    /// attempt as a whole did not: the declared dialect is read off the raw
+    /// bytes before parsing, so a document that fails to parse still has one
+    /// to report. `Some` overwrites the recorded value, `None` leaves
+    /// whatever was already known in place — a transport failure has no
+    /// document to sniff and must not erase an earlier success's answer.
+    fn record_failure(
+        &self,
+        feed: &str,
+        http_status: Option<u16>,
+        error: String,
+        dialect_declared: Option<String>,
+        last_fetch_ms: i64,
+        armed_until: Instant,
+    );
+}
+
+/// `feeds.last_error` for a `304` whose window was evicted while the request
+/// was in flight — see [`FeedCache::record_not_modified`]. Names no particular
+/// bound, since any of the three can produce the state: `max_bytes` and
+/// `max_entries` through `Inner::evict`, and the `max_observations` backstop
+/// through `Inner::evict_observations`. Carries no response content of any
+/// kind: the condition is entirely local to this cache, and the string is a
+/// fixed literal.
+pub const WINDOW_EVICTED_ON_REVALIDATION: &str = "revalidated (304) but the cached window had already been evicted from the feed \
+     cache; the next attempt refetches it unconditionally";
+
+/// The negative-cache TTL for a feed that just failed: `clamp(ttl / 4, 30s,
+/// 300s)`. Shorter than the success TTL so a dead feed is retried sooner
+/// than a healthy one's normal refresh, but still bounded away from zero
+/// when `ttl_seconds` is 0 (always-live) so a failing always-live feed is not
+/// retried on literally every scan.
+pub fn failure_fuse(ttl: Duration) -> Duration {
+    (ttl / 4).clamp(Duration::from_secs(30), Duration::from_secs(300))
+}
+
+/// How many times `max_entries` the last-resort whole-entry bound
+/// (`max_observations`) is set to, before the [`MIN_OBSERVATIONS`] floor is
+/// applied. `max_entries` bounds feeds holding a window, sized by the
+/// operator for a realistic number of concurrently fetched feeds; feeds that
+/// are failing or being revalidated hold no window at all and so don't count
+/// against it, yet still occupy a map slot. 8x gives that traffic
+/// comfortable headroom over the windowed bound before the backstop below
+/// ever engages, while still capping unbounded growth in the number of
+/// distinct feed keys ever seen (`feed` is a plain `&str` — nothing here
+/// restricts it to a known subscription list).
+const MAX_OBSERVATIONS_MULTIPLIER: usize = 8;
+
+/// Floor under `max_observations`, so a caller-supplied `max_entries` of `0`
+/// (window caching fully disabled — a coherent configuration: `feeds`
+/// health is still worth tracking with no window cache at all) cannot turn
+/// the backstop into "discard the observation the instant it's inserted."
+/// 8 mirrors `1 * MAX_OBSERVATIONS_MULTIPLIER` — the headroom the smallest
+/// *window-caching* configuration (`max_entries == 1`) already gets — so
+/// disabling window caching entirely doesn't collapse observation tracking
+/// below what the smallest windowed configuration would have.
+const MIN_OBSERVATIONS: usize = 8;
+
+/// A cached window plus the byte count it was charged against the budget
+/// with, so eviction can subtract exactly what was added.
+struct WindowEntry {
+    window: CachedWindow,
+    bytes: usize,
+}
+
+/// One feed's full cache entry: an always-present observation, and an
+/// optional window that the byte/entry budget can evict independently.
+struct Entry {
+    observation: FeedObservation,
+    window: Option<WindowEntry>,
+    armed_until: Instant,
+}
+
+struct Inner {
+    map: HashMap<String, Entry>,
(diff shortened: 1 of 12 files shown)


crates/skardi/src/sources/providers/rss/parse.rs
```diff
@@ -0,0 +1,1105 @@
...
+        title: feed.title.as_ref().map(|t| t.content.clone()),
+        site_url: preferred_link(&feed.links).map(|l| l.href.clone()),
+        description: feed.description.as_ref().map(|t| t.content.clone()),
+    };
+
+    let mut items = Vec::with_capacity(feed.entries.len());
+    let mut skipped_without_identity = 0;
+    for entry in &feed.entries {
+        match extract_entry(entry) {
+            Some(row) => items.push(row),
```
BtXin: [P1] Enforce the declared `(feed, guid)` key before building the window

Every extracted entry is appended without checking whether an earlier entry produced the same `guid`. A feed containing duplicate IDs, or two guid-less entries with the same fallback link, therefore exposes duplicate `(feed, guid)` rows. In the documented archive statement both rows miss the anti-join and are inserted together, causing the SQLite `PRIMARY KEY (feed, guid)` constraint to fail and blocking the sync. The design also says link fallback deduplication collapses to link identity. Please deduplicate deterministically in feed order, record a conformance note for dropped duplicates, and add duplicate-id and duplicate-link tests.

crates/skardi/src/sources/providers/rss/sanitize.rs
```diff
@@ -0,0 +1,713 @@
...
+    debug_assert_eq!(rest[0], b'&');
+    let tail = &rest[1..];
+
+    for name in [&b"amp;"[..], b"lt;", b"gt;", b"apos;", b"quot;"] {
+        if tail.starts_with(name) {
+            return Some(1 + name.len());
+        }
+    }
+
+    // `&#[0-9]{1,7};` or `&#x[0-9A-Fa-f]{1,6};` (XML allows only a lowercase `x`).
```
BtXin: [P2] Preserve numeric references with leading zeros

The fixed digit window intentionally rewrites valid references such as `&#000000169;` to literal entity text whenever the ampersand-repair rung runs for another defect in the document. That contradicts the stated contract that every sanitation rung is a byte-level no-op over valid references. The scan can stay bounded without imposing this byte-count restriction: consume a bounded run of leading zeros, validate the significant value/digits, and require the terminator. Please add decimal and hexadecimal leading-zero cases to the conservativeness test.

BtXin (changes requested):
Requesting changes on two data-contract violations in the parsing layer. Duplicate feed identities can break the documented archive pipeline, and the sanitation rung knowingly mutates a class of valid numeric references instead of preserving the byte-level conservativeness contract.

crates/skardi/src/sources/providers/rss/cache.rs
```diff
@@ -0,0 +1,1038 @@
...
+#[derive(Debug, Clone, Default)]
+pub struct FeedObservation {
+    pub last_fetch_ms: Option<i64>,
+    pub last_status: FeedStatus,
+    pub http_status: Option<u16>,
+    pub last_error: Option<String>,
+    pub dialect: Option<String>,
+    pub dialect_declared: Option<String>,
+    pub conformance_notes: Option<String>,
+    pub title: Option<String>,
```
bakey, line 178: **[P2] Feed-controlled observation strings are unbounded in a store that is deliberately never byte-bounded — the exact class the `unknown:{root}` cap already fixed, applied to only one of its members.**

`conformance.rs:29-33` states the rule precisely: a feed-controlled string retained in a `FeedObservation` must be capped, *because* `MemoryFeedCache` meters `RecordBatch` bytes only and never byte-bounds observations. The root-element name got the cap. But the same struct carries `title`, `description`, `site_url`, `conformance_notes`, and `dialect_declared` — all feed- or server-controlled, none capped anywhere in this PR:

- `FeedMeta.title`/`description` come from `feed.title.content` verbatim (`parse.rs:227-229`); a single `<title>` can be `max_response_bytes`-sized and lands here whole.
- `content_type_family_note` (`conformance.rs:200-202`) interpolates the served `Content-Type` essence — an attacker-server-controlled header — into a note string with no truncation, while its sibling `unknown:{name}` in the same file is truncated.
- `sniff_declared_dialect`'s cap protects `dialect_declared` only on the `unknown:` arm; fine — but it shows the intent.

Worst case is bounded by `max_response_bytes × max_observations` rather than unbounded, and feed *keys* are operator-configured, so this is resource-consistency rather than a DoS hole — but the module already committed to the invariant "no feed-controlled string enters the observation store uncapped", and today it holds for exactly one string out of six. Suggest routing the observation's feed-controlled fields through `truncate(…, MAX_ERROR_CHARS)` (or a dedicated larger cap for title/description if 512 chars is too aggressive for real titles) at the point they enter `FeedObservation`, and truncating the essence in `content_type_family_note` like its sibling.

crates/skardi/src/sources/providers/rss/parse.rs
```diff
@@ -0,0 +1,1105 @@
...
+    pub dialect_declared: Option<String>,
+    pub repairs: Vec<Repair>,
+}
+
+/// One rung of the ladder: the transform and the repair it records.
+type Rung = (fn(&[u8]) -> (Vec<u8>, bool), Repair);
+
+/// Sanitize `bytes` with the applicable rungs, then parse them into a feed.
+pub fn parse_with_ladder(bytes: &[u8]) -> Result<ParseSuccess, ParseFailure> {
+    let family = detect_family(bytes);
```
bakey: **[P3] The document family is decided once, on raw bytes — so a UTF-16BE JSON Feed is classified XML and the ampersand rung then corrupts its string content.**

`detect_family` sees UTF-16LE JSON by accident of byte order (`{` = `7B 00`, first byte matches) but not UTF-16BE (`00 7B` — `0x00` is not ASCII whitespace, so the loop returns `Xml` at the first byte). Verified: `b"\xFE\xFF" + '{"title":"Fish & Chips"}'.encode('utf-16-be')` → family `Xml`. From there the XML rung set runs: rung 1 transcodes correctly (BOM is authoritative), but rung 3 then escapes the naked `&` *inside a JSON string literal* — the precise corruption the `DocFamily::Json` arm exists to prevent (`parse.rs:63-64`) — and feed-rs happily parses the result, so `items.title` silently becomes `Fish &amp; Chips` with a `sanitation: escaped-naked-ampersands` note attributing the repair to a document that needed none.

JSON Feed 1.1 mandates UTF-8, so the input is non-conforming — but rescuing non-conforming documents *without corrupting them* is this module's charter (AC16's spirit, if not its letter, since the input isn't well-formed bytes). Cheapest fix with the current structure: re-run `detect_family` on the rung-1 output and pick the remaining rungs from the *post-transcode* family; alternatively teach `detect_family` the two-byte `00 7B` / `7B 00` patterns after a UTF-16 BOM so both endiannesses classify the same. Either way a `utf16be_json_feed_title_survives_the_ladder` test would pin it.

crates/skardi/src/sources/providers/rss/parse.rs
```diff
@@ -0,0 +1,1105 @@
...
+    from_media: bool,
+}
+
+/// The item's primary attachment.
+///
+/// `feed-rs` folds RSS 2.0 `<enclosure>` and MediaRSS into `media`, but *not*
+/// Atom `rel="enclosure"` links or JSON Feed attachments — both of those stay in
+/// `links` (a JSON attachment arrives `rel`-less but carries a `media_type`,
+/// which is what distinguishes it from the item's own URL).
+fn enclosure(entry: &Entry) -> Enclosure {
```
bakey, line 418: **[P3] Second and subsequent Atom/JSON Feed attachments are silently dropped, while the equivalent MediaRSS leftovers are preserved — the same concept survives or vanishes depending on dialect.**

`enclosure()` surfaces the first attachment; `remaining_media` then preserves every *other* MediaRSS content under `extensions_json.media`. But leftover attachment **links** — Atom's additional `rel="enclosure"` links, JSON Feed's `attachments[1..]` (the spec's own example use case is one episode in multiple audio formats) — are never revisited: `extensions_json` iterates `entry.media` only, so they exist in the feed-rs model and reach nothing.

Two consequences: (a) the module doc's contract for `extensions_json` — "whatever the `feed-rs` model exposes beyond the pinned columns" — doesn't hold for links-shaped attachments; (b) cross-dialect inconsistency for identical content: an RSS feed with two MediaRSS contents keeps both (one as the enclosure, one in extensions), while the same podcast published as JSON Feed with two attachments keeps only one, with nothing in any column recording that the second existed.

Suggest folding leftover attachment links (`is_attachment(l)` minus the one surfaced, when `!from_media`) into the same `extensions_json.media`-style array — they already carry the url/content-type/length triple the media objects serialize — or, if deferring, narrowing the module-doc claim and noting the asymmetry where `enclosure()` documents the fold.

crates/skardi/src/sources/providers/rss/cache.rs
```diff
@@ -0,0 +1,1038 @@
...
+        self.window_order.retain(|k| k != feed);
+        self.window_order.push_front(feed.to_string());
+    }
+
+    /// Move `feed` to the front of the whole-entry LRU list. Called on every
+    /// access or update, whether or not `feed` currently holds a window, so
+    /// an observation-only entry that keeps getting queried or refreshed is
+    /// exactly as protected from the `max_observations` backstop as a
+    /// windowed one.
+    fn touch_entry(&mut self, feed: &str) {
```
bakey, line 472: **[P3, perf] Every cache access pays two `VecDeque::retain` passes plus a `String` allocation — O(total entries) per touch, quadratic across a scan.**

`touch_entry` (and `touch_window` when a window exists) runs `retain` over the whole order list and re-allocates the key on every `snapshot`/record call, and `snapshot` fires once per feed per scan. With `max_observations = 8 × max_entries`, a scan over N subscribed feeds does O(N × 8·max_entries) work: negligible at the defaults, but `max_entries` is operator-sized to the subscription count, so a 10k-feed deployment pays ~10⁸ comparisons and 20k allocations per full sweep of `feeds`/`items` — for a cache whose actual hit work is a couple of `HashMap` probes.

A cheap mitigation that keeps the hand-rolled-LRU precedent: store a monotonic `last_used: u64` tick on each entry and evict by min-scan — eviction is the rare path and becomes O(n) only when a bound is actually exceeded, while the hot path (every snapshot) drops to a field write. Not blocking; worth a `// perf:` note at minimum so the shape is a recorded decision rather than an accident.

bakey (commented):
Reviewed phase 2 end to end (sanitize / parse / convert / conformance / schema / cache + the semantics overlay) from business correctness, consistency, and performance angles, after reading the PR motivation and the phase-1 base. Overall: the defensive depth here is exceptional — the depth-scan bypass catalogue, the conservativeness contract, and the 304-vs-eviction race handling are all pinned by tests that explain themselves. Four findings below, none overlapping the existing review; the first is the only one I'd want resolved before merge, and it's an application of a rule this PR itself already established.

Angle coverage, for the record: correctness → findings 1-2; consistency → finding 3 (cross-dialect data loss) and the overlay test (clean); performance → finding 4 plus a pass over the ladder's copy behavior (bounded by `max_response_bytes`, acceptable) and `find_sub` usage (linear in practice, fine).

crates/skardi/src/sources/providers/rss/conformance.rs
```diff
@@ -0,0 +1,414 @@
...
+
+use super::error::{MAX_ERROR_CHARS, truncate};
+use super::sanitize::{DocFamily, find_sub};
+
+const ATOM_1_0_NS: &str = "http://www.w3.org/2005/Atom";
+const ATOM_0_3_NS: &str = "http://purl.org/atom/ns#";
+const JSON_FEED_VERSION_MARKER: &[u8] = b"jsonfeed.org/version/";
+
+/// The dialect the document declares, sniffed lexically.
+pub fn sniff_declared_dialect(bytes: &[u8], family: DocFamily) -> Option<String> {
```
BtXin, line 18: [P3] Dialect sniffing runs on raw pre-rung bytes, so valid UTF-16 feeds report NUL-laden garbage

`sniff_declared_dialect` is fed the original bytes (parse.rs:50), so a perfectly valid BOM'd UTF-16LE RSS 2.0 feed — one that parses fine after rung 1 — gets `dialect_declared = "unknown:\0?\0x\0m\0l…"`: `first_start_element` finds the `<` byte of the UTF-16 `<?xml`, the `rest.starts_with(b"<?")` check fails on the interleaved 0x00, and `from_utf8_lossy` keeps the NULs, which land in the user-facing `feeds.dialect_declared` column (capped at 512 chars but containing control characters). Sniff after rung 1's transcode, or skip the sniff when a UTF-16 BOM is present.

crates/skardi/src/sources/providers/rss/convert.rs
```diff
@@ -0,0 +1,929 @@
...
+/// is roughly 2.5x, not a wide margin. 203 is an empirical bound over those
+/// shapes, not a proof — and the whole argument presupposes the scan really is
+/// an upper bound on open-tag depth, which it only became once the stale
+/// `foreign` undercount was fixed (see the `foreign` counter in
+/// [`max_open_tag_depth`]).
+const MAX_HTML_DEPTH: usize = 100;
+
+/// Convert an HTML fragment to Markdown.
+pub fn html_to_markdown(html: &str) -> String {
+    if max_open_tag_depth(html) > MAX_HTML_DEPTH {
```
BtXin, line 52: [P3] Depth guard overcounts unclosed `<p>`/`<li>` — common legacy feed HTML silently degrades to stripped text

`max_open_tag_depth` never pops on html5ever's implied end tags, so `"<p>para".repeat(150)` — ordinary unclosed-paragraph HTML that real feeds emit — scans as depth 150 against a real DOM depth of ~4 and trips `MAX_HTML_DEPTH`, degrading the whole document to tag-stripped text (all Markdown structure lost). The safe direction of the tradeoff is documented, but nothing acknowledges or tests this shape; the only balanced-content guard test uses closed `<p>x</p>`. Auto-closing the standard implied-end-tag set (`p`, `li`, `td`, `tr`, `dd`, `dt`, `option`) in the scanner would remove the most likely real-world degrade.

crates/skardi/src/sources/providers/rss/parse.rs
```diff
@@ -0,0 +1,1105 @@
...
+///
+/// `feed-rs` fills a missing id itself: from a content hash when there is a link
+/// to hash, and from a **random UUID** otherwise. The random case is unusable as
+/// identity — the same item would get a new `(feed, guid)` on every scan, breaking
+/// window identity and idempotent archiving — and both cases mask the
+/// "no identity at all" entry that spec decision 5 says to skip and count. Handing
+/// `feed-rs` a generator that yields nothing leaves `entry.id` empty unless the
+/// feed really supplied one, so the Field Mapping fallback chain (id → link →
+/// skip) is what decides identity, deterministically.
+fn parser() -> feed_rs::parser::Parser {
```
BtXin: [P3] Convention: import crate types at the top

`feed_rs::parser::Parser` / `feed_rs::parser::Builder` here, and `encoding_rs::Encoding` / `encoding_rs::UTF_16LE`/`UTF_16BE`/`UTF_8`/`WINDOWS_1252` in sanitize.rs, are written as full inline paths in signatures and bodies. Project rule: crate types are imported via `use` at the top of the file.

BtXin (commented):
Sweep 2 (head 6312d4d). Prior-thread status:

- **[P1] parse.rs duplicate `(feed, guid)` — NOT FIXED.** Still no dedup anywhere on the `extract()` → `build_items_batch` path; two id-less entries sharing a `<link>` both get `guid = link`. Made worse since the last round: `docs/rss/semantics.yaml:186` now *promises* "Stable item identity, unique within a feed" and :150 sells `(feed, guid)` as the item identity. Either enforce the key (last-wins or first-wins with a conformance observation) or retreat the doc claim — and add a duplicate-guid test either way. This is the one remaining P1 on this PR.
- **[P2] leading-zero numeric refs — converted to a documented "accepted tradeoff"** (sanitize.rs:388-394) rather than fixed. Note the annotation contradicts the rung's own contract ("by contract (spec AC16), a byte-level no-op on well-formed input" — `&#000000169;` is well-formed XML). If the tradeoff stands, the AC16 contract wording needs the same carve-out; as written the module disagrees with itself.
- **[P2] unbounded observation strings — PARTIALLY fixed.** Count backstop (`max_observations`), capped `unknown:{root}`, capped in-cache `last_error` all landed. But strings arriving through `record_success` (title/description/conformance_notes) are still stored byte-unbounded — the store remains count-bounded, never byte-bounded. Confirm the phase-3 engine caps these upstream, or cap at the cache boundary.
- **[P3] UTF-16 family detection — PARTIALLY fixed.** UTF-16LE-no-whitespace now classifies as JSON, but UTF-16BE (first byte of `{` is 0x00) and any UTF-16 with leading whitespace (the 0x00 half of the space code unit is hit first) still classify as XML, so the ampersand rung corrupts legal `&` inside JSON strings. Reproduced against the exact `detect_family` logic.
- **[P3] 2nd+ Atom/JSON-Feed attachments — NOT FIXED**, and semantics.yaml:272-274 now documents the leftover keys as exactly "media, source, rights, language", cementing the asymmetry with MediaRSS.
- **[P3 perf] double `VecDeque::retain` per cache touch — NOT FIXED**, but both deques are now hard-bounded so the absolute cost is small. Fine to keep deferring.

Also verified clean this round: post-sanitation DTD-guard re-run with evasion tests, decode-based misleading-declaration check, deterministic-id parser so feed-rs UUIDs never become guids, `record_not_modified` re-arm after eviction, schema/semantics drift test, no `.unwrap()` violations, correct lock-poisoning recovery.

crates/skardi/src/sources/providers/rss/parse.rs
gracexmatin: Fixed in a557008.

Deduplication now happens in `extract()`, after the id → link fallback chain has already resolved, so duplicate ids and colliding link fallbacks are covered by the same check.

- Dedup is deterministic in feed document order: the first entry to claim a guid keeps it (`seen_guids: HashSet` as a match guard); later claimants are dropped and counted in a new `ExtractedFeed::duplicate_identity` field. Note that under the usual newest-first feed layout, the first occurrence in document order is also the freshest render.
- Dropped duplicates leave a `duplicate-identity: <n>` conformance note, emitted alongside the existing `entries-without-identity: <n>` note and added to the note shapes documented in `docs/rss/semantics.yaml`.

crates/skardi/src/sources/providers/rss/sanitize.rs
gracexmatin: Fixed in a557008.

The numeric-reference scan in `valid_reference_len` is now a two-segment constant-bound scan: first a bounded run of leading zeros (new constant `MAX_LEADING_ZEROS = 16`, well beyond any real encoder's fixed-width padding), then the existing 7-digit (decimal) / 6-digit (hex) window over the remaining significant digits, then the required `;` terminator. Valid references like `&#000000169;` now pass through rung 3 untouched instead of being rewritten to literal entity text.

- The scan stays constant-bound (≤ 17 + 7 + 1 bytes), so the original bounded-scan property is unchanged; a reference whose zero run exceeds the cap is still treated as naked and escaped — the bound remains, it is just leading-zero-aware now rather than a total-digit count. One implementation detail: the over-cap 17th zero is itself a valid digit and would be absorbed by the significant-digit window, so an over-cap zero run is rejected explicitly, keeping the "beyond the cap ⇒ escaped" semantics true.
- The new acceptance set is a strict superset of the old one (the old window allowed at most 7 digits total, far under the 16-zero cap), so there is no newly-escaping regression direction; all-zero references such as `&#000;` remain accepted — value-range validation stays feed-rs's job, the rung is lexis-only.
- The "Accepted tradeoff" comment block is gone, replaced by a statement of the new bound — that tradeoff contradicted AC16 (every rung is a byte-level no-op on well-formed input) and was never necessary.

crates/skardi/src/sources/providers/rss/parse.rs
gracexmatin: Fixed in 5e9ac01.

`parse_with_ladder` now runs rung 1 first (it applies to both families anyway) and re-runs `detect_family` on the transcoded bytes, so rungs 2 and 3 are chosen from what the document turned out to be rather than from what its byte order made it look like. Took the re-detect option over teaching `detect_family` the `00 7B` / `7B 00` patterns because it generalizes: any document rung 1 rewrites — a mislabelled encoding, not just BOM'd UTF-16 — is now classified after normalization rather than before.

- The pre-sanitation DTD guard and the declared-dialect sniff keep reading the raw bytes and so keep the raw family (now named `raw_family`); the post-sanitation guard moves to the post-transcode family, which is what makes it skip a document that only looked like XML before rung 1 ran. `a_mislabelled_utf16_documents_subset_is_refused_before_the_rungs_run` and `utf16_doctype_with_internal_subset_is_refused_after_reencoding` are unaffected — both are XML in either reading.
- No behavior change for UTF-8 input: rung 1 is a byte-level no-op there, so the second detection reads the same bytes as the first. The two readings can only diverge on a document rung 1 rewrote, which is the case being fixed.
- Tests: `utf16be_json_feed_title_survives_the_ladder` as you suggested — the title keeps its naked `&` and `repairs` is exactly `[ReencodedToUtf8]`, so both the corruption and the false `escaped-naked-ampersands` attribution are pinned. Added the symmetric `utf16le_json_feed_title_survives_the_ladder`: it passes today only by accident of byte order, so it is worth holding the two orders to one answer.

**Unsolved, adjacent — flagging rather than fixing here.** `sniff_declared_dialect` is still called on the raw bytes, and it has the same raw-bytes-are-not-the-document problem in a different place: for a UTF-16 XML document `first_start_element` scans byte-by-byte for `<`, so it reads the root element name with NULs interleaved (UTF-16LE `<rss` is `3C 00 72 00 73 00 73 00`) and `dialect_declared` becomes `unknown:r\0s\0s\0` — a NUL-carrying value in the column. For the UTF-16 JSON case in this finding it fails more quietly: the ASCII `jsonfeed.org/version/` marker is not found in UTF-16 bytes, so the column is NULL. The new test asserts that NULL so that closing the gap has to update the line. Moving the sniff after rung 1 would fix both under the same principle as this change, but it is a separate surface (the sniff feeds a different column and has its own "works on bytes too broken to parse" charter), so it is left for its own change rather than folded in here.

BTW I have noticed that the unsolved problem is mentioned below by BtXin, so it will be fixed in the following comment.

crates/skardi/src/sources/providers/rss/parse.rs
gracexmatin: Fixed in 5e9ac01.

Fixed by folding both sources into one concept rather than adding a second array. The `media` key becomes `attachments`: every attachment past the one already in the `enclosure_*` columns, one flat object per downloadable file carrying `url`/`content_type`/`size`, wherever `feed-rs` kept it.

MediaRSS carries per-object metadata a link cannot (`title`, `description`, `thumbnails`, `duration_secs`), and one object may hold several renditions under a single set of it. Flattening copies that metadata onto each rendition instead of nesting them, so every element is self-describing and the array has one shape.

- Two edges worth naming. The skip direction inverts between the two sources: `enclosure()` only reaches for a link when `media` had nothing to give, so when `from_media` is true **every** attachment link is left over rather than all but the first — getting that backwards would silently drop the first link of any entry carrying both kinds, and a test asserts both directions. And a media object whose renditions were all consumed still emits an element carrying its metadata alone, preserving the existing behaviour of not dropping such objects.
- `extensions_json`'s doc no longer claims "whatever the feed-rs model exposes beyond the pinned columns" — you were right that it did not hold; it now states the named set it actually projects. `enclosure()` documents that it decides the columns and never whether a file is recorded at all. `docs/rss/semantics.yaml` gains the new key and its shape, and `enclosure_url` gains a pointer that an entry may carry several attachments and only the first reaches those three columns.

crates/skardi/src/sources/providers/rss/cache.rs
gracexmatin: Fixed in 5e9ac01.

Every feed- or server-controlled string is now bounded at one boundary rather than at each writer: `FeedObservation::capped()`, applied where an observation enters the store. Two caps, because the fields are two kinds of text — `title` and `description` are the feed's own prose and take a new `MAX_FEED_TEXT_CHARS` (4096, where 512 would cut legitimate channel descriptions); `site_url`, `dialect_declared`, `conformance_notes`, and `last_error` take `MAX_ERROR_CHARS`. `dialect` is deliberately absent: its only writer returns a `&'static str` from a closed set. `last_error` is included even though its writers already cap it, so the guarantee at this boundary is total rather than inherited from every upstream remembering to.

- Applied at both cache write paths, not just `record_success`. `record_failure` writes `error` and `dialect_declared` straight into a retained observation, so without it the struct's invariant would have been aspirational. The invariant now lives on `FeedObservation` itself, and says that adding a feed-controlled field means adding it to `capped()` too.
- One correction to the finding: `content_type_family_note` cannot interpolate an attacker-controlled essence. `essence` only reaches the format string after matching one of eight exact literals — everything else hits `_ => return None` — so the note has always been bounded at ~60 characters. I added the truncation anyway, and the comment says plainly that it is a no-op today: it exists so that widening that match (a prefix arm, a wildcard, a vendor `+xml` rule) cannot quietly put an unbounded header into the column, and so no single note can consume the whole allowance `capped()` gives the joined string.

crates/skardi/src/sources/providers/rss/cache.rs
gracexmatin: Fixed in 5e9ac01.

Both `VecDeque<String>` order lists are gone. Recency is a tick recorded on the entry itself, so a touch is a `u64` write instead of a scan of every other feed's key. The order is only ever *read* when a bound is exceeded, so it is recorded in O(1) and derived by a min-scan in `evict`/`evict_observations`.

- Two ticks rather than one. `window_last_used` and `last_used` track different notions of "used". `record_failure` on a feed that still holds a window touches the entry but not the window (the stale-error path), and `snapshot` touches the window only when it actually served one. Collapsing them into a single field would silently reorder eviction.
- `max_entries` bounds the number of windowed feeds, which the deque used to answer with `.len()`. That is now an incrementally maintained `windowed` count, so the check stays O(1) instead of becoming a scan on every write — it changes in exactly two places (`store_window`, `drop_window`), and `evict_observations` credits through `drop_window` rather than keeping a second copy of the arithmetic.
- Ticks are strictly increasing, so no two accesses compare equal and the eviction order is total — not a partial order broken by `HashMap` iteration order.

crates/skardi/src/sources/providers/rss/conformance.rs
gracexmatin: Fixed in 04a89e6.

The sniff now splits by path, on the same principle as the family fix: each path sniffs the bytes it is actually about to act on.

- The main path sniffs the transcoded bytes, beside the post-transcode family decision — the post-sanitation guard refusal, the strict-parse failure, and the success all share that one sniff. What the sniff is *for* is unchanged: rung 1 converts encodings, it does not parse, so a document too broken to parse is exactly as sniffable as before.
- The pre-sanitation guard's refusal branch keeps a raw-bytes sniff, because it refuses before the rungs run and there is nothing else to sniff. It cannot produce the NUL-laden form: that guard fires on a literal ASCII `<!DOCTYPE`, which UTF-16's interleaved NULs hide — a UTF-16-carried subset is caught by the post-sanitation guard instead, after the transcode. The comment states that dependency.
- This also fixes the quieter JSON half: the ASCII `jsonfeed.org/version/` marker is invisible in UTF-16 bytes, so a BOM'd UTF-16 JSON Feed's `dialect_declared` was NULL. The earlier family-fix test pinned that NULL with a note that closing the gap would have to update the line — it now asserts `json-feed-1.1`.
- New test: a valid BOM'd UTF-16 RSS 2.0 feed, both byte orders, asserting `dialect_declared = "rss-2.0"` with transcoding as the only repair — the exact document that used to report `unknown:\0?\0x\0m\0l`.

crates/skardi/src/sources/providers/rss/convert.rs
gracexmatin: Fixed in 04a89e6.

The scan now pops the sibling-closing implied end tags: a new `implied_end_pops` encodes html5ever's rules — `p` closed by every sibling-closer on its way in (`li`/`dd`/`dt` explicitly, the table tags via the cell they close), `li` by `li`, the definition and cell pairs by either sibling, `tr` by the next row (popping the open cell first), `option` by `option`. `"<p>para".repeat(150)` scans as depth 1 and converts as 150 paragraphs.

- The never-under-count obligation holds two ways. The pops consult only the top of the stack, so an element html5ever would not close — an `li` under a nested `<ul>`, anything shielded by a container — keeps its full count, and `"<ul><li>".repeat(60)` still trips the ceiling (pinned by a test). And where the scan's stack disagrees with html5ever's, the stale entry is one the scan itself pushed for a token html5ever ignored outright (`<td>` outside any table), so popping it retires the scan's own surplus while the push that follows restores it — the estimate stays an upper bound.
- Suspended in foreign content like the void-element skip, and for the same reason: `foreign` is only trusted where believing it too readily over-counts.
- The 203 sweep figure in `MAX_HTML_DEPTH`'s doc predates the pops; a caveat records that they lower the scan only on shapes html5ever flattens the same way — an argument, not a re-measurement.
- Tests: the exact review shape (depth exactly 1, and the Markdown output keeps paragraph structure rather than degrading to a single run of text), the rest of the sibling set in their legacy shapes (`li` runs, `dt`/`dd`, `tr`/`td` rows, `option`, `li` closing the `p` left open inside the previous `li`), and the counter-direction: genuinely nested lists still count and still degrade.

crates/skardi/src/sources/providers/rss/parse.rs
gracexmatin: Fixed in 04a89e6.

`use encoding_rs::{Encoding, UTF_8, UTF_16BE, UTF_16LE, WINDOWS_1252};` at the top of `sanitize.rs` (which previously had no top-level imports at all) and `use feed_rs::parser::{Builder, Parser};` in `parse.rs`, with the ten inline paths shortened. The three `feed_rs::parser::parse(…)` calls in tests stay qualified: it is a free function rather than a type, and a bare `parse` in a file named `parse.rs` full of its own `parse_*` functions would cost more clarity than the prefix does.

BtXin (approved):
Verification sweep (head 2bc29fe). **All nine open findings are genuinely fixed, each with test coverage** — approving.

- **[P1] duplicate `(feed, guid)`**: `extract()` dedupes via `HashSet`, first-in-document-order wins, policy documented, `duplicate-identity: {n}` conformance note emitted, three tests (explicit dup, link-fallback collision, negative case); semantics.yaml's uniqueness claim now holds.
- **[P2] leading-zero refs**: bounded zero-run scan (`MAX_LEADING_ZEROS = 16`) ahead of the significant-digit window — `&#000000169;` survives; edge cases hand-checked (empty ref, non-digit after zeros, over-cap, `&#0;`).
- **[P2] observation caps**: `FeedObservation::capped()` (4096 chars for title/description, 512 for the rest) enforced at both write paths, char-boundary-safe, multi-byte pinned.
- **[P3] UTF-16 family**: family + dialect sniff now run post-transcode; both byte orders tested, `dialect_declared` NUL-free.
- **[P3] attachments**: 2nd+ Atom/JSON-Feed/MediaRSS attachments fold into `extensions_json.attachments`; the media-vs-link skip inversion verified correct and pinned.
- **[P3 perf] LRU**: tick-based, O(1) on record, true-LRU eviction, accounting cross-checked by test.
- **[P3] depth scan**: `implied_end_pops` verified pair-by-pair against html5ever semantics — no under-count path found; legacy unclosed-`<p>` runs convert as paragraphs while genuinely deep DOMs still trip the guard (incl. an ~1000-deep adversarial battery).
- Convention + doc-link fixes verified (featureless `cargo doc -D warnings` clean, run locally with CI's exact flags).

Local suite: 182/183 rss tests pass (1 deliberately ignored); clippy clean in `providers/rss/*`.

Four P4 nits for whenever (none blocking): `MemoryFeedCache`'s doc links private `Entry` (warns only under an all-features doc build, which CI doesn't run today); a C0 control char inside a root element name can still reach `dialect_declared` via `unknown:<name>` (length-capped, pre-existing class); BOM-less UTF-16 remains lexically undetectable (pre-existing, non-conforming input); `record_not_modified` doesn't bump `window_last_used`, so revalidation alone doesn't refresh eviction recency (reads do).

PR #182, opened 2026-08-02
feat(sources): RSS provider 3/4 — freshness engine, partition-per-feed exec, table providers, registration + public egress seam

gracexmatin:
Phase 3 of 4 stacked PRs splitting #175 (plan Tasks 11-14). Base: `rss-m1-phase2`; retarget after its merge.

## Scope
- **RssEngine**: the per-feed freshness state machine (TTL hit → serve cached; expired → conditional GET; 304 → re-arm; failure → record `last_error` capped at 512 chars, keep the stale window). Sole production consumer of `egress`/`fetch`/`cache` — their `dead_code` allowances drop here. Feed-key discipline: every cache key originates from the engine's own subscription list, never from a predicate value.
- **RssScanExec**: one partition per subscription, concurrency bounded by `max_concurrent`, LIMIT gates further fetch launches post-acquire, scan deadline degrades a partition to zero rows + warn instead of failing the query; `reset_state` overridden so re-execution (e.g. `RecursiveQueryExec`) starts clean
- **Table providers**: exact pushdown + pruning for single-feed equality, `IN` lists, and single-column equality disjunctions — the shape DataFusion's simplifier rewrites short `IN` lists into, so `WHERE feed IN ('a','b')` prunes at every list length
- **register_rss_tables**: one source = one catalog (`<name>.main.feeds` / `<name>.main.items`), zero network I/O at registration. Production constructs `Arc::new(AllowAll)` — no destination filtering, the deliberate OSS stance.
- **Public egress seam**: `register_rss_tables_with_policy` is a public (non-test) entry point that accepts an `Arc<dyn EgressPolicy>`, and the egress trait/types (`EgressPolicy`, `AllowAll`, `EgressReason`, `EgressDenied`) are re-exported — so an embedder (Skardi Cloud, or an operator running Skardi as a library) can implement a destination filter and inject it in place of `AllowAll`. OSS still ships only `AllowAll`; a denying policy in-tree is `#[cfg(test)]`. An engine test injects a denying `DenyList` and asserts the refusal reaches `feeds.last_error` end-to-end.

## Verification
`cargo test -p skardi --features rss` passes at this phase (1018 tests, includes non-rss crate tests); the committed tree compiles standalone.


crates/skardi/src/sources/providers/rss/engine.rs
```diff
@@ -0,0 +1,2147 @@
...
+            http_status,
+            error.clone(),
+            dialect_declared,
+            now_ms(),
+            arm(Instant::now(), failure_fuse(self.ttl)),
+        );
+        tracing::warn!(
+            source = %self.source_name,
+            feed = %sub.name,
+            url = %sub.url,
```
BtXin, line 581: [P1] Redact the URL from degraded-feed warnings

This `warn` event includes the complete configured feed URL on every failure. The same PR documentation notes that URLs may contain private query tokens but claims they appear only at `debug`; in practice ordinary info/warn deployments will export them. Reqwest transport error strings can also carry the current or redirected URL into `last_error`. Please log only a sanitized host/path identifier here and strip userinfo, query, and fragment from transport errors before storing or logging them.

crates/skardi/src/sources/providers/rss/engine.rs
```diff
@@ -0,0 +1,2147 @@
...
+            }
+            Ok(FetchOutcome::Fetched {
+                body,
+                http_status,
+                etag,
+                last_modified,
+                content_type,
+            }) => {
+                let bytes = body.len();
+                match parse_feed_document(&body, content_type.as_deref()) {
```
BtXin, line 463: [P1] Move feed parsing off Tokio worker threads

`parse_feed_document` synchronously performs sanitation, feed parsing, DOM construction, and HTML-to-Markdown conversion over attacker-authored input of up to the configured response cap. Because it runs inside the async serve future, several pathological feeds can occupy the runtime workers and stall unrelated server work; the surrounding `timeout_at` cannot interrupt synchronous CPU work once this call starts. Please move this stage to a bounded blocking/CPU executor and preserve explicit complexity/depth limits and cancellation semantics.

crates/skardi/src/sources/providers/rss/engine.rs
```diff
@@ -0,0 +1,2147 @@
...
+        // only answer "still current" about rows this process no longer holds
+        // — the state `record_not_modified` has to record as an error
+        // (`cache.rs:561-577`). A full body is what actually refills the
+        // window.
+        let validators = snapshot.window.as_ref().map(|window| Validators {
+            etag: window.etag.clone(),
+            last_modified: window.last_modified.clone(),
+        });
+
+        match self.fetcher.fetch(&sub.url, validators.as_ref()).await {
```
BtXin, line 421: [P1] Coalesce refreshes or generation-check cache commits

An expired snapshot is taken before the permit, and concurrent scans can independently fetch the same feed and commit in completion order. A slower response representing an older window can overwrite a newer successful fetch; a `304` can also re-arm and label whatever window is currently stored even though its validators came from the earlier snapshot. The cache can therefore regress or claim that a window was revalidated when that response never validated it. Please add per-feed singleflight, or attach a generation to the snapshot and reject/reconcile stale success, failure, and 304 commits. Add a concurrent 200/200 and 200/304 ordering test.

crates/skardi/src/sources/providers/rss/engine.rs
```diff
@@ -0,0 +1,2147 @@
...
+                    Err(failure) => self.degrade(
+                        sub,
+                        Some(http_status),
+                        parse_error_message(failure.stage, &failure.reason),
+                        failure.dialect_declared,
+                        bytes,
+                    ),
+                }
+            }
+            Err(error) => {
```
BtXin, line 498: [P2] Do not serve cached rows after an egress denial

All `FetchError` variants enter normal stale degradation, so `FetchError::Egress` serves an existing window as `stale-error`. The design explicitly requires a policy refusal to produce zero rows, and a dynamic policy or DNS result can deny a destination after this cache was warmed. Please special-case egress denial so the denied subscription contributes no item rows while health records the refusal, and test denial after a successful cached fetch.

crates/skardi/src/sources/providers/rss/engine.rs
```diff
@@ -0,0 +1,2147 @@
...
+        Self {
+            source_name,
+            subscriptions,
+            by_name,
+            fetcher,
+            cache,
+            // `RssConfig::validate` rejects `max_concurrent: 0`; the floor
+            // keeps a directly constructed config from producing a
+            // semaphore that parks every fetch forever.
+            semaphore: Arc::new(Semaphore::new(config.max_concurrent.max(1))),
```
BtXin, line 290: [P2] Cap `max_concurrent` before constructing the semaphore

Config validation rejects only zero, but Tokio documents that `Semaphore::new` panics when the permit count exceeds `Semaphore::MAX_PERMITS`. A valid YAML integer above that threshold therefore aborts registration instead of returning a configuration error, contrary to the no-panics production rule. Clamp the effective value to the subscription count and/or `MAX_PERMITS`, or reject it during validation, and add an extreme-value regression test.

BtXin (changes requested):
Requesting changes on the execution layer. The current implementation leaks secret-bearing URLs at warning level, runs attacker-controlled parsing synchronously on Tokio workers, permits stale/out-of-order concurrent cache commits, serves cached rows after an egress denial despite the design contract, and exposes a configuration-triggered semaphore panic.

crates/skardi/src/sources/providers/rss/exec.rs
```diff
@@ -0,0 +1,1027 @@
...
+        match self {
+            Self::Feeds => "feeds",
+            Self::Items => "items",
+        }
+    }
+}
+
+/// State every partition of one scan shares, built once in
+/// [`RssScanExec::new`].
+struct ScanShared {
```
BtXin, line 134: [P3] `ScanShared` carries LIMIT + deadline state across `execute()` calls of one plan object

The spent-LIMIT counter and the deadline (stamped `Instant::now()` at plan *construction*, exec.rs:207) live in the plan, not the stream. A plan object executed twice without an intervening `reset_state()` serves zero rows the second time, and a plan executed after `scan_timeout` has elapsed since planning degrades every partition to zero rows + warn. The `reset_state` override covers `RecursiveQueryExec`, and the test at exec.rs:873 pins the behavior as expected — but correctness now depends on every re-executing operator calling `reset_state`; a plan cached or re-collected outside that path silently shrinks. Deliberate per the docs, so P3 — but worth a doc note on the provider that this leaf is single-execution unless reset.

BtXin (commented):
Sweep 2: head `f656780` **predates the previous review comments and has not moved since** — all five prior threads are open verbatim at the same lines:

- **[P1] engine.rs:581 URL in degraded-feed warn — open.** `degrade()` still logs `url = %sub.url` at warn; config validation only checks scheme, so userinfo/query tokens pass and get logged whole. (The "Feed URLs are safe to log" comment predates the review — it's a rationale, not a fix.)
- **[P1] engine.rs:463 synchronous parse on Tokio worker — open.** `parse_feed_document` still runs sanitation/parse/DOM/markdown inline in the serve future, with the politeness permit held across it, and zero `spawn_blocking` anywhere under `providers/rss/`. Compounding: `max_response_bytes` validation rejects only `0`, so the blocking work is operator-unbounded.
- **[P1] engine.rs:421 concurrent-scan double-fetch — documented, not fixed.** The module doc now declares "No in-flight coalescing … a documented future extension" and `record_success` remains last-writer-wins with no generation check. If the team decision is to defer, say so on the original thread so it's an explicit call rather than an open P1.
- **[P2] engine.rs:498 egress denial serves stale rows — open, and now half-pinned.** All `FetchError` variants including `Egress` still route through `degrade()`. The new test `egress_blocked_feed_degrades_like_unreachable` pins only the no-cached-window case; the contested case — denial *after* a window was cached → rows still served — remains reachable and untested.
- **[P2] engine.rs:290 `max_concurrent` uncapped before `Semaphore::new` — open.** `.max(1)` floors, nothing ceilings; a YAML value above `Semaphore::MAX_PERMITS` panics at registration. Conspicuous because the same commit *does* clamp `MAX_TTL` and `MAX_SCAN_TIMEOUT`.

Stack hygiene: this branch is also based on the pre-fix phase 1 — the proxy/OPML fix commits (`c3a8326`…`649b226`) are not in its history. Needs a rebase once phase 1 settles.

One more registration-time note: `mod.rs` `register_catalog` silently replaces an existing catalog — a source named `datafusion` passes duplicate-name validation (which only checks inter-source duplicates) and would shadow every table-level source. Pre-existing pattern shared with sqlite/open_connector, so not new debt from this PR, but worth a guard while the registration code is hot.

Verified clean this round: projection pushdown incl. empty projection, pushdown/prune consistency through the single `feed_filter` classifier, permit released on cancellation (RAII, tested), no std-Mutex across await in new code, cache byte+entry bounds, catalog published last, statistics correctly Absent, no `.unwrap()` violations.

gracexmatin:
Cross-referencing a phase-1 review item (PR #180, [thread](https://github.com/SkardiLabs/skardi/pull/180#discussion_r3719442110)) whose fix lands in this engine PR.

**Redirected feeds permanently lose conditional GET.** `CachedWindow` (in the phase-2 cache) stores the validators but not the URL they belong to, and the engine caches them under the feed's *original* URL. For the ubiquitous http→https 301 feed, the etag/last_modified captured by the fetcher belong to the *final* hop B, cached under the original URL A. The next scan sends `If-None-Match: etag(B)` to A, A just 301s again, and the hop to B carries no validators by design — so a redirected feed re-downloads in full every scan and never 304s. Safe (a mismatched etag only ever yields a full 200), just wasteful, and it hits a very common redirect shape.

**Decided fix:** store the final landing URL alongside the validators and send the conditional request straight to it, so the etag and the address line up and 304 works again — paired with periodically re-probing the original URL to catch a redirect target that has drifted (301 is nominally permanent but drifts / gets misused in practice). The re-probe interval is still open; flagging here so it's designed into the engine's cache/freshness model rather than bolted on later.

crates/skardi/src/sources/providers/rss/engine.rs
gracexmatin, line 421: Cross-referencing a phase-1 review item (PR #180, [thread](https://github.com/SkardiLabs/skardi/pull/180#discussion_r3719442116)): **the scan deadline must wrap this `fetch()` call.**

The fetcher grants each hop a fresh retry budget, so one `fetch()` is bounded worst-case at ~300s (6 hops × (3 attempts × 10s request_timeout + 2 waits × 10s MAX_RETRY_WAIT)) — 5× the default `scan_timeout_seconds: 60`. By design the fetcher only bounds each request/hop, never the whole fetch; the total deadline belongs to this layer.

Right now `scan_timeout` is stored on `RssEngine` with only a getter — this `await` is bare, so `scan_timeout_seconds` is still enforced by nothing. A malicious or misbehaving server can drag a single feed fetch out to ~300s via slow redirects + retries. Please wrap the whole call, e.g. `tokio::time::timeout(self.scan_timeout, self.fetcher.fetch(&sub.url, validators.as_ref()))` (not a per-request timeout — the whole future), cancelling the entire fetch when the deadline fires.

gracexmatin:
Cross-reference from the phase-1 review (#180): `max_concurrent` was reconciled there as **a bound on total fetch parallelism per source, not a per-host bound**. One shared semaphore caps in-flight feed fetches (each fetch is internally sequential), so feeds sharing a host can receive up to `max_concurrent` concurrent requests — there is no per-host accounting. The design spec (§Fetcher and its YAML example) was corrected accordingly, and baseline host-level politeness is understood to rest on honoring `Retry-After` plus TTL pacing.

Two things for this phase to weigh when it lands here:

1. **Wording to align.** `engine.rs:219` and `:382` describe the semaphore as a "Politeness bound" — the same per-host framing #180 removed as over-promising. Suggest rewording to "total fetch-parallelism bound (not per-host)" so the engine code does not re-assert the guarantee the spec just dropped.

2. **A real per-host cap is an open engine-phase decision, not promised.** If we do want proactive per-host politeness, this semaphore is where it would go — but the design question has to be settled first: count by **hostname** vs **resolved IP** vs **CDN**? Per-hostname misses shared-CDN hosts (Substack/Feedburner/Cloudflare front thousands of feeds behind a few IPs); per-IP tangles with the egress/DNS-resolver layer and can throttle unrelated feeds sharing a CDN IP. Until that is resolved we should not re-advertise a per-host guarantee.

crates/skardi/src/sources/providers/rss/engine.rs
gracexmatin: Fixed, in three parts — the leak had two more channels beyond the `warn!` field itself:

1. **The `warn!` field.** `url` is dropped from the `rss feed degraded` event entirely, rather than replaced with a sanitized identifier: `source` + `feed` already locate the subscription, and the full URL is one `feeds.url` lookup away. Pinned by `a_degraded_feed_emits_a_warning_naming_the_feed_and_reason`, which now asserts the field is absent.

2. **Unnamed subscriptions.** The default name *was* the URL itself, so a tokened URL would have kept leaking through the `feed` field (and the `feed` column) regardless of (1). The default is now the URL stripped of userinfo, query, and fragment (`config::redact_url`). Two subscriptions differing only by query now fail validation with a hint to set explicit `name:` values.

3. **Error strings.** Transport reasons are built from `Error::without_url()` plus the folded `source()` chain. Stripping alone would not have been enough: reqwest's `Display` names only the error kind and the URL — the actual cause lives in `source()` — so today's strings had the URL as most of their information. They now read e.g. `transport error: error sending request: client error (Connect): tcp connect error: Connection refused`, which carries at most a host or ip:port, never query or userinfo. The two `InvalidUrl` sites that quoted full URLs (the redirect-join message and `parse_and_check`) are redacted the same way; the redirect `Location` stays verbatim — it is the server's own header and the datum being diagnosed.

With (1)–(3) the documented claim ("at `info` and above, no feed URL is emitted") is now actually true. The `docs/rss.md` wording that must follow (unnamed names are the stripped URL; what `last_error` may now contain) is tracked on the phase-4 PR: https://github.com/SkardiLabs/skardi/pull/183#issuecomment-5230573208


Landed in ce789d9f49f718de8b0c04d6ed8a0395a0e71d04.

crates/skardi/src/sources/providers/rss/engine.rs
gracexmatin: Fixed. `parse_feed_document` now runs on the blocking pool (`parse_off_worker` in engine.rs), and the serve awaits its `JoinHandle` — the runtime workers stay free during a parse, and that await is a genuine suspension point, so the exec layer's `timeout_at` can finally fire mid-parse instead of waiting out the synchronous stretch.

The politeness permit is held across the parse await, so at most `max_concurrent` (default 6) parses run per source. One honest gap: a parse abandoned by the fuse (below) keeps burning on its blocking thread *outside* the permit — a running thread cannot be preempted. Its ignition rate is capped by the failure path re-arming the feed's TTL to `failure_fuse = (ttl/4).clamp(30s, 300s)` — at most one abandoned parse per feed per fuse window — and Tokio's 512-thread blocking-pool cap is the last-resort backstop.

The await is wrapped in a 10s `PARSE_TIMEOUT`. On expiry the serve synthesizes a `"timeout"` `ParseFailure` and takes the existing degrade path — `feeds.last_error`, the degraded-feed `warn`, negative caching.

10s is deliberately far under the 60s scan default, so the diagnosable failure always fires before the undiagnosable one; it is also generous headroom (a legitimate `max_response_bytes`-sized document parses well under a second) — anything reaching it is hostile, broken, or a super-linear corner in a parsing stack we don't own. A `JoinError` (a panicking parse) synthesizes `"panic"` the same way — the engine's no-`Err` contract holds.

Worth calling out explicitly: an abandoned parse occupies one blocking thread until it finishes naturally, and its result is discarded. During that window the real concurrent burn can exceed the permit's book value; in the worst case (an unknown super-linear corner in the parsing stack × several hostile feeds timing out repeatedly), zombie threads accumulate at roughly `feeds × burn_duration / fuse_window`. The nature of that risk is thread occupancy and capacity loss, not a return of this finding's failure mode: blocking threads are OS-preempted, so the workers and query handling keep making progress at their fair share — what leaks is wasted CPU, itself bounded by parse being linear in a capped body.


Landed in ce789d9f49f718de8b0c04d6ed8a0395a0e71d04.

gracexmatin:
**Design note: per-feed singleflight (request coalescing) — deferred; recorded here for a future decision.**

Context: the cache-commit race finding (https://github.com/SkardiLabs/skardi/pull/182#discussion_r3710280999) offered two remedies. We took generation-checked commits; per-feed singleflight — the first scan to find a feed expired becomes the *leader* and fetches, concurrent scans *wait* and serve from the refreshed cache — remains the documented future extension (engine.rs module doc, "No in-flight coalescing, but generation-checked commits"). This comment is the fuller trade-off record so a future decision starts here instead of rediscovering it.

**What it buys**

1. **Politeness.** N concurrent scans issue exactly one `GET` per feed per expiry instead of up to N. This is the one cost the generation gate does not remove. It matters most for always-live (`ttl_seconds: 0`) sources under many concurrent agent queries — the configuration where duplicates multiply.
2. **One agreed window.** Concurrent scans serve identical rows; today each racing serve emits its own legitimate read, so two simultaneous queries can briefly see different windows.
3. **No wasted work.** The losing fetch+parse (bounded, but real bytes and CPU — its commit is dropped by the generation gate today) is never spent, and the duplicate holds no politeness permit, freeing slots for other feeds under contention.

**What it costs**

1. **Cancellation-safe leader hand-off — the hard part.** The leader is a future that can be dropped at any await (scan deadline, LIMIT gate, client disconnect). Waiters must detect leader death and either elect a replacement or all revert to fetching; getting this wrong strands waiters as deadlocked partitions until their scan deadline. This class of code is notoriously easy to get subtly wrong and needs tests for every cancellation interleaving.
2. **Cross-query coupling.** A partition of query B awaits query A's in-flight fetch, inheriting A's tail latency (worst case: A's remaining retry/redirect budget). Today B's latency is its own.
3. **Semantics to specify.** A waiter served off a leader's `304` is consuming a revalidation whose validators it never held; window-evicted-during-flight edges need re-deciding for waiters.
4. **Pure addition, no retirement.** The generation gate stays regardless — it is the correctness backstop under leader hand-off races too — so singleflight adds an in-flight registry (per-feed channel map with drop-guarded lifecycle) without simplifying anything else.

**When to revisit**

Duplicate `GET`s visible in feed-server logs, rate-limit hits attributable to concurrent scans, or deployments pairing `ttl_seconds: 0` with high query concurrency. **Not** cache correctness — the generation gate owns that as of this PR.

**Rough shape if adopted:** engine-held `Mutex<HashMap<feed, watch::Receiver>>`; leader inserts before acquiring its politeness permit and removes via a drop guard; waiters subscribe, and on a channel closed without a completion marker re-enter the serve from a fresh snapshot (the cache then either has the leader's commit or the waiter becomes the new leader).

crates/skardi/src/sources/providers/rss/engine.rs
gracexmatin: Fixed, via the generation option. `FeedSnapshot` now carries a `generation` (a cache-wide monotonic counter; `0` = no entry, never re-issued, so entry eviction/recreation cannot alias). All three `record_*` methods take the `expected_generation` of the snapshot their fetch was decided from and check it under the lock. A mismatched commit is stale and is dropped, with one reconcile rule: a stale *success* still supersedes a failure-labelled state — dropping it would leave the feed negatively cached for up to the failure fuse despite a successful response in hand, which is worse than either racing response winning.

What each ordering now does. 200/200: the first commit wins and the slower one is dropped — the window can no longer regress wholesale. 200/304, either order: the `304` never stamps `revalidated` on (nor re-arms) a window its validators did not come from. Failure/success, either order: a slow failure cannot stamp `stale-error` over — or arm the fuse against — a fresher success, and a slow success is not starved by a fast failure. Each racing serve still answers its own query from its own read (both are legitimate reads of the feed); only the cache commit is arbitrated.


Landed in ce789d9f49f718de8b0c04d6ed8a0395a0e71d04.

BtXin (approved):
Verification sweep (head ce789d9). **The three P1s are genuinely fixed with tests — approving.** The two P2s remain open; neither is severe, so they don't block, but both need a follow-up (details below).

Verified fixed:
- **[P1] URL in logs**: `degrade()`'s warn drops the URL field entirely (with a test asserting `fields.get("url").is_none()`); the only remaining URL log is the documented `debug!` in `serve_feed`. Redaction goes deeper than asked: unnamed subscriptions get `redact_url` names (userinfo/query/fragment stripped), `transport_reason` rebuilds reqwest errors with `.without_url()`, and both `InvalidUrl` sites redact. Four tests pin it.
- **[P1] parse off workers**: `parse_off_worker` wraps `parse_feed_document` in `spawn_blocking` under a 10s `PARSE_TIMEOUT` fuse; a panicked parse degrades the feed (`stage: "panic"`) instead of poisoning anything; cancellation mid-parse detaches harmlessly with a bounded burn (body is size-capped upstream). Holding the politeness permit across the parse is now the documented concurrency bound. Both fuse paths tested.
- **[P1] cache commit races**: cache-wide monotonic `commit_counter` (first value 1, so recreated entries can't false-match), generation checked and committed under a single `MutexGuard` — no check-to-commit window; stale commits rejected with one justified exception (a stale *success* may supersede an error state). Race behavior pinned at both cache and engine level (`a_slower_concurrent_fetch_cannot_regress_the_window`, stale-304 end-to-end). Double-fetch without coalescing is now an explicitly documented accepted trade-off with a revisit trigger — that closes the thread as far as I'm concerned.

Still open (follow-ups, not blockers):
- **[P2] egress denial serves stale rows**: `FetchError::Egress` still routes through generic degradation, so a feed denied by an injected policy keeps serving its last window as `stale-error`. On reflection the data was fetched while permitted, so this is a retention/freshness semantics question rather than an exfiltration hole — but it's still undecided-by-silence: the only test covers the cold-feed case. Please either purge-on-denial or document + test the serve-after-denial behavior as intended.
- **[P2] `max_concurrent` semaphore ceiling**: still floor-only (`.max(1)`); an absurd YAML value passes `validate()` and panics inside `Semaphore::new` at server startup. One-line clamp (mirroring `MAX_TTL`) or a validate() ceiling.

Fix-commit review found nothing severe; one nit: `JoinError`'s Display can quote a panic payload into `feeds.last_error` (bounded by `MAX_ERROR_CHARS`, same class as the documented structural-token channel). Local test run didn't finish in budget — CI (green on this head) is the arbiter.

crates/skardi/src/sources/providers/rss/engine.rs
gracexmatin: Fixed. `FetchError::Egress` no longer enters the stale degradation path: the engine routes it to a new cache commit, `record_egress_denial`, which — unlike `record_failure` — **drops the cached window and its validators** and records the refusal under `last_status = 'error'`. The design's requirement (the failure-mode table; acceptance criterion 15) now holds unconditionally: a refused subscription contributes zero item rows, warm cache or not, with the refusal named in `feeds.last_error`.

Purging the window is the load-bearing part, not just skipping the serve. A denial arms the failure fuse like any failure, and within the fuse every subsequent scan takes the zero-network cache-hit path — special-casing only the denial's own serve would have the very next scan resurrect the stale window from cache without ever consulting the fetcher, the one component that knows the destination is refused. With the window gone, `Error`'s contract ("no window cached") makes those within-fuse scans serve zero rows with zero requests, and once the fuse expires a re-allowed fetch rebuilds the window with an unconditional `GET` (its validators went with it).

Denials pass the same commit-generation gate as every other commit (drop-always): a stale denial cannot purge a window a concurrent fetch just committed under an allowing verdict.

Tested with a verdict that flips after a successful fetch (`a_denial_after_a_warm_cache_serves_zero_rows_not_stale`): zero rows despite the warm cache, the refusal in `feeds`, no re-poke within the fuse.

Landed in 9d1bd5eb65bb3d6f94d963ce624f509f708538e8.

crates/skardi/src/sources/providers/rss/engine.rs
gracexmatin: Fixed, via the clamp. `with_parts` now clamps the effective value to `Semaphore::MAX_PERMITS` and warns when it does (`configured_max_concurrent` / `effective_max_concurrent` fields), so any YAML-expressible integer now reaches a working semaphore instead of tokio's assert. Same clamp-and-warn treatment — and the same test shape — as the existing `ttl_seconds` and `scan_timeout_seconds` clamps: this is the third instance of the raw-config-integer-meets-narrower-API-domain pattern, and all three now look alike.

One deliberate deviation from the suggestion: the ceiling is **not** the subscription count. Permits are shared across concurrent scans of the same source, so feed count does not bound useful concurrency — a 3-feed source at `max_concurrent: 100` legitimately lets two overlapping scans' six serves all proceed today, and clamping to 3 would silently queue the second scan. `MAX_PERMITS` is a no-op for every legitimate config, which is the property a safety clamp should have.

The regression test builds an engine at `usize::MAX` and asserts construction survives, the semaphore actually hands out permits (a serve completes), and the warning carries both values (`an_absurd_max_concurrent_is_clamped_rather_than_panicking`).

Landed in fb875ee2e87318466004cb981fb361e954c6e505.

gracexmatin:
Re: [Sweep 2](https://github.com/SkardiLabs/skardi/pull/182#pullrequestreview-4862976159)'s compounding point on the parse item — `max_response_bytes` validation rejecting only `0`, leaving the blocking parse work operator-unbounded:

Addressed in [ef26fe2](https://github.com/SkardiLabs/skardi/commit/ef26fe2f8f2cf802ec07a31d067a719db5cfa3c4). The parse fuse now scales with the configured cap — ten seconds per 5 MiB unit, partial units rounded up, floored at one unit, capped at an hour; the default config keeps the original 10s — so an operator who raises `max_response_bytes` has licensed proportionally more parse *time* along with the parse *work*, instead of a constant fuse misfiring on exactly the larger documents the raised cap now permits (parse is linear in the body, so the time budget scales linearly with the input budget). Registration warns when the scaled fuse meets `scan_timeout_seconds` — the point where parse timeouts would start surfacing as traceless scan-deadline drops instead of diagnosable degrades — with the remedy named (`raise scan_timeout_seconds`). Pinned by `the_parse_fuse_tracks_max_response_bytes` and `a_parse_fuse_past_the_scan_timeout_warns_at_construction`.

crates/skardi/src/sources/providers/rss/exec.rs
gracexmatin: Doc note added. Agreed on the analysis, including the severity: the state has nowhere else to live — one execution's partitions are just N `execute(i)` calls on the same plan object, and DataFusion offers no other home for cross-partition state — so `reset_state` is the only re-execution signal there is, and the behavior itself stays.

What changed: `table.rs`'s module doc (the provider whose `scan()` mints the plan, i.e. where a consumer would look) now carries the contract — "The plan `scan()` returns is single-execution": a second execution of the same plan object finds its own satisfied LIMIT, a plan executed more than `scan_timeout_seconds` after construction finds its deadline passed, both silently, and any caller that caches and re-collects a physical plan outside DataFusion's own `reset_state` path must call `reset_state` first or the scan silently shrinks. `ScanShared` in `exec.rs` carries a back-pointer to that section, so the two statements cannot drift apart.

Landed in 36ca267d2f0c8e5707ef07eefe23c988028bfff7.

crates/skardi/src/sources/providers/rss/engine.rs
gracexmatin: Enforced — one layer up from where this comment points. The exec layer wraps each partition's *entire* serve in `timeout_at` against the scan deadline (exec.rs's module doc: "enforced here and nowhere else"), and the `fetch()` await sits inside that wrapper. `fetch()` is async throughout — socket I/O, the sleeps between retry attempts, every redirect hop — so the outer deadline can fire at any of its await points: the ~300s worst case is truncated at `scan_timeout_seconds`, and a slow-loris server cannot drag a partition past its budget. Wrapping at exec is strictly broader than wrapping the fetch call alone: it also bounds the parse wait and the batch build. Pinned by `the_scan_deadline_degrades_one_partition_to_zero_rows` and `an_absurd_scan_timeout_is_clamped_rather_than_overflowing` (plus the `scan_deadline_clamps_to_the_ceiling` unit test).

gracexmatin:
**Re: [redirected feeds losing conditional GET](https://github.com/SkardiLabs/skardi/pull/182#issuecomment-5202724232)** (cross-referenced from [#180](https://github.com/SkardiLabs/skardi/pull/180#discussion_r3719442110)) — implemented here in [3766e0a](https://github.com/SkardiLabs/skardi/commit/3766e0a3874a406aa736b2409a54359b98501b3b), the decided fix plus the open interval question settled.

**Landing URL.** `FetchOutcome::Fetched` reports a `final_url` (the hop the body came off), `CachedWindow` stores it as `fetched_from` beside the validators it belongs to, and the engine aims its conditional request there. `NotModified` needs no landing URL, and that is a property rather than an omission: a `304` can only answer a hop that actually sent validators, validators go out on the first hop only, and an unconditional `304` is already mapped to `FetchError::Status` — so a `NotModified`'s URL is always the one requested.

**Re-probe interval — the question that was left open.** `ttl × 24`, clamped to [6h, 7d] (default TTL → 6h), in the same shape as `failure_fuse`. What sets the magnitude is that a probe costs *exactly one un-optimized fetch*: starting from the configured URL spends an extra round trip on the redirect and cannot carry the landing URL's validators, so that one fetch pays for a full body where the others get a `304`. Probing every N refreshes therefore keeps (N−1)/N of the saving — at 24 the feed keeps ~96% of it while drift surfaces within hours instead of never. The probe is unconditional by construction: presenting the landing URL's validators *to* the configured URL is the original bug in reverse. A content refresh carries the drift clock forward rather than resetting it, or a frequently-updated feed would never probe at all; the clock rides on the window, where eviction can only lose it harmlessly (a feed with no window refetches from its configured URL anyway, which is itself a probe).

Also recorded as a deliberate non-feature: a **cheap probe** — request the configured URL without following it, and when the `Location` still points at the stored landing URL, stop there and revalidate as usual. One round trip, no body, which would make the interval nearly cost-free. It needs a "probe without following" mode in the fetcher, so it is documented next to `reprobe_interval` for whenever the probe fetches show up as a real cost.

**Identity is unchanged**, and one option deliberately not taken: `items.feed_url` and `feeds.url` stay the *configured* URL — the value an operator wrote and pruning matches against — because where the bytes came from is cache state, not a row value. We considered surfacing the landing URL as a `feeds.final_url` column (diagnostic: which subscriptions redirect, and spotting drift) and left it out of this change. It stays available on demand: it is an additive nullable column, so it needs no surface-version bump and can land whenever operators actually want that visibility — say so and it is a small follow-up.

Tests: the regression itself (`a_redirected_feed_revalidates_at_its_landing_url` — scan 2 goes straight to the landing URL, conditionally, and gets its `304`; scan 3 proves the landing URL survives a revalidation), drift adoption and the unconditional probe, the clock-carry-forward, the interval clamp, and the fetcher-level `final_url` contract in both the redirected and un-redirected cases.

PR #183, opened 2026-08-02
feat(sources): RSS provider 4/4 — server wiring, acceptance suites, fixture corpus, M2 docs

gracexmatin:
Phase 4 of 4 stacked PRs splitting #175 (plan Tasks 15-20). Base: `rss-m1-phase3`; retarget after its merge. **The tree after this PR is byte-identical to branch `add_RSS_Feed`** — all four phases together are exactly that branch's content.

## Scope
- **Server wiring**: `DataSourceType::Rss`, the typed `rss:` block on `data_sources[]` (parses without the feature; registration is what names the missing build capability), catalog-hierarchy enforcement, rss sources rejected as job destinations (read-only). The CLI needs no wiring: after #170 it is a thin HTTP client and rss sources are configured on the server.
- **Fixture corpus** (Task 17): committed feed documents across dialects + field-reviewed golden Markdown outputs — the regression ratchet that turns a feed-rs upgrade's behavior drift into a test failure instead of a silent SQL change
- **Mock-HTTP integration suite** (Task 18): acceptance criteria at the SQL surface — pruning, TTL, 304 revalidation, per-feed fault isolation, the short-IN-list disjunction case pinned from both sides. (No reserved-range SSRF tests here: egress is exercised via the injected-policy seam tests in phases 1 and 3, since OSS ships `AllowAll`.)
- **E2E composition** (Task 19): federated join across sources, idempotent two-INSERT archive pipeline, citability after the live window rolls, candle-embedding composition (skips gracefully without `SKARDI_TEST_EMBED_MODEL`)
- **M2 docs** (Task 20): `docs/rss.md` (states OSS does not sandbox fetch egress — operator/infra-layer or injected-policy control), sample context YAML, README source-table row. Plus the **Skardi Cloud egress governance design** doc.

## Verification
`cargo test -p skardi --features rss` passes (1054 tests). `git diff` against `add_RSS_Feed` is empty.


crates/server/src/pipeline_handlers.rs
```diff
@@ -379,7 +379,10 @@ pub async fn get_data_sources(
             | DataSourceType::Influxdb
             | DataSourceType::Clickhouse
             | DataSourceType::OpenConnector
-            | DataSourceType::Dynamodb => None,
+            | DataSourceType::Dynamodb
+            // RSS has no path: its feed URLs live in the typed `rss:` block,
+            // not in `path`.
+            | DataSourceType::Rss => None,
```
BtXin, line 385: [P1] Enumerate the RSS catalog tables in `/data_source`

Adding RSS as a pathless source is not enough for discovery: the code below still asks the default `datafusion.public` schema for one table named after the source (`news`). RSS actually registers `news.main.feeds` and `news.main.items`, so the endpoint returns a fake `news` table with an empty schema. The qualified `news.main.*` entries in the bundled semantics overlay therefore never reach agents or `skardi schema`, missing the design requirement that agents discover the health and absence semantics from the schema. Please make this endpoint enumerate catalog-mode inner tables, merge semantics by qualified name, and add an endpoint-level RSS test.

docs/rss.md
```diff
@@ -0,0 +1,1063 @@
...
+  reduced to its text content. `<template>` content is dropped rather than
+  reduced to text — the converter walks a parsed tree, and its contents are not
+  in it.
+- It **never fails a feed**: pathological HTML degrades to text content, not to
+  an error.
+- Plain-text values pass through untouched — **byte-exact, including anything
+  tag-shaped in them.** See the caveat below.
+- **The source HTML is not retained.**
+
+> **The claim is "no HTML tag survives as markup", not "no raw HTML is
```
BtXin, line 644: [P2] Reconcile this narrowed claim with the normative design

This section correctly documents what the implementation does, but the normative RSS design and its acceptance criteria still promise that no raw HTML is stored. The converter intentionally preserves HTML-looking attribute text, and `text/plain` bodies pass through byte-exact, so consumers following only the design may enable raw HTML rendering under a guarantee the provider does not provide. Please choose one contract before landing: escape such text to uphold the design, or update the normative design, acceptance criteria, semantics, and security wording to state the narrower guarantee consistently.

BtXin (changes requested):
Requesting changes on final server integration and contract reconciliation. The RSS catalog registers successfully, but the agent discovery endpoint cannot expose either inner table or the bundled qualified semantics. The final documentation also narrows the normative no-raw-HTML guarantee without reconciling the design acceptance criteria. At stack level, the observability section promises provider metrics for fetches, retries, bytes, rows, duration, and terminal categories, but the implementation currently emits tracing events only; either add the instruments or narrow that requirement before merge.

docs/sample_data/rss_context.yaml
```diff
@@ -0,0 +1,43 @@
+# Example context for the RSS/Atom data source (`type: rss`).
+# Reference: docs/rss.md
+#
+# Try it — registration performs no network I/O and `feeds` is a pure state
+# read, so this works offline and fetches nothing:
+#
+#   cargo run -p skardi-cli --features rss -- \
```
BtXin, line 7: [P2] The "Try it" command cannot run at all

Three independent failures: (a) `crates/cli/Cargo.toml` has no `[features]` section, so cargo aborts with "Package `skardi-cli` does not have the feature `rss`" before anything executes; (b) the `Query` subcommand accepts only `-e/--sql`, `-f/--file`, `--max-rows`, `--table` — there is no `--ctx` flag; (c) the CLI is a thin HTTP client ("It holds no local query engine", main.rs header), so "this works offline and fetches nothing" is impossible — a running server loaded with this context is required. It also contradicts docs/rss.md's own header ("the CLI is a thin HTTP client; only the server needs the feature"). This is the doc's only runnable example — please rewrite it as: start the server with this context (`--features rss` on the *server* crate), then `skardi-cli query -e "SELECT name, last_status FROM news.main.feeds"`.

BtXin (commented):
Sweep 2 (head 34515ef). Prior-thread status:

- **[P1] pipeline_handlers.rs `/data_source` discovery — NOT FIXED.** `get_data_sources` still builds `tables: vec![TableInfo { name: data_source.name, … }]` and `get_table_schema` still resolves only `datafusion.public.<name>`, so an RSS source lists a single table named after the source with an *empty schema* (the lookup fails into the warn + empty-Vec arm); `feeds`/`items` are never enumerated. The only in-diff change is the `| DataSourceType::Rss => None` pathless arm. There is also no server-level test of `GET /data_source` with an RSS source configured — such a test would have caught this.
- **[P2] docs/rss.md:644 vs normative design — PARTIALLY reconciled.** rss.md is now internally coherent and test-pinned ("no HTML tag survives as markup", not "no raw HTML stored", with `plaintext_typed_content_is_stored_byte_exact_tags_included`). But `docs/superpowers/specs/2026-07-22-rss-feed-support-design.md` is byte-identical to phase 3 and still asserts the stronger claim in six places (e.g. line 94 "stored content carries no executable markup", line 534 / AC18 "Markdown with no raw HTML", and line 558, which describes rss.md as saying the exact sentence rss.md now disclaims). The cross-document half of the ask is still due.

Other notes:

- **[P3] `cancellation_stops_further_fetches` flake window.** The only margin between "mock observed request 1" and `query.abort()` is the scripted 500ms response delay — a CI scheduler stall >~490ms lets feed `a` complete, releases the single permit, launches `/b.xml`, and fails the `paths().len() == 1` assertion. Every other timing test in the suite has verified headroom; this one could use a longer scripted delay or an explicit abort-before-release barrier.
- Stack hygiene: this branch (via phase 3) is based on the pre-fix phase 1 — the proxy/OPML fix commits are not in its history; needs the stack rebase.

Verified clean this round: feature gating compiles both ways with the featureless registration error named and pinned; Rss correctly absent from writable/connection-string arms and present in catalog-supported; docs-vs-code spot checks all pass (retry table, redirect cap, Retry-After clamp, cache budget, MAX_ERROR_CHARS, deny_unknown_fields, categories NULL semantics); acceptance tests assert real wire behavior with zero real-network access and ephemeral ports; lock poisoning handled correctly; all new `.unwrap()`s are in test code.

crates/server/src/config.rs
```diff
@@ -1672,6 +1748,42 @@ async fn register_data_source(
...
+                tracing::info!(
+                    "Registering RSS source: {} (hierarchy_level: {:?})",
+                    source.name,
+                    source.hierarchy_level
+                );
+
+                // Config presence, catalog-only, and read-only are re-checked
+                // inside the provider — the single enforcement point shared
+                // with the CLI (`sources/providers/rss/mod.rs:226-258`).
+                skardi::sources::providers::rss::register_rss_tables(
```
BtXin, line 1787: [P3] Convention + stale architecture claim

`skardi::sources::providers::rss::register_rss_tables(…)` is a full inline crate path — the project rule is `use`-imports at the top (a `#[cfg(feature = "rss")] use …` satisfies it; yes, the pre-existing `documents` arm has the same flaw, but new code shouldn't copy it). The comment above this arm also claims the enforcement point is "shared with the CLI" — the CLI has had no registration path since #170 made it a thin HTTP client; only the server front-end reaches `register_with_policy`.

docs/rss.md
```diff
@@ -0,0 +1,1063 @@
...
+
+`stale-error` means the TTL expired, the refetch failed, and the **previous**
+window is being served. The rows are real — they are the last complete window
+that parsed — but they are not current. `feeds.last_error` and
+`feeds.last_fetch` say why and as of when. Everything else in the scan is
+unaffected: one dead feed among fifty never fails the scan.
+
+### Reading `last_error`
+
+`last_error` is bounded at 512 characters (the provider's `MAX_ERROR_CHARS`; the
```
gracexmatin, line 979: Cross-referencing a phase-1 review item (PR #180, [round-3 review](https://github.com/SkardiLabs/skardi/pull/180#pullrequestreview)): this `512` literal is the other half of a cross-file contract. Phase 1 defines `MAX_ERROR_CHARS = 512` and has a test (`max_error_chars_is_the_number_the_docs_publish`) that pins the constant to the literal 512 precisely so this doc and the code can't silently drift apart — neither the doc nor the semantics.yaml column description (also `512`, semantics.yaml:85) is Rust, so nothing else would catch a change to one but not the other.

At the phase-1 merge point these docs didn't exist yet, so the phase-1 references were softened to "published later in this stack" rather than citing exact line numbers. Now that the docs land here: if anyone ever changes this 512 (here, in semantics.yaml, or the `MAX_ERROR_CHARS` constant), all three move together, and the phase-1 test is what enforces it. Flagging so the tie is visible from this side too.

gracexmatin:
**Follow-up after rebasing onto `rss-m1-phase3`** — the phase-3 branch picked up URL-redaction changes addressing the P1 review finding "Redact the URL from degraded-feed warnings":

1. The `rss feed degraded` `warn!` (engine.rs) no longer carries a `url` field — pinned by `a_degraded_feed_emits_a_warning_naming_the_feed_and_reason`.
2. An unnamed subscription's default name is now its URL **stripped of userinfo/query/fragment** (`config::redact_url` via `default_name`), so a private query token never becomes the feed's public identity (`feed` column, log fields). Two URLs differing only by query now collide at validation with a hint to set explicit `name:` values.
3. Fetch error strings are redacted: `transport_reason()` drops reqwest's ` for url (…)` clause and folds the `source()` chain in (so `last_error` gains the actual cause and loses the URL); the redirect-join error quotes the current hop redacted (the `Location` stays verbatim); `parse_and_check` no longer echoes the raw URL.

`docs/rss.md` in this branch needs to follow suit:

- **"What gets logged"**: the claim "At `info` and above, no feed URL is emitted" is now actually true — keep it, but note that an *unnamed* subscription's **name** is its URL minus credentials/query/fragment, so a name-shaped URL (without its secrets) can still appear at `info`+.
- Config example comment `# optional; defaults to the URL` → "defaults to the URL minus credentials/query" (config.rs and the design spec already say this).
- The `feeds.name` column description, if it mirrors the spec's table: defaults to the stripped URL.
- **`last_error` — "What it may contain"**: transport errors now carry the reqwest `source()` chain (at most a host or ip:port) and never the request URL.

No test in this branch matches the old error-string formats (checked ` for url`, `is not a valid URL`, `does not resolve against '<full url>'`), so the rebase should be mechanical outside `fetch.rs`/`engine.rs`/`config.rs` conflicts.

BtXin (commented):
Sweep 4 (head fefbac2). **Holding approval on this one** — it's the only PR in the stack with its prior findings still open, plus a base-branch gap:

- **[P1, open] `/data_source` discovery**: unchanged — `get_data_sources` still lists one table named after the source and `get_table_schema` still resolves only `datafusion.public.<name>`, so an RSS source presents a single empty-schema table and `feeds`/`items` are never enumerated. (Still no server-level test of `GET /data_source` with an RSS source, which would pin the fix.)
- **[P2, open] `docs/sample_data/rss_context.yaml` "Try it"**: unchanged — `skardi-cli` has no `rss` feature, `query` has no `--ctx` flag, and the CLI is a thin HTTP client, so the doc's only runnable example still fails at `cargo run`.
- **[stack] This branch does not contain phase 3's fix commit** (`ce789d9` — URL redaction, parse off workers, generation-checked commits): it was rebased onto phase 3's *feature* commit only. As-is, this PR's tree reverts the phase-3 fixes; it needs a rebase onto the current `rss-m1-phase3` tip.
- [P2, partial, carried] the normative design spec still makes the stronger "no raw HTML stored" claim in six places that `docs/rss.md` now explicitly disclaims — the cross-document reconciliation from the earlier round is still due.
- [P3, carried] `cancellation_stops_further_fetches` has the ~490ms flake window (only the scripted 500ms delay separates "request 1 observed" from the abort).

Phases 1–3 are approved as of this sweep; happy to approve this one as soon as the discovery P1, the doc example, and the rebase land.

gracexmatin:
Tracked follow-up from the phase-1 review (#180) — a **post-M1 consolidation pass**, not a change requested in this stack. The stack-phase scoping that produced these duplicates was accepted as deliberate; recording the digest here (top of the stack) so it is not lost once M1 lands.

The RSS provider currently carries several copies of logic that already exists elsewhere in the crate — `util::http`'s own module doc claims this territory, so a consolidation after M1 would pay for itself:

- **`MockFeedServer`** (rss fetch tests) forks ~130 lines of `open_connector/testutil.rs`'s HTTP/1.1 mock server.
- **`backoff` / `MAX_RETRY_WAIT` / `RETRYABLE_STATUSES`** (rss `fetch.rs`) are a third copy of `open_connector`'s retry primitives — and the jitter has already diverged between the copies, which is exactly how three-copies-drift starts.
- **`read_body`**'s size-metering loop (rss `fetch.rs`) mirrors `read_body_bounded`.
- **`truncate`** (rss `error.rs`) is the crate's third string-bounding helper.

Suggested when the pass happens: hoist the shared HTTP-retry/backoff and bounded-read/truncate helpers into `util::http` (or a shared test-util module for the mock), and reconcile the diverged jitter so there is one implementation.

Separately (small, could be done any time): the file-wide `#![allow(dead_code)]` in `fetch.rs` / `egress.rs` could be narrowed to the specific root items that are unreferenced until the engine wires them in, so genuinely-orphaned helpers still warn instead of being blanket-silenced.

crates/server/src/pipeline_handlers.rs
gracexmatin: Fixed in 2c0c481 — all three asks, with one deliberate generalization.

**Enumeration.** The endpoint now branches on `hierarchy_level` rather than special-casing RSS: every catalog-mode source enumerates its real inner tables from its own catalog (`<source>.<schema>.<table>`), sorted at both levels for a deterministic payload. Table-mode sources keep the exact old path. This was a pre-existing blind spot for every catalog-mode source — Open Connector gateways were getting the same fabricated empty-schema table; they now enumerate for free. A missing catalog degrades to an empty `tables[]` with a warning instead of a fake entry.

**Qualified semantics.** Inner-table names are emitted fully qualified (`news.main.feeds`), preserving the payload's existing invariant that `tables[].name` is directly usable in a `FROM` clause. Descriptions merge through the registry's `resolve_table_description` / `resolve_column_description` (qualified entry wins, bare source-name entry is the broad fallback) — these existed but had zero callers until now.

**Test.** `get_data_sources_enumerates_rss_catalog_tables` registers an offline RSS source (`.invalid` host, zero network) and loads the *shipped* `docs/rss/semantics.yaml` rather than a synthetic fixture — so beyond asserting the shape, names, and merged health/absence descriptions, it doubles as a drift tripwire: any overlay column that stops matching the registered Arrow schema, or loses its description on the way through the endpoint, fails here. Also pinned the table-mode single-entry shape in the existing dynamodb test.

Docs updated to drop the known-limitation paragraphs this closes (agent_data_plane.md, semantics.md).

docs/rss.md
gracexmatin: Took the second option — narrowed the normative design to the contract the implementation actually provides. Spec-only change in 4dabc75; no code, fixture, or user-doc edits were needed.

Two things tipped it beyond "spec catches up with code":

1. The spec was internally inconsistent, not merely stale. The Markdown-converter section promised both "output contains no raw HTML" and "plain-text values pass through unchanged" in the same paragraph — mutually incompatible the moment a `type="text"` body carries tag-shaped text. Escaping (option one) couldn't have satisfied the spec as written: it upholds one sentence by breaking the other, and it pollutes the primary consumers (prompts, fts5, embeddings) with transport noise to protect a rendering mode the design forbids anyway.

2. The stronger claim never licensed anything. Even with escaping, enabling raw/inline HTML rendering would remain unsafe — `[x](javascript:…)` still requires scheme filtering — so the renderer rules were always load-bearing. The rewrite makes that explicit: the Security section now states the two rules as MUST (raw/inline HTML disabled; link schemes filtered) and says the storage contract is not a substitute for them.

The canonical phrasing, reused verbatim at every site (goals, converter section, storage chain, Security, test plan, acceptance criterion 18, module-tree comment, docs deliverable): no HTML tag survives as markup; tag-shaped text can survive in two pinned shapes — attribute-derived text keeps its literal `<` (`<a href="#" title="<script>">t</a>` → `[t](# "<script>")`, pinned by a unit test), and plain-text-typed bodies pass through byte-exact (pinned by the `plaintext_typed_markup` fixture). This matches what docs/rss.md, the semantics overlay, and the convert.rs test suite already stated, so the tree now carries one contract everywhere.

docs/sample_data/rss_context.yaml
gracexmatin: Fixed in 98de3bb. All three failures confirmed: the CLI crate has no `[features]` section, the `query` subcommand has no `--ctx` flag, and the CLI holds no local engine, so the single-command offline claim was impossible.

One correction on location: the runnable example lives in `docs/sample_data/rss_context.yaml` (linked from docs/rss.md), not in rss.md itself — rss.md's own header already had the correct build line.

The comment now shows the real two-step flow:

```
cargo run -p skardi-server --features rss -- --ctx docs/sample_data/rss_context.yaml
# then, in another terminal:
cargo run -p skardi-cli -- query -e "SELECT name, last_status FROM news.main.feeds"
```

No connection config is needed: the CLI defaults to `http://127.0.0.1:8080`, which is the server's default port. The offline claim is narrowed to what actually holds — registration performs no network I/O and `feeds` is a pure state read, so the query fetches nothing and the `.invalid` hosts are never contacted.

crates/server/src/config.rs
gracexmatin: Fixed in d0e1e08 — both points, plus two more instances of the same stale claim the grep turned up.

**Convention.** `register_rss_tables` now joins the top-of-file import block like every other provider's register fn, behind `#[cfg(feature = "rss")]` so a no-rss build sees no unused import; the call site drops the inline path. (Left the documents arm as-is per the comment's own framing — new code stops copying the flaw; cleaning up the precedent is a separate change.)

**Stale architecture claim.** "Shared with the CLI" is retired in all three places it lived: this arm's comment, the `register_with_policy` comment in rss/mod.rs ("both front-ends (server and CLI)"), and the `ResolvedSubscription` doc ("the server and CLI can hold it"). The enforcement-point design itself stands — its second entry today is the public `register_rss_tables_with_policy` embedder seam, not the CLI — so the comments now name that instead of deleting the rationale. Also swapped the drifted `mod.rs:226-258` line-range citation for a function-name citation so it can't go stale the same way again.

docs/rss.md
gracexmatin: Verified the whole tie from this side — all six sites agree today.

The constant (`MAX_ERROR_CHARS: usize = 512`, error.rs), the phase-1 pin (`max_error_chars_is_the_number_the_docs_publish`, whose message still carries the "published later in this stack" phrasing from the pre-docs merge point), and the five prose sites: rss.md's Reading-`last_error` section (which itself names the constant and the overlay as carrying its value), rss.md's `unknown:`-prefix cap, and semantics.yaml's three mentions (`last_error` ×2, the `unknown:` fallback). All say 512.

One precision on "the phase-1 test is what enforces it": it mechanically enforces the code side only — the constant can't silently leave 512, which keeps the prose true. A prose-side edit that leaves the constant alone trips nothing; that direction stays on review, which seems like the right trade versus a doc-grepping test. No change needed — thanks for wiring the two ends of the stack together.

crates/server/src/pipeline_handlers.rs
```diff
@@ -399,35 +502,43 @@ pub async fn get_data_sources(
...
-                    e
-                );
-                // Continue with empty schema if table not found or schema retrieval fails
-                Vec::new()
-            }
+        // Catalog-mode sources register a whole catalog named after the
+        // source — their tables live at `<source>.<schema>.<table>`, not
+        // under `datafusion.public.<source>` — so they are enumerated
+        // rather than reported as a single source-named table.
+        let tables = if data_source.hierarchy_level == HierarchyLevel::Catalog {
```
BtXin, line 509: [P2, non-blocking] Catalog branch is keyed on a field the other source types ignore

The generalization is right, but `hierarchy_level == Catalog` is now load-bearing for source types that never opted into it. Nothing rejects `hierarchy_level: catalog` on a non-catalog type: `CATALOG_SUPPORTED_SOURCES` (config.rs:802) only gates the *conflicting-options* check, and there is no inverse guard, so a `csv`/`parquet`/`lance`/`iceberg`/`documents`/`influxdb` source carrying that field passes `validate()` and registers at `datafusion.public.<name>` as usual.

For such a source this branch then calls `enumerate_catalog_tables`, `ctx.catalog(name)` returns `None`, and it takes the warn + empty-`Vec` arm — so the entry regresses from a full table schema to `"tables": []`, silently except for a log line. `skardi schema` renders the same response, so discovery degrades there too. Before this change the field was inert for those types and the schema came back fine.

Narrow (it needs a meaningless field value to reach), but it is a regression on configs that used to work. Two clean fixes:

1. Fall back to `get_table_schema` when `ctx.catalog(&data_source.name)` is `None`, which makes the branch self-correcting for any type; or
2. reject `hierarchy_level: catalog` at validation for types outside `CATALOG_SUPPORTED_SOURCES` — louder, and arguably where this belongs, since the config is meaningless rather than merely unlucky.

I'd take (2) plus (1) as a belt-and-braces, but either alone closes it.

crates/server/src/gui.rs
```diff
@@ -325,36 +372,43 @@ async fn render_semantics_card(
...
-            })
-            .collect::<Vec<_>>()
-            .join("\n"),
-        Ok(_) => r#"<div class="col-row"><span class="col-name no-desc">Schema not available.</span></div>"#.to_string(),
-        Err(_) => r#"<div class="col-row"><span class="col-name no-desc">Schema not available.</span></div>"#.to_string(),
+    // Catalog-mode sources render one section per inner table — heading is
+    // the fully-qualified `FROM`-usable name, with the table's own overlay
+    // description — mirroring what `GET /data_source` reports. Table-mode
+    // sources keep the single "Columns" section.
+    let tables_html = if ds.hierarchy_level == HierarchyLevel::Catalog {
```
BtXin, line 379: [P2, non-blocking] Same branch, same consequence on the dashboard

Mirrors the `/data_source` note: a non-catalog source configured with `hierarchy_level: catalog` takes this path, `enumerate_catalog_tables` returns empty, and the card renders a "Tables" section containing only "Schema not available." instead of the source's columns. Whichever fix lands for the endpoint should cover this call site too — they share `enumerate_catalog_tables`, so a fallback inside that helper (or the validation guard) fixes both at once.

crates/skardi/src/sources/providers/rss/fetch.rs
```diff
@@ -128,7 +128,7 @@ use std::error::Error as StdError;
 use std::fmt::Write as _;
 use std::net::IpAddr;
 use std::sync::Arc;
-use std::time::{Duration, SystemTime, UNIX_EPOCH};
+use std::time::{Duration, SystemTime};
```
BtXin, line 131: [P3, non-blocking] `SystemTime` is now unused outside tests

Hoisting the jitter source correctly took `UNIX_EPOCH` with it, but `SystemTime`'s only remaining use is in `#[cfg(test)] mod tests` (line 1252, `httpdate::fmt_http_date`). Reproduced with `cargo check -p skardi --features rss`:

```
warning: unused import: `SystemTime`
 --> crates/skardi/src/sources/providers/rss/fetch.rs:131:27
```

CI stays green — `cargo check --all` doesn't enable `rss`, and the coverage job builds `--all-features` *with* tests, where the import is live — so this only shows up in a non-test `rss` build, i.e. the shipping one. Drop it to `use std::time::Duration;` here and put `SystemTime` in the test module's own imports.

BtXin (approved):
Sweep 5 (head `c2c45a1`). **Approving** — every item I held on is closed, and the new work this round is clean. Verified locally on the branch tip:

- **[P1, closed] `/data_source` discovery.** `get_data_sources` now branches on `hierarchy_level` and enumerates real inner tables via `enumerate_catalog_tables` (`<source>.<schema>.<table>`, sorted at both levels). Generalizing past the RSS special case was the right call. `get_data_sources_enumerates_rss_catalog_tables` registers through the real `register_data_sources`, resolves against the shipped `docs/rss/semantics.yaml` overlay, and calls the handler — that pins the fix rather than restating it. `gui.rs` renders the same shape on the dashboard card.
- **[P2, closed] `docs/sample_data/rss_context.yaml` "Try it".** Now `cargo run -p skardi-server --features rss -- --ctx …` plus a separate thin-client `query`. Checked all three legs against the code: `rss = ["skardi/rss"]` exists in the server manifest, `--ctx` is the server's own `long = "ctx"`, and the offline claim holds (registration does no I/O, `feeds` is a state read).
- **[P2, closed] Cross-document reconciliation.** The normative design now carries the narrowed "no HTML **tag** survives as markup" contract at both the converter section and AC18, and names the two pinned tag-shaped-text survivals explicitly instead of leaving them as surprises. rss.md and the spec agree.
- **[P3, closed] `cancellation_stops_further_fetches`.** Scripted delay 500ms → 5s with an 8s post-abort observation window; the flake margin is now real headroom.
- **[stack, closed] Rebase.** The branch sits on the current `rss-m1-phase3` tip (`d2c275f`), so the phase-3 fixes are in its history — the tree no longer reverts them.

The new consolidation commit (`c2c45a1`) is conservative and I found no behavior change in it: `truncate_chars` cuts on `char_indices().nth()` so a multi-byte sequence is never split; hoisting only the jitter *source* keeps both spreads and both retry loops local as `util::http`'s own doc requires; and dropping the gateway mock's hardcoded `content-type: application/json` is safe — nothing in `OpenConnectorClient` reads a response content type.

`cargo test -p skardi --features rss` → 1176 passed, 0 failed. `cargo test -p skardi-server --features rss` → 240 passed, 0 failed.

One note that is *not* a finding, for the record: `a_transport_error_names_the_cause_but_never_the_url` fails deterministically for me, but as a correct consequence of the design. `test_fetcher()` injects no policy, so per the module doc proxies stay honored, and my shell exports `http_proxy=http://127.0.0.1:7897` — the connection to the deliberately-closed port goes to the proxy and times out instead of being refused, so the assertion sees `Timeout` rather than `Transport`. Unset the proxy vars and it passes. The test's doc comment says "the test environment sets no proxy variables", which is true of CI but not of every dev machine; worth a `#[cfg(test)]`-only `no_proxy()` or a note someday, not now.

Three small things below — none blocking, all fine as follow-ups.

PR #185, opened 2026-08-03
docs(spec): reframe RSS egress as operator responsibility; add cloud egress design

gracexmatin:
## Summary

Docs-only PR extracted from `add_RSS_Feed`, revising the RSS design's egress stance and adding the companion cloud design:

- **`2026-07-22-rss-feed-support-design.md`** (renamed from `-v2`): OSS no longer promises a default-deny SSRF guard. The fetcher exposes an `EgressPolicy` seam whose only OSS implementation is `AllowAll` — feed fetches reach any routable address, and constraining egress is delegated to the operator's infrastructure or an injected policy. Security, Fetcher, Failure Modes, Testing, Acceptance, Repo Shape, Docs Commitments, and Future Extensions are updated to state this boundary explicitly. Two measured feed-rs fact corrections are carried along.
- **`2026-08-03-rss-cloud-egress-design.md`** (new): specifies the two-layer managed egress governance Skardi Cloud lands on top of the seam — infrastructure controls (egress NetworkPolicy, IMDSv2 hop limit, minimal IAM) plus an injected application policy (reserved-range `EgressPolicy`, per-tenant allowlists, quotas, audit).

The code change implementing the seam stays on `add_RSS_Feed`. The two docs only reference each other, so nothing dangles on main.


docs/superpowers/specs/2026-08-03-rss-cloud-egress-design.md
```diff
@@ -0,0 +1,74 @@
...
+- **`PolicyDns`** — the fetcher's DNS resolver wraps the injected `EgressPolicy`, so an injected policy is enforced at resolution time (rebinding-safe) with no additional Cloud wiring.
+- **Per-redirect re-check** — the fetcher re-runs the policy against each resolved redirect target, so a policy sees redirect hops, not just the initial URL. This is the OSS fetcher's manual redirect loop, which exists regardless of policy.
+- **`FetchError::Egress(EgressDenied)`** — a denied fetch surfaces through the existing error path and degrades that feed exactly like an unreachable one (`feeds.last_status = 'error'`, `last_error` names the reason, zero rows in `items`, other feeds unaffected). Cloud reuses this variant rather than adding its own.
+
+The fetcher constructor accepts the policy as a parameter (`Arc<dyn EgressPolicy>`, defaulting to `AllowAll`); Cloud passes its implementation at construction. No OSS source change is required to add Cloud egress governance — only the injected object.
```
BtXin, line 56: **[P1] Expose a provider-level injection path**

The fetcher constructor is not sufficient for a separate `skardi-cloud` crate to inject this policy. On the companion branch, `rss::egress` and `rss::fetch` are private, production `register_rss_tables` constructs the policy internally, and the only registration function accepting a policy is `#[cfg(test)] pub(crate)`. Export the policy contract and provide a public provider-registration or factory path, then exercise it from an external-crate integration test before claiming that Cloud needs only an injected object. See the current [provider module](https://github.com/SkardiLabs/skardi/blob/ef87eb5349c0676fbd47f9b24bb204fcf61c3bb7/crates/skardi/src/sources/providers/rss/mod.rs#L181-L238).

docs/superpowers/specs/2026-08-03-rss-cloud-egress-design.md
```diff
@@ -0,0 +1,74 @@
...
+- **Egress audit logging.** Every refused fetch — reserved-range hit, allowlist miss, quota trip — is logged with tenant, feed, target, and reason, so an SSRF attempt is an auditable event, not a silent refusal. Successful fetches to unusual destinations can be flagged for review.
+
+## OSS interface contract
+
+Cloud must inject Layer 2 **without forking** the OSS fetcher. OSS therefore guarantees this seam (defined in the OSS provider, `crates/skardi/src/sources/providers/rss/`):
+
+- **`trait EgressPolicy: Send + Sync + Debug`** with `fn check_ip(&self, ip: IpAddr) -> Result<(), EgressReason>`, consulted for every resolved address. The method returns only the *reason* a target is refused (e.g. `"link-local"`); the fetcher pairs that reason with the host and ip into an `EgressDenied`, since `check_ip` sees the address but not the originating host. `EgressReason` is `Cow<'static, str>`. OSS ships `AllowAll` (always `Ok`); Cloud supplies its own implementation.
```
BtXin, line 51: **[P1] Pass request context to Cloud policy hooks**

`check_ip(ip)` can classify an address, but it cannot implement the Layer 2 features promised above: it has no hostname for host allowlists, no tenant or feed identity for audit events, and no request or response lifecycle for frequency and byte quotas. Capturing a tenant in one policy instance still does not supply the host, feed, or transferred bytes. Either introduce request-context and lifecycle hooks, or explicitly place allowlists, quotas, and audit outside `EgressPolicy` and specify those integration points.

docs/superpowers/specs/2026-08-03-rss-cloud-egress-design.md
```diff
@@ -0,0 +1,74 @@
...
+- **`PolicyDns`** — the fetcher's DNS resolver wraps the injected `EgressPolicy`, so an injected policy is enforced at resolution time (rebinding-safe) with no additional Cloud wiring.
```
BtXin, line 52: **[P1] Prevent proxies from bypassing PolicyDns**

A DNS-resolver policy does not inspect the destination when the HTTP client uses a proxy: the client connects to the proxy and the proxy resolves or forwards the feed target. Reqwest enables system proxies by default, and the companion fetcher currently builds its client without `no_proxy()`, so `HTTP_PROXY` or `HTTPS_PROXY` can bypass the reserved-address check and DNS-pinning guarantee. Disable automatic proxies for this client or define a trusted proxy as the enforcement point, and add an internal-target regression test with proxy variables set. See [reqwest proxy behavior](https://docs.rs/reqwest/latest/reqwest/#proxies) and the current [client construction](https://github.com/SkardiLabs/skardi/blob/ef87eb5349c0676fbd47f9b24bb204fcf61c3bb7/crates/skardi/src/sources/providers/rss/fetch.rs#L204-L220).

docs/superpowers/specs/2026-08-03-rss-cloud-egress-design.md
```diff
@@ -0,0 +1,74 @@
...
+
+In a multi-tenant hosted context these matter more than in single-tenant self-hosting: feed URLs are attacker-influenceable *without* even needing to hijack an agent (a tenant is an untrusted principal), the instance-metadata credentials reachable from a shared node are higher-value, and one tenant's SSRF must never reach another tenant's network segment. OSS's "operator is responsible" boundary lands the whole of this on Cloud.
+
+## Layer 1 — Infrastructure controls
+
+Zero application code. These exist in `deploy/` (which does not yet exist in the repository — this design creates it) and in the Cloud provisioning stack.
+
+- **Egress `NetworkPolicy` (default-deny).** The RSS-serving workload's namespace denies egress to RFC 1918 (`10/8`, `172.16/12`, `192.168/16`), CGNAT (`100.64/10`), link-local (`169.254/16`, including the metadata address), unique-local IPv6 (`fc00::/7`), and loopback, allowing only DNS and public-internet egress. This is the backstop that holds even if Layer 2 has a bug: a fetch that the application policy failed to refuse still cannot leave the namespace toward an internal target.
```
BtXin, line 33: **[P1] Close same-pod and same-node gaps in Layer 1**

A standard Kubernetes `NetworkPolicy` cannot provide the stated backstop: a pod cannot block access to itself, traffic to and from its local node is always allowed, and enforcement requires a capable network plugin. Consequently this layer cannot claim to deny loopback or every internal target, and the Layer-1-first rollout leaves `AllowAll` exposed to same-pod and same-node services. Specify the actual CNI and pod topology, add node or host-firewall and metadata controls, and verify them in the deployed cluster. The hop-limit claim is also topology-dependent rather than a universal container denial. See the [Kubernetes exceptions](https://v1-35.docs.kubernetes.io/docs/concepts/services-networking/network-policies/) and [AWS container guidance](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-IMDS-new-instances.html).

docs/superpowers/specs/2026-07-22-rss-feed-support-design.md
```diff
@@ -458,7 +460,7 @@ Content is stored as Markdown. The chain from wire to storage: XML transport enc
 | Scenario | Behavior |
 |---|---|
 | Feed down / DNS failure | Partition serves stale cached window — rows stamped `window_status = 'stale-error'` — or zero rows if never fetched (no rows to stamp; `feeds` is the only signal); other partitions unaffected; `feeds` row records error; tracing warns |
-| URL blocked by egress policy | Host resolves to a reserved range (loopback/link-local/private/CGNAT/ULA), or a redirect targets one; fetch refused before connect; `feeds.last_status = 'error'`, `last_error` names the egress block; zero rows in `items`; other feeds unaffected |
+| URL refused by an injected egress policy | Only when a non-default `EgressPolicy` is wired (OSS default `AllowAll` refuses nothing): fetch refused before connect, on the initial URL or a redirect target; `feeds.last_status = 'error'`, `last_error` names the denial reason; zero rows in `items`; other feeds unaffected |
```
BtXin, line 463: **[P2] Preserve the cached-window failure contract**

This says an egress denial always produces `feeds.last_status = 'error'` and zero item rows, but the companion engine routes every `FetchError` through `degrade()`, which serves an existing cached window as `stale-error`. Only a cold-cache denial produces `error` and zero rows. Decide whether a policy denial intentionally purges cached content; otherwise update this failure mode, Cloud line 54, and acceptance criterion 15 to distinguish cold and warm caches, with tests for both. See the current [degradation path](https://github.com/SkardiLabs/skardi/blob/ef87eb5349c0676fbd47f9b24bb204fcf61c3bb7/crates/skardi/src/sources/providers/rss/engine.rs#L498-L510).

BtXin (changes requested):
Requesting changes for four security/design blockers and one contract inconsistency. The proposed Cloud policy is not externally injectable through the provider API, its check_ip-only contract cannot express the promised tenant/host/quota controls, system HTTP proxies can bypass PolicyDns, the NetworkPolicy layer overstates Kubernetes isolation guarantees, and egress-denial behavior conflicts with the existing stale-cache contract. Details are inline. CI is green, and the document rename leaves no references to the old filename.

PR #186, opened 2026-08-04
feat(sources): Feishu source pack (milestone 5.4) — six tables, digit-string epoch + boolean-pushdown engine support

bakey:
## Summary

Milestone 5.4: the **Feishu source pack** — six tables over the `feishu.*` read actions (OAuth user_access_token), the chat-history-first wave: `chats`, `messages`, `chat_members`, `tasks`, `wiki_spaces`, `wiki_nodes`.

**Draft until the real-workspace verification (skill phase 4) lands.** The gateway declares every feishu items schema loose (zero declared properties), so no column is protected by the fingerprint gate — the coverage-gap pin records all columns as passthrough-only, and column truth can only be settled by scanning real rows. The spec entry stays unticked and the fixtures are marked synthetic drafts until that pass replaces them with redacted live captures.

## Wire contract (live-reconciled, gateway v1.3.3)

- 396 feishu actions enumerated live; the six chosen actions' contracts captured verbatim to `fixtures/feishu/contracts/` and fingerprint-pinned.
- Executor source read end to end: the feishu list executors rebuild the pagination **envelope** (camelCase `$.items` / `$.pageToken` — provider `page_token` normalized to null at end — / `$.hasMore`) while passing item objects through **raw** (snake_case, epoch digit-string timestamps). No post-pagination filtering anywhere, so cursor termination is undamaged; in-band `code != 0` is consumed by the executors → `error_path: None` for every table.
- All six tables' exact generated input sets validated to the credential wall (calibrated 400 `invalid_input` vs 403 `authorization_failed`, default connection).

## Design decisions

- **Cursor stability**: `chats`/`messages` pin `sortType: ByCreateTimeAsc` — the activity-ordered default reshuffles rows mid-scan (skips/duplicates); creation order is immutable.
- **`messages` is per-chat history**: `containerIdType` pinned to `chat`, the chat a required `containerId` binding resource (the `block_children` shape). One pushdown: `create_time >=` → `startTime`, inclusive epoch seconds rendered as a **digit string** under the strict schema, `Inexact`. `endTime` deliberately unmapped: exclusive + flooring = dropped rows.
- **`tasks`** pins `type: my_tasks` (declared, not inherited) and omits `completed` — Feishu's state=all; `completed` equality pushes Inexact.
- **`wiki_nodes` lists one level** (children of optional `parentNodeToken`, space root default); full-tree traversal is documented client-side recursion.

## Engine extensions (each backward-compatible, engine-level tested)

- `FieldType::TimestampMillisStringUtc` / `TimestampSecondsStringUtc` (`timestamp_ms_string_utc` / `timestamp_s_string_utc`): Feishu's epochs are digit STRINGS (im millis, wiki seconds) — neither existing timestamp type accepts them. Strict digit-string rule; overflow keeps its own diagnosis.
- `ValueFormat::EpochSecondsString` (`epoch_seconds_string`): epoch-seconds flooring rendered as a JSON string for schemas that TYPE epoch inputs as strings (`startTime`, minLength 1) — a number there is a hard 400.
- **Simplified-boolean pushdown normalization**: DataFusion's simplifier hands pushdown `completed` / `NOT completed` instead of `= true/false`, so boolean Eq mappings were unreachable; `translate_one` now normalizes both spellings back to the equality they mean.

## Verification

- `cargo test -p skardi --lib` → **822 passed** (16 new pack tests + 3 engine tests).
- Pack tests: fixture conversions (nulls/omissions/nesting/extras + admission-gate schema mismatch), fingerprint sync + coverage-gap pins, two-page cursor e2e with ordering pin and null-token termination, digit-string filter pushdown e2e (exact declared inputs, nothing else), required+optional resource forwarding, pre-HTTP resource enforcement, LIMIT early stop, UDTF parity, drift refusal.
- Note the shared pins: chats/messages/chat_members share one declared envelope schema (identical hash), as do the two wiki tables — the same shared-contract situation as Notion's search tables, recorded in the module doc.

## Remaining before ready-for-review

Phase-4 real-data pass against a live Feishu workspace (needs an operator-created Feishu custom app + OAuth authorization through the gateway): per-table scans through skardi-server, both-direction diff of real row keys vs mapped columns, fixture re-derivation as redacted live captures, then spec tick with counted verification.

---

## Phase-4 real-workspace verification (update, 2026-08-04)

Live pass against a real Feishu workspace (OAuth user token via the gateway's authorization-code flow; gateway v1.3.3). **tasks / wiki_spaces / wiki_nodes are now fully verified; the three im tables remain drafts pending the test app's bot capability.**

**Two contract defects caught by real rows (the reason this phase exists):**

1. **`tasks` had an always-NULL draft column with a pushdown on it.** Real rows carry no `completed` boolean — completion is `status` (todo|done) + `completed_at`, whose "not completed" spelling is the digit string `"0"` (epoch-zero sentinel, documented at the column). The Inexact pushdown on that column would have re-trimmed **every** row to zero. Dropped the column and mapping, mapped `status` instead; the action's `completed` input stays deliberately unmapped (a status string cannot render into a boolean input without a value transform the engine deliberately lacks).
2. **Feishu wiki answers its final page with `has_more: false` beside a NON-empty `page_token`** (`"0||…"`; page 2 returns an empty page with yet another non-empty token). Null-token cursor termination therefore refetches a finished scan and dies as `PaginationLoop`. New engine field **`has_more_path`** on the Cursor strategy (optional, backward-compatible): when declared it is the authoritative termination; `hasMore: true` without a usable token fails as `PaginationCursorInvalid`, a non-boolean signal as the new `PaginationHasMoreInvalid`. Declared on all six feishu tables; 4 engine tests plus a live-shape e2e pin both sides.

**Live evidence (skardi-server → real gateway → real workspace):**
- Registration passes the fingerprint gate against LIVE discovery for all bound tables.
- `tasks`: 9 real rows (6 todo / 3 done); **every mapped column non-NULL somewhere** (`due` on 2 of 9 rows); `GROUP BY status` + sentinel check: done → real 2026 instants, todo → 1970-01-01 sentinel exactly as documented.
- `wiki_spaces`: completes in ONE request on the real `hasMore:false` + non-empty-token page — the scan that previously would have looped.
- `wiki_nodes`: 3 real nodes; epoch-second digit strings decode to real instants (2021-11-05T12:18:46Z); required `spaceId` resource forwarded; gains `creator`/`url` columns from live reconciliation (`wiki_spaces` gains `open_sharing`).
- Fixtures for the three verified tables are now **redacted live captures** (ids/names/urls synthetic, structure and envelope verbatim — including the non-empty final pageToken).

**Remaining before ready-for-review:** the im tables (`chats`/`messages`/`chat_members`) — Feishu gates im APIs on the app's **bot capability** (error 232025) independently of OAuth scopes. Once enabled on the test app, the same live pass completes for them.

**Upstream notes (to file on oomol-lab/open-connector):** (a) OAuth authorization URLs request the union of ALL ~400 feishu actions' scopes with no config-level narrowing — Feishu rejects the request (20027) unless the app enables every one; verified locally with a narrowed scope list. (b) wiki executors pass Feishu's non-empty final `page_token` through instead of normalizing it to the null the envelope contract implies.

Full suite: `cargo test -p skardi --lib` → **827 passed**.

### Update: chats / chat_members verified; messages page_size corrected (1d780ae)

Bot capability enabled on the test app unblocked the im probes. **Five of six tables are now fully live-verified end to end** (registration through live discovery + engine scans + redacted-live-capture fixtures): `chats` (8 real chats; gains `chat_mode`/`chat_status` from reconciliation; the assistant chat legitimately lacks `owner_id` — nullable proven live), `chat_members` (wire carries exactly the four mapped columns), plus the previously verified `tasks`/`wiki_spaces`/`wiki_nodes`.

Third live-only contract defect caught: **Feishu's `im/v1/messages` caps `page_size` at 50 on the wire** (99992402 above it) while the gateway schema declares a maximum of 100 — corrected to 50, pinned by the e2e.

`messages` rows remain the one unverified surface: with all OAuth scopes granted AND bot capability on, reads still answer 99991679 — Feishu's third gate, **tenant-admin approval of the high-sensitivity `im:message.*.get_as_user` permissions**, pending on the test app. Once approved, the same live pass completes and this PR leaves draft.

### Final: all six tables live-verified — ready for review (f82d1de)

`messages` cleared its last gate (the `im:message:readonly` scope — see the pack doc's operational findings for why the gateway-declared `get_as_user` scopes are not what this path needs). Live e2e: **86 messages over two real cursor pages, zero duplicate ids (2020→2026)**; `create_time >= startTime` pushdown narrowing a live scan to 15 rows; field-family coverage across 8 real chats (replies' `root_id`/`parent_id`, a topic group's `thread_id`, an @-mention's `mentions`, system messages' empty-string sender ids); fixture re-derived as a redacted live capture with one row per family. Third wire-vs-schema defect recorded: `im/v1/messages` caps `page_size` at 50 (99992402) against a declared max of 100 — corrected and pinned.

Milestone 5.4 ticks with counted verification: 270 open_connector tests / 17 pack-scoped / 827 full suite. Every fixture in the pack is now a redacted live capture.

### DIFF
--- a/crates/skardi/src/sources/providers/open_connector/filters.rs
+++ b/crates/skardi/src/sources/providers/open_connector/filters.rs
@@ -50,8 +50,18 @@ pub enum ValueFormat {
     /// precision. Flooring widens a *lower* bound (a superset — safe under
     /// [`Fidelity::Inexact`] re-filtering); for an upper-bound provider
     /// parameter it would narrow the fetch and drop rows, so epoch-seconds
-    /// mappings are for lower-bound inputs (`ts_from`-style) only.
+    /// mappings are for lower-bound inputs (`ts_from`-style) only — the
+    /// pack loader REJECTS any other operator (and any fidelity but
+    /// Inexact) at load time. Pre-epoch instants clamp to `0`: still a
+    /// lower-bound superset, and providers with unsigned epoch inputs
+    /// (Feishu 400s a `-` sign) never see a negative.
     EpochSeconds,
+    /// [`ValueFormat::EpochSeconds`] rendered as a JSON *string* of decimal
+    /// digits, for providers whose strict input schemas type their epoch
+    /// inputs as strings (Feishu's `startTime`, `minLength: 1`) — a JSON
+    /// number there is a hard 400. Same flooring semantics, so the same
+    /// lower-bound-only rule applies.
+    EpochSecondsString,
 }
 
 /// One allowlisted pushdown rule: `column <operator> literal` → `input_field: literal`.
@@ -113,6 +123,39 @@ pub fn translate_filters(filters: &[Expr], mappings: &[FilterMapping]) -> Transl
 
 /// Translate one filter, returning `(input_field, value, fidelity)` on a match.
 fn translate_one(filter: &Expr, mappings: &[FilterMapping]) -> Option<(String, Value, Fidelity)> {
+    // DataFusion's simplifier rewrites boolean equality to the bare
+    // column (`completed = true` → `completed`) or its negation
+    // (`completed = false` → `NOT completed`) before pushdown, so an Eq
+    // mapping on a boolean column would otherwise never fire. Normalize
+    // both spellings back to the equality they mean.
+    //
+    // Invariant: these two arms fabricate a Boolean literal without
+    // checking the mapped column's declared type, because only a
+    // boolean-typed column can appear as a bare predicate in a
+    // type-checked DataFusion plan — an Eq mapping on a non-boolean
+    // column is reachable solely through the BinaryExpr arm below, where
+    // the literal comes from the query.
+    match filter {
+        Expr::Column(column) => {
+            let mapping = mappings.iter().find(|mapping| {
+                mapping.column == column.name && mapping.operator == Operator::Eq
+            })?;
+            let value = scalar_to_json(&ScalarValue::Boolean(Some(true)), mapping.value_format)?;
+            return Some((mapping.input_field.to_string(), value, mapping.fidelity));
+        }
+        Expr::Not(inner) => {
+            let Expr::Column(column) = inner.as_ref() else {
+                return None;
+            };
+            let mapping = mappings.iter().find(|mapping| {
+                mapping.column == column.name && mapping.operator == Operator::Eq
+            })?;
+            let value = scalar_to_json(&ScalarValue::Boolean(Some(false)), mapping.value_format)?;
+            return Some((mapping.input_field.to_string(), value, mapping.fidelity));
+        }
+        _ => {}
+    }
+
     let Expr::BinaryExpr(binary) = filter else {
         return None;
     };
@@ -184,25 +227,40 @@ fn scalar_to_json(literal: &ScalarValue, format: ValueFormat) -> Option<Value> {
         // Verbatim arms return None on purpose: the mapping declared no
         // timestamp spelling, so the predicate stays local rather than
         // being pushed in a guessed format.
+        // Epoch renderings clamp pre-epoch instants to 0 (see the
+        // ValueFormat docs): a negative would render with a `-` sign that
+        // strict digit-string schemas 400 on, failing the whole scan where
+        // no pushdown at all would have succeeded — and 0 is still a
+        // superset for the lower bounds these formats are restricted to.
         ScalarValue::TimestampSecond(Some(v), _) => match format {
             ValueFormat::Verbatim => None,
             ValueFormat::Rfc3339 => DateTime::from_timestamp(*v, 0).map(rfc3339),
-            ValueFormat::EpochSeconds => Some(Value::from(*v)),
+            ValueFormat::EpochSeconds => Some(Value::from((*v).max(0))),
+            ValueFormat::EpochSecondsString => Some(Value::from((*v).max(0).to_string())),
         },
         ScalarValue::TimestampMillisecond(Some(v), _) => match format {
             ValueFormat::Verbatim => None,
             ValueFormat::Rfc3339 => DateTime::from_timestamp_millis(*v).map(rfc3339),
-            ValueFormat::EpochSeconds => Some(Value::from(v.div_euclid(1000))),
+            ValueFormat::EpochSeconds => Some(Value::from(v.div_euclid(1000).max(0))),
+            ValueFormat::EpochSecondsString => {
+                Some(Value::from(v.div_euclid(1000).max(0).to_string()))
+            }
         },
         ScalarValue::TimestampMicrosecond(Some(v), _) => match format {
             ValueFormat::Verbatim => None,
             ValueFormat::Rfc3339 => DateTime::from_timestamp_micros(*v).map(rfc3339),
-            ValueFormat::EpochSeconds => Some(Value::from(v.div_euclid(1_000_000))),
+            ValueFormat::EpochSeconds => Some(Value::from(v.div_euclid(1_000_000).max(0))),
+            ValueFormat::EpochSecondsString => {
+                Some(Value::from(v.div_euclid(1_000_000).max(0).to_string()))
+            }
         },
         ScalarValue::TimestampNanosecond(Some(v), _) => match format {
             ValueFormat::Verbatim => None,
             ValueFormat::Rfc3339 => Some(rfc3339(DateTime::from_timestamp_nanos(*v))),
-            ValueFormat::EpochSeconds => Some(Value::from(v.div_euclid(1_000_000_000))),
+            ValueFormat::EpochSeconds => Some(Value::from(v.div_euclid(1_000_000_000).max(0))),
+            ValueFormat::EpochSecondsString => {
+                Some(Value::from(v.div_euclid(1_000_000_000).max(0).to_string()))
+            }
         },
         _ => None,
     }
@@ -254,6 +312,20 @@ mod tests {
             input_field: "ts_from",
             fidelity: Fidelity::Inexact,
             value_format: ValueFormat::EpochSeconds,
+        },
+        FilterMapping {
+            column: "create_time",
+            operator: Operator::GtEq,
+            input_field: "startTime",
+            fidelity: Fidelity::Inexact,
+            value_format: ValueFormat::EpochSecondsString,
+        },
+        FilterMapping {
+            column: "completed",
+            operator: Operator::Eq,
+            input_field: "completed",
+            fidelity: Fidelity::Inexact,
+            value_format: ValueFormat::Verbatim,
         },
     ];
 
@@ -616,6 +688,99 @@ mod tests {
             assert_eq!(
                 translated.inputs,
                 vec![("ts_from".to_string(), Value::from(1_767_225_600i64))],
+            );
+            assert_eq!(
+                translated.pushdown,
+                vec![TableProviderFilterPushDown::Inexact]
+            );
+        }
+    }
+
+    #[test]
+    fn simplified_boolean_predicates_normalize_back_to_equality() {
+        // DataFusion's simplifier hands pushdown `completed` instead of
+        // `completed = true` and `NOT completed` instead of
+        // `completed = false`; both must still reach the mapped input —
+        // otherwise a boolean Eq mapping is dead code.
+        let translated = translate_filters(&[col("completed")], MAPPINGS);
+        assert_eq!(
+            translated.inputs,
+            vec![("completed".to_string(), Value::from(true))]
+        );
+        assert_eq!(
+            translated.pushdown,
+            vec![TableProviderFilterPushDown::Inexact]
+        );
+
+        let translated = translate_filters(&[Expr::Not(Box::new(col("completed")))], MAPPINGS);
+        assert_eq!(
+            translated.inputs,
+            vec![("completed".to_string(), Value::from(false))]
+        );
+
+        // A bare column without an Eq mapping stays local — and NOT over
+        // anything but a bare column is never translated.
+        let translated = translate_filters(&[col("value")], MAPPINGS);
+        assert!(translated.inputs.is_empty());
+        assert_eq!(
+            translated.pushdown,
+            vec![TableProviderFilterPushDown::Unsupported]
+        );
+        let translated = translate_filters(&[Expr::Not(Box::new(gt("value", 5)))], MAPPINGS);
+        assert!(translated.inputs.is_empty());
+    }
+
+    #[test]
+    fn pre_epoch_instants_clamp_to_zero_never_a_signed_string() {
+        // `create_time >= '1965-01-01'` would otherwise render "-157766400"
+        // — not a digit string, so a strict provider schema (Feishu's
+        // startTime) 400s and the whole scan fails where no pushdown at
+        // all would have succeeded. Zero is still a superset for the
+        // lower bounds these formats are restricted to.
+        let filter = Expr::BinaryExpr(BinaryExpr::new(
+            Box::new(col("create_time")),
+            Operator::GtEq,
+            Box::new(Expr::Literal(
+                ScalarValue::TimestampMillisecond(Some(-157_766_400_000), Some("UTC".into())),
+                None,
+            )),
+        ));
+        let translated = translate_filters(&[filter], MAPPINGS);
+        assert_eq!(
+            translated.inputs,
+            vec![("startTime".to_string(), Value::from("0"))],
+        );
+        assert_eq!(
+            translated.pushdown,
+            vec![TableProviderFilterPushDown::Inexact]
+        );
+    }
+
+    #[test]
+    fn epoch_seconds_string_mappings_render_digit_strings() {
+        // Feishu-style startTime: same flooring semantics as epoch_seconds,
+        // but the provider's strict schema types the input as a STRING —
+        // the rendered literal must be a JSON string of digits, never a
+        // number.
+        let epoch_ms = 1_767_225_600_000i64; // 2026-01-01T00:00:00Z
+        for scalar in [
+            ScalarValue::TimestampSecond(Some(epoch_ms / 1000), None),
+            ScalarValue::TimestampMillisecond(Some(epoch_ms), Some("UTC".into())),
+            ScalarValue::TimestampMicrosecond(Some(epoch_ms * 1000), None),
+            ScalarValue::TimestampNanosecond(Some(epoch_ms * 1_000_000), None),
+            // Sub-second precision floors — widening the lower bound,
+            // which Inexact re-filtering trims back.
+            ScalarValue::TimestampMillisecond(Some(epoch_ms + 999), None),
+        ] {
+            let filter = Expr::BinaryExpr(BinaryExpr::new(
+                Box::new(col("create_time")),
+                Operator::GtEq,
+                Box::new(Expr::Literal(scalar, None)),
+            ));
+            let translated = translate_filters(&[filter], MAPPINGS);
+            assert_eq!(
+                translated.inputs,
+                vec![("startTime".to_string(), Value::from("1767225600"))],
             );
             assert_eq!(
                 translated.pushdown,
--- a/crates/skardi/src/sources/providers/open_connector/packs/feishu.rs
+++ b/crates/skardi/src/sources/providers/open_connector/packs/feishu.rs
@@ -1,0 +1,1039 @@
+//! Feishu source pack: stable relational contracts over the Open Connector
+//! `feishu.*` read actions (OAuth user_access_token, cursor pagination).
+//!
+//! **The wire contract is Open Connector's HYBRID shape.** Every feishu
+//! list executor rebuilds the pagination envelope — camelCase `$.items` /
+//! `$.pageToken` / `$.hasMore`, with the provider's `page_token`
+//! normalized to a null `pageToken` at end-of-collection — while passing
+//! the Feishu API's item objects through RAW (`optionalObjectArray(...)`
+//! / `Array.isArray(...)` in the executors, no per-item rebuild). Rows
+//! therefore keep Feishu's snake_case keys and its epoch-digit-STRING
+//! timestamps: millis for im and tasks (`create_time: "1609296809000"`,
+//! `completed_at: "1780322529638"`), seconds for wiki
+//! (`obj_create_time: "1642402790"`) — the reason
+//! `TimestampMillisStringUtc` / `TimestampSecondsStringUtc` exist.
+//! Inputs are Open Connector's camelCase strict schemas (`pageSize`,
+//! `pageToken` with `minLength: 1`, `chatId`, …). Everything below is
+//! reconciled against a live gateway (v1.3.3) and the OC provider
+//! source.
+//!
+//! Design decisions, per the integration design spec and the source-pack
+//! admission gate:
+//!
+//! - **Cursor pagination on every table** (`pageToken` in, top-level
+//!   `$.pageToken` out; page size 100 for chats/chat_members/tasks, 50
+//!   for wiki — and 50 for `messages`, whose REAL wire cap is 50 despite
+//!   the schema's declared 100: Feishu hard-fails larger values with
+//!   99992402, live-verified 2026-08-04), with `$.hasMore` declared as
+//!   the AUTHORITATIVE
+//!   termination signal (`has_more_path`). That is a live-verification
+//!   correction, not a nicety: Feishu's wiki space listing answers its
+//!   final page with `has_more: false` beside a NON-empty `page_token`
+//!   (`"0||…"`, captured 2026-08-04), so the null-token spelling alone
+//!   would refetch a finished scan and die as `PaginationLoop`. With the
+//!   signal declared, `hasMore: true` without a usable token is contract
+//!   drift (`PaginationCursorInvalid`), a non-boolean signal is
+//!   `PaginationHasMoreInvalid` — never a silent truncation. No feishu
+//!   executor filters fetched pages, so the signals are undamaged.
+//! - **Orderings are pinned for cursor stability**: `chats` and
+//!   `messages` pin `sortType: ByCreateTimeAsc` because the API's
+//!   activity-ordered default reshuffles rows mid-scan (skips and
+//!   duplicates across pages); creation order is immutable.
+//! - **`messages` is chat history**: `containerIdType` pinned to `chat`
+//!   (threads are a different action), the chat itself a required
+//!   binding resource — the per-container shape Notion's
+//!   `block_children` established. One filter pushes: `create_time >=`
+//!   → `startTime`, inclusive epoch seconds typed as a digit STRING in
+//!   the strict schema (the reason `ValueFormat::EpochSecondsString`
+//!   exists), `Inexact` so DataFusion re-trims the floored bound.
+//!   `endTime` is deliberately unmapped: it is EXCLUSIVE, and flooring
+//!   an upper bound drops rows.
+//! - **`tasks` pins `type: my_tasks`** (the executor default, declared
+//!   rather than inherited) and omits the action's `completed` input —
+//!   Feishu's state=all. Live rows (2026-08-04) carry NO `completed`
+//!   boolean: completion is `status` (todo|done) plus `completed_at`,
+//!   whose "not completed" spelling is the digit string "0" (epoch-zero
+//!   sentinel, documented at the column). The `completed` input is
+//!   therefore deliberately unmapped — a `status` string cannot render
+//!   into a boolean input without a value transform the filter engine
+//!   deliberately lacks, and a mapping on an always-NULL draft column
+//!   would have re-trimmed every row to zero.
+//! - **`wiki_nodes` lists ONE level** (children of `parentNodeToken`,
+//!   space root when omitted) — the action's own shape; full-tree
+//!   traversal is client-side recursion, documented in the pack doc.
+//! - **In-band provider errors never reach rows**: executors throw on
+//!   Feishu's non-zero `code` envelope, so the gateway returns a failure
+//!   envelope and `error_path` is `None` for every table.
+//! - **Fingerprints are pinned** from the live capture
+//!   (`fixtures/feishu/contracts/`), but the gateway declares every
+//!   item schema LOOSE (`additionalProperties: true`, zero declared
+//!   properties) — so ALL mapped columns ride passthrough outside the
+//!   fingerprint gate, and the coverage-gap pin records that honestly.
+//!   Column truth is therefore settled ONLY by real rows, and ALL SIX
+//!   tables are reconciled against a live workspace (2026-08-04; every
+//!   fixture is a redacted live capture). What the pass changed: chats
+//!   gained `chat_mode`/`chat_status`; tasks lost the nonexistent
+//!   `completed` boolean for `status`/`completed_at`; wiki tables
+//!   gained `open_sharing`/`creator`/`url`; messages' page size dropped
+//!   to the real 50 cap. Two operational findings the pack doc records:
+//!   reading messages under the user identity requires the
+//!   `im:message:readonly` (or `im:message` /
+//!   `im:message.history:readonly`) scope — the `get_as_user` scopes
+//!   the gateway's actions declare are NOT honored for this path
+//!   (99991679 names the real set) — plus the app's bot capability
+//!   (232025). `message_position` (a digit string on every live row) is
+//!   deliberately unmapped: no public Feishu doc pins its semantics.
+//!   Live e2e evidence: 86 messages over two real cursor pages with
+//!   zero duplicate ids; the `create_time >=` pushdown narrowing a
+//!   live scan; wiki's non-empty final token terminating cleanly.
+
+use std::sync::OnceLock;
+
+use crate::sources::providers::open_connector::error::OpenConnectorError;
+use crate::sources::providers::open_connector::source_pack::SourcePack;
+
+use super::loader;
+
+static PACK: OnceLock<Result<SourcePack, String>> = OnceLock::new();
+
+/// The Feishu pack, parsed once from the embedded YAML asset.
+pub fn pack() -> Result<&'static SourcePack, OpenConnectorError> {
+    loader::builtin("feishu.yaml", include_str!("feishu.yaml"), &PACK)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::sources::hierarchy::HierarchyLevel;
+    use crate::sources::providers::open_connector::action_registry::fingerprint_schema;
+    use crate::sources::providers::open_connector::json_to_arrow::RowConverter;
+    use crate::sources::providers::open_connector::row_path::RowPath;
+    use crate::sources::providers::open_connector::source_pack::SourcePackTable;
+    use crate::sources::providers::open_connector::testutil::{
+        EnvVarGuard, MockGateway, MockResponse, discovery_ok, envelope_ok,
+        fingerprint_uncovered_columns,
+    };
+    use crate::sources::providers::open_connector::{
+        OpenConnectorConfig, OpenConnectorGateways, register_open_connector_tables,
+        register_open_connector_udtfs,
+    };
+    use arrow::array::{Array, BooleanArray, StringArray, TimestampMillisecondArray};
+    use arrow::record_batch::RecordBatch;
+    use datafusion::prelude::SessionContext;
+    use serde_json::{Value, json};
+
+    /// Look up a table by short name; the assets are test-pinned to parse.
+    fn table(short: &str) -> &'static SourcePackTable {
+        pack()
+            .expect("embedded asset is test-pinned to parse")
+            .tables
+            .iter()
+            .find(|t| t.id.rsplit('.').next() == Some(short))
+            .unwrap_or_else(|| panic!("table {short}"))
+    }
+
+    /// Discovery serving the live-captured contracts, so every mock
+    /// registration exercises the fingerprint gate's pass side.
+    fn feishu_discovery(path: &str) -> MockResponse {
+        let output_schema = if path.ends_with("feishu.list_chats") {
+            include_str!("fixtures/feishu/contracts/list_chats.json")
+        } else if path.ends_with("feishu.list_messages") {
+            include_str!("fixtures/feishu/contracts/list_messages.json")
+        } else if path.ends_with("feishu.list_chat_members") {
+            include_str!("fixtures/feishu/contracts/list_chat_members.json")
+        } else if path.ends_with("feishu.list_tasks") {
+            include_str!("fixtures/feishu/contracts/list_tasks.json")
+        } else if path.ends_with("feishu.list_wiki_spaces") {
+            include_str!("fixtures/feishu/contracts/list_wiki_spaces.json")
+        } else if path.ends_with("feishu.list_wiki_nodes") {
+            include_str!("fixtures/feishu/contracts/list_wiki_nodes.json")
+        } else {
+            r#"{"type": "object"}"#
+        };
+        MockResponse::ok(&discovery_ok("{}", output_schema, true, None))
+    }
+
+    // ── Contract tests: bundled fixtures are the build-time conversion
+    // contract (null-bearing, nested, empty, extra upstream fields, and a
+    // schema mismatch per the admission gate). DRAFT status: synthetic,
+    // re-derived from live captures in the real-data phase. ─────────────
+
+    fn convert_fixture(table: &SourcePackTable, fixture: &str) -> RecordBatch {
+        let page: Value = serde_json::from_str(fixture).expect("fixture parses");
+        let rows = RowPath::parse(table.row_path)
+            .expect("row path")
+            .rows(&page, 1)
+            .expect("row array");
+        RowConverter::new(table.fields)
+            .expect("converter")
+            .convert(rows, 1)
+            .expect("fixture converts")
+    }
+
+    fn utf8<'a>(batch: &'a RecordBatch, name: &str) -> &'a StringArray {
+        batch
+            .column_by_name(name)
+            .unwrap_or_else(|| panic!("column {name}"))
+            .as_any()
+            .downcast_ref()
+            .expect("Utf8 column")
+    }
+
+    fn boolean<'a>(batch: &'a RecordBatch, name: &str) -> &'a BooleanArray {
+        batch
+            .column_by_name(name)
+            .unwrap_or_else(|| panic!("column {name}"))
+            .as_any()
+            .downcast_ref()
+            .expect("Boolean column")
+    }
+
+    fn millis<'a>(batch: &'a RecordBatch, name: &str) -> &'a TimestampMillisecondArray {
+        batch
+            .column_by_name(name)
+            .unwrap_or_else(|| panic!("column {name}"))
+            .as_any()
+            .downcast_ref()
+            .expect("Timestamp column")
+    }
+
+    #[test]
+    fn chats_fixture_converts_live_shapes() {
+        // Redacted live capture (2026-08-04): an assistant chat without an
+        // owner, an internal group with an EMPTY description, and a topic
+        // chat.
+        let batch = convert_fixture(table("chats"), include_str!("fixtures/feishu/chats.json"));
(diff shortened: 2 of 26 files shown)


crates/skardi/src/sources/providers/open_connector/packs/fixtures/feishu/messages.json
```diff
@@ -0,0 +1,95 @@
+{
+  "items": [
+    {
+      "body": {
+        "content": "{\"template\": \"{to_chatters} joined {group_name}\", \"from_user\": [], \"to_chatters\": [\"pyh\"], \"divider_text\": {}}"
```
gracexmatin: Real member name in the nested JSON: `to_chatters: ["pyh"]`.

crates/skardi/src/sources/providers/open_connector/packs/fixtures/feishu/messages.json
```diff
@@ -0,0 +1,95 @@
...
+        "id_type": "open_id",
+        "sender_type": "user",
+        "tenant_key": "t-0001"
+      },
+      "update_time": "1681116815304",
+      "updated": false
+    },
+    {
+      "body": {
+        "content": "{\"template\": \"{to_chatters} joined {group_name}\", \"from_user\": [\"陈桦Grace\"], \"to_chatters\": [\"朱天梁\"], \"divider_text\": {}}"
```
gracexmatin: Real member names in the nested JSON: `from_user: ["陈桦Grace"]`, `to_chatters: ["朱天梁"]`.

gracexmatin (commented):
**P1**: real member names survived redaction in `fixtures/feishu/messages.json`. The `msg_type: "system"` rows keep real names inside the JSON-encoded `body.content`: `to_chatters: ["pyh"]`, `from_user: ["陈桦Grace"]` / `to_chatters: ["朱天梁"]`. The mechanical audit walks the outer JSON tree only, so strings one decode level deeper never surface — same family as the "titles hide in URL slugs" blind spot.

Fix: swap the three names for counter-style placeholders (cheap — no test pins them; `feishu.rs:236` asserts only the `template` key), and extend the audit to decode-and-recurse into JSON-shaped string values. Worth folding into the skill as a #190 follow-up.

Is it worthy to fix?

Userneima:
Tried this pack end to end as a first-time user today (2026-08-05) and got all six tables returning real rows from my own workspace. Notes are ordered by how much time each cost me; every item ends with a concrete suggestion, and there's a short list at the bottom of the ones I'm happy to open as PRs myself.

**What ran.** Real workspace, read-only:

| table | rows | latency |
|---|---|---|
| `chats` | 5 | 569 ms |
| `wiki_spaces` | 2 | 537 ms |
| `tasks` | 0 | 252 ms |
| `chat_members` | 4 (8 in another chat) | 602 ms |
| `wiki_nodes` | 3 (4 in another space, 2 at depth 2) | 645 ms |
| `messages` | 367 (full history) | 4008 ms |

Three things I verified independently:

- **Cursor pagination holds up.** Full history of one chat: **367 messages, ~8 pages, zero duplicate ids**, spanning 7-14 → 8-05 in correct order. Wider coverage than the 86 rows / 2 pages in the PR description.
- **The `create_time >=` pushdown genuinely fires.** Same chat: 4113 ms / 367 rows (8 pages) unfiltered, vs 1182 ms / 72 rows (2 pages) with `create_time >= 2026-08-04`. Six fewer pages, 3.5× faster — the predicate really reaches Feishu rather than being applied locally.
- **`tasks` returning 0 rows is real.** I tried `assigned` / `created` / `followed` against the native action; Feishu answers `1470400: Invalid Param 'type'. Only 'my_tasks' is supported.`

Timestamp decoding is correct and `ORDER BY create_time DESC` works. `messages` being visibly slower tracks with per-page cursor scanning at the 50-row wire cap.

---

## 1. Behind a zero-trust corporate VPN, the egress guard blocks everything and gives no clue

**Symptom.** After clicking "authorize", the browser showed only this, and the gateway log only added an `errorCode`:

```json
{"error":{"code":"oauth_token_exchange_failed","message":"OAuth token request failed."}}
```

No status, no provider response, no hostname, no address. The failure took 36 ms — too fast for a real round trip, which was the only tell.

**Root cause.** A zero-trust VPN (aTrust / EasyConnect class) maps external domains into `198.18.0.0/15`, a range in `reservedIpv4Cidrs`. `isBlockedIpv4` checks the reserved list **unconditionally**, so `allowPrivateNetwork` cannot open it. `open.feishu.cn` resolved to `198.18.0.196` and `providerFetch` refused before the request left the process. Side finding: the shared `providerFetch` singleton is constructed without `allowPrivateNetwork`, so `OOMOL_CONNECT_ALLOW_PRIVATE_NETWORK` never reaches provider egress at all — not sure whether that's intended.

**Suggestion (two parts, both of which I can open as a PR).**

First, put enough in the error to locate the problem:

```diff
-      throw policy.createError(`${fieldName} must not resolve to private or reserved IP addresses`);
+      throw policy.createError(
+        `${fieldName} must not resolve to private or reserved IP addresses ` +
+          `(resolved ${entry.address}; if this host is reached through a corporate ` +
+          `VPN that uses a reserved range, add it to OOMOL_CONNECT_EGRESS_SKIP_IP_CHECK_HOSTS)`,
+      );
```

Second, a deployment-level hostname allowlist matching exact hosts or `.suffix`, leaving every other check (scheme, port, no redirects, response size cap) untouched:

```ts
// src/core/guarded-fetch.ts
function egressHostAllowlisted(hostname: string): boolean {
  const raw = process.env.OOMOL_CONNECT_EGRESS_SKIP_IP_CHECK_HOSTS;
  if (!raw) return false;
  const host = hostname.toLowerCase().replace(/\.$/, "");
  return raw.split(",").map(s => s.trim().toLowerCase()).filter(Boolean).some(p =>
    p.startsWith(".") ? (host === p.slice(1) || host.endsWith(p)) : host === p);
}

// inside the loop that validates resolved addresses
const skipIpCheck = egressHostAllowlisted(hostname);
if (!skipIpCheck && isBlockedIpAddress(entry.address, policy.allowPrivateNetwork)) { throw ... }
```

Started with `OOMOL_CONNECT_EGRESS_SKIP_IP_CHECK_HOSTS=".feishu.cn,.larksuite.com"`. That's how I got the whole OAuth flow working on a local copy.

## 2. OAuth failures discard the provider's response on at least three paths

**Symptom.** Three completely different failures throw the same `"OAuth token request failed."`: the fetch throwing, the body being unreadable, and `readTokenErrorMessage` extracting nothing. On a first failure there's no way to tell a network problem from a credential problem from a response-format problem.

**Root cause (with incidental evidence).** I added logging to the non-2xx path and **my patch never fired** — the real throw site was earlier, in the fetch layer. If the person patching it picks the wrong site, the paths really are indistinguishable.

**Suggestion.** Give each throw site its own context:

```diff
   } catch (error) {
-    throw input.createError(
-      error instanceof Error && error.name === "TimeoutError"
-        ? "OAuth token request timed out."
-        : "OAuth token request failed.",
-    );
+    if (error instanceof Error && error.name === "TimeoutError") {
+      throw input.createError("OAuth token request timed out.");
+    }
+    const cause = (error as { cause?: { code?: string } })?.cause?.code;
+    throw input.createError(
+      `OAuth token request failed at transport level: ` +
+        `${(error as Error)?.name}: ${(error as Error)?.message}` +
+        (cause ? ` (cause: ${cause})` : ""),
+    );
   }
```

Plus: when the response is non-2xx and no structured message can be extracted, attach the status and a truncated body.

## 3. The gateway-version requirement is a hard prerequisite but reads as a footnote

**Symptom.** Registration fails with `Open Connector action 'feishu.list_chats' was not found on the gateway`, which reads like a config typo.

**Root cause.** These six actions only exist on newer upstream. My local gateway (a July branch) exposed **7** feishu actions, all docs/bitable; the six this pack needs live in `im-user-actions.ts` / `task-actions.ts` / `wiki-actions.ts`, and current upstream exposes **396**. "Reconciled against a live gateway (v1.3.3)" is a statement of fact, and a reader won't realize it's a floor.

**Suggestion (two parts).**

First, a self-check command in the doc so a reader who fails it doesn't go further:

```bash
curl -s -H "Authorization: Bearer <runtime token>" \
  "<gateway>/v1/actions?service=feishu&limit=500" \
  | python3 -c "import json,sys; d=json.load(sys.stdin); \
    ids=[i['id'] for i in d['data']['items']]; print(len(ids)); \
    print([n for n in ['feishu.list_chats','feishu.list_messages','feishu.list_chat_members','feishu.list_tasks','feishu.list_wiki_spaces','feishu.list_wiki_nodes'] if n not in ids])"
# expect: 396 (or more), then an empty list []
```

Second, when an action is missing, look up whether the gateway has the service at all and how many actions it exposes, so the error can point somewhere:

```
Open Connector action 'feishu.list_chats' was not found on the gateway
(the gateway exposes 7 actions for service 'feishu'; this pack needs 6 that
were added upstream later — the gateway may be too old)
```

## 4. Authorization requests 164 scopes; six read-only tables need 10

**Symptom.** Feishu rejects the whole request (20027) unless the app has enabled every one. That's a hard stop for a fresh app.

**Root cause.** `feishuOAuthScopes` is `offline_access` plus the union of `providerPermissions` across all 28 action modules. Measured on the live authorize URL: **164 scopes** (statically, deduped `providerPermissions` is 59; the real request also folds in `requiredScopes`), including `calendar:calendar.event:delete`, `mail:user_mailbox.message:send` and `im:message.send_as_user` — destructive write scopes, to read six tables.

**Suggestion.** The doc already tells deployers to expect to narrow this themselves, so it saves everyone the same detour to just show the edit:

```ts
// src/providers/feishu/definition.ts
const feishuOAuthScopes = [
  "offline_access",
  "im:chat:read",                       // chats
  "im:chat.members:read",               // chat_members
  "im:message:readonly",                // messages (next three also messages)
  "im:message.group_msg:get_as_user",
  "im:message.p2p_msg:get_as_user",
  "im:message.reactions:read",
  "task:task:read",                     // tasks
  "wiki:space:retrieve",                // wiki_spaces
  "wiki:node:retrieve",                 // wiki_nodes
];
```

I verified the authorize URL then requests exactly those 10. Also, the Feishu console supports bulk import/export of scopes, so shipping the same list as pasteable JSON in the doc turns "click through a dozen sections" into one paste:

```json
{"scopes":{"tenant":[],"user":["offline_access","im:chat:read","im:chat.members:read",
"im:message:readonly","im:message.group_msg:get_as_user","im:message.p2p_msg:get_as_user",
"im:message.reactions:read","task:task:read","wiki:space:retrieve","wiki:node:retrieve"]}}
```

## 5. The copyable binding example omits the tables that need a resource

**Symptom.** The Binding block lists `tables: [chats, tasks, wiki_spaces]`, so I added `wiki_nodes` to the same list and startup failed with `binding 'team' is missing required resource input 'spaceId'`.

**Root cause.** The Tables table does document `spaceId (required)` — but the copyable example comes first, and people copy before they read.

**Suggestion.** Mark it in the example itself, and add a sentence about ordering:

```yaml
bindings:
  # the three tables that need no resource
  - name: team
    source_pack: feishu
    tables: [chats, tasks, wiki_spaces]
  # each of the following needs one resource, so each needs its own binding
  - name: handbook
    source_pack: feishu
    resource: { spaceId: "<from querying wiki_spaces>" }
    tables: [wiki_nodes]
  - name: standup
    source_pack: feishu
    resource: { containerId: "<from querying chats>" }
    tables: [messages]
  - name: standup_members
    source_pack: feishu
    resource: { chatId: "<from querying chats>" }
    tables: [chat_members]
```

> Note the ordering: `spaceId` / `containerId` / `chatId` all come from querying `wiki_spaces` / `chats` first, so a first ctx can only cover the three unbound tables — you come back and add the rest once you have the ids.

## 6. Four smaller ones, each with a suggestion

- **`POST /query` hides the cause.** It returns `SQL query execution failed; see server logs for details` while the real cause (`HTTP 403: authorization_failed: Connect feishu with OAuth first.`) is already in the error chain. → Suggest surfacing the tail of the chain in the response, at least outside production mode.
- **Health-check failures dump a whole HTML page into the log.** With something else on the gateway port (a dev server, in my case), the registration error embedded an entire Next.js document. → Suggest truncating to ~200 characters, or reporting status + content-type only.
- **`open_connector_query` returns 500 for a bad argument.** The second argument is the table name (`feishu.tasks`), not the action id (`feishu.list_tasks`); getting it wrong reports `source pack 'feishu' has no table 'list_tasks'`, which is user input error. → Suggest 4xx, and a sentence in the doc about the distinction (I got it wrong on my first try).
- **Wording on `tasks` / `type`.** "pins `type: my_tasks`" reads like a design trade-off; it's the only value the API accepts. → Suggest "this is the only value Feishu accepts", so readers don't assume it's tunable.

---

## Three things I can open as PRs

If the shapes look right, I can take these so you don't have to:

1. Resolved address in the guard error + the hostname allowlist (both parts of item 1)
2. `error.name` / `cause.code` on the OAuth transport-level error (item 2)
3. Docs: the gateway self-check command, the narrowed scope list and pasteable JSON, comments in the binding example (the doc parts of items 3, 4, 5)

**One thing I could not verify**: whether a **fresh** app with only those 10 scopes works end to end. The app I authorized with had been used by other tools and already had 133 scopes granted, so the narrow-scope path is only verified as far as "the authorize URL requests exactly 10".

Full walkthrough with every error verbatim, timings and the diagnostic commands is in our internal getting-started notes — happy to paste any section here.

crates/skardi/src/sources/providers/open_connector/packs/fixtures/feishu/messages.json
abbccdda: **Blocker (PII):** this "redacted live capture" ships two real personal names inside the JSON-encoded `body.content` payload: `"from_user": ["陈桦Grace"], "to_chatters": ["朱天梁"]` — and line 5 carries the handle `"pyh"`. Combined with the unredacted real capture timestamps, these are identifiable group-membership events of real people from the live workspace, contradicting the redaction claim in the spec tick, the pack doc, and the module doc. The redaction pass scrubbed top-level ids/names but never re-opened the nested JSON-encoded `content` strings.

Since `f82d1dea` and earlier commits already contain the file, please amend/squash the branch history before merge — editing only the tip leaves the names in git history.

docs/open-connector-feishu.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+## Tables
+
+| Table | Action | Resources | Pagination | Filter pushdown |
+|---|---|---|---|---|
+| `chats` | `feishu.list_chats` | — | cursor, 100/page | — |
+| `messages` | `feishu.list_messages` | `containerId` (required) | cursor, 50/page | `create_time >=` → `startTime` (inexact) |
+| `chat_members` | `feishu.list_chat_members` | `chatId` (required) | cursor, 100/page | — |
+| `tasks` | `feishu.list_tasks` | — | cursor, 100/page | `completed =` → `completed` (inexact) |
```
abbccdda: **Stale row — this pushdown no longer exists.** The live pass deliberately dropped the `completed` column and its mapping (the yaml's comment at `feishu.yaml:147-152` explains why: real rows carry no `completed` boolean, and a status string can't render into a boolean input). This row should be `—`, and the design note at line 93 ("omits `completed` unless the query pushes it") needs the same fix — as written the doc promises server-side narrowing that never happens.

crates/skardi/src/sources/providers/open_connector/filters.rs
```diff
@@ -52,6 +52,12 @@ pub enum ValueFormat {
     /// parameter it would narrow the fetch and drop rows, so epoch-seconds
     /// mappings are for lower-bound inputs (`ts_from`-style) only.
     EpochSeconds,
+    /// [`ValueFormat::EpochSeconds`] rendered as a JSON *string* of decimal
+    /// digits, for providers whose strict input schemas type their epoch
+    /// inputs as strings (Feishu's `startTime`, `minLength: 1`) — a JSON
+    /// number there is a hard 400. Same flooring semantics, so the same
+    /// lower-bound-only rule applies.
+    EpochSecondsString,
```
abbccdda, line 64: **Minor:** the lower-bound-only rule is documented here (and on `EpochSeconds`) but unenforced. A future pack declaring `op: eq` with an epoch-seconds format on a sub-second column would floor the literal and push an equality at a *different* instant — the provider then drops rows the predicate matches, which `Inexact` re-filtering cannot recover. `validate_table` in loader.rs already does cross-field checks and would be the natural place to reject `Eq` + epoch-seconds formats. Pre-existing gap for `EpochSeconds`, but this PR doubles the surface.

crates/skardi/src/sources/providers/open_connector/filters.rs
```diff
@@ -188,21 +220,29 @@ fn scalar_to_json(literal: &ScalarValue, format: ValueFormat) -> Option<Value> {
             ValueFormat::Verbatim => None,
             ValueFormat::Rfc3339 => DateTime::from_timestamp(*v, 0).map(rfc3339),
             ValueFormat::EpochSeconds => Some(Value::from(*v)),
+            ValueFormat::EpochSecondsString => Some(Value::from(v.to_string())),
```
abbccdda: **Minor:** a negative epoch (e.g. `create_time >= '1965-01-01'`) renders as `"-157766400"` — not a digit string, so Feishu's strict schema 400s and the whole scan fails where no pushdown would have succeeded. Clamping negatives to `"0"` is strictly better: still a superset for a lower bound, still Inexact-safe. (The same latent issue exists for `EpochSeconds`, but this string variant's stricter provider schema makes the 400 concrete.)

crates/skardi/src/sources/providers/open_connector/filters.rs
```diff
@@ -113,6 +119,32 @@ pub fn translate_filters(filters: &[Expr], mappings: &[FilterMapping]) -> Transl
 
 /// Translate one filter, returning `(input_field, value, fidelity)` on a match.
 fn translate_one(filter: &Expr, mappings: &[FilterMapping]) -> Option<(String, Value, Fidelity)> {
+    // DataFusion's simplifier rewrites boolean equality to the bare
+    // column (`completed = true` → `completed`) or its negation
+    // (`completed = false` → `NOT completed`) before pushdown, so an Eq
+    // mapping on a boolean column would otherwise never fire. Normalize
+    // both spellings back to the equality they mean.
+    match filter {
+        Expr::Column(column) => {
```
abbccdda, line 139: **Minor:** this normalization keys only on `(column.name, Operator::Eq)` and unconditionally fabricates `Boolean(true)` — nothing checks the mapped column is boolean-typed. Unreachable through a type-checked DataFusion plan today (a bare non-boolean column can't appear as a predicate), but a schema/mapping mismatch in a future pack would silently push `name=true`. A one-line comment stating the invariant, or a field-type cross-check in `validate_table`, would close it.

crates/skardi/src/sources/providers/open_connector/packs/feishu.rs
```diff
@@ -0,0 +1,959 @@
...
+        let batch = convert_fixture(table("tasks"), include_str!("fixtures/feishu/tasks.json"));
+        assert_eq!(batch.num_rows(), 3);
+        assert_eq!(utf8(&batch, "status").value(0), "todo");
+        assert_eq!(utf8(&batch, "status").value(2), "done");
+        // Feishu spells "not completed" as "0" — the epoch-zero sentinel
+        // the YAML documents; a done task carries a real instant.
+        assert_eq!(millis(&batch, "completed_at").value(0), 0);
+        assert_eq!(millis(&batch, "completed_at").value(2), 1_780_322_529_638);
+        assert_eq!(millis(&batch, "created_at").value(0), 1_775_202_331_888);
+        // A present JSON null `due` is SQL NULL; members survive as
```
abbccdda: **Minor:** this comment pins the wrong claim — no row in `fixtures/feishu/tasks.json` carries a `due` key at all, so the assertion passes because *absence* yields NULL, not because "a present JSON null" does. Either add a row with `"due": null` (or better, a real `due` object — the live pass saw it non-NULL on 2 of 9 rows) or reword to "an absent `due`".

crates/skardi/src/sources/providers/open_connector/packs/feishu.yaml
```diff
@@ -0,0 +1,221 @@
...
+      - { name: thread_id, path: thread_id, type: utf8, nullable: true }
+      # sender carries id/id_type/sender_type/tenant_key — a small stable
+      # object, kept opaque; mentions is an array of user references.
+      - { name: sender, path: sender, type: json, nullable: true }
+      # body.content is Feishu's message payload: itself a JSON-encoded
+      # string whose inner schema varies BY msg_type — opaque by design,
+      # searchable as text.
+      - { name: content, path: body.content, type: utf8, nullable: true }
+      - { name: mentions, path: mentions, type: json, nullable: true }
+      - { name: upper_message_id, path: upper_message_id, type: utf8, nullable: true }
```
abbccdda, line 117: **Minor:** `upper_message_id` appears in zero fixture rows (same for `due` at line 166). Under this pack's own doctrine — loose items schemas, "real rows are the only column truth", no fingerprint protection — these two columns have no evidence trail at all. Worth capturing a row that carries them, or annotating them as doc-derived so the gap is explicit.

crates/skardi/src/sources/providers/open_connector/packs/feishu.rs
```diff
@@ -0,0 +1,959 @@
...
+            .as_object()
+            .expect("input object")
+            .keys()
+            .map(String::as_str)
+            .collect();
+        keys.sort_unstable();
+        assert_eq!(
+            keys,
+            vec!["pageSize", "type"],
+            "the status predicate is not pushed and no `completed` input is invented"
```
abbccdda, line 780: **Minor (coverage):** this test asserts the `pageSize` *key* exists but not its value (100). The wire-asserted pins are chats=100, messages=50, wiki_nodes=50; tasks (100), wiki_spaces (50), and chat_members (100) are never value-asserted. One-line additions to existing tests — cheap insurance given `page_size` is exactly where the third live contract defect surfaced.

crates/skardi/src/sources/providers/open_connector/pagination.rs
```diff
@@ -297,6 +356,16 @@ impl Pagination {
...
 
                 let Some(next) = next else {
+                    // With a declared has-more signal saying `true`, a missing
+                    // cursor is drift — stopping here would silently truncate
+                    // the scan the provider says is unfinished.
+                    if self.has_more_path.is_some() {
+                        return Err(OpenConnectorError::PaginationCursorInvalid {
+                            path: path.as_str().to_string(),
+                            page: self.page,
+                            found: "null or absent while the has-more signal is true".to_string(),
```
abbccdda: **Nit:** an empty-string cursor also maps to `None` above (the `Ok(Value::String(_))` arm) and reaches this error, so the text should read "null, empty, or absent while the has-more signal is true".

crates/skardi/src/sources/providers/open_connector/pagination.rs
```diff
@@ -267,6 +295,37 @@ impl Pagination {
                 Ok(more)
             }
             PaginationStrategy::Cursor { .. } => {
+                // The authoritative has-more signal, when the pack declares
+                // one, is consulted FIRST: some providers return a non-empty
+                // cursor on the final page (Feishu wiki spaces answer
+                // `has_more: false` with `page_token: "0||…"`), so the
+                // cursor's spellings alone would refetch a finished scan.
+                if let Some(path) = &self.has_more_path {
```
abbccdda, line 303: **Nit (docs):** a declared `has_more_path` makes the signal mandatory on *every* page — correct for Feishu (live-verified always-present), but providers that omit it when false (the Slack `response_metadata` pattern the cursor arm below tolerates) would hard-fail mid-scan rather than get a bind-time hint. Worth a sentence in `docs/open-connector.md` so future pack authors only declare it for providers that always emit the signal.

crates/skardi/src/sources/providers/open_connector/packs/feishu.yaml
```diff
@@ -0,0 +1,221 @@
+# Feishu source pack. Wire contract is Open Connector's HYBRID shape,
+# reconciled against a live gateway (v1.3.3): every list executor
+# rebuilds the pagination ENVELOPE (camelCase `$.items` / `$.pageToken` /
+# `$.hasMore`, with `page_token` absent-or-empty normalized to a null
+# `pageToken` at end-of-collection) while passing the Feishu API's item
+# objects through RAW — snake_case keys, timestamps as epoch DIGIT
+# STRINGS (millis for im, seconds for wiki), the reason the
```
abbccdda: **Nit:** "millis for im, seconds for wiki" omits tasks, which are also millis (13-digit fixture values, correctly declared `timestamp_ms_string_utc`). The per-column declarations are right; just this summary sentence (and the same phrasing in feishu.rs's module doc) is incomplete.

crates/skardi/src/sources/providers/open_connector/packs/fixtures/feishu/tasks.json
```diff
@@ -0,0 +1,230 @@
...
+          "zh_tw": ""
+        }
+      },
+      "parent_task_guid": "",
+      "repeat_rule": "",
+      "source": 1,
+      "status": "todo",
+      "subtask_count": 0,
+      "summary": "Redacted summary",
+      "task_id": "70004",
```
abbccdda, line 60: **Nit (redaction self-consistency):** `task_id: "70004"` here but the row's own `url` points at `guid-0005` while its `guid` is `guid-0002`; and `task_id: "70008"` appears on two *different* tasks (lines 130 and 217). Harmless, but for a fixture whose whole value is "internally consistent live capture", self-consistent redaction would be better.

abbccdda (commented):
Thorough review pass: correctness, security, maintainability, plus a cross-check against the milestone 5.4 spec entry. Inline findings below — one blocker (PII in a fixture), one doc/code mismatch, the rest minors/nits.

**Adversarially verified and held up:**
- `has_more_path` termination matrix (`false`+token → clean terminate with no spurious `PaginationLoop`; `true`+null/empty/absent token → typed error, never silent truncation; non-boolean/absent signal → `PaginationHasMoreInvalid`). No constructible input loops or truncates, and `has_more_path: None` takes the exact pre-existing code path — the backward-compat claim is real.
- Epoch flooring direction: `div_euclid` floors toward −∞, so a `>=` bound only ever widens the fetch and `Inexact` re-filtering trims — the inclusive-semantics claim is sound.
- Boolean-pushdown normalization respects SQL three-valued logic (`col` ≡ `col = true`, `NOT col` ≡ `col = false`, NULL cells excluded in all spellings).
- Digit-string timestamp parsing rejects empty/signed/whitespace/RFC-3339/overflow inputs distinctly; seconds→millis uses `checked_mul`; nothing parses as zero.
- No injection paths: resources, filter literals, and page tokens all flow through structured `serde_json` bodies; pagination is bounded by `max_pages` + deadline + response-size caps; new error variants carry only path/page/JSON-kind.
- Fixtures vs declarations: every declared column path (except the two flagged inline) exists in fixture rows with the declared type; timestamp scale choices match fixture magnitudes; the wiki fixture really does carry `hasMore: false` beside a non-empty token with a test pinning one-request termination; shared-fingerprint contract files are byte-identical as claimed.
- **Counted verification reproduced locally**: `cargo test -p skardi --lib` → 842 passed; `sources::providers::open_connector` → 285; `packs::feishu` → 17. Matches the spec tick exactly.

**Spec conformance:** the 5.4 entry accurately describes what shipped (three live-caught contract defects, four engine extensions, live e2e evidence). Two gaps: the "every fixture is a redacted live capture" claim is contradicted by the PII finding until fixed, and the two upstream gateway defects (scope-union OAuth URL, wiki non-empty final `page_token` passthrough) live only in docs/PR body — worth filing as issues on oomol-lab/open-connector so they don't get lost.

One discussion point, no action needed now: `completed_at`'s epoch-zero sentinel decodes to 1970-01-01 with no NULL affordance (documented at the column, but `MIN(completed_at)` gives 1970). If more Feishu-style sentinels appear, a `zero_as_null` column flag may be warranted.

**Suggested merge order:** scrub the fixture PII (history rewrite, not just tip edit), fix the stale doc row, merge; the rest can fast-follow.

crates/skardi/src/sources/providers/open_connector/packs/fixtures/feishu/messages.json
bakey: Fixed in c6ece9a — `to_chatters: ["pyh"]` → `["member-0001"]`. The branch history was rewritten (squashed to a single commit and force-pushed), so the name is gone from every reachable commit, not just the tip. A tripwire test (`fixtures_stay_redacted_one_json_level_deep`) now decodes the nested `body.content` payloads and asserts every membership entry is a `member-NNNN` placeholder — the exact blind spot this finding exposed.

crates/skardi/src/sources/providers/open_connector/packs/fixtures/feishu/messages.json
bakey: Fixed in c6ece9a — `from_user: ["陈桦Grace"]` → `["member-0002"]`, `to_chatters: ["朱天梁"]` → `["member-0003"]`. History squashed + force-pushed so the names are out of every reachable commit; the new tripwire test also asserts NO CJK text survives in any feishu fixture (the real names were Chinese), so this class of leak now fails the build.

bakey: Fixed in c6ece9a, all three parts: (1) the nested names are `member-NNNN` placeholders now; (2) the real capture timestamps in messages.json are coarsened (last six digits zeroed, ~±17 min) so the rows no longer encode identifiable membership events — the one test pinning an exact value was updated; (3) the branch history is REWRITTEN per your ask — squashed to the single commit c6ece9a and force-pushed, so f82d1de and every other commit carrying the names is no longer reachable from the branch. Durable fix beyond this instance: `fixtures_stay_redacted_one_json_level_deep` walks all seven fixtures asserting no CJK anywhere, then decodes each message's `body.content` and asserts `from_user`/`to_chatters` entries match `member-NNNN` — the audit now reaches the decode level the manual pass missed.

docs/open-connector-feishu.md
bakey: Fixed in c6ece9a — the tasks row now reads `—` and the design note says what the wire actually does: `type: my_tasks` is pinned, the `completed` input is ALWAYS omitted (state=all), completion on the wire is `status: todo|done` + `completed_at`, filter on `status` locally. Good catch: the row was a leftover from the pre-verification draft that the live pass invalidated.

crates/skardi/src/sources/providers/open_connector/filters.rs
bakey: Implemented in c6ece9a, in `validate_table` as you suggested — and one notch stronger than the ask: epoch formats (`epoch_seconds`, `epoch_seconds_string`) now require BOTH a lower-bound operator (gt/gt_eq) AND `fidelity: inexact` at load time. The second half matters for the same reason as the first: flooring widens the fetch, so an Exact claim would surface rows in `[floor(a), a)` that the predicate excludes and DataFusion never re-filters. Targeted errors for each; `epoch_formats_are_lower_bound_inexact_only` covers both rejections plus the legal Feishu shape. (The YAML op enum only admits eq/gt/gt_eq today, so eq was the one declarable bad operator — the check still guards future upper-bound additions.)

crates/skardi/src/sources/providers/open_connector/filters.rs
bakey: Fixed in c6ece9a — all four timestamp granularities clamp to `.max(0)` under both epoch formats, with the rationale at the clamp site and in the `ValueFormat` docs (still a lower-bound superset; strict digit-string schemas never see a `-` sign). `pre_epoch_instants_clamp_to_zero_never_a_signed_string` pins `create_time >= '1965-01-01'` → `startTime: "0"`, Inexact. Combined with the load-time lower-bound rule from the sibling thread, the clamp is provably only ever applied to lower bounds.

crates/skardi/src/sources/providers/open_connector/filters.rs
bakey: Fixed in c6ece9a with the invariant comment (your first option): the normalization arms now state explicitly that they fabricate a Boolean without checking the mapped column's type because only a boolean-typed column can appear as a bare predicate in a type-checked plan — a non-boolean Eq mapping is reachable solely through the BinaryExpr arm, where the literal comes from the query. I went with the comment rather than a validate_table cross-check on purpose: 'Eq mapping ⇒ boolean column' would be wrong as a load rule (string-equality mappings are legitimate future pushdowns), and the hazard you identified is specifically the bare-column arm, which the comment now documents at the exact site.

crates/skardi/src/sources/providers/open_connector/packs/feishu.rs
bakey: Fixed in c6ece9a — reworded to the truth: no fixture row carries a `due` key at all (the live pass saw it on 2 of 9 rows; none survived the redacted subset), so the assertion pins ABSENT → SQL NULL. I didn't inject a synthetic `"due": null` row: the fixture's whole value is being an unedited-shape live capture, and fabricating a key it never carried would quietly break that doctrine. The evidence gap itself is now explicit in the YAML (see the doc-derived annotations thread).

crates/skardi/src/sources/providers/open_connector/packs/feishu.yaml
bakey: Fixed in c6ece9a with the annotation option — both columns now carry a `DOC-DERIVED` comment in the YAML stating exactly what the evidence trail is: `upper_message_id` appears only on merge-forwarded messages (none captured), `due` was live-seen on 2 of 9 rows but none survived the redacted subset. Capturing carrier rows needs another live pass against the workspace; the annotations make the gap a reviewed fact until then, per this pack's own loose-schema doctrine.

crates/skardi/src/sources/providers/open_connector/packs/feishu.rs
bakey: Fixed in c6ece9a — all six tables now pin the wire value: tasks `assert_eq!(input["pageSize"], 100)`, wiki_spaces `50` (asserted on the terminate-on-hasMore test's one request), chat_members `100` (asserted in the UDTF-parity test's captured input). Agreed on the insurance rationale: page_size is exactly where the third live contract defect surfaced.

crates/skardi/src/sources/providers/open_connector/pagination.rs
bakey: Fixed in c6ece9a — the error text now reads "null, empty, or absent while the has-more signal is true", matching all three spellings that reach that arm.

crates/skardi/src/sources/providers/open_connector/pagination.rs
bakey: Fixed in c6ece9a — docs/open-connector.md's pagination section now documents the contract: a declared `has_more_path` makes the signal mandatory on every page (absence = contract drift, fail loud), so declare it only for providers that always emit it; the omit-when-false pattern (Slack's `response_metadata`) should stay undeclared and terminate via the cursor spellings.

crates/skardi/src/sources/providers/open_connector/packs/feishu.yaml
bakey: Fixed in c6ece9a — both summaries now read "millis for im and tasks, seconds for wiki" (feishu.yaml header and the feishu.rs module doc, the latter with a tasks example value). The per-column declarations were already right, as you noted.

crates/skardi/src/sources/providers/open_connector/packs/fixtures/feishu/tasks.json
bakey: Fixed in c6ece9a — redaction is self-consistent now: every task's `url` guid equals the row's own `guid`, and `task_id`s are unique (70004/70008/70012). Agreed on the principle: a fixture whose value is 'internally consistent live capture' should survive its own cross-references.

bakey:
All 14 findings addressed in **c6ece9a** — every inline thread has its fix note and is resolved. Per the blocker's requirement, the branch history was **rewritten** (squashed to that single commit and force-pushed), so the PII exists in no reachable commit, not just no tip.

Beyond the inline fixes, on the review body's two spec-conformance gaps:
- **Upstream defects are filed**, not just documented: [oomol-lab/open-connector#267](https://github.com/oomol-lab/open-connector/issues/267) (scope-union OAuth URL), [#268](https://github.com/oomol-lab/open-connector/issues/268) (get_as_user scopes), [#269](https://github.com/oomol-lab/open-connector/issues/269) + [PR #271](https://github.com/oomol-lab/open-connector/pull/271) (pageSize 50 cap), [#270](https://github.com/oomol-lab/open-connector/issues/270) (wiki final-page token). docs/open-connector-feishu.md now links all four so the trail isn't PR-body-only.
- **The redaction claim is enforceable now**: `fixtures_stay_redacted_one_json_level_deep` audits all seven fixtures (no CJK anywhere) and decodes nested `body.content` membership arrays down to `member-NNNN` placeholders — the decode-level blind spot can't silently regress.

Also took the validate_table suggestion one step further: epoch formats are load-time restricted to lower-bound operators AND `inexact` fidelity (an Exact claim over a floored-therefore-widened fetch would surface rows the predicate excludes), plus the negative-epoch clamp with its regression test.

On the `completed_at` epoch-zero discussion point: agreed a `zero_as_null` column flag is the right shape if a second sentinel appears — noted, deliberately not speculatively built now.

New counts: **845** lib tests green (was 842; +tripwire, +clamp, +load-rule), 290 open_connector, 24 feishu. fmt/clippy clean.

bakey:
@Userneima — thank you; this is the most useful kind of review a pack can get, and all three of your independent verifications (pagination depth, the pushdown actually firing on the wire, the tasks `type` behavior) now back claims the PR itself could only assert from one workspace. Disposition of every item, in your order:

**1. Egress guard vs zero-trust VPNs** — verified all three of your findings against current gateway `main` (unconditional `reservedIpv4Cidrs` check in `isBlockedIpv4`; no resolved address in the guard error; `providerFetch` constructed without `allowPrivateNetwork`, so the env switch never reaches provider egress). Filed as [oomol-lab/open-connector#275](https://github.com/oomol-lab/open-connector/issues/275) with credit and your two patch shapes — **please do open the PR you offered**; you have the environment that can actually verify it, and both parts look right to me (the allowlist skipping only the IP check while keeping scheme/port/redirect/size guards is the correct scope). The pack doc now documents the symptom (the ~36ms failure is the tell) and links the issue (21043bd).

**2. OAuth failures discarding context** — agreed, and your "my patch never fired" evidence is exactly why per-throw-site context matters. That's upstream code too; your offered PR (`error.name`/`cause.code` on the transport path + status/truncated-body on the non-2xx path) is welcome — happy to reference it from #275 as the diagnosability sibling.

**3. Gateway version floor** — fixed in 21043bd: the doc now states it as a hard prerequisite with your self-check command verbatim. The engine-side error enrichment (action count for the service when an action is missing) is a good idea beyond this PR — noted for the OC provider backlog.

**4. Scope narrowing** — fixed in 21043bd: the doc ships your 10-scope narrowed list and the pasteable console JSON, linked to [#267](https://github.com/oomol-lab/open-connector/issues/267). Your caveat is honest and now quoted: the narrow list is verified as far as "the authorize URL requests exactly 10"; a fresh-app end-to-end run would close that.

**5. Binding example** — fixed in 21043bd: the example now covers all six tables with required resources annotated inline and the query-first ordering note. You're right that people copy before they read.

**6. Four smaller ones** — the wording fix (`my_tasks` is the only value Feishu accepts, 1470400) is in 21043bd, doc + yaml comment. The three engine diagnostics items are real but beyond this PR's scope; filed as [#194](https://github.com/SkardiLabs/skardi/issues/194) (/query error chain), [#195](https://github.com/SkardiLabs/skardi/issues/195) (health-check HTML truncation), [#196](https://github.com/SkardiLabs/skardi/issues/196) (open_connector_query 4xx + doc sentence), each crediting this report.

On your PR offers: yes to all three — 1 and 2 belong upstream (issue #275 gives them a home), and if you'd rather we take the doc parts of 3/4/5, they're already in 21043bd so you can skip those.

BtXin (approved):

PR #191, opened 2026-08-05
feat(etl): runtime prerequisites for the skardi-etl generator (chunk_parts, json_pack, packs iterator)

bakey:
## Scope change: the generator moved to skardi-cloud

This PR originally carried the full skardi-etl generator M1. Per the follow-up product decision, the generator itself (config/recipes/plans/dialects/bundle/validation, the `skardi-etl` binary, golden bundles, and both ingest e2e suites) now lives in **[skardi-cloud#232](https://github.com/SkardiLabs/skardi-cloud/pull/232)** as `code/rust/crates/skardi-etl`, consuming this repo as a pinned git dependency.

What REMAINS here is what the engine must run — the milestone-1a runtime prerequisites the generated SQL depends on:

- **`chunk_parts` UDF** (`List<Struct<chunk_idx Int32, chunk_text Utf8>>`) with the plannability pin: DataFusion cannot plan `UNNEST … WITH ORDINALITY` (apache/datafusion#11419), so the working spelling — projection-position `UNNEST(chunk_parts(...)) AS part` in a subquery with field access outside — is pinned by `sql_chunk_parts_unnest_yields_ordered_indexed_rows`, including per-source ordinal restart and overlap-0 reassembly end to end.
- **`json_pack` UDF** — the only SQL-callable JSON encoder: insertion-ordered keys (byte-deterministic output), injection-boundary tests round-tripping quotes/backslashes/control chars/non-ASCII byte-exact through a real JSON parser, JSON scalar set + Timestamp (epoch ms) + `List<Utf8>` support, non-finite floats refused.
- **`SourcePackRegistry::packs()`** — deterministic name-sorted iteration, with the enumeration pin test.
- Both UDFs registered on both server session contexts.

All prior review threads (the five P1s + the rustdoc P2) were addressed in this branch's history and the fixes travelled to skardi-cloud with the migration; the threads remain resolved with their replies.

## Verification

839 lib tests green with `--features chunking` (822 default); fmt + clippy clean; docs build clean under CI's `-D warnings` flags.


crates/skardi/src/etl/dialects/sqlite.rs
```diff
@@ -0,0 +1,780 @@
...
+         \x20 arrow_cast(u.{id}, 'Utf8')                              AS source_id,\n\
+         \x20 {title}                                                 AS title,\n\
+         \x20 u.part['chunk_text']                                    AS content,\n\
+         \x20 -- BIGINT: the INTEGER destination column reads back as Int64 and\n\
+         \x20 -- the executor's type preflight is exact (chunk_idx is Int32).\n\
+         \x20 CAST(u.part['chunk_idx'] AS BIGINT)                     AS chunk_index,\n\
+         \x20 {author}                                                AS author,\n\
+         \x20 {created_at}                                            AS created_at,\n\
+         \x20 {metadata}                                              AS metadata,\n\
+         \x20 vec_to_binary(candle('{embedding_model}', u.part['chunk_text'])) AS embedding\n\
```
bakey: **[P1] Honor `embedding.udf` instead of always rendering `candle()`.**

`config.rs` accepts `udf: remote_embed` and requires its provider, but both this ingest expression and the search query are hard-coded to `candle(model, text)`. A valid remote-embed config therefore silently generates a different embedding implementation and omits the provider argument/credential guidance. Please render the expression from `EmbeddingSpec` in both paths (`candle(model, text)` vs `remote_embed(provider, model, text)`) and add a remote-embed generation/plan-check test. If M1 is intended to be Candle-only, reject `remote_embed` during config validation instead of accepting and rewriting it.

crates/skardi/src/etl/bundle.rs
```diff
@@ -0,0 +1,595 @@
...
+fn render_readme(config: &EtlConfig, plan: &HybridPlan, slug: &str) -> String {
+    let catalog = &config.destination.catalog;
+    let dest_path = config.destination.path.as_deref().unwrap_or_default();
+    let model = &plan.search.embedding_model;
+
+    let mut job_runs = String::new();
+    let mut refresh_notes = String::new();
+    for ingest in &plan.ingests {
+        let job = format!("{slug}-ingest-{}", ingest.source_table);
+        job_runs.push_str(&format!("skardi job run {job} -p limit=500\n"));
```
bakey: **[P1] Include `{since}` in the initial command for incremental jobs.**

The GitHub issues job contains `WHERE updated_at >= {since}`, so `since` is a required request field in the current JobExecutor. This generated command only supplies `limit`, and following the flagship README fails synchronously with `missing_parameters: since`. Emit an explicit initial watermark (for example an RFC3339 epoch/backfill start selected by the product contract), then document how subsequent runs advance or overlap it. Please add a test that compares every generated README job command with that job's inferred required parameters.

crates/skardi/src/etl/dialects/sqlite.rs
```diff
@@ -0,0 +1,780 @@
...
+
+    fn ctx_fragment(&self, config: &EtlConfig) -> String {
+        let catalog = &config.destination.catalog;
+        let path = config.destination.path.as_deref().unwrap_or_default();
+        let extensions = config
+            .destination
+            .sqlite
+            .as_ref()
+            .and_then(|s| s.extensions_env.as_deref());
+        let options = match extensions {
```
bakey: **[P1] Do not generate a vec0 destination without an extension option in the ctx fragment.**

`spec.destination.sqlite.extensions_env` is optional, so this branch can emit no `options.extensions_env`. Setting `SQLITE_VEC_PATH` as the README suggests is then insufficient: the SQLite provider only loads env-named extensions when that option is present. The E2E does not catch this because it manually inserts `extensions_env = SQLITE_VEC_PATH` before registration instead of consuming the generated fragment. For SQLite hybrid search, either require `extensions_env` or default the fragment to `SQLITE_VEC_PATH`, and make the E2E register the destination from the generated ctx artifact.

crates/skardi/src/etl/dialects/sqlite.rs
```diff
@@ -0,0 +1,780 @@
...
+        let vec = Self::qualified(config, "documents_vec");
+        let model = &plan.search.embedding_model;
+        let knn_n = plan.search.knn_candidates;
+        let fts_n = plan.search.fts_candidates;
+        format!(
+            "WITH vec AS (\n\
+             \x20 SELECT doc_rowid AS rid, ROW_NUMBER() OVER (ORDER BY _score ASC) AS rk\n\
+             \x20 FROM sqlite_knn('{vec}', 'embedding',\n\
+             \x20     (SELECT candle('{model}', {{query}})),\n\
+             \x20     {knn_n})\n\
```
bakey: **[P1] Deduplication after the fixed candidate limits is not replay-safe.**

KNN returns at most 80 physical rows and FTS at most 60 before `hits` partitions by `doc_id`. After enough `{since}` replays, copies of one highly ranked chunk can fill an entire candidate pool; post-limit dedup then keeps one result, but other documents have already been discarded. This makes the README's "duplicates accumulate harmlessly" claim false and causes recall to degrade as history grows. Please move deduplication ahead of candidate selection through a storage/write strategy that indexes one active copy, or restrict M1 to rebuild semantics until upsert/uniqueness is available. A regression test should replay one document more than the candidate depth and verify that unrelated relevant documents remain discoverable.

crates/skardi/src/etl/bundle.rs
```diff
@@ -0,0 +1,595 @@
...
+         \n\
+         ## Contents\n\
+         \n\
+         - `setup.sql` — destination DDL: the `documents` table plus the engine's\n\
+         \x20 search artifacts and sync triggers. Idempotent (`IF NOT EXISTS`).\n\
+         - `jobs/` — one `kind: job` per source table (chunk + embed inline in SQL).\n\
+         - `pipelines/` — `{slug}-search-hybrid` (RRF) and `{slug}-get-document`.\n\
+         - `ctx.fragment.yaml` — the data-source entry to merge into your `ctx.yaml`.\n\
+         \n\
+         ## Run it (five steps)\n\
```
bakey: **[P1][Security] Add the required first-contact and ACL-flattening warning before ingestion.**

The PRD explicitly states that source ACLs are not preserved: every private repo/channel/page visible to the binding becomes searchable by anyone with destination access, and it requires the generated README to disclose this before the first ingest command. The current README has no ACL warning or the design's promised first-contact checklist. Please add a visible preflight section covering binding/pack availability, sqlite-vec loading, embedding-dimension truth, destination access, and especially the source-ACL-to-destination-access flattening.

crates/skardi/src/etl/config.rs
```diff
@@ -0,0 +1,737 @@
+//! The `kind: etl` generation config — parsing and cross-field validation.
+//!
+//! Strict everywhere (`deny_unknown_fields`): a misspelled key fails
+//! parsing instead of silently disabling what it was meant to set — the
+//! Open Connector config/pack-loader discipline, applied to this crate's
+//! own surface (it is deliberately NOT a repo-wide convention; see the
+//! design's Research Findings).
+//!
+//! Parsing yields a [`RawEtlConfig`] shaped exactly like the YAML;
```
bakey: **[P2] Fix the public rustdoc link to the private `RawEtlConfig`.**

The Docs CI job runs rustdoc with `-D warnings` and currently fails here with `rustdoc::private-intra-doc-links`. Use code formatting without an intra-doc link (`` `RawEtlConfig` ``), avoid naming the private implementation type in public module docs, or make the documented surface public only if that is actually intended.

bakey (commented):
I reviewed this M1 implementation against the latest PRD, technical design, the current JobExecutor/SQLite provider contracts, the generated golden bundles, and the CI results. The overall architecture is strong: the format/dialect split is clean, `chunk_parts` and `json_pack` close real DataFusion gaps, the positional schema-order gate is valuable, and the real JobExecutor hybrid E2E is a meaningful addition.

I found five runtime/product-contract issues that should be addressed before merge, plus the current rustdoc failure:

1. `remote_embed` is accepted by config but silently rendered as `candle()` in both ingest and search SQL.
2. The flagship README's first ingest command omits the required `{since}` parameter and fails at job submission.
3. SQLite hybrid generation permits no `extensions_env`, producing a ctx fragment that cannot load vec0; the E2E masks this by manually injecting the option.
4. Deduplicating only after the fixed KNN/FTS candidate limits lets replayed copies exhaust the candidate pools and suppress other documents.
5. The generated README omits the PRD-required ACL-flattening warning and first-contact checklist.
6. Docs CI currently fails on a public rustdoc link to the private `RawEtlConfig` type.

The inline comments include concrete fixes and the missing tests. I would also add coverage for remote embedding rendering, required README job parameters, registration through the generated ctx fragment, and replay counts beyond the 80/60 candidate depths.

crates/skardi/src/etl/dialects/sqlite.rs
bakey: Fixed in 730e6f0 — valid finding, the config's promise and the rendered SQL had diverged. `EmbeddingSpec::call_expr` is now the single renderer for both sides: the ingest projection emits `vec_to_binary(remote_embed('<provider>', '<model>', …))` and the search pipeline's query vector uses the same call — they share one function so they cannot diverge again. The plan-check registers a `remote_embed` stub (same volatile List<Float32> shape as the candle stub), `remote_embed_configs_pass_all_four_gates` runs the full generation path, `remote_embed_configs_render_remote_embed_in_both_ingest_and_search` pins both renderings (and asserts no `candle(` leaks), and the README's first-contact checklist now surfaces the provider's `${PROVIDER}_API_KEY` env for the server.

crates/skardi/src/etl/bundle.rs
bakey: Fixed in 730e6f0 — confirmed: the executor's submit-time parameter validation rejects any missing request field, so the old README command failed synchronously exactly as you describe. Incremental job commands now carry an explicit epoch first-backfill watermark (`-p since="1970-01-01T00:00:00Z"`) with an inline comment, and the Refresh section documents advancing it: later runs pass the last watermark, chosen to OVERLAP (replayed rows now REPLACE their previous copies at write time — see the candidate-pool thread — so overlap is safe and a gap loses rows). The requested test exists: `readme_job_commands_carry_every_inferred_required_parameter` loads every generated job through the real loader and asserts each README command supplies exactly that job's inferred request-schema fields — a drift in either direction fails.

crates/skardi/src/etl/dialects/sqlite.rs
bakey: Fixed in 730e6f0 — you're right on both counts: the provider only loads env-named extensions when the ctx option is present, and the e2e was masking that by hand-wiring the option. The fragment now ALWAYS emits `options.extensions_env` (the config's override, else the `SQLITE_VEC_PATH` default — same convention as the demo), `validate_ddl` resolves the same env so generate-time validation and the shipped fragment agree, and the e2e now parses `ctx.fragment.yaml` and registers the destination from the artifact's own name/path/options (asserting the default is present) — a fragment that couldn't load vec0 now fails the e2e directly. I chose defaulting over requiring the field to keep the zero-config G1 path; the config override still wins (`ctx_fragment_defaults_the_extension_env_when_none_is_configured` pins both).

crates/skardi/src/etl/dialects/sqlite.rs
bakey: Fixed in 730e6f0 — this was the sharpest finding: post-limit dedup genuinely degrades recall as replay history grows, and the README's 'accumulate harmlessly' claim was false. Fixed at the storage layer, per your first option: a `documents_bi_replace` BEFORE INSERT trigger deletes the incoming doc_id's previous copy from the table AND both mirrors before every insert (mirror cleanup is inlined, not chained through DELETE triggers, so it never depends on recursive-trigger settings; fts rows are inserted with rowid = rid so replacement can delete them by integer key; a doc_id index backs the lookup). The executor keeps its plain append path — no ON CONFLICT needed — and the index holds exactly one live copy per doc_id, so the 80/60 candidate pools structurally cannot fill with replayed history. Read-time dedup stays as defense in depth. The requested regression coverage: `validate_ddl` now executes a replay smoke on every generate when vec0 is loadable (double insert → exactly 1 row in table + each mirror), and the e2e runs the generated job three times plus 100 direct re-inserts of one doc_id — beyond both candidate depths — asserting one live copy, mirrors in lockstep, and other documents still discoverable. One honest residue documented in the README: a document that SHRINKS to fewer chunks leaves stale tail chunks until a rebuild (doc_id embeds chunk_index); full row-set replacement per (source_table, source_id) arrives with v2 upsert.

crates/skardi/src/etl/bundle.rs
bakey: Fixed in 730e6f0 — the README now opens with a '⚠ Source permissions are NOT preserved' callout ahead of any command: everything the binding can see (private repos/channels/pages included) lands in the destination and becomes searchable by anyone with destination or pipeline access. It's followed by the design's first-contact checklist: source binding existence/authorization/scope, sqlite-vec loadability via the fragment's extensions_env, embedding dimensions being DECLARED-not-verified (mismatch surfaces on first ingest → rebuild), remote-embed API-key env when applicable, and 'destination access = search access'. `readme_discloses_acl_flattening_before_the_first_ingest_command` pins both the content and the ordering (warning strictly before the first ingest command); both golden READMEs carry it.

crates/skardi/src/etl/config.rs
bakey: Fixed in bc5b58d — the intra-doc link is now plain code font (`RawEtlConfig` stays private; the public docs no longer link it). Verified locally with CI's exact flags (`RUSTDOCFLAGS="--cfg docsrs -D warnings" cargo doc --no-deps -p skardi`) for both the default and `chunking` feature sets — zero warnings.

bakey:
All six findings addressed (fixes in 730e6f0; the rustdoc one was already fixed in bc5b58d) — each inline thread has the details and is resolved.

The four coverage additions you asked for all landed: remote-embed rendering (`remote_embed_configs_pass_all_four_gates` + the dialect-level rendering test), required README job parameters (`readme_job_commands_carry_every_inferred_required_parameter` compares each command against the job's loader-inferred request schema), registration through the generated ctx fragment (the e2e now parses `ctx.fragment.yaml` and registers the destination from it), and beyond-candidate-depth replay (100 re-inserts of one doc_id > both the 80/60 depths, one live copy asserted, plus a replay smoke inside `validate_ddl` itself).

The candidate-pool finding drove the one real semantic change: replay is now replace-on-insert at the storage layer (BEFORE INSERT trigger + doc_id index + rowid-pinned fts rows), so the 'accumulate + read-time dedup' model is gone — read-time dedup remains only as defense in depth. Goldens regenerated; 885 lib (chunking) / 857 (default) / 3 bin tests + the e2e green, including the vec0-loaded validate_ddl branch.

bakey:
Scope note: the generator implementation moved to [skardi-cloud#232](https://github.com/SkardiLabs/skardi-cloud/pull/232) (commit 88f21cc trims this branch to the engine-side runtime prerequisites; the skardi-cloud crate pins that rev as its git dependency). Review fixes from this PR's round 1 travelled with the migration.

PR #192, opened 2026-08-05
feat(sources): Gmail source pack (milestone 5.5)

gracexmatin:
## Summary

The first Google Workspace pack: **`gmail`** — `threads`, `messages`, `drafts`, `labels`, `filters` as stable SQL tables over the Open Connector `gmail.*` read actions (Google OAuth, page-token cursor pagination). Everything is reconciled against a live gateway (**v1.3.4**, commit `2410fbe`) and the OC executor source (`src/providers/gmail/`) — never the Gmail API docs.

**Live-verified end to end against a real mailbox (2026-08-05)** — see Verification below.

### `packs/gmail.yaml` (new)
The declarative pack. Wire facts it encodes:
- The gmail executors **rebuild rows Slack-style**: camelCase identity (`threadId`/`messageId`), `From`/`To`/`Subject` headers flattened into `sender`/`to`/`subject`, epoch-millis `internalDate` re-emitted as an RFC 3339 `messageTimestamp`. Mapping Gmail's own field names would have produced all-NULL columns. `list_labels`/`list_filters` are the two raw-passthrough exceptions.
- **Cursor pagination** on `threads`/`messages`/`drafts` (`pageToken` in, top-level `$.nextPageToken` out). The executors never post-filter rows and re-emit the token as an explicit `null` at end-of-collection; absent and `""` also terminate. Loop/invalid-cursor fail loudly.
- **`messages` pins `detail: summary`, page size 100** — the executor hydrates every listed message with a metadata `messages.get` (batches of 10), so a page of N rows is N+1 Gmail calls; 500 would be a 501-request burst. `ids` is too thin for a table, `full` is unbounded bodies (deferred to a future content-oriented surface).
- **No filter pushdown anywhere**: Gmail's `q` is a free-text search language no SQL predicate maps to faithfully; `labelIds` is an AND-semantics array a scalar `column op literal` cannot represent. `query`/`labelIds`/`includeSpamTrash` are **optional resources** instead (Slack `channelId` precedent), passed through verbatim with their YAML types.
- The default listing is Gmail's own (excludes SPAM/TRASH) — documented, not pinned away, because `list_threads` offers no spam/trash input: pinning `includeSpamTrash: true` on `messages` alone would make the two tables describe different mailboxes.
- `to` → `to_addresses` (TO is a reserved SQL keyword); header-derived absents stay `""`, never NULL; `error_path: None` everywhere (executors consume provider errors; the gateway failure envelope carries the code).
- Excluded with rationale: `search_threads` (strict subset of `list_threads`), `list_history` (checkpoint/incremental-sync API, not a collection), `list_forwarding_addresses` (no output schema to fingerprint; `settings.sharing` scope), `get_profile` (scalar endpoint).

### Engine extension: loader `single_page` strategy (`packs/loader.rs`)
`list_labels`/`list_filters` declare **no pagination inputs at all** — their strict schemas would 400 an injected page key, and each returns the complete collection in one response. The engine's `PaginationStrategy::SinglePage` predates this PR but was unreachable from YAML. Added as a braced variant (`SinglePage {}`) so `deny_unknown_fields` keeps rejecting stray keys — a misspelled pagination field fails loudly. Backward-compatible; pass and refusal sides both tested.

### `packs/gmail.rs` (new)
Module doc carrying every design decision above with its rationale, the `OnceLock` accessor, and the test suite (below). Registry entry in `source_pack.rs`, `mod` line, loader parse-all list.

### Fixtures (`packs/fixtures/gmail/`)
Five contract captures (`contracts/*.json`, saved verbatim from live discovery on gateway v1.3.4) and per-table page fixtures covering the admission-gate categories: null-bearing (`historyId: null`), omitted-key absents (system labels without visibility/color), nested paths (`message.messageId`), extra-field passthrough, empty lists, schema-mismatch (valid first row + number-where-Utf8), plus an inline null-parent converter pin. **Page fixtures are provider-shaped synthetic pages until phase 4 re-derives them as redacted live captures** (Notion precedent); the module doc and spec say so explicitly.

### Docs
`docs/open-connector-gmail.md` (binding YAML, per-table reference, behavior bullets, scopes — `filters` needs `gmail.settings.basic`, `drafts` is declared under `gmail.compose` — and rate-limit notes on the hydrated `messages` scan), `docs/open-connector.md` status paragraph, tasks-spec 5.5 entry ticked with the verification blurb (5.6 = later waves).

## Verification

- **Contract-level live reconciliation (done)**: every action ID, input key, and bound probed on a live gateway v1.3.4 — wrong keys 400 as `invalid_input` before credentials; every table's exact generated input set (pagination params, `detail` pin, resource keys) reaches the credential wall (`authorization_failed`, 403); `maxResults` bounds `[1,500]` and the `detail` enum are enforced by the real gateway. Fingerprints pinned from this capture; a Python replication of `canonical_json`+BLAKE3 was sanity-checked against Notion's known pin before computing the five gmail pins (CI's sync test re-verifies through the real `fingerprint_schema`).
- **Phase 4 real-mailbox pass (done, 2026-08-05)**: skardi-server registered the binding against LIVE discovery (all five fingerprint pins matched); full scans of all five tables through `POST /query`; **every mapped column of every table extracted a real non-NULL value** (messages: 7/7 columns on 29/29 rows; labels' `color` via a seeded user label; drafts/filters via seeded rows); real multi-page cursor chaining observed at the wire (`maxResults: 1` chained three pages, token at `$.nextPageToken`) and real final-page null-token termination; `query`/`labelIds`/`includeSpamTrash` resources observed narrowing real listings (a `labelIds: [INBOX]` binding excluded the DRAFT message the unscoped binding sees); UDTF parity spot-checked live. The live pass settled two synthetic guesses (SENT/INBOX/DRAFT/STARRED/UNREAD omit visibility fields, CHAT carries them; a fresh draft's `threadId` equals its `messageId`) and **caught an upstream bug**: a zero-filter mailbox fails `list_filters` with `internal_error` — Gmail answers `settings/filters` with an empty body that the executor's unconditional `response.json()` does not tolerate (verified: creating one filter makes the call succeed). Documented in the pack doc; fix belongs in open-connector.
- **Fixtures are redacted live captures** from that pass (real envelope keys, field presence/absence, timestamp spellings; synthetic identifiers, placeholder text), mechanically audited against an allowlist; executor-guaranteed absent spellings the capture lacked are pinned by an inline converter test; the schema-mismatch fixture stays synthetic and says so.
- **Tests**: 284 open_connector tests (CI-confirmed; counted by `cargo test -p skardi --lib sources::providers::open_connector`; **23 new**) — loader single_page pass/refusal; registry pin; five fixture contract suites plus schema-mismatch (full error identity: column/page/row/found-kind) and null-parent pins; fingerprint sync + coverage-gap pins (all `messages` columns ride the `anyOf` passthrough — pinned); drift-refusal e2e; and per-declaration e2e: three cursor tables each pinning its own exact wire key sets per page and a distinct termination spelling (null / absent / `""`), single-page one-request pins for `labels`/`filters`, optional-resource forwarding with YAML type fidelity and per-table withholding, no-pushdown guard with row identity, LIMIT early-stop, pagination-loop refusal, gateway failure-envelope surfacing (`authorization_failed` + action named, never a row-path error), UDTF parity.
- Local runs: `cargo fmt` only, per repo convention — CI is the verdict on tests/clippy.

## Deliberate deferrals

- Message bodies / attachments (`detail: full`) — future content-oriented surface.
- `googlecalendar` / `googledrive` / `googletasks` packs — later waves (spec 5.6).
- Upstream fix for the zero-filter `list_filters` crash — filed against open-connector; the pack doc carries the caveat until it lands.


abbccdda:
rebase the PR please

crates/skardi/src/sources/providers/open_connector/packs/gmail.yaml
```diff
@@ -0,0 +1,119 @@
...
+    # detail=summary pins the bounded row shape: `ids` carries no
+    # metadata worth a table, `full` hydrates entire decoded bodies and
+    # attachment trees (unbounded row size). Bodies belong to a future
+    # content-oriented surface, not a listing table.
+    fixed_inputs:
+      detail: summary
+    resources:
+      optional: [query, labelIds, includeSpamTrash]
+    columns:
+      - { name: message_id, path: messageId, type: utf8, nullable: false }
```
abbccdda, line 63: **Non-blocking (by design, worth being explicit).** `message_id`/`thread_id` are `nullable: false`, but `fetch_emails` declares its row items as an `anyOf` whose third branch (`{type: object, additionalProperties: true}`) requires nothing, and the fingerprint coverage walker does not descend into the `anyOf` — so these columns sit outside the registration-time gate (pinned honestly by `fingerprint_coverage_gap_is_pinned`).

In practice the non-null guarantee holds only because the pack pins `detail: summary`, and the summary branch *does* require both fields. Any drift fails loud at scan with a non-null conversion error (never silent corruption), which matches the pack's loud-on-mismatch philosophy. Flagging that the guarantee rests on the `detail: summary` pin rather than the schema gate — fine as-is, just an accepted tradeoff to keep in mind if `detail` ever becomes configurable.

crates/skardi/src/sources/providers/open_connector/packs/gmail.yaml
```diff
@@ -0,0 +1,119 @@
...
+    columns:
+      - { name: id, path: id, type: utf8, nullable: false }
+      - { name: name, path: name, type: utf8, nullable: true }
+      - { name: type, path: type, type: utf8, nullable: true }
+      - { name: message_list_visibility, path: messageListVisibility, type: utf8, nullable: true }
+      - { name: label_list_visibility, path: labelListVisibility, type: utf8, nullable: true }
+      - { name: color, path: color, type: json, nullable: true }
+
+  filters:
+    action: gmail.list_filters
```
abbccdda, line 107: **Non-blocking (top e2e reliability caveat).** A zero-filter mailbox makes `SELECT … FROM filters` hard-fail with the gateway's `internal_error` (Gmail returns an empty body the upstream executor's `response.json()` won't tolerate). This is an upstream open-connector bug, not a Skardi defect, and it's honestly documented in the module doc + `open-connector-gmail.md`.

Practical impact: `filters` is effectively a second-class table until the upstream fix — a broad discovery query or any binding including `filters` on a fresh mailbox breaks, and the mitigation ("bind only for mailboxes with ≥1 filter") is runtime-state-dependent. Nothing to change here; worth tracking the upstream fix so this edge closes.

crates/skardi/src/sources/providers/open_connector/packs/gmail.yaml
```diff
@@ -0,0 +1,119 @@
...
+    columns:
+      - { name: message_id, path: messageId, type: utf8, nullable: false }
+      - { name: thread_id, path: threadId, type: utf8, nullable: false }
+      - { name: label_ids, path: labelIds, type: utf8_list, nullable: true }
+      - { name: subject, path: subject, type: utf8, nullable: true }
+      - { name: sender, path: sender, type: utf8, nullable: true }
+      # Wire key `to`; renamed because TO is a reserved SQL keyword and a
+      # column requiring quotes on every reference is hostile to use.
+      - { name: to_addresses, path: to, type: utf8, nullable: true }
+      - { name: message_timestamp, path: messageTimestamp, type: timestamp_ms_utc, nullable: true }
```
abbccdda, line 71: **Nit (cosmetic).** `timestamp_ms_utc` reads as epoch-milliseconds, but this column's wire value is an RFC 3339 *string* (the executor's re-emission of `internalDate`). It's the shared engine type and it converts correctly (tested), so no change needed — just a slightly misleading name for a future reader.

abbccdda (commented):
Review: correctness/security/maintainability + design-spec admission gate.

Strong, ship-quality pack. No correctness or security defects found in the changed code; the `single_page` engine extension, cursor termination, column mappings, fingerprint gate, and registration are all sound, and the admission gate is satisfied end to end. Local `gmail::tests` run green (21 passed). Two non-blocking reliability observations and one nit inline — all already documented; noting them for visibility.

crates/skardi/src/sources/providers/open_connector/packs/gmail.yaml
gracexmatin: Accurate read of the gate mechanics — one refinement to the failure surface, which is narrower than "summary or bust":

All three `anyOf` branches were checked against the pinned capture (`fixtures/gmail/contracts/fetch_emails.json`): both the summary **and** full branches `require` all seven mapped fields, including `messageId`/`threadId`. Only the ids branch requires nothing. So the non-null guarantee rests on `detail` being pinned to any non-ids shape, not on `summary` specifically — if the pin ever moved to `full`, the schema would still carry the guarantee. The only value that would strand the columns is `detail: ids`, and that shape is already excluded by design in the module doc ("carries no metadata worth a table").

Descending the `anyOf` in the coverage walker wouldn't close the gap either: registration can't know which branch the runtime `detail` selects, so the walker could only take the intersection of the branches' `required` sets — and the ids branch makes that intersection empty. Same result, which is why the pin (`fingerprint_coverage_gap_is_pinned`) is the honest treatment rather than a workaround.

Agreed on fine-as-is. If `detail` ever becomes configurable, the yaml's nullability (and this pin) are the two places that would need revisiting — noting that here so the thread is findable.

crates/skardi/src/sources/providers/open_connector/packs/gmail.yaml
gracexmatin: Confirmed diagnosis and framing. The crash mechanics: Gmail answers `settings/filters` with an empty body when the mailbox has zero filters, and the upstream executor's unconditional `response.json()` throws, which the gateway surfaces as `internal_error`. Skardi deliberately doesn't paper over it: that failure envelope is indistinguishable from a genuine internal error, so synthesizing an empty page on `internal_error` would mask real failures — the loud-on-mismatch rule applies.

Tracking: an upstream issue is drafted against oomol-lab/open-connector (empty-body tolerance in the filters executor) and will be linked here once filed. When the fix ships, the unwind points are exactly the two caveat notes (module doc + `docs/open-connector-gmail.md`) — nothing in the pack's contract changes.

crates/skardi/src/sources/providers/open_connector/packs/gmail.yaml
gracexmatin: The name describes the Arrow storage, not the wire: `timestamp_ms_utc` is the engine's Timestamp(Millisecond, UTC) column type, and its converter accepts an RFC 3339 string **or** an epoch-millis number (json_to_arrow: "RFC 3339 string or epoch-millis number ↔ Arrow Timestamp(Millisecond, UTC)") — github's `*_at` columns ride the same spelling over RFC 3339 wires. Renaming the shared spelling to something wire-shaped would misdescribe that dual acceptance, so the name stays; added a clarifying comment at the column (788e8fb) so the next reader doesn't have to dig into the converter to learn this.

abbccdda (commented):
## Review: spec-conformance + end-to-end reliability

Read against the design spec's source-pack admission gate (`docs/superpowers/specs/2026-07-11-open-connector-integration-design.md:727-740`) and the tasks-spec 5.4 entry. No local tests or builds were run — CI is the verdict on those, per the review request.

**Overall this is strong work.** The wire contract is reconciled against the gateway rather than the Gmail API docs, the live phase-4 pass is real verification (it settled two synthetic guesses and caught an upstream bug), the exclusion rationales are individually defensible, and the one engine extension is minimal and correctly shaped — a braced `SinglePage {}` so `deny_unknown_fields` keeps rejecting stray keys is the right call, since a unit variant would silently ignore them.

Findings below, most substantive first. I've skipped two areas already covered in earlier threads (the `anyOf` coverage gap, and the zero-filter `list_filters` crash) except where I have something additive — the `anyOf` thread in particular turns out to *strengthen* the nullability finding, so I've built on it there rather than restating it.

One thing worth flagging up front: **7 findings, 1 of which I'd want addressed before merge** (the `messages` nullability gap). The rest are docs/tests/caveats that could equally land as follow-ups.

crates/skardi/src/sources/providers/open_connector/packs/gmail.yaml
```diff
@@ -0,0 +1,123 @@
...
+      - { name: subject, path: subject, type: utf8, nullable: true }
+      - { name: sender, path: sender, type: utf8, nullable: true }
+      # Wire key `to`; renamed because TO is a reserved SQL keyword and a
+      # column requiring quotes on every reference is hostile to use.
+      - { name: to_addresses, path: to, type: utf8, nullable: true }
+      # The wire value is an RFC 3339 string (the executor re-emits
+      # Gmail's epoch-millis `internalDate`); `timestamp_ms_utc` names the
+      # Arrow storage — Timestamp(ms, UTC) — and its converter accepts
+      # RFC 3339 or epoch-millis wires alike, same as github's *_at columns.
+      - { name: message_timestamp, path: messageTimestamp, type: timestamp_ms_utc, nullable: true }
```
abbccdda, line 90: **Correctness (worth changing).** Building on the `anyOf` thread above rather than repeating it — I re-checked all three branches of the pinned capture and can confirm the refinement: branch 0 (`summary`) and branch 1 (`full`) both `require` all seven mapped fields; only branch 2 (ids) requires nothing.

That makes the captured contract's guarantee *stronger* than this YAML claims. `label_ids`, `subject`, `sender`, `to_addresses` and `message_timestamp` are declared `nullable: true`, yet under the pinned `detail: summary` the executor always emits them (`""` / `[]` fallbacks, per the module doc). The combination is what bites: `json_to_arrow.rs:256` turns a missing key into a **silent NULL** for a nullable column, and `messages` is the one table in this pack with no fingerprint coverage to catch drift at registration. So an upstream rename of, say, `sender` produces exactly the all-NULL column the module doc says mapping Gmail's own field names would have caused — silently, at scan time, no error anywhere.

`nullable: false` on those five converts that into a loud `missing key` conversion failure (`json_to_arrow.rs:259-261`). It is what the captured contract asserts, it is what the pack's loud-on-mismatch rule wants, and it costs nothing today — the phase-4 pass extracted non-NULL values for 7/7 columns on 29/29 rows. `thread_id`/`message_id` already do exactly this; the other five are looser than the contract they were pinned from.

Worth noting this doesn't introduce false failures: a genuinely empty header is a *present* key holding `""`, which still converts. And `message_timestamp`'s `""` fallback already fails conversion loudly by design (gmail.rs:68-72), so non-null changes nothing there.

The convention elsewhere (identity non-null, rest nullable — cf. `slack.yaml:23-31`) is safe precisely *because* the fingerprint gate covers those tables. Here it doesn't, so the convention leaves the gap open.

crates/skardi/src/sources/providers/open_connector/packs/gmail.yaml
```diff
@@ -0,0 +1,123 @@
...
+    # `verbose` is deliberately never sent: omitted means the summary
+    # shape (the executor checks `=== true`), and the hydrated variant
+    # costs one threads.get per row. `query` scopes the listing with
+    # Gmail's own search syntax — a free-text language no SQL predicate
+    # maps to faithfully, hence a resource, not a filter.
+    resources:
+      optional: [query]
+    columns:
+      - { name: thread_id, path: threadId, type: utf8, nullable: false }
+      - { name: snippet, path: snippet, type: utf8, nullable: true }
```
abbccdda, line 40: **Docs accuracy / data-minimization posture.** `snippet` is Gmail's own body excerpt — roughly the first ~100 characters of the thread's message text. That is message content, but the pack's docs state the opposite twice: "`full` hydrates entire decoded bodies and attachment trees ... Bodies belong to a future content-oriented surface" (gmail.yaml:55-57) and "**`messages` pins `detail: summary`** — the bounded row shape (no bodies, no attachment trees)" (`docs/open-connector-gmail.md`).

Both claims are true *of `messages`*, but they read as the pack's overall posture, and `threads.snippet` quietly ships truncated body text under `gmail.readonly`. That matters concretely for anyone sizing PII exposure or justifying the scope grant from these docs — a reader would conclude no body content is reachable through this pack, and they'd be wrong.

No change to the mapping: `snippet` is genuinely useful and it's what the executor returns. One sentence in the docs' caveat list noting that `threads.snippet` is a body excerpt would scope the "no bodies" claim to `messages` where it actually holds.

crates/skardi/src/sources/providers/open_connector/packs/gmail.yaml
```diff
@@ -0,0 +1,123 @@
...
+    fingerprint: 6e41cae2d3e764d7a5a92c220c4db2eb6c9b999eb3fc9b13f5bb514eabb877f1
+    pagination:
+      strategy: cursor
+      cursor_input: pageToken
+      next_cursor_path: "$.nextPageToken"
+      page_size_input: maxResults
+      page_size: 500
+    # `verbose` deliberately never sent, as on threads: the summary row
+    # is the draft id plus its message identity; hydrated drafts cost a
+    # drafts.get per row and carry unbounded bodies.
```
abbccdda, line 104: **Reliability (documented trade-off, but the cross-table effect isn't).** `drafts` declares no resources, so a binding's `query` reaches `threads` and `messages` and is silently dropped for `drafts`: `table.rs:94` filters by `declares_resource`, and the registration-time `UnknownResourceKey` guard (`mod.rs:200-206`) only fires when *no* bound table accepts the key — a partially-applicable key passes clean.

Concretely, the binding in `docs/open-connector-gmail.md` (`query: "newer_than:90d"`) yields 90-day `threads` and `messages` next to **all-time** `drafts`, with no error and nothing in the schema signaling it. Join `drafts` against `messages` in that binding and you're joining two different mailbox slices. The e2e at `gmail.rs:838-906` pins this as intended, and the docs' Resources column does list the per-table sets, so it's doc-visible — just not doc-*explained*.

Two things would close it:
1. A caveat bullet stating that resources scope only the tables declaring them, so a partially-applicable resource makes tables in the same binding describe different slices. That's the same hazard the `includeSpamTrash` bullet already calls out for `threads` vs `messages` — this generalizes it.
2. If the gateway's `list_drafts` input schema accepts `q` / `includeSpamTrash` (Gmail's own `drafts.list` does), declaring them here removes the inconsistency outright.

On (2) I can't check from the repo: the committed contracts are output schemas only (consistent with every other pack), so the four design decisions resting on input schemas — the `detail` pin, the per-table resource sets, `single_page`, and "`list_threads` offers no spam/trash input" — aren't re-verifiable here after an upstream bump. Not this PR's convention to change, but this pack leans on it harder than its predecessors.

crates/skardi/src/sources/providers/open_connector/packs/gmail.yaml
```diff
@@ -0,0 +1,123 @@
...
+    pagination:
+      strategy: cursor
+      cursor_input: pageToken
+      next_cursor_path: "$.nextPageToken"
+      page_size_input: maxResults
+      # 100, not the schema's 500 ceiling: the executor hydrates every
+      # listed message with a metadata `messages.get` (batches of 10),
+      # so a page of N rows costs N+1 Gmail calls — the page size bounds
+      # that burst against the per-user quota.
+      page_size: 100
```
abbccdda, line 56: **Bounds interaction worth one doc line.** The 100-row page size is well-reasoned for the hydration burst, but it also silently lowers this table's default scan ceiling: with the default `max_pages: 100` (`config.rs:59`), `messages` hits `ScanBoundsExceeded` at **10,000 rows**, where `threads`/`drafts` at 500/page reach 50,000. So a full-mailbox `SELECT *` fails 5× sooner here — after roughly 100 × 101 ≈ 10,100 Gmail calls.

The docs point at raising `max_pages`/`max_rows`, but on *this* table `max_pages` is the one knob that multiplies the hydration burst rather than just the row budget. Worth stating the effective ceiling and steering readers to `query`/`LIMIT` first.

---

Secondary, and **pre-existing** — the same sentence appears in the github/slack/notion/feishu docs, so not this PR's doing: "`429` responses are retried with backoff inside the scan deadline" (`docs/open-connector-gmail.md`) looks wrong for the failures Gmail actually produces. Execute runs under `RetryPolicy::NonIdempotent`, which retries only an HTTP 429 (`client.rs:110`). A Gmail-origin rate limit reaches Skardi via the executor's `assertGmailResponse` throw — the same path this PR documented as surfacing as `internal_error` for empty filters — and that is *not* retried. Since this is the pack's most quota-hungry surface, the claim is worth either confirming against the gateway's status mapping for a provider 429, or softening.

crates/skardi/src/sources/providers/open_connector/packs/gmail.rs
```diff
@@ -0,0 +1,1141 @@
...
+            r#"{"type": "object"}"#
+        };
+        MockResponse::ok(&discovery_ok("{}", output_schema, true, None))
+    }
+
+    // ── Contract tests: provider-shaped fixture pages derived from the
+    // executor source and the captured contracts (synthetic until the
+    // phase-4 live pass re-derives them as redacted captures). They pin
+    // the conversion contract per the admission gate: null-bearing,
+    // null-parent, nested, extra-field, and schema-mismatch. ─
```
abbccdda: **Stale comment.** This still says the page fixtures are "synthetic until the phase-4 live pass re-derives them as redacted captures" — but phase 4 landed in 060726d2, and both the module doc (gmail.rs:122-129) and every individual test comment ("Redacted live capture (2026-08-05)") now say the opposite. A reader hitting this block first will distrust the fixtures the PR worked hardest to make real.

Keep the one genuine exception when rewording: `messages_type_mismatch.json` *is* still synthetic, and correctly says so.

crates/skardi/src/sources/providers/open_connector/packs/gmail.rs
```diff
@@ -0,0 +1,1141 @@
...
+        assert_eq!(inputs[0]["maxResults"], 500);
+        assert_eq!(input_keys(&inputs[1]), vec!["maxResults", "pageToken"]);
+    }
+
+    #[tokio::test]
+    async fn single_page_tables_issue_exactly_one_request_with_no_inputs() {
+        // labels and filters declare the single_page strategy: one POST
+        // each, an EMPTY input object (no pagination keys — the strict
+        // schema would 400 them — and no userId), and no second request
+        // no matter what the response carries.
```
abbccdda: **The test doesn't cover its own claim.** The comment promises "no second request **no matter what the response carries**", but neither mock body carries a `nextPageToken` — so `inputs.len() == 1` passes on a response that has nothing to ignore. The property actually claimed here (a `SinglePage` table ignores a token if one shows up) is never exercised.

This is the only engine change in the PR, and this is its key safety property, so it's worth the one-line fix: add `"nextPageToken": "tok-2"` to the `list_labels` mock body at :796-801. Then `inputs.len() == 1` asserts something real about `advance()` returning `Ok(false)` (`pagination.rs:379`) instead of asserting something about the mock. As written, a regression that made `SinglePage` start following tokens would still pass this test green.

crates/skardi/src/sources/providers/open_connector/packs/gmail.rs
```diff
@@ -0,0 +1,1141 @@
...
+            gateway,
+            "SKARDI_TEST_OC_GMAIL_SINGLE",
+            "labels, filters",
+            "",
+        )
+        .await;
+
+        let batches = collect(&ctx, "SELECT id FROM saas.mail.labels ORDER BY id").await;
+        assert_eq!(column_values(&batches, "id"), vec!["INBOX", "Label_1"]);
+        let batches = collect(&ctx, "SELECT id FROM saas.mail.filters").await;
```
abbccdda: **Nit (cheap guard).** `to` was renamed to `to_addresses` for reserved-keyword hygiene (gmail.yaml:68-70), but `criteria` and `action` are never selected by bare name anywhere: this e2e selects only `id`, and `filters_fixture_converts_with_opaque_json` reads the columns off a converted `RecordBatch` rather than through SQL. `ACTION` is a sqlparser keyword, and `type` is only cleared by SQL because `udtf_parity_for_labels` happens to select it.

Widening this to `SELECT id, criteria, action FROM saas.mail.filters` costs nothing and pins that both names parse as identifiers in DataFusion — exactly what the `to_addresses` rename was protecting against.

crates/skardi/src/sources/providers/open_connector/packs/gmail.yaml
gracexmatin: All five flipped to nullable: false in 1144529; the module doc's nullability bullet is rewritten to match (it previously stated the convention this diverges from). Verified before flipping: the live-capture fixture, the absent-spellings pin, the type-mismatch fixture, and every e2e mock row carry all seven keys, and "" is a present key so nothing converts differently today.

One refinement to the framing: the fingerprint is a whole-schema BLAKE3 hash, so a declared rename inside the anyOf still fails registration. What nothing catches is emitted-key drift with an unchanged declaration — and non-null is the only guard for that case, which strengthens rather than weakens the conclusion.

crates/skardi/src/sources/providers/open_connector/packs/gmail.yaml
gracexmatin: Caveat added in 1144529: threads.snippet is Gmail's excerpt of the latest message's text, returned under gmail.readonly, and the "no bodies" posture is messages', not the pack's. Also annotated the yaml column and reworded the exclusions bullet to "Full message bodies".

crates/skardi/src/sources/providers/open_connector/packs/gmail.yaml
gracexmatin: (1) added as a caveat, with the example binding flagged inline (drafts: not narrowed).

(2) is settled against upstream source: list_drafts declares only maxResults/pageToken/verbose (src/providers/gmail/actions.ts:335-343), so there is nothing to declare — the strict schema would 400 an injected q. Gmail's own drafts.list does accept q, so exposing it is an upstream feature request, not a pack change.

On the aside: the other three input-schema-dependent decisions were re-verified against the same file — detail (enum with default: "summary"), list_threads declaring no spam/trash input, and labels/filters declaring no pagination inputs.

crates/skardi/src/sources/providers/open_connector/packs/gmail.yaml
gracexmatin: Ceiling documented in the messages caveat in 1144529 (10,000 rows at the default max_pages, with query/LIMIT steering, since max_pages is the burst-multiplying knob on this table).

The 429 aside doesn't hold, though — confirmed against the gateway source at the deployed v1.3.4: assertGmailResponse throws ProviderRequestError(429) (gmail/executors.ts:1073), toProviderExecutionError classifies status 429 as rate_limited (provider-runtime.ts:731 @v1.3.4), and mapExecutionErrorStatus re-emits rate_limited as the gateway's own HTTP 429 (runtime-api.ts:248 @v1.3.4) — exactly the one status NonIdempotent retries. The empty-filters internal_error is a different path: a response.json() TypeError falling to the fallback arm, not a ProviderRequestError. So the sentence stays, now with the verified mechanism noted in the doc — and the same sentence in the github/slack/notion/feishu docs is correct for the same reason.

crates/skardi/src/sources/providers/open_connector/packs/gmail.rs
gracexmatin: Reworded in 1144529: the header now states the fixtures are redacted live captures (2026-08-05) and names messages_type_mismatch.json as the deliberate synthetic exception (a live wire can't be made to produce a schema mismatch).

crates/skardi/src/sources/providers/open_connector/packs/gmail.rs
gracexmatin: Added in 1144529. The labels body now carries "nextPageToken": "tok-2"; filters keeps none — so the pair pins both shapes: a carried token must be ignored (advance() → Ok(false)) and an absent one still terminates. The comment now describes what the test actually asserts.

crates/skardi/src/sources/providers/open_connector/packs/gmail.rs
gracexmatin: Widened to SELECT id, criteria, action in 1144529, with a comment noting the ACTION-keyword hazard it pins.

abbccdda (commented):
## Review: Gmail source pack (milestone 5.4)

Strong pack. The wire contract is reconciled against the gateway rather than the Gmail API docs, the exclusions carry reasons, the fixtures are real redacted captures, and the phase-4 live pass is the kind of verification most connector PRs skip. `single_page` is the minimum viable engine extension and is correctly braced so `deny_unknown_fields` still bites. I found no security issues: no provider credentials touch Skardi, request inputs and row values stay out of logs and out of error messages (the mismatch test pins `found: "number"`, never the value), registration rejects undeclared resource keys, and the read-only posture holds.

Verified locally: `cargo test -p skardi --lib sources::providers::open_connector` → **312 passed, 0 failed**; the 21 `packs::gmail` tests all pass.

Findings below are ordered by production impact.

### 1. `messages`' all-non-null decision rests on a claim that isn't true — inline on `gmail.yaml`

The round-2 flip to `nullable: false` on all seven columns is justified by *"these columns ride outside the fingerprint gate (anyOf row items)"*. They don't. `fingerprint_schema` hashes `canonical_json` of the **whole** output schema, which recurses through `anyOf` like any other array. I deleted `to` from `contracts/fetch_emails.json`'s summary branch and ran the pack's own pin: `2a656f…` → `07276e…`. That drift fails at **registration**, same as every other pack.

What's actually uncovered is `fingerprint_uncovered_columns`' walker (`testutil.rs:172`) — a test helper. The pin is reporting a tooling limitation and four documents read it as a contract gap.

Meanwhile the cost is real and asymmetric: one message whose `messageTimestamp` is the executor's documented `""` fallback, or one row a failed hydration `messages.get` leaves partial, fails `ConversionFailed` for the entire `messages` scan — and with no filter pushdown there is no `WHERE` that excludes it. This PR also ships the counter-evidence that declared `required` isn't a runtime guarantee: `fixtures/gmail/drafts.json` (a live capture) violates `contracts/list_drafts.json`'s own `required` list, and the pack compensates there with `nullable: true`.

### 2. Silent truncation is the specified behavior of `single_page` — inline on `loader.rs`

`SinglePage::advance()` returns `Ok(false)` unconditionally, and the labels test now deliberately serves a `nextPageToken` to pin that it's *ignored*. That's the one place in the engine where a short answer is returned as complete, against the admission gate's "complete and terminating pagination" and against the fail-don't-truncate posture argued everywhere else in this PR. An optional `next_cursor_path` used purely as an assertion would turn the premise into something enforced.

### 3. Input-schema drift is ungated, and gmail is the most exposed pack — inline on `gmail.rs`

`fingerprint_schema` hashes the output schema only, and `ActionMetadata::input_schema()` has zero callers in the crate. These actions are `additionalProperties: false` strict and the pack generates six input keys; all were live-reconciled, none are pinned in the repo (`contracts/*.json` are output schemas). A gateway rename of `pageToken` passes registration and 400s every scan. Capturing input schemas next to the output ones and asserting the generated key sets against them would put the PR body's live reconciliation under CI.

### 4. The doc's binding example contradicts its own caveat — inline on `docs/open-connector-gmail.md`

The example `tables:` includes `filters`, which line 121-127 documents as failing on any mailbox with zero filters (most of them). Good find and honestly documented — the example should just agree with it.

### 5. Milestone number collision — inline on the tasks spec

Feishu is already 5.4; Gmail should be 5.5 (and the Review notes' "5.1–5.3 (GitHub, Slack, Notion)" omits Feishu).

### Minor

- **`README.md:339`** — the Open Connector row still advertises only the GitHub and Slack packs. Pre-existing (Notion and Feishu are missing too), but this PR adds a pack plus a guide and is a natural moment to fix the roster.
- **No empty-collection test for the single-page tables.** `optional_resources_forward_verbatim_and_only_where_declared` covers empty pages for threads/messages/drafts; `labels`/`filters` have none — and an empty `filters` collection is precisely the shape that's known-broken upstream. A one-line mock returning `{"filters": []}` would pin that the *engine* side is fine and localize the bug to the gateway.
- **Registration failure mode isn't documented**: binding `labelIds` while `tables:` omits `messages` fails registration ("every supplied resource key must be declared by at least one bound table", `mod.rs:201`). Worth a sentence next to the "resources narrow only the tables that declare them" bullet, since that bullet primes the reader to expect silent withholding.

### What checks out

- Fingerprint pins recomputed through the same `fingerprint_schema` registration uses; drift refusal pinned at registration, not the scan.
- Each of the three cursor tables pins its own exact input key set and a distinct end-of-collection spelling (explicit null / absent / empty string). `verbose` and unbound resources provably never reach the wire.
- No-pushdown guard asserts row identity, not just that the scan succeeded. Pagination-loop and failure-envelope tests assert the error *names* the action and the cause, and that the misleading row-path error is not what surfaces.
- Nullability on the other four tables is more lenient than the contract's `required`, which is the safe direction.
- `timestamp_ms_utc` genuinely accepts both the RFC 3339 and epoch-millis wires the comment claims (`json_to_arrow.rs:99`).
- Resource/fixed-input/pagination namespace collisions are rejected by the loader; gmail's declarations are clean.

### Mergeability: 8 / 10

Nothing here is a merge blocker on its own, and the engineering underneath is well above the bar for this codebase. The score is held back mainly by #1 — not because non-null is indefensible, but because the reasoning recorded in four places is wrong, and a future maintainer inheriting it will make the next decision on a false premise. #1 and #4 are small edits worth doing before merge; #2, #3 and the minors are fine as fast-follows.

crates/skardi/src/sources/providers/open_connector/packs/gmail.yaml
```diff
@@ -0,0 +1,132 @@
...
+    resources:
+      optional: [query, labelIds, includeSpamTrash]
+    columns:
+      # Non-null across the board: both non-ids branches of the captured
+      # contract `require` all seven fields, and the executor's fallbacks
+      # keep every key present ("" headers, [] labelIds). These columns
+      # ride outside the fingerprint gate (anyOf row items — see the
+      # coverage-gap pin), so non-null is the tripwire: emitted-key drift
+      # fails the scan loudly instead of silently going all-NULL.
+      - { name: message_id, path: messageId, type: utf8, nullable: false }
```
abbccdda, line 78: **The all-non-null bet on `messages` turns one odd row into a dead table — and neither premise behind it holds.**

Round 2 flipped these five columns to `nullable: false` on two stated grounds. I don't think either survives:

**1. "These columns ride outside the fingerprint gate (anyOf row items)."** They don't. `fingerprint_schema` hashes `canonical_json` of the **entire** discovered output schema (`action_registry.rs:178` → `util/json.rs:24`), and that walker recurses through every array and object — `anyOf` branches included. Drop `to` from the summary branch upstream and the canonical string changes, so the hash changes, so registration fails with exactly the mismatch `drifted_contract_fails_registration_not_the_scan` already pins.

I verified this rather than reasoning about it. Deleting `to` from `contracts/fetch_emails.json`'s `anyOf[0]` (both `properties` and `required`) and re-running the pack's own `pinned_fingerprints_match_the_reconciled_contracts` — i.e. through the real `fingerprint_schema`, not a replica:

```
gmail.messages: pinned Some("2a656f5fcab965512e0ce1b6382466cda2bd79dc28924aae460e054ad3ea29bb"),
                contract fixture hashes to 07276e71c10fe276823e4052de54fdeb05a900b2972765a0935bbc0a8e5cdcfa
```

The fingerprint moves. That drift fails at registration, exactly like every other pack's.

What actually stops at `anyOf` is `fingerprint_uncovered_columns` (`testutil.rs:172`), a *test helper's* path walker. The coverage pin is reporting a tooling limitation, and this comment reads it as a contract gap.

**2. "Both non-ids branches of the captured contract `require` all seven fields."** True of the declaration — but this PR also ships the counter-example. `fixtures/gmail/drafts.json` is a redacted **live** capture whose `message` object carries `messageId`/`threadId` and nothing else, while `contracts/list_drafts.json` requires all seven inside `message`. Same executor family, same declaration style, and the pack already had to compensate there with `nullable: true`.

So what non-null actually buys is a tripwire against *executor-vs-contract divergence* — the one class where this pack has already demonstrated divergence happens, and the one class no declaration-level gate can catch. The cost is asymmetric: one message whose `messageTimestamp` is the executor's `""` fallback (a spelling the module doc at line 68-72 says the executor can emit), or one row a failed hydration `messages.get` leaves partial, fails `ConversionFailed` for the **entire** `messages` scan. With no filter pushdown there is no `WHERE` that excludes the bad row — the table is simply unusable until upstream changes.

Suggestion: keep `message_id`/`thread_id` non-null (real identity, and the drift tripwire that matters), and return `label_ids`/`subject`/`sender`/`to_addresses`/`message_timestamp` to nullable. Per (1), silent all-NULL drift on those is still caught at registration by the fingerprint.

crates/skardi/src/sources/providers/open_connector/packs/gmail.rs
```diff
@@ -0,0 +1,1154 @@
...
+//!   headers, `[]` labelIds). Since `messages` rides outside the
+//!   fingerprint gate (next bullet), non-null is what turns emitted-key
+//!   drift into a loud `missing key` failure instead of a silently
+//!   all-NULL column. Everything else stays nullable.
+//! - **Fingerprints are pinned** from a live gateway capture
+//!   (`fixtures/gmail/contracts/`, gateway v1.3.4). `fetch_emails`
+//!   declares its row items as an `anyOf` (ids | summary | full) the
+//!   coverage walker does not descend, so every `messages` column rides
+//!   outside the fingerprint gate — the coverage-gap pin records that
+//!   honestly; drift there surfaces at scan time per conversion rules.
```
abbccdda: **The "outside the fingerprint gate" framing is inaccurate, and it's repeated in four places.**

`fingerprint_schema` hashes the canonicalized **whole** output schema (`action_registry.rs:178`, `util/json.rs:24`), which recurses into `anyOf` branches like any other array. A summary-branch rename or a dropped `required` entry *does* move the fingerprint and *does* fail registration — I confirmed by deleting `to` from `anyOf[0]` and watching this pack's own `pinned_fingerprints_match_the_reconciled_contracts` report `2a656f…` → `07276e…`. What stops at `anyOf` is `fingerprint_uncovered_columns`' walker (`testutil.rs:172`) — a test helper, not the gate.

For every other pack, the coverage pin means something real: those paths live under `additionalProperties` passthrough, so the declared schema genuinely says nothing about them and upstream can rename them without moving the hash. Gmail's `messages` columns are *declared* — inside branches the helper can't follow. Same pin, categorically different meaning, and the doc collapses the two.

The same sentence appears in `gmail.yaml` (messages column block), `docs/open-connector-gmail.md:134-139`, and the tasks-spec 5.4 entry, and it's load-bearing: it's the stated reason for the round-2 nullability change. Worth correcting in all four, and worth renaming `fingerprint_coverage_gap_is_pinned` to say what it pins — paths the *static coverage walker* cannot resolve. As written, the next maintainer will read "drift there surfaces at scan time per conversion rules" and reason away from a guarantee the engine actually gives at registration time.

Follow-up worth considering: teach `fingerprint_uncovered_columns` to descend `anyOf`/`oneOf` (a path is covered when every branch declares it). Gmail's pin then becomes `&[]` honestly, and the helper stops misreporting for the next pack with a branched item schema.

crates/skardi/src/sources/providers/open_connector/packs/fixtures/gmail/drafts.json
```diff
@@ -0,0 +1,12 @@
+{
+  "drafts": [
+    {
+      "id": "r-4000000000000000001",
+      "message": {
```
abbccdda, line 5: Flagging this fixture as the evidence behind my comment on `messages`' nullability.

This is a redacted **live** capture, and its `message` object carries `messageId` and `threadId` only. `contracts/list_drafts.json` declares `drafts[].message` with `required: [messageId, threadId, labelIds, subject, sender, to, messageTimestamp]` — all seven. So the executor's real output does not satisfy its own declared contract, and the pack correctly compensates with `nullable: true` on `message_id`/`thread_id`.

That is exactly why I'd push back on `messages` deriving `nullable: false` from *"both non-ids branches of the pinned contract `require` all seven fields"*: this PR ships a same-family, same-gateway counter-example showing `required` is a declaration, not a runtime guarantee. Worth a line in the module doc either way — the drafts/messages asymmetry is currently unexplained, and the next reader will wonder why the same reasoning produced opposite nullability.

crates/skardi/src/sources/providers/open_connector/packs/loader.rs
```diff
@@ -460,6 +460,15 @@ enum PaginationDoc {
...
     },
+    /// One request is the complete collection: for actions whose strict
+    /// input schema declares no pagination keys at all (Gmail
+    /// `list_labels` / `list_filters`), where injecting a page or cursor
+    /// parameter would be rejected as `invalid_input`. The scan issues a
+    /// single request and never advances. A braced variant (not a unit
+    /// one) so `deny_unknown_fields` keeps rejecting stray keys — a
+    /// misspelled pagination field must fail loudly, not ride along
+    /// ignored.
+    SinglePage {},
```
abbccdda: **`single_page` makes silent truncation the specified behavior of the one strategy that can truncate.**

`SinglePage::advance()` returns `Ok(false)` unconditionally (`pagination.rs:379`), and `single_page_tables_issue_exactly_one_request_with_no_inputs` now deliberately serves `nextPageToken: "tok-2"` on the labels body to pin that it is *ignored*.

That sits badly against everything else this PR argues for. The design spec's admission gate opens with "complete and terminating pagination" (`2026-07-11-open-connector-integration-design.md:731`); the engine elsewhere fails rather than shortens — `ScanBoundsExceeded` instead of a truncated answer, `PaginationLoop` instead of a spin; the pack doc sells "fail-don't-truncate" explicitly. `single_page` is the one place a short answer is returned as if it were complete.

The strategy's premise is "one request **is** the complete collection". When the wire contradicts that premise, silence is the wrong response: a mailbox that outgrows a single labels response yields a quietly short table and no signal anywhere.

Suggestion — keep the strategy, add an assertion rather than an assumption: let `single_page` accept an optional `next_cursor_path`, used only to check that the premise held. If it resolves to a non-null, non-empty value, fail the scan with a targeted error naming the action. Gmail's two tables declare `$.nextPageToken`, and the test flips from "the token is ignored" to "the token is refused" — a materially stronger pin, and one that keeps the variant braced-and-strict as intended.

crates/skardi/src/sources/providers/open_connector/packs/gmail.rs
```diff
@@ -0,0 +1,1154 @@
...
+        } else if path.ends_with("gmail.list_drafts") {
+            include_str!("fixtures/gmail/contracts/list_drafts.json")
+        } else if path.ends_with("gmail.list_labels") {
+            include_str!("fixtures/gmail/contracts/list_labels.json")
+        } else if path.ends_with("gmail.list_filters") {
+            include_str!("fixtures/gmail/contracts/list_filters.json")
+        } else {
+            r#"{"type": "object"}"#
+        };
+        MockResponse::ok(&discovery_ok("{}", output_schema, true, None))
```
abbccdda, line 239: **Nothing gates input-schema drift, and gmail is the pack most exposed to it.**

`discovery_ok("{}", output_schema, …)` passes an empty input schema here — and that faithfully mirrors production. `fingerprint_schema` hashes the **output** schema only (`action_registry.rs:178`), and `ActionMetadata::input_schema()` has **zero** callers anywhere in the crate (`grep -rn "input_schema()" open_connector/` returns only its own definition). The registration gate says nothing whatsoever about inputs.

For most packs that's a modest hole. Here it isn't. The module doc's own framing is that these actions are `additionalProperties: false` strict and "a wrong key is a hard 400", and the pack generates six distinct input keys across the five tables — `pageToken`, `maxResults`, `detail`, `query`, `labelIds`, `includeSpamTrash`. All six were reconciled against live gateway v1.3.4, and none of them are pinned in the repo: `contracts/*.json` capture output schemas only. A gateway upgrade that renames `pageToken`, narrows the `maxResults` bound below 500, or tightens the `detail` enum passes registration cleanly and then 400s **every scan** — precisely the class of failure the fingerprint gate exists to pull forward to registration time. The design spec assumes the registry holds input schemas for this purpose (`…-integration-design.md:164`).

Cheap version, in scope for this PR: capture each action's input schema next to its output schema in `contracts/`, and add one test asserting that each table's generated input key set is accepted by it — required keys present, no key rejected under `additionalProperties: false`, `detail: summary` inside the enum, page sizes inside the declared bounds. That converts the live reconciliation described in the PR body into something CI re-checks.

Gating input drift at registration is engine work and fairly a follow-up — but even a comment recording that the gate is output-only would stop the next pack author from assuming otherwise.

docs/open-connector-gmail.md
```diff
@@ -0,0 +1,161 @@
...
+      open_connector:
+        runtime_token_env: OPEN_CONNECTOR_TOKEN
+        bindings:
+          - name: mail                   # schema name in SQL
+            source_pack: gmail
+            resource:                    # all optional; omit for the default listing
+              query: "newer_than:90d"        # Gmail search syntax; threads + messages (drafts: not narrowed)
+              labelIds: [INBOX]              # messages only
+              includeSpamTrash: false        # messages only
+            tables: [threads, messages, drafts, labels, filters]
```
abbccdda: **The copy-paste binding example includes `filters`, which this same doc says fails on most mailboxes.**

Line 121-127 documents that a mailbox with zero filters fails the `filters` scan with the gateway's `internal_error`, and advises: *"bind `filters` only for mailboxes that have at least one filter until then."* But the binding example — the block people actually copy — lists `filters` in `tables:`.

Most Gmail mailboxes have no filters at all, so the documented starting configuration produces one broken table on first run, and the reader only finds out 80 lines later. Good catch on finding this live and documenting it honestly; the example should just agree with the caveat.

Suggestion: drop `filters` from the example's `tables:` and mark it, e.g.

```yaml
            tables: [threads, messages, drafts, labels]
            # filters: add once the mailbox has >= 1 filter — see caveats below
```

docs/superpowers/specs/2026-07-11-open-connector-integration-tasks.md
```diff
@@ -305,7 +305,74 @@ event carrying the scan identity and error.
       tests (`cargo test -p skardi --lib sources::providers::open_connector`,
       post-merge with 5.3), 17 pack-scoped (`… packs::feishu`), 842 full
       library suite.
-- [ ] 5.5 Later waves per the design rollout (Google Workspace, Discord, HubSpot, Jira, …) through the source-pack admission gate
+- [x] 5.4 Gmail pack (Google OAuth, page-token cursor pagination): `threads`,
```
abbccdda: **Milestone number collides with Feishu's.**

Line 287 is already `- [x] 5.4 Feishu pack`, so this entry makes two 5.4s. The Review notes below (line 386) compound it: *"Milestones 1–4 and 5.1–5.3 (GitHub, Slack, Notion packs) are merged"* skips Feishu entirely, which is merged and is 5.4.

Gmail should be 5.5, "Later waves" (line 375) renumbered to 5.6, and the Review notes line should read 5.1–5.4 (GitHub, Slack, Notion, Feishu).

crates/skardi/src/sources/providers/open_connector/packs/gmail.yaml
gracexmatin: Both points taken; the suggestion is implemented in ab6e4c0 — `label_ids`/`subject`/`sender`/`to_addresses`/`message_timestamp` back to nullable, `message_id`/`thread_id` kept non-null.

Two corrections for the record, since the reasoning outlives the diff. Neither cost example actually distinguishes the two settings. A `""` `messageTimestamp` is a *present* key, and `collect_cells_described` (json_to_arrow.rs:509-514) routes only JSON null and missing keys to NULL — `""` reaches the converter and fails identically under `nullable: true`; the module doc said as much before the flip. And a failed hydration cannot leave a partial row: `hydrateInBatches` awaits `Promise.all`, so one bad `messages.get` rejects the whole request into a failure envelope.

Nor does keeping identity non-null carry the tripwire for the other five — `summarizeGmailMessage` is a single object literal, so an upstream that renames `sender` leaves `messageId`/`threadId` untouched and nothing fires.

So the revert rests on the general asymmetry you named — no filter pushdown means no `WHERE` to route around a dead table, too blunt for descriptive columns — rather than on those instances. The accepted consequence is recorded instead of glossed: the yaml and the pack doc both say an all-NULL `subject`/`sender`/`labelIds` is a contract alarm to investigate, never mail without senders.

crates/skardi/src/sources/providers/open_connector/packs/gmail.rs
gracexmatin: Correct, and corrected in all four places (6a8a865), with the test renamed to `columns_the_coverage_walker_cannot_resolve_are_pinned` (32a31c0).

7b51057 goes after the source rather than the copies: `fingerprint_uncovered_columns`' own doc comment made exactly this claim, which is where I took it from. It now states the two situations it conflates and classifies every pack against its captured contract — github, slack and feishu genuinely undeclared under `additionalProperties` (feishu declares its item schemas wholly loose); gmail purely the walker's blind spot; and notion a **mix** worth knowing about: most of its search-backed paths are declared inside the same kind of branch, while `is_archived` and `public_url` really are not. The helper's doc now tells the reader not to take a non-empty list as "outside the gate" without checking the pack's own pin comment.

The `anyOf`/`oneOf` descent is noted as a follow-up — it belongs in the helper rather than in any one pack, so I've left it out of this PR.

crates/skardi/src/sources/providers/open_connector/packs/fixtures/gmail/drafts.json
gracexmatin: The counter-example is well taken, and it is now load-bearing in the module doc rather than incidental: it is cited as the concrete proof that `required` is a declaration and not a runtime guarantee, which is precisely the drift class no declaration-level gate catches. I traced it upstream to be sure — `listDrafts`' non-verbose path returns `message: { messageId, threadId }` literally, while the contract describes that object as a "Normalized Gmail message", so the divergence is structural rather than a capture artefact.

The unexplained asymmetry you flagged is fixed in 004aed0. It survives the nullability revert (messages' identity is non-null, drafts' is not), and the reason is structural rather than a change of mind: `messages` carries `messageId`/`threadId` at the row's top level, whereas `drafts` reaches them through a `message` parent — and a nullable column behind a JSON-null parent must become SQL NULL rather than fail, the admission gate's null-parent category, pinned by `null_parent_on_a_nested_path_becomes_sql_null`. Non-null there would convert a nulled-out parent into a dead scan.

crates/skardi/src/sources/providers/open_connector/packs/loader.rs
gracexmatin: Implemented in b745dcf, as suggested. `single_page` takes an optional `next_cursor_path`, read only to check the premise and never to fetch: absent, null and empty-string continuations terminate as before — the same three spellings the cursor strategy accepts — and anything else fails as `SinglePageIncomplete`. gmail's `labels`/`filters` declare `$.nextPageToken`, and the test flips from "the token is ignored" to "the token is refused", with an e2e that also pins that the refusal happens after exactly one request rather than by following it.

One addition beyond the suggestion: the path is parsed in `PaginationStrategy::validate()`, so a pack typo fails at registration instead of leaving the check quietly unarmed until some scan happens to exercise it. `open_connector_scan` passes `None` and keeps the historic behaviour — its raw actions declare no pagination contract to check against, and "one request, one page" is already its documented contract.

One deliberate deviation: the error names the path and page rather than the action, following its three sibling pagination errors (`Pagination` holds no action context). Happy to plumb it if you'd rather have the action named.

For the record, the unconditional `Ok(false)` predates this PR — `pagination.rs` was untouched on this branch until b745dcf. What this PR did was make the strategy reachable from YAML, which is exactly why the behaviour surfaced here.

crates/skardi/src/sources/providers/open_connector/packs/gmail.rs
gracexmatin: Verified and acted on: `input_schema()` really does have zero callers, the committed contracts really are output-only, and the design spec at :164 assumes the registry holds both.

1b6e4be does the cheap version. The input halves are captured under `contracts/inputs/` and a test asserts each table's generated key set against them — strictness, every key declared, required keys supplied, page sizes inside the declared bounds, `detail: summary` inside the enum. Provenance is the same capture rather than a reconstruction: re-running discovery against a local gateway v1.3.4 returned all five *output* schemas byte-identical to the committed ones before I saved the input halves.

03f2f3e then narrows a claim I overstated first time round. Both sides of that test are committed artifacts, so it locks declarations against the capture and catches input drift on **re-capture**, not live. Real protection needs an `expected_input_fingerprint` compared at registration, symmetric with the output pin — recorded as an open engine follow-up in the source-pack skill, with the interim rule that inputs are re-checked on every gateway bump.

docs/open-connector-gmail.md
gracexmatin: Fixed in 6a8a865: the example's `tables:` drops `filters`, with an inline note that it is bindable once the mailbox has at least one filter, pointing at the caveat below.

docs/superpowers/specs/2026-07-11-open-connector-integration-tasks.md
gracexmatin: Fixed in 6a8a865: gmail is 5.5, later waves 5.6, and the Review notes now read "Milestones 1–4 and 5.1–5.4 (GitHub, Slack, Notion, Feishu packs) are merged". The PR title and body were updated to match, so a squash merge will not reintroduce 5.4.

gracexmatin:
## Follow-up: input-schema contracts for the other four packs

Recording this here so it does not get lost with the thread it came out of. The input-schema gap raised on `gmail.rs:231` is **engine-wide, not a gmail trait** — gmail only got the pin because it was the pack under review. What follows is what I verified while implementing it, plus the obstacle a rollout has to clear first.

### The gap

`fingerprint_schema` hashes the **output** schema alone (`action_registry.rs:178`), and `ActionMetadata::input_schema()` has zero callers in the crate. Registration therefore says nothing about the keys a pack *sends*. Every provider's actions are `additionalProperties: false` — I probed one action per provider against a live gateway v1.3.4 (`github.list_repository_issues`, `slack.list_conversations`, `notion.search`, `feishu.list_chats`, `gmail.fetch_emails`), all strict — so a renamed or retyped input key is a hard 400 on **every scan** of the affected table, for any pack, while registration stays green.

### Exposure per pack

Distinct action-input keys each pack can put on the wire, from its yaml:

| pack | input keys | filter pushdown |
|---|---|---|
| feishu | 10 | yes |
| github | 9 | yes |
| slack | 8 | yes |
| notion | 7 | yes |
| gmail | **6** | no |

The four packs with pushdown also have a *query-dependent* input surface — a translated predicate contributes keys a fixed audit of the yaml does not see — so their exposure is both wider and less static than gmail's.

Only gmail currently pins the input halves (`contracts/inputs/` + `generated_inputs_are_accepted_by_the_captured_input_contracts`, 1b6e4be).

### The obstacle: capture provenance

Each pack's contracts were captured against a different gateway, and two record no version at all:

| pack | recorded capture version |
|---|---|
| slack | v1.3.1 |
| feishu | v1.3.3 |
| gmail | v1.3.4 |
| github | *none recorded* |
| notion | *none recorded* |

Capturing input halves at v1.3.4 for a pack whose output halves came from v1.3.1 would put the two halves of one "contract" at different upstream revisions — worse than the current honest gap. Two ways out:

1. **Capture each pack at its own recorded version.** Faithful, but impossible as-is for github and notion, which have no version to reproduce.
2. **Re-baseline every pack on one gateway version.** Re-takes the output fingerprints too, so it is a larger change — but it leaves all five internally consistent and gives github and notion a recorded provenance they currently lack.

### Ceiling worth knowing before anyone copies the gmail approach

The per-pack test locks two *committed* artifacts against each other, so it catches input drift when the contracts are **re-captured** after a gateway bump — not while a gateway drifts underneath a deployed binary. The fix that actually closes it is an `expected_input_fingerprint` compared at registration, symmetric with the output pin (recorded as an open follow-up in `docs/superpowers/skills/source-pack/references/contract-reconciliation.md`). That single engine change would cover all five packs plus every future one, and would reduce the per-pack work to a one-line pin.

Suggested order, therefore: the engine change first, then one capture pass — rather than four more copies of the interim measure.

crates/skardi/src/sources/providers/open_connector/packs/gmail.yaml
```diff
@@ -0,0 +1,143 @@
...
+    pagination:
+      strategy: cursor
+      cursor_input: pageToken
+      next_cursor_path: "$.nextPageToken"
+      page_size_input: maxResults
+      # 100, not the schema's 500 ceiling: the executor hydrates every
+      # listed message with a metadata `messages.get` (batches of 10),
+      # so a page of N rows costs N+1 Gmail calls — the page size bounds
+      # that burst against the per-user quota.
+      page_size: 100
```
abbccdda, line 56: **LIMIT never reaches the wire — and this is the first table where that costs a 101-call burst.**

The page-size reasoning here is right, but it bounds the *page*, not the *query*. DataFusion's pushed-down LIMIT is threaded into `limit_remaining` and used to truncate the batch and stop pagination (`exec.rs:533-541`), but `Pagination::apply` always injects the pack's fixed `page_size` (`pagination.rs:232-234`). The limit never influences the request.

So `SELECT ... FROM saas.mail.messages LIMIT 1` still sends `maxResults: 100`, and because this executor hydrates every listed message with a metadata `messages.get`, the gateway makes **101 Gmail calls (~505 quota units) to answer a one-row query**. `open_connector_query('saas', 'gmail.messages', '{}')` has the same shape, and it is the surface most likely to be typed ad hoc by someone exploring.

Every earlier pack pays only the listing cost for this, which is why it has never mattered. `messages` is the first table where a page fetch has N+1 amplification, so it is where the gap first bites — and it bites hardest exactly where the docs send people ("narrow with `query` or `LIMIT`", gmail-doc:113).

Engine-side fix is small: inject `min(limit_remaining, page_size)` when a `page_size_param` is declared. One interaction to be careful with — `PageNumber`'s heuristic termination compares `rows_in_page >= per_page` against the *declared* size (`pagination.rs:288`), so the effective size has to be threaded there too, or the shrink applied to `Cursor` only. Safe for gmail either way: the executors never post-filter rows, and a short page can only end a scan that LIMIT was about to end anyway.

Until that lands, one line in the doc beside the existing `max_pages` note would help: `LIMIT` bounds rows returned, not Gmail calls made.

crates/skardi/src/sources/providers/open_connector/packs/gmail.yaml
```diff
@@ -0,0 +1,143 @@
...
+    # detail=summary pins the bounded row shape: `ids` carries no
+    # metadata worth a table, `full` hydrates entire decoded bodies and
+    # attachment trees (unbounded row size). Bodies belong to a future
+    # content-oriented surface, not a listing table.
+    fixed_inputs:
+      detail: summary
+    resources:
+      optional: [query, labelIds, includeSpamTrash]
```
abbccdda, line 64: **Nothing type-checks binding resource values, and gmail is the first pack where they aren't all strings.**

`labelIds` (array) and `includeSpamTrash` (boolean) are the first non-string resources any pack declares — github, slack, notion and feishu bind strings only. Resource values keep their YAML types verbatim (`config.rs:294-298`, pinned by `resource_values_keep_their_yaml_types` and by this pack's forwarding e2e), and `validate()` rejects only nulls (`config.rs:321-325`). Nothing ever compares a supplied value against the action's declared input type.

Consequence: `labelIds: INBOX` — the natural single-label spelling, and the one YAML makes easiest to write — passes config validation, passes registration, and then fails **every** scan of `saas.mail.messages` as `invalid_input` against the `additionalProperties: false` schema, at query time, with nothing pointing at the binding. Same for `includeSpamTrash: "true"`.

The fix is already sitting there: `ActionMetadata::input_schema()` is populated from discovery at registration (`action_registry.rs:44,57`) and still has zero callers. Checking each supplied resource against `properties[key].type` at registration turns this into a startup error naming binding and key. That is strictly stronger than `generated_inputs_are_accepted_by_the_captured_input_contracts`, which is a good addition but locks two committed artifacts to each other and by construction cannot see a user-supplied value.

Follow-up rather than a blocker — and it is the same open engine gap the input-contract thread already identified. Worth noting there that the gap has a second half (values, not just keys) that gmail is the first pack to expose.

crates/skardi/src/sources/providers/open_connector/packs/loader.rs
```diff
@@ -242,7 +242,7 @@ fn validate_table(table: &SourcePackTable) -> Result<(), String> {
         } => std::iter::once(cursor_param)
             .chain(page_size_param)
             .collect(),
-        PaginationStrategy::SinglePage => Vec::new(),
+        PaginationStrategy::SinglePage { .. } => Vec::new(),
```
abbccdda, line 245: **`validate_table` guards filter and fixed inputs against pagination params, but never resources.**

`pagination_params` is built here and then checked against filter inputs (`:295`) and fixed inputs (`:315`). Resource names are never checked against it — even though `exec.rs:451-460` seeds the request from the resource object **first** and applies pagination **last**. A pack declaring a resource named `pageToken` or `maxResults` would have it silently overwritten on every page, which is precisely the failure mode the comment at `:226-231` says this block exists to prevent ("exec.rs applies pagination LAST, so a collision would silently overwrite...").

gmail does not hit it — `query`/`labelIds`/`includeSpamTrash` vs `pageToken`/`maxResults` — so this is prevention, not a bug report. But this PR is the one adding a strategy whose `pagination_params` is empty, which makes the asymmetry that much easier to overlook next time. One loop beside the fixed-input check closes it:

```rust
for key in table.required_resources.iter().chain(table.optional_resources) {
    if pagination_params.contains(key) {
        return Err(format!(
            "{id}: resource '{key}' collides with a pagination input, which is \
             applied last and would overwrite it"
        ));
    }
}
```

crates/skardi/src/sources/providers/open_connector/packs/gmail.rs
```diff
@@ -0,0 +1,1437 @@
...
+//! - **`to` → `to_addresses`**: TO is a reserved SQL keyword; a column
+//!   requiring quotes on every reference is hostile to use. The value is
+//!   the raw `To` header string (display names and all), exactly as the
+//!   executor emits it.
+//! - **Header-derived fields spell "absent" as `""`**, never null (the
+//!   executor's `readHeader` fallback): empty subjects/recipients stay
+//!   empty strings in SQL. `messageTimestamp` always carries an RFC 3339
+//!   instant in practice (Gmail's `internalDate` is set on every real
+//!   message); the executor's `""` fallback for a hypothetical
+//!   dateless message would fail conversion loudly rather than corrupt
```
abbccdda, line 76: **This verdict on `""` contradicts the nullability bullet 40 lines below it — and `nullable: true` cannot split the difference.**

Here: an executor-emitted `""` `messageTimestamp` "would fail conversion loudly rather than corrupt the column — the same loud-on-mismatch rule every pack follows." At `:109-116` the pack rejects exactly that outcome for the other descriptive columns: "`messages` has no filter pushdown, so a single unconvertible row takes the entire table down with no `WHERE` to route around it."

Both cannot be the pack's position, and the mechanics do not let nullability mediate. `""` is a **present** key, so `collect_cells_described` (`json_to_arrow.rs:507-520`) hands it to the converter regardless of `nullable`; `parse_timestamp` (`:546-552`) accepts RFC 3339 or a JSON number and returns `None` for `""`, so it lands in `fail` as `ConversionFailed`. One dateless message renders `saas.mail.messages` permanently unqueryable, with no predicate available to exclude it — the exact failure round 3 flipped these five columns to nullable to avoid. The message is mis-descriptive too: `found: "string"` against `expected: "RFC 3339 timestamp or epoch millis"` reads as a type error rather than an empty value.

I do not think it is likely — `internalDate` is set on every real Gmail message and the live pass saw real values on 29/29 rows — so this is not a blocker. But the two paragraphs should agree. Either mitigate (let a nullable timestamp column read `""` as NULL, or map the column as `utf8` and cast in SQL), or confirm upstream cannot emit `""` on this path and say *that* — rather than presenting the whole-table failure as the desired behaviour in one bullet and as the rejected one in the next.

docs/superpowers/specs/2026-07-11-open-connector-integration-tasks.md
```diff
@@ -305,7 +305,84 @@ event carrying the scan identity and error.
...
+      guesses (which system labels omit visibility fields; a fresh
+      draft's threadId equals its messageId) and caught one upstream
+      bug: a zero-filter mailbox fails `list_filters` with
+      `internal_error` (Gmail returns an empty body the executor's
+      `response.json()` does not tolerate) — documented in the pack doc,
+      fix belongs upstream. Fixtures are redacted live captures from
+      that pass (mechanically audited against an allowlist); executor
+      absent-spellings the capture lacked are pinned by an inline
+      converter test.
+      **Verification**: 284 open_connector tests (counted by `cargo test
```
abbccdda, line 373: **The test count is stale — 284 was true a few commits ago, the head is 316.**

I ran the stated command at `0ce42a8`:

```
cargo test -p skardi --lib sources::providers::open_connector
test result: ok. 316 passed; 0 failed; 0 ignored; 843 filtered out
```

and the pack-scoped slice:

```
cargo test -p skardi --lib sources::providers::open_connector::packs::gmail
test result: ok. 24 passed; 0 failed; 1135 filtered out
```

So both numbers in this block drifted: 284 → **316**, and "23 new" → **24** in the gmail module alone, before counting the loader `single_page` pass/refusal test, the registry pin in `source_pack.rs`, and `a_declared_single_page_premise_is_checked_not_assumed` in `pagination.rs`. The count was accurate when written; the round-3 additions (input-contract pins, the premise check, the empty-collection e2e, the live-continuation refusal) landed after it.

Cosmetic, but this repo treats these counts as load-bearing enough to have a commit dedicated to correcting one (`fd5043d`, "correct the 5.4 test count to the CI-confirmed 283"), and the same drift would otherwise ride into the merge commit. Worth refreshing both figures, and adding the two new e2e names the list does not yet mention (`a_single_page_table_refuses_a_live_continuation_token`, `single_page_tables_scan_an_empty_collection_cleanly`) — they are among the more valuable tests in the PR.

abbccdda (commented):
Reviewed at `0ce42a8` (post round-3), against the integration design spec's source-pack admission gate. I went through the pack YAML, the engine extension, the conversion/pagination paths it rides, and the committed contracts; the earlier rounds' threads (nullability, the coverage-walker framing, `single_page` truncation, input-schema drift, the milestone number) all read as genuinely settled at this head, so I didn't re-litigate them.

**This clears the admission gate.** Terminating pagination on all five tables — and `single_page` is now the only strategy in the engine that *asserts* its own premise instead of assuming it, which is a net improvement to the engine, not just to gmail. Deterministic schemas with fingerprints pinned to committed live captures and a refusal-side e2e. Read-only throughout. Scopes, rate limits, the SPAM/TRASH default, the hydration cost, and the per-table filter/limit behaviour all documented. Fixture categories complete (empty, null-bearing, null-parent, nested, extra-field, schema-mismatch). The phase-4 real-mailbox pass is the part I'd single out: it settled two synthetic guesses and caught an upstream defect, which is exactly what live verification is for.

Five findings below, none blocking. Two are engine gaps that gmail is simply the first pack to make reachable; three are local. Ordered by production impact.

**Verification I ran:** `cargo test -p skardi --lib sources::providers::open_connector` at this head — **316 passed, 0 failed**, of which `…packs::gmail` is **24 passed, 0 failed**. Plus a read of `exec.rs` / `pagination.rs` / `json_to_arrow.rs` / `config.rs` / `table.rs` / `action_registry.rs` to check the claims the pack's comments make about engine behaviour. Every claim I checked held except the two noted in findings 1 and 4.

abbccdda (approved):
Approval with comments, nice job!

PR #193, opened 2026-08-05
feat(udf): chunk_parts + json_pack — runtime functions for generated etl bundles

bakey:
## Why

Generated etl bundles ([skardi-cloud#232](https://github.com/SkardiLabs/skardi-cloud/pull/232)) run their ingest jobs on **this** server. The etl design's runtime audit (skardi-cloud `design_docs/skardi_etl_generator.md`, Research Findings) concluded that today's engine carries everything those bundles need **except two small UDFs** — and the PRD's contract is "everything after generation uses existing skardi commands; the generator adds no new runtime". Without these functions, a generated ingest job fails on a stock server at `chunk_parts`.

This is the engine-side remainder of what #191 carried, isolated: **no generator code, no etl coupling** — three self-contained additions with their full test suites (~725 lines, 6 files).

## What

- **`chunk_parts(mode, text, size[, overlap]) → List<Struct<chunk_idx Int32, chunk_text Utf8>>`** (feature `chunking`, registered beside `chunk`). The 0-based split ordinal is assigned *inside* the UDF, in one deterministic pass per value — necessary because DataFusion cannot plan `UNNEST … WITH ORDINALITY` (apache/datafusion#11419). The plannability pin test locks the projection-position `UNNEST(chunk_parts(...)) AS part` spelling that does plan, asserting per-source-row ordinal restart and overlap-0 reassembly end to end. Also generally useful: inline chunked ingestion with stable ordinals, no external pipeline.
- **`json_pack(key, value, …) → Utf8`** — the SQL-callable JSON **encoder** (DataFusion core has none through 54.x; `datafusion-functions-json` is read-side only). `serde_json` does the encoding, which makes the UDF an injection boundary: adversarial strings (quotes, backslashes, control chars, non-ASCII) round-trip byte-exact through a real JSON parser in tests. Keys are non-null literals, insertion-ordered (deterministic output); values cover the JSON scalar set + Timestamp (epoch ms) + `List<Utf8>`; non-finite floats refuse; duplicate keys are an error. Registered on both server session contexts.
- **`SourcePackRegistry::packs()`** — name-sorted deterministic enumeration of the built-in packs (the map itself is private and unordered), with a completeness pin test.

## Verification

839 lib tests green with `--features chunking` (822 default); `cargo fmt --check`, zero new clippy warnings, docs build clean under `-D warnings`, skardi-server builds. The consumer-side proof lives with the generator in skardi-cloud#232: a plan-check that plans every generated statement against these exact registrations, and two hybrid e2e suites (real executor ingest → RRF search) that execute them.

## Relation to #191

#191 (closed) carried the full generator; the generator itself now lives in skardi-cloud#232. This PR is only the runtime functions the engine must own for those bundles to run — once merged, skardi-cloud will repin its git dependency and drop its temporarily vendored copies of these UDFs.


crates/server/src/config.rs
```diff
@@ -423,6 +423,9 @@ pub async fn load_server_config(args: CliArgs) -> Result<ServerConfig> {
     // Register chunk UDF (text-splitter wrapper for inline ingestion)
     #[cfg(feature = "chunking")]
     register_chunk_udf(&mut session_ctx);
+    // Register json_pack UDF (SQL-side JSON encoding; the etl generator's
+    // metadata/frontmatter serialization boundary)
+    skardi::util::json_pack::register_json_pack_udf(&mut session_ctx);
```
bakey: [P1] This registers `json_pack` unconditionally, but `chunk_parts` is only registered through the `#[cfg(feature = "chunking")]` call immediately above. `skardi-server` has `default = []`, and the dev Docker workflow passes no `FEATURES`, so the plain server command currently documented by skardi-cloud#232 will still fail at `chunk_parts`. Please close the artifact contract before treating the runtime gap as fixed: either ship the ETL path on a server built with `rag`/`chunking` and update the generator/deployment docs to invoke that exact artifact, or make the required capability part of the default server build.

crates/server/src/server.rs
```diff
@@ -193,6 +193,9 @@ pub async fn setup_app_state(config: ServerConfig) -> Result<AppState> {
     // Register chunk UDF (text-splitter wrapper for inline ingestion)
     #[cfg(feature = "chunking")]
     register_chunk_udf(&mut session_ctx);
+    // Register json_pack UDF (SQL-side JSON encoding; the etl generator's
+    // metadata/frontmatter serialization boundary)
+    skardi::util::json_pack::register_json_pack_udf(&mut session_ctx);
```
bakey: [P1] Please add a server-level regression test for this wiring, under `chunking`: build the planning and runtime contexts through `load_server_config`/`setup_app_state`, then assert that both `chunk_parts` and `json_pack` are present (ideally execute a minimal query or generated ingest job). The current UDF tests register functions directly, so they would remain green if either of these production registration calls were later removed; that is the exact failure mode this PR is intended to prevent.

crates/skardi/src/util/json_pack.rs
```diff
@@ -0,0 +1,351 @@
...
+            }
+            keys.push(key);
+        }
+
+        let mut builder = StringBuilder::new();
+        for row in 0..number_rows {
+            // Argument order IS object order — serde_json::Map preserves
+            // insertion order (default feature set), which is what makes
+            // the generator's output deterministic byte-for-byte.
+            let mut object = serde_json::Map::with_capacity(keys.len());
```
bakey, line 128: [P2] Argument-order preservation currently works only because another dependency (`bson`) transitively enables `serde_json/preserve_order`; Skardi itself declares plain `serde_json = "1.0"`. If that feature unification changes, `Map` falls back to key ordering and silently violates the load-bearing contract above. Please enable `preserve_order` explicitly on Skardi's dependency and change the test to use a non-lexicographic key sequence such as `z, a, m` (the current `id, title, updated` assertion passes under either ordering).

bakey (commented):
The core UDF implementations look sound: `chunk_parts` assigns stable per-input ordinals, and `json_pack` uses structured serialization with good NULL/error coverage. I am requesting changes because the runtime delivery contract is not closed yet. In particular, the default and dev server artifacts still omit the feature-gated `chunk_parts`, and there is no server-level regression test for the two production SessionContext registration paths. Please also make the JSON key-order dependency explicit and add user-facing documentation for both SQL UDFs (`docs/chunk.md` currently covers only `chunk`, and `json_pack` has no user guide). Docs/Test CI are green; the Codecov patch check is currently failing.

crates/server/src/config.rs
bakey: Fixed in 3c670cb — you're right that the gap wasn't closed while the delivering artifact stayed feature-gated. Went with your second option, scoped to the server binary: `chunking` is now in **skardi-server's default features**, so the plain `cargo build -p skardi-server` and the no-FEATURES Docker build both carry `chunk_parts` (the Cargo.toml comment records the rationale — the etl PRD's contract is that a STOCK server runs generated bundles). The skardi LIBRARY keeps `default = []` so downstream lib consumers are unaffected, and `--no-default-features` still opts a server build out. docs/chunk.md's build-flags note now leads with this. skardi-cloud#232's generated README needs no change — its five-step path already invokes the plain server command, which now works.

crates/server/src/server.rs
bakey: Added in 3c670cb: `production_contexts_register_the_etl_runtime_udfs` (crates/server/src/server.rs) goes through both REAL entry points — a pipeline whose SQL calls both UDFs must plan during `load_server_config` (schema inference runs against the planning context, so a dropped registration fails the load), and the same statement must plan AND execute through `setup_app_state`'s runtime context, with the `json_pack` output value asserted. Exactly the failure mode you named: the UDF unit tests register directly and stay green if either production call disappears; this one doesn't. It runs by default now that chunking is a default server feature.

crates/skardi/src/util/json_pack.rs
bakey: Fixed in 3c670cb — good catch, and slightly scarier than stated: without `preserve_order`, serde_json's `Map` is a BTreeMap, so losing the transitive feature wouldn't just reorder keys — it would silently break the etl golden bundles' byte-determinism the moment bson stops unifying it. `preserve_order` is now explicit on skardi's own serde_json dependency (with a LOAD-BEARING comment in Cargo.toml), the in-code comment no longer claims "default feature set", and the new `key_order_is_argument_order_not_lexicographic` test pins the contract with `z, a, m` — keys where insertion order disagrees with lexicographic order, so a BTreeMap fallback fails it (the existing `id, title, updated` assertion indeed passes under either ordering).

bakey:
All three findings addressed in 3c670cb (threads replied + resolved):

1. **Delivery contract closed**: `chunking` is now a DEFAULT feature of the skardi-server binary (library default stays `[]`), so the plain cargo build and the no-FEATURES Docker artifact run generated bundles out of the box — the option that keeps the etl PRD's "stock server" promise without changing library consumers.
2. **Wiring regression test**: `production_contexts_register_the_etl_runtime_udfs` exercises both real paths — pipeline planning through `load_server_config` and execution through `setup_app_state`'s runtime context — so removing either production registration call now fails a test that registering-directly unit tests cannot catch.
3. **Key order made explicit**: `preserve_order` on skardi's serde_json (was transitively supplied by bson), honest comments, and an ordering pin using non-lexicographic keys.

Also added the requested user docs: `chunk_parts` joined docs/chunk.md (with the plannable UNNEST spelling and the ordinality rationale) and `json_pack` got docs/json_pack.md (contract + injection-boundary rationale); README links both. Suites: 823/840 (skardi default/chunking) + 147 (server, wiring test included) green; fmt/clippy/docs clean. The new server test also adds coverage on the previously-uncovered registration lines, which should help the Codecov patch check.

crates/skardi/src/sources/providers/open_connector/source_pack.rs
```diff
@@ -227,6 +236,16 @@ mod tests {
...
     }
 
+    #[test]
+    fn packs_iterates_every_builtin_in_name_order() {
+        // The enumeration surface the etl generator's contract suite and
+        // `recipes` listing depend on: complete and deterministic — the
+        // backing map is unordered, so the sort here is load-bearing.
+        let registry = SourcePackRegistry::builtins().expect("embedded assets parse");
+        let names: Vec<&str> = registry.packs().map(|p| p.name).collect();
+        assert_eq!(names, vec!["github", "mock", "notion", "slack"]);
```
BtXin: **Critical — guaranteed post-merge test failure.**

This pin expects `["github", "mock", "notion", "slack"]`, but main (since 2291b19, #186) adds `feishu` to `builtins()`. This PR doesn't touch `builtins()`, so git merges cleanly — and then `packs_iterates_every_builtin_in_name_order` fails on the merged tree (5 packs, `feishu` sorting first). CI runs `nextest --all-features`, so main goes red.

Fix: rebase onto main and expect `["feishu", "github", "mock", "notion", "slack"]`. Ideally also split the assertion — sortedness + count-vs-`builtins()` + membership — so the sortedness pin survives future pack additions and only the completeness list needs touching.

crates/skardi/src/util/json_pack.rs
```diff
@@ -0,0 +1,381 @@
...
+        ScalarValue::Int64(v) => v.map_or(Value::Null, Value::from),
+        ScalarValue::UInt8(v) => v.map_or(Value::Null, Value::from),
+        ScalarValue::UInt16(v) => v.map_or(Value::Null, Value::from),
+        ScalarValue::UInt32(v) => v.map_or(Value::Null, Value::from),
+        ScalarValue::UInt64(v) => v.map_or(Value::Null, Value::from),
+        ScalarValue::Float32(v) => float_value(v.map(f64::from))?,
+        ScalarValue::Float64(v) => float_value(*v)?,
+        // Timestamps render as epoch milliseconds — a number, not a
+        // formatted string, so consumers never parse a locale.
+        ScalarValue::TimestampMillisecond(v, _) => v.map_or(Value::Null, Value::from),
```
BtXin, line 166: **Important — advertised paths with zero test coverage.**

The three non-millisecond timestamp conversions here (seconds ×1000 with overflow guard, micro/nano `div_euclid`) and the `List<Utf8>` path below (lines 198-233) are exercised by no test, yet `docs/json_pack.md` documents both (the `tags` example, "epoch milliseconds"). These are exactly the conversions with sign/rounding/overflow subtleties.

Suggest SQL-level tests: a `List<Utf8>` column (including a null element and a null list), TimestampSecond + TimestampNanosecond, and the unsupported-nested-type error path.

crates/skardi/src/util/json_pack.rs
```diff
@@ -0,0 +1,381 @@
...
+            })
+        })?,
+        ScalarValue::TimestampMicrosecond(v, _) => {
+            v.map_or(Value::Null, |us| Value::from(us.div_euclid(1000)))
+        }
+        ScalarValue::TimestampNanosecond(v, _) => {
+            v.map_or(Value::Null, |ns| Value::from(ns.div_euclid(1_000_000)))
+        }
+        other => {
+            return Err(DataFusionError::Execution(format!(
```
BtXin, line 193: **Important — `List<Utf8>` literals are rejected, contradicting the documented contract.**

`docs/json_pack.md` says a value may be "column or literal" including `List<Utf8>`, but only the `ColumnarValue::Array` path handles lists; `scalar_to_value` has no `ScalarValue::List` arm, so it falls through to this error. Concretely, `json_pack('tags', make_array('a','b'))` const-folds `make_array` to a `ScalarValue::List` and fails with "unsupported value type List(...)".

Either handle `ScalarValue::List` with Utf8 values here, or narrow the doc to "List<Utf8> columns".

crates/skardi/src/util/json_pack.rs
```diff
@@ -0,0 +1,381 @@
...
+        if args.is_empty() || !args.len().is_multiple_of(2) {
+            return Err(DataFusionError::Execution(format!(
+                "json_pack expects a non-empty, even number of arguments \
+                 (key, value, ...); got {}",
+                args.len()
+            )));
+        }
+
+        // Keys: non-null Utf8 literals, unique, in argument order.
+        let mut keys: Vec<String> = Vec::with_capacity(args.len() / 2);
```
BtXin, line 92: **Important — `Utf8View` unhandled, and this codebase already knows view types occur.**

`open_connector/filters.rs:164-170` explicitly handles `ScalarValue::Utf8View` ("DataFusion 52 can carry string literals as view types"), but key matching here accepts only `Utf8`/`LargeUtf8`, and `scalar_to_value` has no `Utf8View` arm — a view-typed string fails with "unsupported value type Utf8View" instead of encoding.

(Chunking's `read_text_column` has the same gap — only downcasts to `StringArray` — but that's pre-existing from `chunk` and fine as a tracked follow-up. This file is new code and the fix is a few match arms.)

crates/server/src/config.rs
BtXin: **Minor — style rule**: project guidelines require `use` imports at the top of the file, never full crate paths inline. Same in `server.rs:198`. Both files already import a dozen `skardi::` items with `use`.

crates/skardi/src/model/chunking/mod.rs
```diff
@@ -169,57 +194,212 @@ impl ScalarUDFImpl for ChunkingUDF {
...
+
+    fn return_type(&self, _arg_types: &[DataType]) -> DfResult<DataType> {
+        Ok(DataType::List(Arc::new(Field::new(
+            "item",
+            DataType::Struct(chunk_part_fields()),
+            true,
+        ))))
+    }
+
+    fn invoke_with_args(&self, args: ScalarFunctionArgs) -> DfResult<ColumnarValue> {
```
BtXin, line 231: **Minor — ~55 lines duplicated with `chunk`'s argument parsing** (mod.rs:143-173): identical except the UDF name string (arity check, mode/size/overlap parsing, `size == 0`, `overlap >= size`). Since the helpers already take a `udf: &str` prefix, a single `parse_chunk_args(udf, &args) -> (mode, texts, size, overlap)` would remove the duplication and guarantee the two UDFs' argument contracts can't drift.

BtXin (changes requested):
# Review: chunk_parts + json_pack runtime UDFs

**Verdict: request changes — sound implementation, but merging as-is guarantees a red main** (see the inline comment on `source_pack.rs`), and two advertised `json_pack` capabilities ship untested or broken.

**Verification performed** at the PR head (`3c670cb`) in an isolated worktree: `json_pack` tests (4 passed), chunking tests incl. `chunk_parts` and the UNNEST plannability pin (17 passed), the packs pin test (1 passed), and the server wiring test `production_contexts_register_the_etl_runtime_udfs` (1 passed). Clippy surfaced errors only in files this PR never touches (pre-existing). Every changed file was read in full, including the pre-existing `chunk` UDF and both server registration sites.

## Strengths

- **Ordinal-in-UDF design is correct and well-reasoned** (`chunking/mod.rs:311-346`): one deterministic pass per value — genuinely the only plan-stable option on DataFusion 52 (`UNNEST … WITH ORDINALITY` is unplannable, apache/datafusion#11419), and the module doc records why.
- **The plannability pin test pins what it claims** (`mod.rs:937-1001`): executes the exact `UNNEST(chunk_parts(...)) AS part` spelling end to end, asserting per-source-row ordinal restart and overlap-0 reassembly — precisely what generated ingest SQL depends on.
- **Single source of truth for the struct schema**: `chunk_part_fields()` feeds both `return_type` and `StructBuilder` — no declared-vs-built Arrow drift possible.
- **`json_pack`'s injection test is genuinely adversarial** (`json_pack.rs:290-327`): quotes, backslashes, control chars, a NUL byte, non-ASCII, round-tripped byte-exact through a real parser.
- **The `preserve_order` hardening is real, not cargo-cult**: insertion order previously held only via transitive feature unification through `bson`; the `z, a, m` key-order test is deliberately chosen to fail under a BTreeMap fallback.
- **The server wiring test closes a real gap** (`server.rs:463-521`): exercises both production registration sites; unit tests alone would stay green if either production call were deleted. The runtime `SessionContext` is the same Arc handed to `JobExecutor`, so ETL jobs see both UDFs.
- **Error handling**: no raw `.unwrap()` in production code; the two `.expect()`s document true invariants; i32 ordinal overflow and timestamp→ms overflow both guarded.
- **Good strictness decisions**: duplicate keys error, non-finite floats refuse, keys must be non-null literals so object shape is never data-driven.
- **No generator/etl coupling**; docs are accurate against the code.

## Minor (not anchored inline)

7. **Scalar-text broadcast relies on constant folding** — `read_text_column` (`chunking/mod.rs:394-423`, pre-existing) returns a length-1 vec for scalar text and never consults `args.number_rows` (unlike `json_pack`, which does this correctly). Pre-existing `chunk` behavior inherited by `chunk_parts`; worth a comment or explicit broadcast.
8. **No overlap > 0 assertion for `chunk_parts`** — both reassembly tests use overlap 0 while the docs claim identical overlap semantics to `chunk`. A small overlap test asserting ordinals stay dense would pin it.
9. **`packs()` exposes the `mock` pack** in enumeration — pre-existing decision, but any user-facing coverage listing iterating `packs()` will now display it; consider filtering or flagging at the enumeration surface.
10. **Test-count nit**: the PR claims "839 lib tests"; the lib harness at the PR head reports 1042. Likely a counting-scope difference (nextest/workspace) — informational only.

## Recommendations

1. Rebase onto main and fix the packs pin test — the only merge blocker.
2. In the same pass, add the `Utf8View` arms and the `List<Utf8>`/timestamp tests — small, and they land naturally with the rebase.
3. Convert the two inline `skardi::util::json_pack::...` paths to `use` imports — trivial, and it's a written project rule.
4. File a follow-up for chunking's text-column array-type coverage (StringView/LargeUtf8/dictionary) rather than expanding this PR.

crates/skardi/src/sources/providers/open_connector/source_pack.rs
bakey: Fixed in 1f94ef4 — you're right this was a guaranteed post-merge red. Absorbed main (merge commit, keeping review history) and took both halves of your suggestion: sortedness is now asserted independently of the roster (`names.windows(2).all(|w| w[0] < w[1])` — survives future pack additions untouched, and doubles as a no-duplicates check), and completeness is the explicit list including `feishu` — the one line a new pack must extend, with a comment saying why it must.

crates/skardi/src/util/json_pack.rs
bakey: Fixed in 1f94ef4 — fair hit: the docs promised paths no test had ever executed. Added SQL-level coverage for all of it: every timestamp granularity pinned by value (seconds ×1000, micro/nano `div_euclid` floor, and a pre-epoch negative instant keeping its sign), the seconds→millis overflow refusal, `List<Utf8>` with a null element AND a null list, the `make_array` literal spelling, and the non-Utf8-element refusal asserting the error names the TYPE (`Int64`) per the info-discipline rule. 9 json_pack tests now.

crates/skardi/src/util/json_pack.rs
bakey: Fixed in 1f94ef4 — took the first option (handle the literal) rather than narrowing the doc: `make_array` const-folding into `ScalarValue::List` is exactly the spelling an interactive user reaches for first, and the doc's "column or literal" is the right contract. Both spellings now funnel through one `utf8_list_value` conversion (extracted from the array path), so they cannot diverge; a NULL list scalar encodes as JSON null like the column case. Pinned by `utf8_lists_encode_as_json_string_arrays_in_both_spellings`.

crates/skardi/src/util/json_pack.rs
bakey: Fixed in 1f94ef4 — `Utf8View` accepted for keys and values in json_pack (scalar arms alongside `Utf8`/`LargeUtf8`, citing the filters.rs precedent), pinned by `utf8view_keys_and_values_encode_like_classic_strings` (a view-typed key, a view-typed literal value, and a view-CAST column carrying the adversarial string byte-exact). I also went one step past the comment and closed the chunking gap you flagged as a follow-up: `read_text_column` is shared by `chunk` and `chunk_parts`, so both now take view columns and scalars — it was a five-line fix in code this PR already touches, and a tracked follow-up would have outlived its cost.

crates/server/src/config.rs
bakey: Fixed in 1f94ef4 — `register_json_pack_udf` imported via `use` at the top of both files (after the last `skardi::` import), call sites bare.

crates/skardi/src/model/chunking/mod.rs
bakey: Fixed in 1f94ef4 — extracted `parse_chunk_args(udf, &args) -> (mode, texts, size, overlap)`: arity, literal rules, `size > 0`, `overlap < size`, and the text read in one place, with `udf` naming the caller in every diagnostic exactly as the existing helpers do. Both invoke bodies shrink to one call + the mode dispatch, and the two UDFs' argument contracts can no longer drift — the existing per-UDF error tests keep both prefixes pinned.

bakey:
Round-2 findings all addressed in 1f94ef4; each thread has its fix note and is resolved. The critical one (the packs roster going red on the merged tree after #186 landed feishu) is fixed by absorbing main and splitting the assertion — sortedness now survives future pack additions, completeness stays an explicit roster. json_pack gains Utf8View (keys + values) and const-folded `make_array` list-literal support with both spellings funneled through one conversion, plus value-pinned coverage for every timestamp granularity, list nulls, the overflow refusal, and the type-named error path. The ~55 duplicated argument-parsing lines in chunking collapse into one `parse_chunk_args` shared by both UDFs (which also closes the Utf8View gap in the shared text reader — took it now rather than as a tracked follow-up since the helper was already in hand). Verification: 872 lib (chunking) / 855 (default) / 166 server tests green, fmt clean, no new clippy warnings, docs build clean under -D warnings.

BtXin (approved):

PR #198, opened 2026-08-07
feat(sources): Discord source pack (milestone 5.5) — keyset pagination engine support, live-verified

bakey:
## What

Milestone 5.5: the **Discord source pack** — the current OAuth user's guilds, external-account connections, and the public Nitro sticker-pack catalog as SQL tables — plus the engine extension it genuinely needs: `PaginationStrategy::Keyset`. Live-verified end to end against a real Discord account through skardi-server on 2026-08-07.

## Scope: what this provider can and cannot see

Open Connector's `discord` provider is the OAuth **user-identity** surface — its own `get_user` executor rejects any id but `@me`. Channels, messages, and guild members are **bot-token** surface the provider does not carry: those tables cannot exist in this pack by provider scope, not by deferral. `entitlements` IS a deferral: the upstream executor exposes none of the endpoint's documented `before`/`after`/`limit` inputs, making it first-page-only — filed as [oomol-lab/open-connector#283](https://github.com/oomol-lab/open-connector/issues/283) and linked from the pack doc; the table joins when the executor grows the inputs.

## Per module

**`pagination.rs` — `PaginationStrategy::Keyset`** (new). Discord's list executors pass rows through raw with NO pagination envelope; the next request's cursor is a field of the previous page's LAST ROW (`after` = last guild id). Semantics, each with a targeted test:
- **Only an EMPTY page terminates.** Deliberately NOT short-page termination: providers silently clamp page sizes (the Feishu live pass caught a declared 100 clamped to 50 on the wire), and under a clamp every page is "short" — a short-page rule would end a clamped scan after page 1 and read as complete, the silent-truncation class this engine treats as the worst failure. The cost is one extra empty request per scan (the standard keyset tax).
- Missing or non-string cursor on a non-empty page is **typed drift** (`PaginationCursorInvalid` with path/page/JSON-kind), never a quiet stop.
- A repeated cursor fails **with identity only** — not `PaginationLoop`, whose Display quotes the token: for Keyset the token is ROW data (a field a pack can point anywhere), and row values never appear in errors. A test asserts the value is absent from the rendered error.
- `row_cursor_field` takes plain or dotted row-relative paths, validated at construction; `page_size` is a throughput knob only.
- `advance()` gained a `last_row: Option<&Value>` parameter; exec.rs passes the raw row array's last element **before** LIMIT truncation/projection, so the cursor is always the wire page's true last row. Existing strategies never consult it — no behavioral change (all pre-existing call sites updated mechanically).

**`packs/loader.rs`**: `keyset` and explicit `single_page` YAML spellings. Keyset load-time validation: non-empty input names (an empty name would reach the wire as a literal `""` key and fail late as a misdiagnosed loop), cursor/page-size input collision, positive page size — both sides tested.

**`packs/discord.yaml` + `packs/discord.rs`**: 3 tables — `guilds` (keyset `after`/`limit` 200, `with_counts: true` pinned so the `approximate_*` columns are not a per-scan coin flip), `connections` (single_page; the API does not paginate), `sticker_packs` (single_page; one public catalog). `connection_type` renames the SQL-keyword wire key `type`. All three output schemas fingerprint-pinned (BLAKE3) from live captures; the coverage-gap test pins that the loose item schemas leave every column uncovered — real rows are the only column truth, a reviewed fact rather than an implicit one.

## Phase-4 real-data evidence (2026-08-07, real account, local gateway)

- **Registration through LIVE discovery** passed the fingerprint gate; all three captured contracts byte-identical to discovery at verification time.
- **guilds**: 6 real rows; every mapped column extracted a non-NULL value (banner 1/6 non-null — nullable proven both ways). Real keyset walk at `limit: 2`: 3 full pages + the empty terminator, 6/6 unique ids, no boundary drop, ascending-snowflake ordering confirmed on the wire.
- **connections**: 1 real linked-account row; all nine wire keys present and mapped; the `connection_type` rename extracts; `revoked` genuinely ABSENT on a non-revoked row (its non-NULL arm rides a synthetic fixture row — capturing it live would mean revoking a real account link).
- **sticker_packs**: 14 rows, all seven columns non-NULL, `stickers` opaque JSON queryable.
- **The live pass caught a contract defect no mock could** — exactly the class phase 4 exists for: the gateway calls the UNVERSIONED `discord.com/api`, which Discord serves as its legacy default version — there `permissions` is a truncated NUMBER and the full bitfield string lives in `permissions_new`. The draft mapped `permissions` as utf8 and failed conversion on every real row. The column now maps `permissions_new` (keeping the natural name; the legacy number deliberately unmapped), and the version-coupled risk — a future `/api/v10` pin removes that key and sends the column always-NULL past every fingerprint — is documented in the pack doc with an upstream ask pending to pin the version.
- **Operational**: Discord 429s rapid successive calls to `/users/@me/guilds`; the gateway surfaces them loudly (a scan fails, never silently stops). At 200-row pages a scan stays comfortably clear.
- **Fixtures are redacted live captures** (guilds, sticker_packs, connections row 1): real key sets and spellings, synthetic ids/names/hashes/permission bits, lists truncated. A mechanical allowlist tripwire test audits every string leaf of every person-linked fixture — a real guild name or id cannot satisfy any allowlist arm. sticker_packs stays verbatim by design (public catalog) with a pack_id self-consistency assertion.

## Verification

- 307 tests `cargo test -p skardi --lib sources::providers::open_connector` (19 new: 8 keyset engine, 1 loader, 10 pack-scoped `… packs::discord`)
- 864 full library suite (`cargo test -p skardi --lib`); fmt + clippy clean on touched files
- Pack e2e pins exact wire key sets per request (sorted-keys equality, absence included — page 1 carries no cursor), declared constants by VALUE (`limit: 200`, `with_counts: true`), row identity across the page boundary (the exact 201-id sequence), and both sides of the fingerprint gate (drift refuses registration; the pass side runs against the captured contracts in every e2e).

## Deferrals

- `entitlements` — gated on [oomol-lab/open-connector#283](https://github.com/oomol-lab/open-connector/issues/283).
- Upstream API-version pin ask (the `permissions`/`permissions_new` drift risk) — issue pending; the pack doc records the risk and will link it.


bakey:
**Review round 1 — all three findings verified and fixed in eb33172.**

1. **[P2] Keyset error diagnostics were wrong or self-contradictory** — confirmed: everything funneled through `PaginationCursorInvalid`, whose fixed `"is {found}, not a string"` tail rendered an empty-string cursor as *"is a string, not a string"* and misreported a repeated (string-typed) cursor as a type error. Fixed with two new variants: `PaginationKeysetCursorInvalid { path, page, reason }` with per-case wording (absent / *is an empty string* / *is a number, not a string* / caller-bug), and `PaginationKeysetLoop { path, page }`, which names the loop and deliberately withholds the repeated value (row data never appears in errors). Tests now assert the **full rendered message** (`to_string()`), not variant + fragment, including a previously-missing dedicated empty-string case and value-absence assertions.

2. **[P2] Missing per-table/UDTF coverage** — confirmed against the #177/#186 standard. Added: `sticker_packs`' own gateway e2e (own row path + action, exact **empty** input key set, row identity in wire order); `udtf_parity_for_sticker_packs` pinning `open_connector_query('saas', 'discord.sticker_packs', '{}')` to the bound table's schema and formatted rows (test setup now registers UDTFs like the sibling packs); `empty_pages_preserve_every_table_schema` for all three tables — the case connections proved live, since the first scan of a link-less account returned zero rows.

3. **[P3] Doc drift** — all three confirmed and fixed: discord.yaml's header no longer says DRAFT (the live pass is done; per-column annotations record what it settled vs. what stays doc-derived); the request-count math beside the rate-limit warning is now exact (`ceil(n / 200) + 1` — the `+1` is keyset's terminating empty page); the binding example gains `source_pack_version: 1` and an `open_connector_query` example, matching the other pack docs.

Verification after this round: **311** open_connector tests (13 pack-scoped `packs::discord`), **868** full library suite; fmt + clippy clean on touched files.

crates/skardi/src/sources/providers/open_connector/packs/discord.rs
```diff
@@ -0,0 +1,742 @@
...
+        // guild name or id has no way to satisfy any arm. sticker_packs
+        // is deliberately NOT here: it is a public catalog kept verbatim.
+        fn audit(name: &str, value: &Value) {
+            match value {
+                Value::String(s) => {
+                    let allowed = s.starts_with("10000000000000000")
+                        || s.starts_with("Redacted ")
+                        || s == "0123456789abcdef0123456789abcdef"
+                        || s == "fedcba9876543210fedcba9876543210"
+                        || s.bytes().all(|b| b.is_ascii_digit())
```
BtXin: **The redaction tripwire cannot enforce the guarantee it states.**

This all-digits arm exists for the `permissions_new` bitfield strings, but it **subsumes the `s.starts_with("10000000000000000")` arm entirely**: every Discord snowflake is a digit-only string, so a real guild id (`"81384788765712384"`), a real connection id, or a real sku id all satisfy the allowlist unchanged. The test comment claims *"A real guild name or id has no way to satisfy any arm"*, and the PR body repeats it — for ids, the highest-risk person-linked class in these fixtures, that is not true. The ALL_CAPS arm has the same hole for an all-caps guild name, and the empty string passes the digit arm vacuously.

Concrete failure mode: a future fixture re-capture that lands un-redacted real ids passes CI silently. Today's fixtures are clean — the tripwire just isn't the thing keeping them that way.

The fix is cheap: thread the key name through `audit` and scope the permissive arms, with a default-deny tail so a future fixture key can't slip past on shape coincidence.

```rust
fn audit(name: &str, key: &str, value: &Value) {
    match value {
        Value::String(s) => {
            let digits = !s.is_empty() && s.bytes().all(|b| b.is_ascii_digit());
            let allowed = match key {
                // Snowflakes MUST carry the synthetic prefix — a real id is
                // digit-only too, so digit-shape alone proves nothing.
                "id" | "pack_id" | "sku_id" => s.starts_with("10000000000000000"),
                // Bitfields are the only place a bare digit string is legal.
                "permissions" | "permissions_new" => digits,
                "name" => s.starts_with("Redacted "),
                "type" => ["github", "steam"].contains(&s.as_str()),
                "icon" | "banner" => SYNTHETIC_HASHES.contains(&s.as_str()),
                // features[] items arrive under the parent key.
                "features" => !s.is_empty()
                    && s.bytes().all(|b| b.is_ascii_uppercase() || b == b'_'),
                _ => false,
            };
            assert!(allowed, "{name}: {key} = {s:?} is not on the redaction allowlist");
        }
        Value::Array(items) => items.iter().for_each(|v| audit(name, key, v)),
        Value::Object(map) => map.iter().for_each(|(k, v)| audit(name, k, v)),
        _ => {}
    }
}
```

crates/skardi/src/sources/providers/open_connector/packs/discord.yaml
```diff
@@ -0,0 +1,117 @@
...
+      - { name: banner, path: banner, type: utf8, nullable: true }
+      - { name: owner, path: owner, type: boolean, nullable: true }
+      # LIVE-PASS CORRECTION: this endpoint serves TWO permission fields —
+      # legacy `permissions` is a NUMBER (truncated to the bits that fit),
+      # and `permissions_new` is the full bitfield as a decimal string
+      # (it exceeds JSON's safe-integer range). The draft mapped
+      # `permissions` as utf8, which fails conversion on every real row.
+      # The column maps the authoritative string; the legacy number is
+      # deliberately unmapped. Decoding bits stays query-side work.
+      - { name: permissions, path: permissions_new, type: utf8, nullable: true }
```
BtXin: **`nullable: true` makes the one drift you flagged as undetectable actually undetectable — `nullable: false` would catch it.**

The pack doc says of a future `/api/v10` pin: *"this column goes always-NULL, and the loose item schemas mean no fingerprint pin can catch the move."* True of the fingerprint gate — but not of the converter: `json_to_arrow.rs` fails a **non-nullable** column with `missing key` and full row identity (column/path/page/row), which is exactly the loud failure this engine prefers everywhere else.

Declaring `permissions` as `nullable: false` converts the single documented silent-drift path in this pack into a hard `ConversionFailed`. The live evidence supports it — 6/6 real rows carried `permissions_new` non-null, and the legacy API attaches it to every guild object — and it costs nothing today.

Worth taking *because* you've already identified this as the drift nothing else can see. (Same argument applies weakly to `owner`, which `/users/@me/guilds` always sends; your call.)

crates/skardi/src/sources/providers/open_connector/packs/loader.rs
```diff
@@ -275,6 +280,32 @@ fn validate_table(table: &SourcePackTable) -> Result<(), String> {
...
+            page_size,
+            ..
+        } => {
+            // An empty param name would survive to scan time as a literal
+            // `""` input key — a strict gateway's 400 blamed on the
+            // request, or a lax gateway's page-1 refetch misdiagnosed as a
+            // provider loop. The loader's promise is one complete
+            // diagnostic pass at load. (row_cursor_field is validated by
+            // the strategy itself at construction.)
+            if cursor_param.is_empty() || page_size_param.is_empty() {
```
BtXin: **The non-empty-input-name rule is keyset-only, but its rationale isn't.**

The comment above — *"An empty param name would survive to scan time as a literal `\"\"` input key — a strict gateway's 400 blamed on the request, or a lax gateway's page-1 refetch misdiagnosed as a provider loop"* — applies verbatim to `Cursor` and `PageNumber`, and both arms above still admit `""` for `cursor_input` / `page_input` / `page_size_input`.

`pagination_params` is already collected just above the `match` (line ~232), so hoisting one emptiness check over it covers all three strategies in a couple of lines and keeps the loader's stated promise of *one complete diagnostic pass at load*. Alternatively, note in the comment why keyset alone needs it.

BtXin (commented):
Two independent review passes over `eb33172` converged on the same three findings, inlined below. None is a correctness bug against today's wire — the engine logic held up under both passes.

**Verified locally**: `cargo test -p skardi --lib` → 868 passed / 0 failed (311 in `sources::providers::open_connector`; 13 `packs::discord`, 9 `pagination::tests::keyset`). `cargo fmt --check` clean; `cargo clippy -p skardi --lib --all-features` produces zero warnings attributable to the touched files. No `.unwrap()` outside `#[cfg(test)]` in new production code. Branch history scanned — no un-redacted PII in any earlier commit's fixtures.

**Checked and clean:**
- `advance(&envelope, rows.len(), rows.last())` passes the *raw* wire row array's last element — `rows` comes from `row_path.rows()` before projection and before LIMIT truncation — and there is exactly one call site, so no projected/truncated row can leak in as a cursor.
- Keyset errors fire *before* `self.page += 1`, so every diagnostic names the page that actually produced the bad cursor; the full-rendered-message assertions pin the right number.
- Empty-page-only termination, typed drift on missing/empty/non-string cursor, value-free loop error, dotted row-relative traversal, construction-time path validation: no off-by-one or inverted condition.
- `pagination_params` correctly extended, so fixed-input and filter-input collisions are still caught for keyset. `PaginationStrategy` has no other exhaustive match sites needing the new variant.
- `max_pages`/`max_rows` defaults leave ample headroom for keyset's `+1` empty-page tax; the `ceil(n/200)+1` request math in the pack doc is exact at n = 0, 200, 201, 400.

The design calls are right, and the reasoning is load-bearing rather than decorative: empty-page-only termination is correct given the observed Feishu 100→50 clamp, and splitting `PaginationKeysetCursorInvalid`/`PaginationKeysetLoop` out of the shared variants is justified twice over (the "is a string, not a string" self-contradiction, and keeping row data out of error text).

---

### Standing items from the first pass (not correctness bugs, no diff anchor)

- **`discord.yaml` is missing from `builtin_assets_parse_and_validate`** (`packs/loader.rs`, ~line 722). Its docstring claims *"Every shipped asset parses AND passes the same structural checks binding performs"* — that invariant now has a hole. `SourcePackRegistry::builtins()` covers it redundantly, but the test that *names* the invariant no longer enforces it, and pack 7 will inherit the omission. One line.
- **No LIMIT-pushdown e2e for the keyset table.** `feishu`, `notion`, and `slack` each ship a `limit_stops_*_pagination_early` test; `packs::discord` has none. Keyset is where it matters most: `advance()` is skipped entirely when `limit_remaining == Some(0)`, and keyset is the only strategy that would otherwise issue a terminating empty request. Traced it — `SELECT id FROM saas.me.guilds LIMIT 1` correctly makes exactly one POST today. Worth pinning, given the route is aggressively rate-limited.
- **Keyset's exclusive-cursor assumption is undocumented.** The `Keyset` doc comment records the *ordering* assumption but not that the cursor input must be **exclusive**. An inclusive cursor re-serves the boundary row at the head of every page — duplicate rows, and `seen_tokens` never fires because it only inspects each page's *last* row. Discord's `after` is exclusive and the live walk confirmed no duplication, so this is a doc gap. If you want the guard rather than the sentence, it's ~3 lines: page N+1's *first* row cursor must not equal the token just consumed — strictly cheaper than the check already running, and it closes the duplication hazard symmetric to the looping one.
- **Stale comment** at `packs/discord.rs:304`: *"connections will be too once captured"* — connections *was* captured in ccd95af, and the header block ~130 lines above already says so.
- **PR description is stale**: body says 307 tests / 864 suite / "8 keyset engine, 10 pack-scoped"; actual on this tip is **311 / 868 / 9 keyset / 13 pack-scoped**. The in-repo spec entry has the right numbers. Worth syncing, since the body is the reviewer-facing evidence artifact.

crates/skardi/src/sources/providers/open_connector/packs/discord.rs
bakey: Fixed in f2e61f3, taken exactly as sketched: `audit` now threads the key through and every arm is key-scoped with a default-deny tail — `id` must carry the synthetic prefix (digit shape no longer proves anything), digit strings are legal only under `permissions`/`permissions_new` (the latter for real rows, the former because the type-mismatch fixture deliberately plants a string `"0"` there), `name` must carry `Redacted `, and a key the allowlist has never seen fails loudly.

You were right that the old comment's claim was false for ids — the highest-risk class was exactly the one the digit arm waved through, and the empty string passed vacuously. Since the finding was "the tripwire cannot enforce what it states", the fix also has to prove the new one can: four negative probes (a real un-prefixed snowflake, the empty string, an ALL_CAPS name, an unknown key) now assert via `catch_unwind` that the audit actually panics on each leak class a future re-capture could reintroduce.

crates/skardi/src/sources/providers/open_connector/packs/discord.yaml
bakey: Taken in f2e61f3 — `permissions` is now `nullable: false`, and the yaml comment, module doc, and pack doc all state the reasoning you gave: this is the pack's one drift no fingerprint can see, so the converter's `missing key` failure (with column/page/row identity) is the ONLY tripwire, and the live evidence (6/6 rows, legacy API attaches `permissions_new` to every guild object) says it costs nothing.

The change immediately proved itself: the keyset walk test's mock rows had been fabricated with `"permissions": "0"` and no `permissions_new` at all — the exact shape mistake the draft originally made against the live wire — and the non-nullable column failed them with precisely the loud `ConversionFailed { path: "$.permissions_new", found: "missing key" }` you predicted. The mock rows are now wire-faithful (legacy number + bitfield string).

`owner` I left nullable: `/users/@me/guilds` does always send it, but no drift path is identified for it, and if one ever appears the same converter failure names it the moment we tighten. Happy to take it too if you feel strongly.

crates/skardi/src/sources/providers/open_connector/packs/loader.rs
bakey: Fixed in f2e61f3 the way you suggested: the emptiness check is hoisted over the already-collected `pagination_params`, so one check covers Keyset, Cursor, and PageNumber (including Cursor's optional `page_size_input: ""`), and the rationale comment moved with it — the keyset arm now just points up. You're right that nothing about the rationale was keyset-specific; it was only there because keyset was the strategy this pack introduced.

The load-time test now pins all six empty-name spellings (keyset cursor/page-size, cursor cursor/page-size, page_number page/page-size), each asserting the `must be non-empty` diagnostic.

abbccdda (commented):
**Spec-conformance / end-to-end-reliability pass** (design spec `2026-07-11-open-connector-integration-design.md` + the pack review checklist). The engine extension is well-shaped: keyset terminates only on an empty page, every non-terminating failure is typed and value-free, load-time validation covers the new strategy's authoring mistakes, and the collision checks against fixed/filter/resource inputs pick up `Keyset` correctly. `exec.rs` hands `advance()` the raw row array's last element *before* projection and LIMIT truncation, so the cursor is genuinely the wire page's last row — that was the one place this could have gone quietly wrong, and it didn't.

What I'm flagging is mostly coverage-of-the-claims rather than defects in the code: the pack states two things it treats as load-bearing (the `permissions_new` drift tripwire, and the keyset request-count contract) that nothing in the suite pins, plus a doc-parity gap against the other pack guides. Details inline.

One cross-cutting item that has no line to hang on: **the PR description's verification numbers are stale**. It says 307 tests / 19 new / 864 full suite; the spec entry updated in the same branch (`2026-07-11-open-connector-integration-tasks.md`, commit `f2e61f3`) says 311 / 23 / 868, and counting the diff confirms 23 new (9 keyset engine + 1 loader + 13 `packs::discord`). The checklist's own rule — *"every test count in docs/spec matches a fresh cargo run"* — should apply to the PR body too, since it is what a reviewer reads first. The body also still describes the branch as of the pre-review-round commits (it does not mention the round-1/round-2 diagnostics, per-table e2e parity, or the key-scoped redaction tripwire).

crates/skardi/src/sources/providers/open_connector/packs/discord.yaml
```diff
@@ -0,0 +1,123 @@
...
+      # `permissions` as utf8, which fails conversion on every real row.
+      # The column maps the authoritative string; the legacy number is
+      # deliberately unmapped. Decoding bits stays query-side work.
+      # NON-nullable on purpose: the legacy API attaches permissions_new
+      # to every guild object (6/6 live rows), and this mapping is the
+      # pack's one version-coupled drift — if the gateway ever pins
+      # /api/v10 (no permissions_new), the loose item schemas mean no
+      # fingerprint can see the move, so the converter's missing-key
+      # failure (with column/page/row identity) is the ONLY tripwire.
+      - { name: permissions, path: permissions_new, type: utf8, nullable: false }
```
abbccdda, line 69: **The pack's single documented tripwire for its highest-risk drift has no test.**

This mapping is called out three times — here, in the module doc (`discord.rs:55-67`), and in the pack doc (`docs/open-connector-discord.md:93-102`) — as *version-coupled*: if the gateway ever pins `/api/v10`, `permissions_new` disappears, no fingerprint can see the move because the item schemas are loose, and (quoting the yaml) "the converter's missing-key failure ... is the ONLY tripwire".

Nothing in the suite exercises that tripwire. `guilds_type_mismatch.json` happens to omit `permissions_new`, but the assertion in `a_number_where_utf8_is_declared_fails_with_row_identity` pins `column == "id"`, so the arm that would actually fire on the v10 drift is never observed. The checklist is explicit that a deliberate gap needs "a module-doc rationale AND a guard test", and that "a gate whose failure arm no test exercises is dead code until proven otherwise".

Suggested: one test converting a wire-faithful guild row with `permissions_new` removed (everything else valid), asserting `ConversionFailed { column: "permissions", .. }` with page/row identity — and, ideally, that the offending row's other values do not appear in the message. That is ~10 lines and it turns the doc's claim into an enforced one.

crates/skardi/src/sources/providers/open_connector/packs/fixtures/discord/guilds_type_mismatch.json
```diff
@@ -0,0 +1,11 @@
+{
+  "guilds": [
+    {
+      "id": 100000000000000001,
```
abbccdda, line 4: **This fixture carries two independent defects, so the test that reads it can pass for the wrong reason.**

The row is both (a) `id` as a number where `utf8` is declared and (b) missing `permissions_new`, which the now-non-nullable `permissions` column requires. Which one surfaces depends on the converter's field iteration order; `a_number_where_utf8_is_declared_fails_with_row_identity` asserts `column == "id"` and is therefore coupled to the column order in `discord.yaml`. Reorder the columns (or fix `id` while leaving the fixture otherwise as-is) and the test either breaks for an unrelated reason or keeps passing while testing a different drift than its name claims.

Make the mismatch fixture valid in every respect except the one declared defect — add `"permissions_new": "0"` — and let the missing-key case live in its own test (see the comment on `discord.yaml:69`). One fixture, one defect.

crates/skardi/src/sources/providers/open_connector/packs/discord.rs
```diff
@@ -0,0 +1,783 @@
...
+                    .collect::<Vec<_>>()
+            })
+            .collect();
+        let expected: Vec<String> = (1..=201).map(|i| format!("g-{i:04}")).collect();
+        assert_eq!(ids, expected, "both pages scanned, boundary row intact");
+
+        let inputs = execute_inputs(&gateway);
+        // Three requests: two row pages plus the terminating empty page —
+        // the standard keyset tax.
+        assert_eq!(inputs.len(), 3, "two row pages plus the empty terminator");
```
abbccdda, line 641: **Two coverage gaps around the new strategy, both of which the sibling packs already close for their own strategies.**

1. **No `LIMIT` + keyset e2e.** `feishu.rs` has `limit_stops_cursor_pagination_early`; there is no keyset equivalent. This matters more here than it did there, because keyset is the strategy whose contract explicitly includes an extra terminating request: a `SELECT ... FROM saas.me.guilds LIMIT 5` must issue exactly **one** gateway request (the LIMIT-satisfied path in `exec.rs:559-566` sets `done` and skips `advance()` entirely, so the empty terminator is never fetched). Nothing pins that today, and it is exactly the interaction a future refactor of the done/advance ordering would break — silently, in the direction of extra provider traffic against an endpoint the doc says 429s aggressively.

2. **The keyset failure arms never reach the gate through the public entry point.** `PaginationKeysetCursorInvalid` and `PaginationKeysetLoop` are covered thoroughly at the `Pagination::advance` unit level (nice tests — the rendered-message assertions and the "value never appears" checks are the right shape), but no scan-level test serves a page whose last row lacks `id`, or a gateway that re-serves the same last row, and watches the *query* fail. The checklist calls this out specifically: "the failing input must reach the gate THROUGH THE PUBLIC ENTRY POINT, not by calling the guard function directly". The mock in this test already branches on `after`, so a second `MockGateway` that returns `Some("g-0200") => vec![guild_row("g-0200")]` gives you the loop arm end to end in a few lines.

crates/skardi/src/sources/providers/open_connector/pagination.rs
```diff
@@ -376,6 +439,81 @@ impl Pagination {
...
                 Ok(true)
             }
+            PaginationStrategy::Keyset {
+                row_cursor_field, ..
+            } => {
+                // ONLY an empty page ends the keyset walk. A short page is
+                // NOT termination: page-size clamping providers make every
+                // page short, and a short-page rule would read a clamped
+                // scan as complete after page 1 — silent truncation.
+                if rows_in_page == 0 {
```
abbccdda, line 482: **The empty terminator spends a `max_pages` unit, so a keyset table's real capacity is `max_pages - 1` full pages.**

`exec.rs:439` guards with `pagination.page() > max_pages` *before* the fetch. After N full pages the counter is N+1, so a collection that is exactly `max_pages × page_size` rows fails with `ScanBoundsExceeded { bound: "max_pages" }` on the terminator request rather than completing. Cursor and page-number strategies don't have this: their end-of-collection signal arrives inside page N.

This fails loudly, which is the right side of the tradeoff, so I'm not asking for a behavior change — but it is a real asymmetry between strategies that a reader of `max_pages: 100` would not predict, and the pack doc's `ceil(n / 200) + 1` note (`docs/open-connector-discord.md:105-108`) stops just short of saying that the `+1` counts against the same budget. Worth one sentence in this doc comment and one in the pack doc (and it is cheap to pin with a test: `max_pages: 1` over a single full page must fail rather than silently return page 1).

docs/open-connector-discord.md
```diff
@@ -0,0 +1,126 @@
...
+  `/api/v10` — where `permissions` *is* the string and
+  `permissions_new` does not exist — this mapping breaks, and the loose
+  item schemas mean no fingerprint pin can catch the move. The column
+  is therefore declared **non-nullable** (the legacy API attaches
+  `permissions_new` to every guild object; 6/6 live rows carried it):
+  the drift surfaces as a hard `ConversionFailed: missing key` with
+  full column/page/row identity instead of a silently always-NULL
+  column. Upstream issue pending (the gateway should pin an API
+  version); this doc links it once filed.
+- **Rate limits are tight**: rapid successive calls to
```
abbccdda: **Doc parity with the other pack guides, against the admission gate in the design spec.**

`open-connector-github.md`, `-slack.md`, `-notion.md`, and `-feishu.md` all carry a `## Rate limits and freshness` section; this guide folds rate limits into a design-note bullet and says nothing about **freshness or caching**. The checklist names cache-key contents as an operational surprise that must be documented ("cache-key contents (LIMIT is load-bearing)"), and the design spec's gate asks for "filter and limit behavior documented per table" — the table matrix documents filter pushdown as `—` (correct), but limit behavior appears nowhere.

Also missing versus the gate's "stable-table, built-in UDTF, and raw-action examples": there is no `open_connector_scan` example. (`-feishu.md` omits it too, so this is a standing gap rather than a regression — but the gate lists it, and `discord.list_my_guilds` is a clean one-liner to show.)

A short `## Rate limits and freshness` section mirroring feishu's — 429 behavior, `ceil(n/200) + 1` requests per scan and how that interacts with `max_pages`, live-by-default, and the LIMIT-in-cache-key rule — closes all of it.

crates/skardi/src/sources/providers/open_connector/packs/fixtures/discord/contracts/list_sticker_packs.json
```diff
@@ -0,0 +1,6 @@
+{
```
abbccdda, line 1: **Worth stating in the pack doc: for this table the fingerprint gate cannot refuse anything.**

The captured contract is a bare `{type: object, properties: {}, additionalProperties: true}` — a schema every possible output satisfies. The pin is therefore honest bookkeeping, not a gate: an upstream change to this action can only be caught by `fingerprint_schema` seeing a *different* declared schema, and `the_fingerprint_coverage_gap_is_every_column` passes here vacuously (`descend()` walks into `null` and every column reports uncovered because the traversal found nothing, not because the schema declared nothing about them).

That is consistent with what the yaml comment says ("even the row_path rests on the wire, not the contract"), and `drifted_contract_fails_registration_not_the_scan` covers the gate's fail side via `guilds` — so no change is strictly required. But the pack doc's status blurb currently says "registration through LIVE discovery passed the fingerprint gate" for all three tables, which reads stronger than it is for `sticker_packs`. One clause distinguishing "pinned and verified" from "pinnable but unfalsifiable" would keep the doc and the artifact telling the same story.

abbccdda (approved):
Approval upon addressing review comments!

crates/skardi/src/sources/providers/open_connector/packs/loader.rs
```diff
@@ -242,8 +242,22 @@ fn validate_table(table: &SourcePackTable) -> Result<(), String> {
         } => std::iter::once(cursor_param)
             .chain(page_size_param)
             .collect(),
+        PaginationStrategy::Keyset {
+            cursor_param,
+            page_size_param,
+            ..
+        } => vec![cursor_param, page_size_param],
```
abbccdda, line 249: **Pagination inputs are still never checked against declared resources.**

The comment above this block says the request-input namespace is shared by "resources, fixed inputs, filter inputs, and pagination parameters" and that `exec.rs` applies pagination **last** — but the validations below only cover *filter vs pagination* and *fixed-input vs pagination*. Nothing rejects a pagination param that collides with a declared resource, and the new `Keyset` arm inherits that gap.

Concrete scenario: a future pack declares `resources: { required: [after] }` (an entirely plausible input name for a resource-scoped list) together with `pagination: { strategy: keyset, cursor_input: after, ... }`. The pack loads clean. At scan time `exec.rs` builds `input` from the resource first (`exec.rs:451`), then `self.pagination.apply(&mut input)` (`exec.rs:460`) overwrites `after` from page 2 onward — the scan silently walks a different collection than the binding asked for, with no error anywhere. The same shape applies to `page_input`/`page_size_input` under the other strategies.

Since this function's stated promise is "one complete diagnostic pass at load", the resource-vs-pagination pair looks like the missing fourth check.

crates/skardi/src/sources/providers/open_connector/packs/discord.yaml
abbccdda, line 69: **`nullable: false` here makes an upstream API-version move take out the whole table, not just this column.**

The reasoning for mapping `permissions_new` is right, and the tripwire argument is fair — but the blast radius deserves a second look. `RowConverter` fails the *page*, not the cell: a missing non-nullable key returns `ConversionFailed { column: "permissions", … }` from `convert()` (`json_to_arrow.rs:259-261`), which aborts the scan before any batch is emitted.

Concrete scenario: the gateway pins `/api/v10` (or Discord finally retires the unversioned default it has already deprecated). `permissions_new` disappears. From that moment **every** query against `me.guilds` fails — including `SELECT id, name FROM me.guilds`, which never touches `permissions` and would otherwise be entirely unaffected, and including the projection-pushdown path (the converter runs before `batch.project`, `exec.rs:510` vs `516`). A pipeline that only reads guild names goes down with it.

That may still be the intended trade (loud > silently-NULL), but it is a strictly larger failure than "the column errors": the alternative — `nullable: true` plus the pinned fingerprint/doc note — degrades one column instead of removing the table. Worth stating explicitly in the pack doc which of the two was chosen and why, since the current note only argues against the always-NULL column, not for a table-wide outage.

docs/open-connector-discord.md
abbccdda: **The keyset tax and this rate-limit note point at each other — worth confirming how the gateway relays a provider 429.**

This bullet says rapid successive calls to `/users/@me/guilds` return 429 and that the gateway "surfaces it as a loud scan failure". Keyset makes *every* scan issue at least two back-to-back calls to exactly that route (the terminating empty page), and they go out with no deliberate spacing — the next `fetch` is issued as soon as the previous batch is polled.

Whether that is safe depends entirely on the relay shape:
- If the gateway returns **HTTP 429**, `client.rs` retries it under `RetryPolicy::NonIdempotent` (`allows_status_retry`, `client.rs:106-110`) honouring `Retry-After` — fine.
- If the gateway instead consumes Discord's 429 and returns a **failure envelope / 5xx** (which is what "surfaces as a loud scan failure" reads like, and is consistent with the pack declaring no `error_path` because "the executors consume Discord's error responses themselves"), then nothing retries it: `SELECT * FROM me.guilds` on an ordinary account with ≤200 guilds fails on the *terminator* request, after every row was already fetched.

The live pass (limit 2, 4 rapid requests) is decent evidence it is the first case, but the doc currently asserts the second. Please pin down which it is and make the bullet say so — as written, the pack documents a failure mode its own pagination strategy guarantees to hit twice per scan.

crates/skardi/src/sources/providers/open_connector/packs/discord.yaml
bakey: Fixed in b7c0fd5. The tripwire is now enforced exactly as suggested: `a_missing_permissions_new_key_fails_naming_the_permissions_column` converts a wire-faithful guild row (built from the e2e's `guild_row`, so it stays wire-faithful by construction) with only `permissions_new` removed, and asserts `ConversionFailed { column: "permissions", page: 3, row: 0 }` — plus that the row's other values (the id) never appear in the rendered message. The doc's three claims now have a test that IS the claim.

crates/skardi/src/sources/providers/open_connector/packs/fixtures/discord/guilds_type_mismatch.json
bakey: Fixed in b7c0fd5. The fixture is now single-defect: `"permissions_new": "0"` added (and the legacy `permissions` key set to a wire-faithful number), so the only declared defect is `id` as a number — `a_number_where_utf8_is_declared_fails_with_row_identity` can no longer pass for the wrong reason under a column reorder. The missing-key case lives in its own test built from the e2e's `guild_row` helper (see the thread on discord.yaml:69), so there is no second fixture to drift either.

crates/skardi/src/sources/providers/open_connector/packs/discord.rs
bakey: Both gaps fixed in b7c0fd5.

1. `limit_stops_keyset_pagination_after_one_request`: `SELECT id FROM saas.me.guilds LIMIT 5` against a gateway serving only full pages asserts exactly ONE execute request — the LIMIT-satisfied path sets `done` before `advance()` would prepare the next request, so neither the empty terminator nor a full walk ever goes out. This is now the pinned interaction you described.

2. Both failure arms now reach the gate through the public entry point: `a_repeated_last_row_cursor_fails_the_scan_as_a_loop` uses exactly the `Some("g-0200") => vec![guild_row("g-0200")]` mock you sketched and asserts the QUERY fails naming the cursor field with the row value absent from the message; `an_empty_cursor_field_fails_the_scan_with_identity` covers the invalid-cursor arm — deliberately with an empty STRING rather than a missing key, because a missing `id` trips the non-nullable converter before pagination ever sees the row (the test's comment records that ordering, so the choice reads as reasoned rather than accidental).

crates/skardi/src/sources/providers/open_connector/pagination.rs
bakey: Fixed in b7c0fd5, on all three of your suggested surfaces: the keyset strategy doc now states the terminator spends a `max_pages` unit and that a keyset table's practical capacity is `max_pages − 1` full pages (naming the asymmetry against cursor/page-number, whose end signal arrives inside page N); the pack doc's new Rate limits section says the `+1` counts against the same budget and what to do about it; and `max_pages_budget_includes_the_keyset_terminator` pins the loud side — `max_pages: 1` over exactly one full page fails `ScanBoundsExceeded` on the terminator instead of completing. Agreed no behavior change: loud-on-the-boundary is the right side, it just needed to be a documented, pinned choice.

docs/open-connector-discord.md
bakey: Fixed in b7c0fd5. The guide now carries a `## Rate limits and freshness` section mirroring the siblings: the verified 429 behavior (see the thread on line 103 — it turned out to be retried-then-loud, not loud-first), the `ceil(n/200) + 1` request cost and its interaction with the `max_pages` budget, live-by-default, and the LIMIT-in-cache-key rule (with the pointer that a satisfied LIMIT also skips the terminator, pinned by the new e2e). The gate's raw-action example is in too: an `open_connector_scan('saas', 'discord.list_my_guilds', …, '$.guilds')` one-liner right after the binding examples, linking the allowlist requirement.

crates/skardi/src/sources/providers/open_connector/packs/fixtures/discord/contracts/list_sticker_packs.json
bakey: Fixed in b7c0fd5, doc-side as suggested: the status blurb now flags `sticker_packs` inline ("for `sticker_packs` that pin is honest bookkeeping, not a gate — see Design notes"), and a new design note spells out why — the captured contract is `{type: object, properties: {}, additionalProperties: true}`, a schema every output satisfies, so the pin can only catch the gateway DECLARING a different schema, never rows changing shape under the same declaration; the row shape rests on the live pass and the conversion tests, and the gate's refusal arm is exercised via `guilds`. "Pinned and verified" vs "pinnable but unfalsifiable" is now the artifact's own distinction.

crates/skardi/src/sources/providers/open_connector/packs/loader.rs
bakey: Fixed in b7c0fd5 — you are right that this was the missing fourth pair of the shared namespace. The loader now rejects any pagination input naming a declared resource (required or optional), with the error explaining the mechanism ("pagination is applied last and would overwrite the binding's value from page 2 onward"). The rejection table gains both shapes from your comment: the keyset `after`-as-required-resource case verbatim, and a page_number twin (`perPage` as an optional resource) so the check demonstrably covers every strategy rather than the new arm only.

crates/skardi/src/sources/providers/open_connector/packs/discord.yaml
bakey: Addressed in b7c0fd5 by making the choice explicit rather than changing it. The pack doc's permissions bullet now states the blast radius in your terms: the converter fails the PAGE, so on the v10 drift every `me.guilds` query goes down — including `SELECT id, name` and the projection-pushdown path — not just the column. And it states why that trade was picked: the alternative (nullable: true + doc note) fails only when someone finally READS `permissions`, possibly weeks after the move and far from the cause, while the table-wide outage points at the cause on the first query after it. With the tripwire now enforced by its own test (the thread on line 69), the loud path also carries the precise diagnostic the doc promises. If operational experience later says one column's availability shouldn't gate the table, flipping to nullable is a one-line change with the reasoning already recorded on both sides.

docs/open-connector-discord.md
bakey: Pinned down in b7c0fd5 — and you were right to push: the doc asserted the wrong shape. Verified in the gateway's own source rather than guessed: the Discord executor maps a provider 429 to `ProviderRequestError(429)` → execution error code `rate_limited` (`provider-runtime.ts::toProviderExecutionError`), and the gateway's runtime API maps `rate_limited` → **HTTP 429** — pinned by the gateway's own `runtime-api` test table (`["rate_limited", 429]`). So it is your first case: skardi's client retries HTTP 429 with bounded backoff on every call class (for POST execute it is the one retryable status under `RetryPolicy::NonIdempotent`), the keyset back-to-back pair rides the retry budget, and only retry exhaustion surfaces as a loud failure. The bullet now says exactly that, with the "surfaces as a loud scan failure" claim scoped to exhaustion. (The no-`error_path` note stays true — that mechanism is about IN-BAND provider errors inside a 200 envelope, which these executors do consume; the 429 path is HTTP-status-level and never reaches it.)

PR #199, opened 2026-08-08
docs(superpowers): graph engine bypass design — read-only Cypher as SQL tables (Neo4j/Kuzu, AGE candidate) with planning-time schemas and agent-first discovery

bakey:
## Summary

Design proposal for **read-only graph/Cypher support** in Skardi via a dedicated graph engine bypass, rather than implementing a Cypher parser or planner inside DataFusion.

The design follows the same bypass pattern already proven by the Open Connector integration: Skardi keeps SQL as its primary language, exposes graph queries through a table-returning UDTF, and delegates graph storage/traversal/optimization to mature external engines (Neo4j or Kuzu). A single SQL query can `JOIN` graph results with Postgres, Lance, CSV, or Open Connector tables.

This revision explicitly adds an **Agent and LLM interaction** section, because Skardi's primary consumer is an agent runtime backed by a large language model, not just a human SQL user.

## What this PR contains

- A full design document at `docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md`
- SQL surface: `cypher_query(connection, cypher, params, columns)` UDTF, implemented in `crates/skardi/src/sources/providers/graph/udtf.rs` as a DataFusion `TableFunctionImpl` that delegates Cypher execution to a backend `GraphClient`
- Optional catalog surface: `type: graph` data source with pre-declared Cypher views from context YAML
- Backend abstraction: `GraphClient` trait (`graph/client.rs`) hiding Neo4j Bolt vs Kuzu details
- Result flattening: Cypher nodes / relationships / paths mapped to stable Arrow `STRUCT` columns
- Security model: read-only by construction, parameterized queries only, credentials via environment variables
- **Agent workflow section**: discovery → Cypher generation by LLM → execution through Skardi → JSON consumption
- `graph_schema(connection)` introspection UDTF so agents can discover labels, relationship types, and properties before generating Cypher
- Agent-friendly error messages with actionable backend error codes and bounded message snippets
- Milestones: Neo4j read-only UDTF → Kuzu backend → YAML catalog views → future write path

## Why not native Cypher in DataFusion

DataFusion is a relational SQL engine. Building a full Cypher frontend would require parsing a new query language, planning graph traversals, and implementing graph-aware physical operators — months of engine work. The bypass approach reuses proven external graph engines and keeps Skardi focused on its federated SQL mission.

## Agent / LLM interaction model

The design is explicit about the agent workflow:

1. **Discovery** — the agent queries Skardi metadata (`/data_source` and `graph_schema`) to learn what graph sources and graph content exist.
2. **Generation** — the agent asks its LLM to translate natural-language intent into Cypher, using the discovered labels, relationship types, and stable `STRUCT` result shapes.
3. **Execution** — the agent sends Cypher to `cypher_query` or a YAML-bound catalog view.
4. **Consumption** — Skardi returns JSON rows; the agent passes them back to the LLM or answers the user directly.

Read-only enforcement engine-side is therefore a safety feature: an agent-generated mutating Cypher is rejected with a clear, actionable error, and the agent can fall back to explaining that writes are not yet supported.

## Key design decisions

- **UDTF-first**: `cypher_query` is the primary ad-hoc interface; YAML views provide stable catalog tables.
- **Backend-agnostic SQL surface**: same `cypher_query` works for Neo4j and Kuzu through a `GraphClient` trait.
- **Deterministic Arrow schema**: nodes become `STRUCT<id, labels, properties>`; relationships become `STRUCT<id, start_id, end_id, type, properties>`; paths become `List<STRUCT>` or flattened via `UNWIND`.
- **Planning-time-stable schema, no probe**: the optional 4th `columns` argument declares output columns and Arrow types (`'{"user_name": "Utf8"}'`); without it, every row is one `record: Utf8` JSON-text column. No backend call happens at planning, empty results keep the schema, and dynamically-typed properties cannot break a fixed column mid-scan.
- **`graph_schema` introspection**: gives agents the metadata needed to write correct Cypher without hand-crafted prompts.
- **No credentials in Skardi**: `username_env` / `password_env` reference environment variables; tokens never appear in YAML, logs, or errors.
- **Read-only guard**: string-level rejection of mutating Cypher clauses (`CREATE`, `SET`, `DELETE`, `MERGE`) before queries reach the backend.

## Documentation structure

The doc follows the repo's established design-spec format (same shape as the Open Connector and RSS designs):

- Status / Date / Branch
- Summary
- Motivation
- Research Findings
- Goals / Non-goals
- Decisions
- Alternatives Considered
- High-level Architecture (mermaid)
- Detailed Design (module layout, YAML config, UDTF signature, error handling, testing)
- **Agent and LLM interaction** (discovery, generation, execution, consumption)
- Milestones
- Risks and Open Questions
- References

## Review focus

- Is the UDTF bypass the right level of abstraction for graph support?
- Should YAML catalog views be part of milestone 1 or deferred?
- ~~Is execution-time schema probing acceptable~~ Resolved: probes are rejected (planning-time network I/O, empty-result schema drift, first-row inference unsound for dynamically-typed properties). The design now declares columns via the 4th argument or falls back to one `record` JSON-text column. Open: is the single-column JSON fallback ergonomic enough, or should it require column declarations?
- Is `graph_schema` sufficient for agent discovery, or do we also need per-label / per-relationship-type property statistics?
- Should the read-only guard reject `CALL` (procedures) as well, since some procedures are mutating?
- Neo4j Bolt vs HTTP — which protocol should the driver prefer?
- Should Kuzu be embedded in milestone 1, or kept as a server-mode backend only?

## Related work

- Open Connector Integration Design (`docs/superpowers/specs/2026-07-11-open-connector-integration-design.md`) — the UDTF-bypass precedent this proposal follows.
- Graph database sources listed in `README.md:434` as a planned capability.

### DIFF
--- a/docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
+++ b/docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
@@ -151,8 +151,8 @@ Kuzu's Rust API can be embedded, avoiding a separate process. This is attractive
 ```mermaid
 flowchart LR
     SQL["SQL query"] --> DF["DataFusion"]
-    DF --> Catalog["graph.main.user_posts<br/>(YAML-declared view)"]
-    DF --> UDTF["cypher_query('neo4j', 'MATCH ...')"]
+    DF --> Catalog["graph.main.user_posts"]
+    DF --> UDTF["cypher_query UDTF"]
     Catalog --> Client["GraphClient"]
     UDTF --> Client
     Client --> Neo4j["Neo4j Bolt / HTTP"]
--- a/docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
+++ b/docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
@@ -34,7 +34,7 @@ A Cypher `RETURN` clause produces rows of heterogeneous values. Each column can 
 | scalar (string, int, float, bool, null) | native Arrow scalar | trivial mapping |
 | node | `STRUCT<id: string, labels: List<Utf8>, properties: Utf8>` | `properties` is JSON text; stable column set independent of label |
 | relationship | `STRUCT<id: string, start_id: string, end_id: string, type: Utf8, properties: Utf8>` | direction preserved |
-| path | `List<STRUCT<...>>` or flattened rows | paths are ordered sequences of alternating nodes and relationships |
+| path | `STRUCT<nodes: List<node STRUCT>, relationships: List<rel STRUCT>>` | a path alternates node/relationship, and the two have different shapes — an Arrow `List` needs ONE element type, so a path is two parallel typed lists: relationship *i* connects node *i* to node *i+1* |
 | list / map | `List<...>` or JSON text | nested values JSON-serialized where no stable schema exists |
 
 Because Cypher can return dynamic structures (`RETURN n, r, p`), the UDTF cannot infer a deterministic schema from the query string alone at planning time. Three candidates were evaluated and the first two rejected:
@@ -51,14 +51,15 @@ Because Cypher can return dynamic structures (`RETURN n, r, p`), the UDTF cannot
 | Protocol | Bolt v4/v5 (official `neo4rs`), HTTP JSON | native Rust API / C API |
 | Credentials | username/password or token | file-based access |
 | Cypher dialect | mature, rich | subset, rapidly growing |
+| Schema model | schema-optional; property types vary per node | **schema-full**: node/rel tables with declared, typed columns |
 | Best for | existing production graphs | embedded analytics, local-first |
 
-Both speak Cypher. The Skardi adapter should be backend-agnostic at the SQL surface, with a small per-backend driver trait.
+Both speak Cypher. The Skardi adapter should be backend-agnostic at the SQL surface, with a small per-backend driver trait. Kuzu's static schema is an upgrade the design should not squander: its `graph_schema` output is exact (read from the catalog, not sampled), and a future optimization can convert Kuzu properties as typed columns instead of JSON text — the uniform JSON-text spelling is the floor both backends meet, not a ceiling.
 
 ### Existing patterns in Skardi
 
 - `open_connector_query` and `open_connector_scan` already prove the UDTF-bypass pattern: a SQL function wraps an external HTTP action and returns rows.
-- The Open Connector design deliberately keeps credentials in the gateway. The graph design mirrors this: credentials live in environment variables referenced by the data source config, never in Skardi YAML or logs.
+- The Open Connector design keeps provider credentials in the gateway — Skardi holds only a runtime token. The graph bypass has **no gateway in the middle**: Skardi speaks Bolt directly, so it must hold the database credential itself. What carries over is the hygiene, not the topology: YAML references environment variables by *name* and never carries values, and the credential is never logged, serialized, or quoted in errors.
 - `TableProvider` scans, projection pushdown, and `LIMIT` pushdown are well-established; the graph UDTF can reuse the same machinery.
 
 ## Goals
@@ -68,9 +69,9 @@ Both speak Cypher. The Skardi adapter should be backend-agnostic at the SQL surf
 - Map Cypher result values (scalars, nodes, relationships, paths, lists, maps) into Arrow rows with a planning-time-stable schema — declared columns when provided, one JSON-text `record` column otherwise.
 - Allow `type: graph` catalog data sources with pre-declared Cypher views, registered from context YAML.
 - Enable federated `JOIN`s between graph results and existing Skardi sources.
-- Keep provider credentials out of Skardi process memory/logs/YAML.
+- Keep credentials out of YAML, logs, and error messages. (Unlike Open Connector there is no gateway to hold them: Skardi speaks Bolt directly, so the credential itself necessarily lives in process memory — read once from environment variables at registration, never serialized.)
 - Prove the design against a real Neo4j or Kuzu instance in phase-4 live verification.
-- Provide a `graph_schema(connection)` introspection UDTF so agents can discover labels, relationship types, and properties before generating Cypher.
+- Provide a `graph_schema(connection)` introspection UDTF so agents can discover labels, relationship types, and property names/types before generating Cypher.
 
 ## Non-goals
 
@@ -116,17 +117,34 @@ Both speak Cypher. The Skardi adapter should be backend-agnostic at the SQL surf
   ```rust
   #[async_trait]
   trait GraphClient: Send + Sync {
-      async fn execute(&self, cypher: &str, params: Value) -> Result<Vec<GraphRow>>;
+      /// Runs read-only Cypher inside a backend-enforced read transaction,
+      /// bounded by the per-source timeout and row cap.
+      async fn execute(
+          &self,
+          cypher: &str,
+          params: Value,
+          bounds: QueryBounds, // { timeout, max_rows }
+      ) -> Result<GraphRowStream>;
   }
   ```
+- `execute` returns a **stream** of `GraphRow`s, not a buffered `Vec`: the
+  row cap and SQL `LIMIT` early-stop both work by ceasing to consume it,
+  and a whole-graph `RETURN` never has to fit in memory before the cap
+  fires. Milestone 1 may buffer internally up to `max_rows` for
+  simplicity — the trait shape is what must not change.
 - A `GraphRow` is a vector of `GraphValue` (scalar, node, relationship, path, list, map).
 - Conversion from `GraphValue` to Arrow arrays is centralized and backend-agnostic.
+- Node identity: the `id` field carries Neo4j 5's `elementId()` (a
+  string). Against a Neo4j 4 server the numeric id is stringified; the
+  driver pins which server versions are supported and the docs state
+  that ids are opaque tokens for joining within one scan, not stable
+  keys across backends or upgrades.
 
 ### Schema handling
 
 There is no schema probe anywhere in the design: planning performs no network I/O, and every query has a planning-time-stable schema.
 
-- **YAML views:** the user declares the output schema explicitly. Skardi validates at registration that the Cypher query returns columns compatible with the declared schema (one live validation call at registration — registration is allowed to do network I/O; query planning is not).
+- **YAML views:** the user declares the output schema explicitly. Skardi validates at registration that the Cypher query returns columns compatible with the declared schema (one live validation call at registration — registration is allowed to do network I/O; query planning is not). An unreachable backend **fails registration hard**, the same contract as Open Connector's health check: a server that starts is a server whose declared views work.
 - **Ad-hoc UDTF with declared columns:** an optional fourth argument declares the output columns and their Arrow types as a JSON object (`'{"user_name": "Utf8", "post_title": "Utf8"}'`). The declared object is the planning-time schema; each returned `GraphValue` is converted against its declared type, and a mismatch fails with a typed error carrying column name, row index, expected type, and found JSON kind. Conversion is page-atomic — rows are buffered into one batch, converted as a unit, then emitted — so a mid-scan type failure never emits a partially-converted batch (batches already emitted before the failure may have been consumed downstream; this matches the Open Connector page-atomic semantics).
 - **Ad-hoc UDTF without declared columns:** every row is returned as one `record: Utf8` column containing the whole Cypher record as canonical JSON text (column names from `RETURN` become keys of the JSON object). Empty results are empty batches with the same one-column schema; downstream extraction uses SQL JSON functions on `record` (a small `json_get(record, '$.key')` UDF is included so `WHERE` clauses can filter on extracted keys without leaving SQL).
 
@@ -134,7 +152,15 @@ There is no schema probe anywhere in the design: planning performs no network I/
 
 - Nodes and relationships are returned as `STRUCT` columns with stable fields, not exploded into dynamic columns per label/type. This keeps the schema predictable across rows.
 - Properties are returned as JSON text (`Utf8`) so the schema is independent of which keys happen to be present in the sampled rows.
-- Paths are returned as `List<STRUCT>` by default; a YAML view may opt to flatten a path into multiple rows using `UNWIND` in Cypher.
+- A path cannot be one `List<STRUCT>`: its elements alternate between the
+  node shape and the relationship shape, and an Arrow list has exactly one
+  element type. A path column is therefore
+  `STRUCT<nodes: List<node STRUCT>, relationships: List<rel STRUCT>>` —
+  two parallel typed lists where relationship *i* connects node *i* to
+  node *i+1* (`nodes` is always one longer than `relationships`; a
+  zero-hop path is one node and an empty relationship list). Callers who
+  want row-per-hop flatten with Cypher `UNWIND`/`relationships(p)` in the
+  query or view.
 
 ### Credentials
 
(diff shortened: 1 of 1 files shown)


docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
```diff
@@ -0,0 +1,491 @@
...
+- **Primary enforcement — the backend refuses writes.** Neo4j: every query
+  runs in an explicit READ-access-mode transaction (Bolt carries the access
+  mode in `BEGIN`/`RUN` metadata, and the *server* rejects writes inside a
+  read transaction — Community edition included). Deployments should
+  additionally use least-privilege credentials (a reader role where RBAC is
+  available). If the pinned `neo4rs` version does not expose transaction
+  access mode, wiring it (or upstreaming it) is a milestone-1 prerequisite,
+  and a read-only database user is the documented deployment fallback in
+  the meantime. Kuzu: the database is opened with `read_only` set, which
+  the engine enforces.
```
abbccdda: **Security boundary depends on an unverified capability — resolve before M1, not during it.** The doc correctly reframes the keyword guard as UX and makes the backend read-transaction the real boundary — but then admits (here and in Risk #2, L481) that it's unconfirmed whether the pinned `neo4rs` exposes transaction access mode. If it doesn't, M1's only actual enforcement is a guard the doc itself calls "bypassable by construction," plus an operator-provided read-only DB user (out of Skardi's control). That's the single highest-risk item in the design. Recommend making the `neo4rs` access-mode spike a **precondition** for starting M1, and if it can't be wired/upstreamed, promoting "read-only DB user" from "documented fallback" to a hard registration requirement (refuse to register a `neo4j` source without proof of a read-only role, or fail closed).

docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
```diff
@@ -0,0 +1,491 @@
...
+  string). Against a Neo4j 4 server the numeric id is stringified; the
+  driver pins which server versions are supported and the docs state
+  that ids are opaque tokens for joining within one scan, not stable
+  keys across backends or upgrades.
+
+### Schema handling
+
+There is no schema probe anywhere in the design: planning performs no network I/O, and every query has a planning-time-stable schema.
+
+- **YAML views:** the user declares the output schema explicitly. Skardi validates at registration that the Cypher query returns columns compatible with the declared schema (one live validation call at registration — registration is allowed to do network I/O; query planning is not). An unreachable backend **fails registration hard**, the same contract as Open Connector's health check: a server that starts is a server whose declared views work.
```
abbccdda: **"Fails registration hard" couples whole-server startup to an external graph backend's availability.** Open Connector's health check talks to a gateway Skardi controls; here the backend is a shared external Neo4j/Kuzu. A transient blip in that graph DB would then block `skardi-server` from starting *at all*, taking down every unrelated Postgres/CSV/Lance source with it. That's a heavier blast radius than the precedent it cites. Consider degraded registration (register the view, mark it unhealthy, surface the error on first scan and in `/data_source`) instead of hard-failing the process — or at minimum call out this coupling explicitly as an accepted operational trade-off.

docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
```diff
@@ -0,0 +1,491 @@
...
+  ```sql
+  -- Declared columns: typed multi-column output.
+  SELECT user_name, post_title
+  FROM cypher_query('neo4j', '
+      MATCH (u:User)-[:POSTED]->(p:Post)
+      WHERE u.id = $userId
+      RETURN u.name, p.title
+      LIMIT 10
+  ', '{"userId": "u-123"}',
+     '{"user_name": "Utf8", "post_title": "Utf8"}');
```
abbccdda: **The `columns` type vocabulary breaks the repo's universal convention.** This uses Arrow PascalCase (`"Utf8"`), but every existing source that parses type names from config uses friendly lowercase — `dynamodb.rs:2357` (`"string"|"str"|"utf8"|"int"|"integer"|"bigint"`), `mongo/mod.rs:535`, `seekdb/mod.rs:618`, `llm_extract/mod.rs:904`. There is no Arrow-name→`DataType` parser anywhere in the tree today, so this would introduce a second, inconsistent spelling that agents and humans both have to learn. Please either adopt the existing lowercase vocabulary or explicitly justify the divergence — and enumerate the exact accepted set (the doc only ever shows `Utf8`).

docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
```diff
@@ -0,0 +1,491 @@
...
+              MATCH (u:User)-[:POSTED]->(p:Post)
+              RETURN u.name AS user_name, p.title AS post_title
+            schema:
+              - name: user_name
+                type: Utf8
+                nullable: true
+              - name: post_title
+                type: Utf8
+                nullable: true
+```
```
abbccdda: **Nullability is expressible in YAML views but not in the ad-hoc `columns` arg** — the JSON object (L138) is name→type only, while views carry `nullable: true` here. Arrow fields need a nullability bit, so the ad-hoc path needs a stated default (presumably `nullable: true` for everything, matching these examples). Worth making explicit so the two schema-declaration paths don't silently disagree on nullability.

docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
```diff
@@ -0,0 +1,491 @@
...
+
+```sql
+cypher_query(
+    connection TEXT,       -- references a registered graph data source name
+    cypher TEXT,           -- read-only Cypher query
+    params TEXT optional,  -- JSON object of query parameters
+    columns TEXT optional  -- JSON object declaring output columns and Arrow types;
+                           -- omitted → one `record: Utf8` JSON column per row
+) RETURNS TABLE(...)
+```
```
abbccdda, line 391: **Two constraints on this signature should be documented, since agents generate these calls.** (1) `params` and `columns` are positional-optional, so declaring `columns` (arg 4) forces passing `params` (arg 3) — the doc has no example of "columns, no params." State whether callers must pass `'{}'`/`NULL` for `params` in that case, and (given `strict_string_arg` rejects NULL for schema-shaping args) whether `params` accepts NULL. (2) Per `udtf_args.rs`, `columns` must be a **string literal** at plan time — a subquery/column-ref will fail planning. An LLM emitting `cypher_query(conn, q, p, cols_from_cte)` won't obviously know that; call it out.

docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
```diff
@@ -0,0 +1,491 @@
...
+  any network round-trip, the wrapper rejects Cypher containing mutating
+  or escape-hatch keywords on **word boundaries** (`CREATE`, `SET`,
+  `DELETE`, `DETACH`, `REMOVE`, `MERGE`, `DROP`, plus `CALL` — procedures
+  can mutate without any write keyword — and `LOAD CSV` — it makes the
+  *graph server* fetch arbitrary URLs). This exists to hand an agent a
+  fast, actionable error naming the blocked keyword; it is deliberately
+  conservative and will false-positive on keyword-shaped string literals
+  (`RETURN 'DELETE'`) — an accepted tax, since the backend read mode
+  behind it is what actually guarantees read-only. Word-boundary matching
+  keeps identifiers like `created_at` out of the blast radius.
```
abbccdda: **Guard rejects `CALL`, but `graph_schema` (the agent-discovery centerpiece) is implemented *with* `CALL db.schema.nodeTypeProperties()` (L402/L444).** These don't collide if the guard only screens user-supplied Cypher on `cypher_query` and `graph_schema` issues its own fixed internal query — but the doc never says that. Please state that the keyword guard applies to caller Cypher only and that internally-issued introspection/`db.schema.*` procedures are exempt, so a reader doesn't conclude introspection is self-blocked.

docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
```diff
@@ -0,0 +1,491 @@
...
+
+- **YAML views:** the user declares the output schema explicitly. Skardi validates at registration that the Cypher query returns columns compatible with the declared schema (one live validation call at registration — registration is allowed to do network I/O; query planning is not). An unreachable backend **fails registration hard**, the same contract as Open Connector's health check: a server that starts is a server whose declared views work.
+- **Ad-hoc UDTF with declared columns:** an optional fourth argument declares the output columns and their Arrow types as a JSON object (`'{"user_name": "Utf8", "post_title": "Utf8"}'`). The declared object is the planning-time schema; each returned `GraphValue` is converted against its declared type, and a mismatch fails with a typed error carrying column name, row index, expected type, and found JSON kind. Conversion is page-atomic — rows are buffered into one batch, converted as a unit, then emitted — so a mid-scan type failure never emits a partially-converted batch (batches already emitted before the failure may have been consumed downstream; this matches the Open Connector page-atomic semantics).
```
abbccdda: **"Page-atomic" is borrowed from Open Connector but the atomic unit isn't defined here.** In Open Connector a "page" is a concrete pagination page from the upstream API; a Bolt/Cypher result stream has no equivalent boundary. This paragraph says rows are "buffered into one batch" while the trait (L168-172) returns a *stream* and M1 "may buffer up to `max_rows`." So is the atomic unit the whole result (≤ `max_rows`, one batch) or a fixed batch size? Define the batch boundary explicitly — it determines both the peak-memory story and exactly how much can be "already emitted" when a mid-scan type mismatch fires (L480).

docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
```diff
@@ -0,0 +1,491 @@
...
+- **Ad-hoc UDTF without declared columns:** every row is returned as one `record: Utf8` column containing the whole Cypher record as canonical JSON text (column names from `RETURN` become keys of the JSON object). Empty results are empty batches with the same one-column schema; downstream extraction uses SQL JSON functions on `record` (a small `json_get(record, '$.key')` UDF is included so `WHERE` clauses can filter on extracted keys without leaving SQL).
```
abbccdda: **`json_get` doesn't exist in the repo yet** — the tree has a custom `json_pack` UDF (`util/json_pack.rs`) but no getter, so "a small `json_get` UDF is included" is net-new work, not reuse. Two things: (1) make the new-UDF cost explicit in a milestone; (2) note that DataFusion's `datafusion-functions-json` provides a `json_get` family — decide whether to depend on it vs. hand-roll, and avoid a name collision / duplicated semantics with the ecosystem function. The single-column JSON fallback's ergonomics hinge entirely on this UDF existing and being good, so it deserves more than a parenthetical.

docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
```diff
@@ -0,0 +1,491 @@
...
+  and a whole-graph `RETURN` never has to fit in memory before the cap
+  fires. Milestone 1 may buffer internally up to `max_rows` for
+  simplicity — the trait shape is what must not change.
+- A `GraphRow` is a vector of `GraphValue` (scalar, node, relationship, path, list, map).
+- Conversion from `GraphValue` to Arrow arrays is centralized and backend-agnostic.
+- Node identity: the `id` field carries Neo4j 5's `elementId()` (a
+  string). Against a Neo4j 4 server the numeric id is stringified; the
+  driver pins which server versions are supported and the docs state
+  that ids are opaque tokens for joining within one scan, not stable
+  keys across backends or upgrades.
```
abbccdda: **"Opaque, for joining within one scan" quietly undercuts a headline selling point.** The summary sells federated `JOIN`s of graph results against other sources. But a SQL statement joining two `cypher_query(...)` calls is two separate scans / transactions, and joining graph output to a Postgres table on a node id spans scans too. If ids are only guaranteed meaningful "within one scan," what exactly is join-safe across the two sides of a federated JOIN? Neo4j `elementId()` is in practice stable across transactions on one server, so this caveat may be stricter than needed — please reconcile the caveat with the federated-JOIN use case (probably: stable within a single server between scans, unstable across servers/upgrades).

docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
```diff
@@ -0,0 +1,491 @@
...
+- Graph-specific SQL extensions (`MATCH`, `()-[]->()` syntax).
+- Embedding the graph engine inside Skardi process by default (Kuzu embedded mode is a later optimization, not the milestone-one default).
+- Transaction or snapshot semantics across graph and relational sources.
+- Full Cypher feature parity on day one (procedures, APOC, GDS are out of scope).
+
+## Decisions
+
+### SQL surface
+
+- **Primary interface:** `cypher_query(connection_name TEXT, cypher TEXT, params JSON OBJECT optional, columns JSON OBJECT optional)` returns a table.
```
abbccdda, line 131: **Minor — JSON `params` loses the int/float distinction Bolt cares about.** Parameters arrive as a JSON string, but JSON has one number type; Cypher/Neo4j distinguish `Integer` from `Float` and some operations/index lookups are type-sensitive. A param meant as `Integer 1` vs `Float 1.0` can't be expressed. Worth a sentence on how numeric params map to Bolt types (e.g. integral JSON numbers → Bolt Integer), or noting the limitation.

abbccdda (commented):
Design review — robustness / correctness / maintainability.

Strong, internally-consistent proposal. I verified the two load-bearing precedents it leans on: the Open Connector UDTF pattern (global registration, planning-time schema, `runtime_token_env` credential hygiene, LIMIT-scoped caching, `strict_string_arg` literal args) and DataFusion's planning-time-schema requirement — both hold, and the doc represents them accurately. The bypass-over-native-Cypher call and the no-probe/declared-schema decision are well argued.

My findings below are mostly gaps to close before Milestone 1 rather than direction changes. The two I'd gate M1 on are (1) the unverified `neo4rs` read-access-mode dependency, since it's the actual security boundary, and (2) the `columns` type-name vocabulary, which diverges from every existing source. The rest are correctness/consistency clarifications. Inline comments follow.

docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
```diff
@@ -0,0 +1,491 @@
...
+- Prove federated `JOIN` between Kuzu and a CSV source.
+
+### Milestone 3 — YAML catalog views
+
+- `type: graph` data source registration.
+- Declared-schema views.
+- Live schema validation at registration.
+- Docs: per-backend guides, examples, and spec entry.
+
+### Milestone 3+ — Apache AGE backend (candidate)
```
BtXin: Maybe move this to milestone 1, let's support PG plugin first, then neo4j.

docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
bakey: Fixed in 1d4cca0, and the milestone reordering (BtXin's comment) dissolves most of the risk: **AGE is now milestone 1**, so the first shipped security boundary is a Postgres `READ ONLY` transaction — native to tokio-postgres, no driver capability in question. For the Neo4j milestone (now M2), your recommendation is adopted verbatim: the neo4rs access-mode spike is a **hard precondition before the milestone starts** (wiring/upstreaming it is that milestone's cost if missing), and registration performs a fail-closed **read-mode proof** — attempt a trivial CREATE inside a read transaction, require the SERVER to refuse it, roll back regardless — so a misconfigured driver/server pair fails at registration, never silently at runtime. The milestone explicitly does not ship on the keyword guard alone.

docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
bakey: Agreed — the blast-radius asymmetry vs Open Connector's gateway is real, and the doc now argues it in place. Fixed in 1d4cca0: **unreachable backend → DEGRADED registration** (views register with their declared, planning-sufficient schemas; the source is marked unhealthy in GET /data_source; the first scan retries validation and fails loudly), while a **reachable backend whose view fails validation still refuses registration** — availability outages and contract violations part ways. Server startup no longer takes a hostage.

docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
bakey: Adopted in 1d4cca0 — no divergence to justify, you're right that a second spelling helps no one. The vocabulary is now the repo's friendly lowercase set, enumerated exactly: `string` (`str`/`utf8`) → Utf8, `int` (`integer`/`bigint`) → Int64, `float` (`double`) → Float64, `bool` (`boolean`), `json` (JSON text verbatim), plus `node`/`relationship`/`path` for the canonical STRUCT shapes. Anything else fails planning with the accepted set listed, and YAML views' `type:` fields use the same vocabulary — one spelling everywhere (the YAML example now reads `type: string`).

docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
bakey: Fixed in 1d4cca0: ad-hoc declared columns are **always nullable** — Cypher can produce null in any position (OPTIONAL MATCH, absent properties), so the JSON object stays name→type with no way to declare otherwise. YAML views default to nullable: true and may declare nullable: false as an authored assertion about their own Cypher. The two paths can't silently disagree because the stricter bit exists only where an author explicitly wrote it.

docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
bakey: Both constraints are now stated in the signature section (1d4cca0), with an example: `columns` (arg 4) requires `params` (arg 3), `'{}'` is the no-parameters spelling, NULL is rejected (schema-shaping args are strict string literals — the open_connector_query precedent you cited), and both `params` and `columns` must be plan-time string **literals**: a column ref, CTE output, or subquery fails planning with a targeted error rather than being evaluated first. The example shows exactly the 'columns, no params' call an agent would want.

docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
bakey: Stated in 1d4cca0, in the guard bullet itself: the keyword guard screens **caller-supplied Cypher only** (text passed to cypher_query or declared in a view); graph_schema issues fixed, engine-authored introspection queries (db.schema.*, ag_catalog reads) that never pass through the guard — so rejecting CALL does not self-block discovery. Thanks for catching that the doc left this inferable-but-unstated.

docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
bakey: Defined in 1d4cca0: the atomic unit is the **conversion batch, which this design defines** since a Cypher stream has no upstream page — a fixed implementation constant (order 1024 rows, never more than max_rows) consumed off the driver stream; each batch converts as a unit before emission; a type mismatch fails the CURRENT batch pre-emission, with earlier batches possibly already consumed (the same trade-off OC accepts). Peak conversion memory is one batch, not one result. The M1 latitude is kept but now has defined semantics: a single batch up to max_rows, in which case nothing at all is emitted before a mid-scan failure.

docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
bakey: Fixed in 1d4cca0 — it's now a named dependency with a decision, not a parenthetical: adopt **datafusion-functions-json's json_get family** (ecosystem names/semantics, no homegrown twin to collide with); verifying the crate against the pinned DataFusion is milestone-1 work; an in-tree minimal getter happens only if the pin conflicts, documented as a subset. Related settlement: with AGE now milestone 1, the JSON-record fallback ships with the **Neo4j** milestone (Bolt needs no declared arity) — on AGE, omitted columns is a targeted error, so the fallback's ergonomics never gate M1.

docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
bakey: You're right that the caveat as written undercut the headline — it was stricter than the engines' actual behavior. Reconciled in 1d4cca0 with the precise contract: an id is **stable for the life of the entity within one database instance**, which makes it join-safe across the scans of a single federated query AND against ids previously exported from the same server (the graph-to-Postgres join). What it is NOT is a durable foreign key: engines may reuse ids after entity deletion, and nothing survives restores, server migrations, or major upgrades — persist business keys from properties for that. Same field now also documents AGE graphids and Kuzu internal ids under the same contract.

docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
bakey: Added in 1d4cca0 as a bullet under the SQL surface: a JSON number with no fraction or exponent binds as Cypher Integer; one with a fraction or exponent binds as Float — write `1.0` to force a Float parameter; strings are never numerically coerced. serde_json preserves the i64/f64 distinction, so the rule is implementable exactly as stated.

docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md
bakey: Adopted in 1d4cca0 — milestones reordered: **M1 = Apache AGE (Postgres plugin), M2 = Neo4j, M3 = Kuzu, M4 = YAML views**. Beyond matching the deployment in hand (GraphRAG data already in Postgres, zero new infrastructure), the reorder also dissolves the other reviewer's highest-risk finding: milestone 1's read-only boundary becomes a Postgres READ ONLY transaction (native to tokio-postgres), and the neo4rs access-mode spike moves to a hard precondition of the now-second Neo4j milestone. The AGE record-fallback arity question is settled as part of this: on AGE, omitted columns is a targeted error; the JSON-record fallback ships with Neo4j where Bolt needs no declared arity.

BtXin (approved):

abbccdda (approved):

PR #200, opened 2026-08-08
fix(udf): LargeUtf8 text columns in chunk/chunk_parts; document json_pack sub-ms truncation

bakey:
Two findings from the skardi-cloud etl-generator review ([skardi-cloud#232 round 4](https://github.com/SkardiLabs/skardi-cloud/pull/232)) that belong in this repo now that #193 landed the UDFs here:

1. **`read_text_column` accepted `LargeUtf8` as a scalar but rejected it as a column** — the array arm downcast only `StringArray`/`StringViewArray`, so a `LargeUtf8` content column failed at job run time with "must be a Utf8 column". A text column must not be stricter than a text literal; `LargeStringArray` joins the downcast chain, with a regression test (split + null passthrough).

2. **`json_pack`'s micro/nano timestamp arms silently floor-divide to epoch milliseconds.** The truncation is by design (the contract is epoch ms) but was undocumented and read as asymmetric against the Second arm's overflow check. The comment now states both: sub-ms precision is deliberately dropped, and only the widening direction needs the overflow guard.

## Verification

- 18 `model::chunking` tests (`--features chunking`), 9 `util::json_pack` tests, clippy clean.
- skardi-cloud's `skardi-etl` will repin to the rev containing this fix (its vendored UDF copies are being deleted in #232 now that #193 is merged).

### DIFF
--- a/crates/skardi/src/model/chunking/mod.rs
+++ b/crates/skardi/src/model/chunking/mod.rs
@@ -379,6 +379,30 @@ fn read_scalar_usize(udf: &str, arg: &ColumnarValue, name: &str) -> DfResult<usi
     Ok(n as usize)
 }
 
+/// Pre-flight for LargeUtf8 COLUMNS: chunk output flows through
+/// i32-offset builders (`StringBuilder`), whose offset overflow is a
+/// PANIC in arrow, not an error — and LargeUtf8 is the one layout whose
+/// contract makes >2 GiB legal input. Chunk output ≈ input bytes (more
+/// with overlap), so reject early with a real error instead of
+/// unwinding mid-batch. (A near-2 GiB Utf8 column plus overlap can
+/// still overflow — that exposure predates LargeUtf8 support and is
+/// shared with every i32-offset producer.) Offsets first-to-last, so a
+/// SLICED array counts its own bytes, not the whole buffer's. A free
+/// function over raw offsets so the boundary is unit-testable without a
+/// 2 GiB allocation.
+fn ensure_large_text_fits_i32(udf: &str, name: &str, offsets: &[i64]) -> DfResult<()> {
+    let total_bytes =
+        offsets.last().copied().unwrap_or_default() - offsets.first().copied().unwrap_or_default();
+    if total_bytes > i64::from(i32::MAX) {
+        return Err(DataFusionError::Execution(format!(
+            "{udf}: '{name}' carries more than 2 GiB of text in one batch; \
+             chunk output is built as 32-bit-offset Utf8 and cannot hold it — \
+             split the input into smaller batches"
+        )));
+    }
+    Ok(())
+}
+
 fn read_text_column<'a>(
     udf: &str,
     arg: &'a ColumnarValue,
@@ -399,32 +423,7 @@ fn read_text_column<'a>(
                 return Ok(view_arr.iter().collect());
             }
             if let Some(large_arr) = arr.as_any().downcast_ref::<LargeStringArray>() {
-                // Pre-flight: chunk output flows through i32-offset
-                // builders (`StringBuilder`), whose offset overflow is a
-                // PANIC in arrow, not an error — and LargeUtf8 is the one
-                // layout whose contract makes >2 GiB legal input. Chunk
-                // output ≈ input bytes (more with overlap), so reject
-                // early with a real error instead of unwinding mid-batch.
-                // (A near-2 GiB Utf8 column plus overlap can still
-                // overflow — that exposure predates LargeUtf8 support and
-                // is shared with every i32-offset producer.)
-                let total_bytes = large_arr
-                    .value_offsets()
-                    .last()
-                    .copied()
-                    .unwrap_or_default()
-                    - large_arr
-                        .value_offsets()
-                        .first()
-                        .copied()
-                        .unwrap_or_default();
-                if total_bytes > i64::from(i32::MAX) {
-                    return Err(DataFusionError::Execution(format!(
-                        "{udf}: '{name}' carries more than 2 GiB of text in one batch; \
-                         chunk output is built as 32-bit-offset Utf8 and cannot hold it — \
-                         split the input into smaller batches"
-                    )));
-                }
+                ensure_large_text_fits_i32(udf, name, large_arr.value_offsets())?;
                 return Ok(large_arr.iter().collect());
             }
             let str_arr = arr.as_any().downcast_ref::<StringArray>().ok_or_else(|| {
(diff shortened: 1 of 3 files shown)


crates/skardi/src/model/chunking/mod.rs
```diff
@@ -400,6 +402,17 @@ fn read_text_column<'a>(
                     })
                     .collect());
             }
+            if let Some(large_arr) = arr.as_any().downcast_ref::<LargeStringArray>() {
```
abbccdda, line 405: **Correctness: accepting LargeUtf8 here routes a 64-bit-offset input into 32-bit-offset output builders, turning a clean error into a panic.**

`build_list_array` (line 457) and `build_parts_array` (lines 267/282) emit output through `StringBuilder` = `GenericStringBuilder<i32>`; in the pinned arrow-array 57.3.1, `GenericByteBuilder::next_offset` is `.expect("byte array offset overflow")` once cumulative appended bytes exceed `i32::MAX` (generic_bytes_builder.rs:87). Chunk output preserves all input bytes — more with `overlap > 0` — so a LargeUtf8 column carrying >2 GiB of text in one batch column (the case the LargeUtf8 layout exists for) now panics mid-batch, aborting the query via task unwind, where main returned `DataFusionError::Execution("…must be a Utf8 column")` at the same spot.

Framing caveats, verified: the panic site isn't strictly new (main already accepted `ScalarValue::LargeUtf8` *scalars*, and a near-2 GiB Utf8 array with overlap could already overflow), and triggering needs >2 GiB of chunk output in a single `RecordBatch` column — plausible for LargeUtf8 document sources, but rare. Still, this change makes the overflow reachable by design for the one input type whose contract is "more than 2 GiB is legal". Cheapest fix: pre-flight the cumulative value-byte count in the LargeUtf8 arm and return an execution error ("text exceeds 2 GiB chunk output limit") instead of unwinding; the full fix is Large output arrays when input is Large.

crates/skardi/src/model/chunking/mod.rs
```diff
@@ -400,6 +402,17 @@ fn read_text_column<'a>(
...
+                        if large_arr.is_null(i) {
+                            None
+                        } else {
+                            Some(large_arr.value(i))
+                        }
+                    })
+                    .collect());
+            }
             let str_arr = arr.as_any().downcast_ref::<StringArray>().ok_or_else(|| {
                 DataFusionError::Execution(format!("{udf}: '{name}' must be a Utf8 column"))
```
abbccdda, line 417: The fallback message now understates the contract this arm accepts (Utf8, LargeUtf8, Utf8View). A user whose column is e.g. `Dictionary(Int32, Utf8)` or `Binary` is told it "must be a Utf8 column" and will reach for an unnecessary `CAST` — the diagnostic misstates the accepted type set that this PR just widened. Suggest: `"{udf}: '{name}' must be a string column (Utf8, LargeUtf8, or Utf8View)"`. The scalar fallback at lines 435-437 ("must be Utf8") has the same problem.

crates/skardi/src/model/chunking/mod.rs
```diff
@@ -400,6 +402,17 @@ fn read_text_column<'a>(
...
+            if let Some(large_arr) = arr.as_any().downcast_ref::<LargeStringArray>() {
+                return Ok((0..large_arr.len())
+                    .map(|i| {
+                        if large_arr.is_null(i) {
+                            None
+                        } else {
+                            Some(large_arr.value(i))
+                        }
+                    })
+                    .collect());
```
abbccdda, line 414: Suggestion (non-blocking): this is the third verbatim copy of the same index/`is_null`/`value` loop (after the StringViewArray and StringArray arms). In arrow 57.3.1, all three array types' `iter()` yields `Option<&'a str>` with identical null, slice-offset, and lifetime behavior (the item borrow stays tied to the input array), so each arm collapses to one line:

```rust
if let Some(large_arr) = arr.as_any().downcast_ref::<LargeStringArray>() {
    return Ok(large_arr.iter().collect());
}
```

(`AsArray`'s `as_string_opt::<i64>()` / `as_string_view_opt()` can also replace the raw downcasts — the repo already uses `AsArray` in the sqlite provider.) As it stands, a fourth string layout means a fourth copy, and any fix to the null-handling pattern has to land in three places or they silently diverge.

crates/skardi/src/util/json_pack.rs
```diff
@@ -171,6 +171,11 @@ fn scalar_to_value(scalar: &ScalarValue) -> DFResult<Value> {
                 )
             })
         })?,
+        // Sub-millisecond precision is TRUNCATED (floor division), by
```
abbccdda, line 174: Wording defect that matters because documenting this contract precisely is the stated purpose of this hunk: `div_euclid` is *floor* division (toward −∞), not truncation (toward zero) — for pre-epoch instants they differ: `(-1500i64).div_euclid(1000) == -2`, whereas "TRUNCATED" implies `-1`. The file's own tests already use the right word ("micros/nanos floor toward −∞", lines 334-335), and since the only pre-epoch test case is a Millisecond pass-through, this comment is the sole statement of the negative micro/nano contract — a consumer coding against "truncated" is off by 1 ms for every pre-epoch timestamp with a sub-ms remainder. Suggest: "Sub-millisecond precision is dropped by floor division (`div_euclid`, rounds toward −∞ — pre-epoch instants round *earlier*, never toward zero), by design: …"

abbccdda (commented):
Reviewed against the two fixes the PR claims (LargeUtf8 column acceptance in `chunk`/`chunk_parts`; json_pack sub-ms documentation), the #193 registration contract, and the surrounding UDF family. The changes do what the PR body says, the regression test is sound (null passthrough + multi-chunk split), and both production SessionContexts — planning (`crates/server/src/config.rs:445-449`) and runtime (`crates/server/src/server.rs:190-194`) — still register `chunk`/`chunk_parts` + `json_pack`, so the #193 end-to-end contract holds. Four inline comments below (one correctness, one diagnostics, two wording/maintainability), plus three notes that can't anchor to the diff:

**1. Docs still declare `text` as exactly `Utf8`.** `docs/chunk.md:27` (`| text | Utf8 | The text to split…`) and the module doc at `crates/skardi/src/model/chunking/mod.rs:67` ("`text`: `Utf8` literal, scalar subquery, or column") weren't updated, so the LargeUtf8/Utf8View acceptance is undocumented behavior: ETL-generator authors reading the doc will keep writing defensive `CAST`s, and a future refactor could remove the `LargeStringArray` arm as "undocumented" and re-break large-string sources. The scalar fallback message at mod.rs:435-437 ("must be Utf8") has the same staleness as the array-arm message flagged inline.

**2. The bug class this PR fixes is still open in every sibling text-consuming UDF** (pre-existing, non-blocking — but the PR's own rationale, "LargeUtf8 columns arrive from large-string sources", applies to them verbatim): `candle` (mod.rs:184 array arm, 199-200 scalar arms), `gguf` (mod.rs:186, 203-204; feature-gated, lower priority), `remote_embed` (mod.rs:244 array arm, 251 scalar arm), and `llm_extract`'s `to_string_array` (mod.rs:1042, 1047) all accept only plain `StringArray` columns / `Utf8` scalars under `Signature::variadic_any`, which does no coercion. `chunk`→embed composition is safe because chunk re-emits plain `Utf8` output, but a LargeUtf8 or Utf8View source column fed *directly* to `remote_embed`/`llm_extract` fails exactly the way `chunk` did before this PR. Suggested follow-up: promote `read_text_column` to a shared util (e.g. `crates/skardi/src/util/`) — or switch these UDFs to `TypeSignature::UserDefined` + `coerce_types` (available in the pinned DataFusion 52.5) so the engine normalizes string layouts before `invoke` and the whole class disappears by construction.

**3. `json_pack` carries the same offset-width asymmetry on its list path.** String *values* are fine (the `Utf8 | LargeUtf8 | Utf8View` scalar match covers columns via `try_from_array`), but lists only handle i32 offsets: the array path downcasts `ListArray` only (json_pack.rs:229), `scalar_to_value` matches `ScalarValue::List` only (:189), and `utf8_list_value` requires `StringArray` elements (:244). A `LargeList<Utf8>` tags column fails with "unsupported value type LargeList(…)" while the semantically identical `List<Utf8>` works. Same follow-up bucket as note 2.

bakey:
**All four inline findings verified and fixed in 3fe3243; notes 1–3 handled as below.**

1. **LargeUtf8 → i32-offset builder panic (correctness)** — confirmed, including your framing caveats (the site predates this PR via `LargeUtf8` *scalars*, and near-2 GiB Utf8 + overlap could already overflow). Took the cheap fix: the LargeUtf8 arm pre-flights the cumulative value-byte count (`value_offsets` last − first, so slices are counted correctly) and returns a targeted execution error — "more than 2 GiB of text in one batch … split the input into smaller batches" — instead of unwinding mid-batch. The pre-existing residual is stated in the comment. No unit test for the guard itself: constructing a >2 GiB `LargeStringArray` in CI is a 2 GiB allocation; the arithmetic is a two-liner under the tested arm. The full fix (Large output when input is Large) is deliberately out of scope for this micro-PR — it changes `return_type` and every consumer's expectations.

2. **Fallback diagnostics** — both messages (array arm and scalar fallback) now name the accepted set: "must be a string column (Utf8, LargeUtf8, or Utf8View)".

3. **Three copies of the null/value loop** — collapsed to `iter().collect()` in all three arms, with a comment noting the shared `Option<&str>` behavior that makes the one spelling safe.

4. **Floor vs truncation** — you're right, and it matters exactly where you said: the comment was the only statement of the negative sub-ms contract. It now says floor with the pre-epoch direction spelled out (−1500 µs → −2 ms, never toward zero), and the granularity test gains `neg_us`/`neg_ns` cases pinning **−2/−1 by value**, so the contract no longer rests on prose.

**Note 1 (doc drift)** — fixed in the same commit: `docs/chunk.md`'s `text` row and the module doc's argument list now declare all three string layouts, so the acceptance is documented behavior a future refactor can't remove as "undocumented".

**Notes 2–3 (sibling UDFs; json_pack's LargeList asymmetry)** — agreed on both substance and bucketing; filed as #202 with your two suggested fixes (shared `read_text_column` util vs `TypeSignature::UserDefined` + `coerce_types`) and a recommendation for the latter, kept out of this micro-PR since skardi-cloud's repin is waiting on it.

abbccdda:
@bakey Could you check and improve on the test coverage?

bakey:
**Codecov follow-up — the 11 uncovered patch lines are closed.**

They were exactly the arms this PR added or reworded. Two moves:

1. **The 2 GiB pre-flight is now testable**: extracted into `ensure_large_text_fits_i32`, a free function over raw offsets, so the boundary is pinned *exactly* — `i32::MAX` passes, one byte past fails naming the remedy — without a 2 GiB allocation (and the sliced-array case is pinned too: a slice counts its own span, not the whole buffer).
2. **Every other new/reworded arm is executed**: Utf8View columns (incl. null passthrough), LargeUtf8/Utf8View *scalars*, typed-NULL scalars of all three layouts ("a literal must never be stricter than a column"), and both widened wrong-type diagnostics (array arm and scalar fallback, each asserting the accepted-set wording).

21 chunking tests (was 18), clippy clean.

abbccdda (approved):
Thanks for the fix!

PR #204, opened 2026-08-11
feat(rss): hostname through the egress seam, external-crate seam proof, honest Layer 1/2 boundaries

gracexmatin:
## Summary

Follow-ups to the [#185 review](https://github.com/SkardiLabs/skardi/pull/185) of the RSS egress design. Of its five findings, two ([proxy bypass](https://github.com/SkardiLabs/skardi/pull/185#discussion_r3709930107), [cached-window failure contract](https://github.com/SkardiLabs/skardi/pull/185#discussion_r3709930115)) were absorbed by the provider stack (#180–#183); this PR carries the three that remained open, one commit each.

- **Prove the egress-injection seam from an external crate** ([review](https://github.com/SkardiLabs/skardi/pull/185#discussion_r3709930097): "expose a provider-level injection path"). `crates/skardi/tests/rss_egress_injection.rs` compiles as its own crate against the public API alone: it implements a denying `EgressPolicy` outside the crate, registers through `register_rss_tables_with_policy`, and observes the refusal in `feeds.last_status`/`last_error` with zero `items` rows — no mock server; the IP-literal target is refused before any connect. The design's interface-contract section now names the public registration path (not the private fetcher constructor) and hangs its "only the injected object" claim on this test; two test-suite headers still claiming the seam is `#[cfg(test)] pub(crate)` are trued up.

- **Pass the hostname through the `EgressPolicy` seam; split Layer 2 by context** ([review](https://github.com/SkardiLabs/skardi/pull/185#discussion_r3709930103): "pass request context to Cloud policy hooks"). `check_ip(&self, ip)` becomes `check(&self, host, ip)`: a bare IP can express reserved-range refusal but not host allowlists, and the policy carries no tenant/feed identity and may be consulted several times per fetch, so it can never carry quotas or attributed audit. The fetcher now passes the hostname at both call sites (resolver and IP-literal hop check), and the design splits Layer 2 by the context each control needs: destination checks (reserved ranges, per-hop host allowlist) in the injected policy; allowlist administration, quotas, and audit in Cloud's orchestration layer, which natively holds the missing context. Byte metering is recorded as an open question — no application seam ever sees a response.

- **State Layer 1's actual coverage boundary** ([review](https://github.com/SkardiLabs/skardi/pull/185#discussion_r3709930112): "close same-pod and same-node gaps in Layer 1"). A standard `NetworkPolicy` cannot block a pod's own loopback, node-local traffic is CNI-specific, enforcement requires a policy-enforcing CNI, and the IMDS hop limit is topology-dependent. The design now says so: Layer 1 contains network-path targets only, with same-pod/same-node covered solely by Layer 2's in-process refusal; provisioning pins and asserts the CNI, constrains pod topology (no `hostNetwork`, no sidecars, no localhost listeners), re-anchors IMDS control on a node-local firewall rule with pod-scoped (IRSA-class) credentials, and lands a deployed-cluster conformance probe that fails a rollout when a documented refusal connects. "Layer 1 first" is re-scoped accordingly: multi-tenant exposure gates on Layer 2's reserved-range policy, not on Layer 1 alone.

The OSS runtime change is the seam signature and its plumbing; behavior under the shipped `AllowAll` is unchanged (it ignores the host; proxy and redirect semantics untouched). Everything else is tests and design docs.


gracexmatin:
**Re the P1 review finding — "Pass request context to Cloud policy hooks"** (resolved in 6deb459):

The finding was correct: the Layer 2 features this design promised were pinned to a seam that could not carry them. `check_ip(ip)` receives a bare IP, so it can implement reserved-range refusal but not host allowlists (no hostname), fetch quotas (no request lifecycle, and the policy may be consulted several times per fetch — each resolved address, each redirect hop), or tenant/feed-attributed audit (no identity, even with a per-tenant policy instance). Of the two remedies the finding offered, this lands mostly the second — placing allowlists, quotas, and audit outside `EgressPolicy` with named integration points — plus one minimal widening of the seam itself.

**Code — one parameter, not new hooks.** The seam's method widens from `check_ip(&self, ip)` to `check(&self, host, ip)`. Both call sites (`check_addrs` on the resolver path, and the fetcher's IP-literal hop check) already held the host, so this is parameter plumbing: no new consultation points, no lifecycle callbacks, and behavior is unchanged under the `AllowAll` default. The parameter exists so a host-based policy — a per-tenant feed allowlist — holds on every redirect hop, not just on the URL vetted at subscription time. That closes the redirect gap the Motivation itself names: a re-registered feed domain 302-ing to a public but never-vetted host passes a reserved-range check and is invisible to a subscription-time check.

**Design doc — Layer 2 split by the context each control needs.** The injected policy now carries only destination checks (reserved-range refusal, per-hop host allowlisting). Allowlist administration, fetch quotas, and egress audit are explicitly re-homed to Cloud's orchestration layer, which natively knows the tenant, the feed, and the scan outcome. Their integration points are named:

- allowlists — validated at subscription time; the same set parameterizes the per-hop policy check;
- quotas — enforced where fetches are initiated (the scan scheduler, per-tenant feed-count caps), not in a policy that cannot tell a fetch from a retry or a redirect hop;
- audit — a policy refusal already surfaces structured through `EgressDenied { host, ip, reason }` → `FetchError::Egress` → `feeds.last_status = 'error'` / `last_error`; the orchestration layer pairs that with the tenant and feed whose scan produced it.

Per-tenant **byte accounting** is recorded as an open question: no application seam sees response sizes (the policy runs before a connection exists), so it is Layer 1 (egress proxy) territory or a future fetcher-level hook.

**Considered and rejected:** full request-context/lifecycle hooks (fetch start/end callbacks, byte counts, tenant identity threaded into the fetcher). They would put permanent OSS API surface in service of Cloud-only accounting that the orchestration layer can already do without any OSS change — and the fetcher is deliberately feed-agnostic.

PR #205, opened 2026-08-12
docs(readme): reposition as a self-improving context framework

BtXin:
## Why

The README sold Skardi as an **"agent data plane."** The phrase is borrowed from cloud infrastructure and does not land with the audience we write for — agent builders, who do not think of themselves as operating a plane. It also undersold what the recent `/query` work made possible: now that every ad-hoc query arrives with declared intent (`ai_context`) and lands in a durable, fail-closed audit ledger (#173), Skardi can observe how an agent actually uses data and grow the agent's toolset out of that.

Secondary problem: 487 lines, with two overlapping essays arguing the same point (`## What is an "agent data plane"?` and `## When does a uniform data plane earn its keep?`) and a Quick Start walking a 20-line hybrid-search statement the reader could not run by copy-paste anyway.

## What

Repositioned as an **open-source self-improving context framework**, with a three-stage loop as the narrative spine:

| Stage | What it is |
| --- | --- |
| **① Observe** | `POST /query` + `ai_context` (`purpose`, `session_id`) → durable audit ledger |
| **② Learn** | read the ledger, group by session and purpose, find what recurs |
| **③ Act** | promote a recurring query into a named pipeline the agent calls as a tool |

One `## The loop` section carries the flywheel and a four-step copy-pasteable sequence, doubling as the entry path — it replaces both deleted essays *and* the top of Quick Start.

- **`data plane` is purged** from the page (0 occurrences); the engine is described as one federated SQL engine over every registered source.
- **Governance is subordinated, not deleted** — audit, lineage and snapshot-as-branch become what makes *acting* on the loop safe, one clause each rather than two sections.
- **The source table survives** with cells trimmed to their doc links. Architecture, Docker, demos and the docs index collapse into `<details>`. The standalone star section goes; the GIF moves next to Community.
- **487 → 305 lines** (37%).

## Accuracy notes

Three places where the obvious copy would have been wrong:

1. **Promotion is marked `in flight` inline.** It is delivered by the [`skardi-query-log`](https://github.com/SkardiLabs/skardi-skills/pull/25) skill, which is held until `--query-audit-db` lands in a tagged release — it is not in v0.5.0. Steps 1, 2 and 4 run today.
2. **Observe is shown with `curl`, not `skardi query`.** The CLI's request builder sends only `sql` and `max_rows` (`crates/cli/src/commands/query.rs:45-49`), so declaring intent is REST-only for now.
3. **The server starts via `cargo run --bin skardi-server`.** Only `skardi-cli` is built and packaged for releases (`.github/workflows/release.yml:119`), so a bare `skardi-server` command would not exist for a reader who downloaded a binary.

## Not in scope

No code changes — every capability claimed as shipped is verified against `0.5.0`. No other doc is rewritten: `docs/agent_data_plane.md` keeps its current name and framing (reachable from the collapsed docs index as "design background") and is addressed in a follow-up docs pass. A roadmap/status page is deliberately absent for now; the one unshipped step is marked inline instead.

## Verification

- `grep -ci "data plane" README.md` → **0**
- Every relative link in the new README resolves to a path that exists
- No dangling `#status` anchor or roadmap reference
- The step-2 request body parses as valid JSON (the first draft had a raw newline inside a JSON string)

Design doc: `docs/superpowers/specs/2026-08-12-readme-self-improving-context-design.md`


Userneima:
One onboarding concern: could we make Docker the primary path in step 1 instead of a parenthetical alternative?

Right now the first copy-paste command is `cargo run --release --bin skardi-server`, which asks a first-time user to compile the server and choose between source and Docker before they have seen Skardi work. We have already run a no-compilation onboarding path end to end: start the server from the Docker image, use the prebuilt CLI, then run the first query.

Because the server is distributed as an image while the CLI has release binaries, I suggest presenting the paths in this order:

1. **Default:** Docker image for `skardi-server` + prebuilt `skardi` CLI
2. **Alternative:** source build in a collapsed or follow-up section

That would keep the first experience copy-pasteable and remove an unnecessary technical choice from the critical path.

BtXin:
> One onboarding concern: could we make Docker the primary path in step 1 instead of a parenthetical alternative?
> 
> Right now the first copy-paste command is `cargo run --release --bin skardi-server`, which asks a first-time user to compile the server and choose between source and Docker before they have seen Skardi work. We have already run a no-compilation onboarding path end to end: start the server from the Docker image, use the prebuilt CLI, then run the first query.
> 
> Because the server is distributed as an image while the CLI has release binaries, I suggest presenting the paths in this order:
> 
> 1. **Default:** Docker image for `skardi-server` + prebuilt `skardi` CLI
> 2. **Alternative:** source build in a collapsed or follow-up section
> 
> That would keep the first experience copy-pasteable and remove an unnecessary technical choice from the critical path.

Will have a separate PR for that. This PR will focus on re-position.

README.md
```diff
@@ -38,449 +41,264 @@
...
+   read the ledger, group by session and purpose,                │
+   find the questions that keep coming back                      │
+     │                                                          │
+     ▼                                                          │
+ ③ ACT                                                          │
+   promote a recurring query into a named pipeline —             │
+   a REST endpoint + shell verb the agent calls by name          │
+     │                                                          │
+     └──────────────────────────────────────────────────────────┘
+                    the toolset grew itself
```
abbccdda: I'd prefer we use cc to draw a triangle diagram for better illustration

BtXin: What about the concept itself? 
I think the ACT should not only stop at the REST pipeline endpoint, WDYT?

README.md
```diff
@@ -3,12 +3,15 @@
 
 <img src="asset/logo.png" alt="Skardi Logo" width="700">
 
-**Skardi is an open-source agent data plane** — parameterized SQL templates served as REST endpoints (and shell verbs) your agent calls as tools, turning *data autonomy* (letting the agent decide what to query and write) into a default you can govern.
+**Skardi is an open-source self-improving context framework.**
```
abbccdda, line 6: After reading the headline, I feel we are missing the goal for the definition of "self-improving", could you clarify that? Is it performance of the system, e2e delivery latency, resource provisioning or something else?

README.md
```diff
@@ -38,449 +41,264 @@
...
-    - name: warehouse
-      type: postgres
-      connection_string: "postgresql://localhost:5432/warehouse"
-      hierarchy_level: catalog           # …or auto-discover every table in the DB.
-                                         # Reference catalog tables in SQL as
-                                         # `warehouse.<schema>.<table>` (3-part name).
+  query: |                                   # the varying window, parameterized
+    SELECT plan, COUNT(*) AS cancels FROM warehouse.public.subs
+    WHERE cancelled_at > now() - ({days} * interval '1 day')
+    GROUP BY plan ORDER BY cancels DESC
```
abbccdda, line 109: **This SQL doesn't plan.** `{days}` is substituted as a typed literal (`params.rs` parses `days=7` as the JSON number `7`), so the engine sees `7 * interval '1 day'` — and DataFusion 52.5.0, the version pinned in `Cargo.lock`, rejects integer↔interval multiplication in either order.

Verified against a standalone `datafusion 52.5.0` binary:

```
OK        SELECT now() - interval '7 days'                 -- step 2, fine
PLAN-FAIL SELECT now() - (7 * interval '1 day')
          Error during planning: Cannot get result type for temporal operation
          Int64 * Interval(MonthDayNano)
PLAN-FAIL SELECT now() - (interval '1 day' * 7)            -- same error, reversed
```

This is the payoff example of the whole page — the artifact the loop produces — so it should be one a reader can paste. Something that does plan, e.g. parameterising the cutoff instead of the window:

```sql
SELECT plan, COUNT(*) AS cancels FROM warehouse.public.subs
WHERE cancelled_at > {cutoff}::timestamp
GROUP BY plan ORDER BY cancels DESC
```

or keep `{days}` and do the arithmetic in seconds. Either way, worth actually running whatever replaces it.

README.md
```diff
@@ -38,449 +41,264 @@
...
+**1 — Serve your data, with the ledger on.** One `ctx.yaml` names your sources
+([full reference](docs/server.md)); `--query-audit-db` turns on the durable
+record that makes the rest of the loop possible.
 
 ```bash
-$ skardi run wiki-search-hybrid -p query="turing machines" -p limit=10  # shell tool, any Bash-tool agent
-$ curl -X POST :8080/wiki-search-hybrid/execute -d '{...}'              # same pipeline, served as REST
+# from a source checkout — or run the Docker image with the same flags
+cargo run --release --bin skardi-server -- \
+  --ctx ctx.yaml --query-audit-db ./audit.db --port 8080
```
abbccdda: **"or run the Docker image with the same flags" isn't true today.** `--query-audit-db` is not in any tagged release — `git show v0.5.0:crates/server/src/config.rs | grep -c query-audit-db` → `0`. The published `ghcr.io/skardilabs/skardi/skardi-server:latest` manifest is only pushed by `release.yml`, which is `workflow_dispatch` on a version tag, so that image will reject the flag with an unknown-argument error rather than starting.

This is the same constraint the PR already acknowledges for step 3 — but it bites step 1 too, which means *no* stage of the loop is reachable from a released artifact right now, only from a source checkout. The step-1 comment says "from a source checkout" and then immediately offers Docker as an equivalent; I'd either drop the Docker clause here or qualify it ("Docker once `--query-audit-db` ships").

README.md
```diff
@@ -38,449 +41,264 @@
...
 ```
```
abbccdda, line 79: **The entry path lost the thing it depends on.** The old Quick Start showed a `ctx.yaml` inline; this rewrite reduces it to "One `ctx.yaml` names your sources ([full reference](docs/server.md))". A reader following the four steps in order now has to leave the page and reverse-engineer the file format before command 1 can run — and `docs/server.md` is a long reference, not a starter template.

It compounds in step 2: `warehouse.public.subs` is a three-part catalog name that only resolves if the reader registered a Postgres source *named* `warehouse` with `hierarchy_level: catalog`. The deleted `ctx.yaml` example taught exactly that, with a comment on that exact line. Without it the naming looks arbitrary.

A six-line `ctx.yaml` (one `postgres` entry with `hierarchy_level: catalog`) restores the sequence's copy-pasteability for maybe 8 lines of page budget — cheap relative to the 182 lines this PR reclaims.

README.md
```diff
@@ -3,12 +3,15 @@
 
 <img src="asset/logo.png" alt="Skardi Logo" width="700">
 
-**Skardi is an open-source agent data plane** — parameterized SQL templates served as REST endpoints (and shell verbs) your agent calls as tools, turning *data autonomy* (letting the agent decide what to query and write) into a default you can govern.
+**Skardi is an open-source self-improving context framework.**
 
-**Federated** · one engine over every source &nbsp;·&nbsp; **Governed** · semantic overlay, lineage, branching &nbsp;·&nbsp; **Agent-native** · REST + shell + MCP-soon
+Point it at your data and your agent can ask anything of it in SQL — declaring *why* it asks. Skardi records every ask, finds the questions that keep coming back, and promotes them into named tools the agent calls next time. The agent you ship on Monday is better at your data by Friday, and you never wrote another integration.
 
+**Observe** · every query, with intent &nbsp;·&nbsp; **Learn** · what recurs across sessions &nbsp;·&nbsp; **Act** · new tools, installed for you
```
abbccdda: **The header states as shipped what the page later marks `in flight`.** "Skardi records every ask, finds the questions that keep coming back, and promotes them into named tools" and "**Act** · new tools, installed for you" are unqualified present tense, as is "better at your data by Friday." But per the design doc, ② is *by hand today* and ③ is *in flight*, blocked on an unreleased flag — a reader only learns this ~90 lines down, after deciding whether to keep reading.

I don't think the answer is to retreat from the positioning; leading with the loop is the right call and the loop is a real differentiator. But the honesty currently lives in the wrong place. One clause in the rhythm line (`**Act** · promote what recurs — *automation in flight*`) or a half-sentence in the intro paragraph would cost almost nothing and moves the caveat to where the claim is made. As written, the reader who checks and finds two-thirds of the flywheel is manual gets a worse impression than the honest version would have given them.

README.md
```diff
@@ -3,12 +3,15 @@
...
+
+<a href="#the-loop">How the loop works</a> •
+<a href="#install">Install</a> •
```
abbccdda, line 14: **Dropping `## Roadmap` orphaned two links outside this diff.** The verification in the design doc ("no dangling reference to the removed roadmap page or status anchor") was run within `README.md` only; repo-wide, two live docs still point at the anchor this PR deleted:

- `docs/agent_data_plane.md:238` — "The full public roadmap … lives in the [main README](../README.md#roadmap)"
- `docs/chunk.md:44` — "Token-based and code-aware splitters are planned follow-ups. See the [roadmap](../README.md#roadmap)."

Both now silently land at the top of the README. Nothing catches this — `docs.yml` only runs rustdoc, there's no markdown link check in CI — so it has to be fixed by hand here. `grep -rn 'README.md#' --include='*.md' .` is the sweep.

README.md
```diff
@@ -38,449 +41,264 @@
...
+
+Reference: [server](docs/server.md) (ad-hoc queries, `ai_context`, the audit
+ledger) · [CLI](docs/cli.md) · [pipelines](docs/pipelines.md) ·
+[jobs](docs/jobs.md) · [semantics](docs/semantics.md) ·
+[federated queries](docs/federated-queries.md) ·
+[embeddings](docs/embeddings/) · [chunking](docs/chunk.md) ·
+[ONNX](docs/onnx_predict.md) · [LLM extraction](docs/llm_extract.md) ·
+[JSON packing](docs/json_pack.md) · [auth](docs/auth/) ·
+[observability](docs/observability.md) ·
+[design background](docs/agent_data_plane.md).
```
abbccdda: **The purge stops at the page boundary.** "data plane" is 0 in the README, but this link — the one doc reachable from the collapsed index under neutral text — opens with `# Why an agent data plane`, and its thesis paragraph ends "**That is what Skardi is.**" So the first reader who follows the design background gets the retired positioning stated more confidently than the new one, plus one of the dangling `#roadmap` links above.

The design doc calls the rename a later pass, which is fine. But a one-line note at the link ("design background — predates the current framing") would stop the contradiction from reading as an oversight, and costs nothing now.

abbccdda (commented):
Reviewed the repositioning against the tree, the `v0.5.0` tag and a live DataFusion 52.5.0 run.

**The reframing itself holds up.** "Self-improving context framework" is grounded in work that actually shipped (`ai_context` + the audit ledger), the two overlapping data-plane essays were genuinely redundant, and the length cut (487 → 305) lands without losing the source table, which is the page's credibility. The mechanical claims I could check are accurate: `/:name/execute` and `/health/:name` routes, `--ctx` / `--pipeline` / `--port` / `--query-audit-db` flags, the `(session_id, created_at DESC)` index, `skardi run -p NAME=VALUE`, the CLI request builder sending only `sql`/`max_rows`, `pg_fts` behind the new "FTS" note on the Postgres row, DDL-always-rejected + `access_mode`, every relative link resolving, and 0 occurrences of "data plane". The `docs/server.md#query-confidentiality` anchor resolves.

What I'd fix before merge, in order: the step-3 SQL does not plan on the pinned DataFusion; the Docker aside in step 1 points at an image that cannot accept `--query-audit-db`; and removing `## Roadmap` orphaned two links that live outside this diff. Details inline.

One structural note that isn't inline-able: the page's four-step sequence is presented as the entry path, but `## Install` sits ~70 lines *below* it, so the reader meets `cargo run --release --bin skardi-server` before being told to clone anything. Worth considering whether Install moves above `## The loop`, or step 1 gets the `git clone` line.

README.md
```diff
@@ -38,449 +41,264 @@
...
+| MongoDB | Full | No | Collections with point lookups | [docs](docs/mongo/) |
+| Redis | Full | No | Hashes mapped to SQL rows | [docs](docs/redis/) |
+| DynamoDB | Full | Yes | Scan + filter pushdown | [docs](docs/dynamodb/) |
+| SeekDB | Full | Yes | MySQL-wire CRUD, FULLTEXT, HNSW | [docs](docs/seekdb/) |
+| ClickHouse | Read | Yes | Columnar OLAP, filter/limit pushdown | [docs](docs/clickhouse/) |
+| Lance | Read + job-write | No | KNN, BM25 FTS; job destination | [docs](docs/lance/) |
+| Apache Iceberg | Read | No | Schema evolution, partition pruning | [docs](docs/iceberg/) |
+| InfluxDB 3 | Read | No | Time series over Arrow Flight SQL | [docs](docs/influxdb/) |
+| S3 / GCS / Azure | Read | No | CSV, Parquet, Lance in object stores | [docs](docs/S3_USAGE.md) |
+| CSV / Parquet / JSON | Read | No | Local or remote files | [docs](docs/server.md) |
```
abbccdda: **`JSON` isn't a supported source type.** `DataSourceType` has `Csv`, `Parquet`, `Postgres`, `Mysql`, `Sqlite`, `Iceberg`, `Mongo`, `Redis`, `Lance`, `Seekdb`, `Influxdb`, `Clickhouse`, `Documents`, `Dynamodb`, `Rss`, `OpenConnector` — no JSON variant, so `type: json` fails config deserialization.

The claim is inherited (the old table had a `JSON / NDJSON` row), so not introduced here — but this PR merges three rows into one and re-points the link, and `docs/server.md` never mentions JSON either (the old row at least pointed at `docs/cli.md`, which also doesn't). Since the row is being rewritten anyway, dropping `/ JSON` is the cheap fix.

README.md
```diff
@@ -38,449 +41,264 @@
...
+     │                                                          │
+     ▼                                                          │
+ ② LEARN                                                        │
+   read the ledger, group by session and purpose,                │
+   find the questions that keep coming back                      │
+     │                                                          │
+     ▼                                                          │
+ ③ ACT                                                          │
+   promote a recurring query into a named pipeline —             │
+   a REST endpoint + shell verb the agent calls by name          │
```
abbccdda: Cosmetic, but this is the visual centrepiece of the new page: the box's right border is ragged. The six prose lines (56, 57, 61, 62, 66, 67) are one column wider than the border lines, because the padding assumes `①`/`▼`/`▶` render double-width. They're East-Asian-ambiguous, so most monospace fonts — GitHub's included — give them one cell, and the `│` on those lines sticks out by one. Under a double-width font the misalignment just moves to the arrow lines instead.

Simplest fix is to drop one space of padding on those six lines and check it in GitHub's preview, or swap `①②③` for `(1)(2)(3)` and `▼▶` for `v`/`>` so every glyph is unambiguously one cell.

README.md
```diff
@@ -38,449 +41,264 @@
...
+**CLI** — pre-built for `x86_64`/`aarch64` Linux and Apple Silicon:
 
 ```bash
-cargo run --bin skardi-server -- \
-  --ctx ctx.yaml \
-  --pipeline pipelines/ \
-  --jobs jobs/ \
-  --port 8080
+curl -fSL "https://github.com/SkardiLabs/skardi/releases/latest/download/skardi-$(uname -m | sed 's/arm64/aarch64/')-$(uname -s | sed 's/Linux/unknown-linux-gnu/' | sed 's/Darwin/apple-darwin/').tar.gz" | tar xz
+sudo mv skardi /usr/local/bin/
```
abbccdda: Nit: the dropped platform table took a warning with it. On an Intel Mac this one-liner expands to `skardi-x86_64-apple-darwin.tar.gz`, which is never published — `release.yml` builds three targets and that isn't one. The reader gets `curl: (22)` piped into `tar`, i.e. a confusing failure rather than a clear "not published" message; the old README said so explicitly. "pre-built for `x86_64`/`aarch64` Linux and Apple Silicon" is technically complete, but the failure mode is silent enough to be worth a trailing clause ("Intel Macs: build from source").

Userneima:
Heads-up on the `## Install` section — this is inherited from `main`, not introduced by this PR, but it went stale a few minutes ago and this PR is the one rewriting that section.

`skardi-skills` [#24](https://github.com/SkardiLabs/skardi-skills/pull/24) just merged. It folds `auto_knowledge_base` and `auto_rag` into a single `auto_context` and retires `skardi_on_sealos`, so the marketplace now publishes two plugins:

```bash
gh api repos/SkardiLabs/skardi-skills/contents/.claude-plugin/marketplace.json \
  --jq '.content' | base64 -d | jq -r '.plugins[].name'
# auto-context
# feishu-connector
```

All three `/plugin install` lines in this branch name plugins that no longer exist, so a reader following the page gets three consecutive failures. The two tree links below the block (`.../tree/main/auto_rag`, `.../tree/main/auto_knowledge_base`) now 404 as well.

Suggested replacement for lines 167–179:

````markdown
**Claude Code** — the fastest path. A skill that stands the whole thing up for
you:

```text
/plugin marketplace add SkardiLabs/skardi-skills
/plugin install auto-context@skardi-skills
```

[`auto_context`](https://github.com/SkardiLabs/skardi-skills/tree/main/auto_context)
turns a folder of documents — or a datastore you already run — into governed,
searchable context served over HTTP by `skardi-server`: hybrid search (vector +
FTS + RRF), defaulting to a local SQLite file the skill creates and owns, or
pointed at Postgres + pgvector, MongoDB, or Lance. For Cursor and other
[Agent Skills](https://agentskills.io/)-compatible hosts, see the
[skardi-skills README](https://github.com/SkardiLabs/skardi-skills#installation).
````

Two notes on the wording: `Skills that` becomes `A skill that` since there is only one now, and the two sentences that separately described "server-backed RAG" and "local, zero-infra" collapse into one — after skardi#170 removed the query engine from the CLI, both paths need a server and the only remaining difference is whose database it is. That was the reason for the merge, so the page reads better for it.

Happy to take this as a separate PR against `main` after this one lands if you'd rather not widen the diff — but the same lines are stale on `main` right now either way, so whichever of us gets there first is fine. `feishu-connector` is deliberately left out; adding it is a separate call, not part of fixing the break.

**Review by Userneima — COMMENTED** [review]
The repositioning and the new loop structure are much clearer, and the `self-improving context framework` direction is aligned. I have three changes to make before merging:

1. **Privacy boundary:** The audit ledger stores raw SQL, including literals. Please add an explicit operator-choice boundary around handing a ledger window to an LLM: use local/redacted analysis, or make sending it to a remote LLM opt-in. As written, it reads like a recommended default and can expose user data.

2. **Dead link:** This PR deletes `docs/agent_data_plane.md`, but `demo/llm_wiki/README.md` still links to it. Please either retain/replace the design-background document or update that inbound link in the same PR.

3. **Shipped vs. in-flight:** The header says Skardi turns recurring intentions into tools and routines, while the automation still depends on the open skardi-skills#25. Could the first-screen wording be made prospective/capability-framed, with the same shipped/in-flight boundary as step 3?

Once these are addressed and CI is green, I am comfortable merging.

**Comment by Userneima** [comment]
The three review points are addressed and CI is green.

One blocker before merge: the branch conflicts with `main` (`mergeable: false`, state `dirty`). Two files, both from work that landed on 8-24, after this PR's merge-base (`97859f8`):

- **`README.md`** — #223 added a OneDrive pack row to the data-source table (`Open Connector` line). This PR's rewrite predates it, so the row is absent on this side. Keep `main`'s version of that row.
- **`docs/jobs.md`** — #224 added the context-precedence note in two places (`mode: server` / `mode: cloud`, pointing at `cli.md — Connecting`). This PR removes 4 lines from the same file. Keep `main`'s additions.

Both sides modify different parts of each file, so keeping both resolves the conflict.

Should I rebase, or will you?

**Review by abbccdda — APPROVED** [review]

File: README.md [hunk 12]
```diff
@@ -38,448 +42,297 @@
...
+file indexed on `(session_id, created_at)`.
 
 ```yaml
-# pipelines/search_hybrid.yaml — declares the SQL once; Skardi infers the params
+# /skardi-query-log --db ./audit.db --dir pipelines/ --port 8080 --restart-cmd '...'
+#   → 4 sessions asked the same churn question with different windows
+#   → wrote pipelines/weekly-churn.yaml, reloaded, /health/weekly-churn ok
 kind: pipeline
-metadata: { name: wiki-search-hybrid }
+metadata: { name: weekly-churn }
```
Userneima: **The step-3 pipeline does not load — two fatals, both before any query runs.** This one wasn't in the earlier round, so it hasn't been addressed yet. Copy-pasted as written, the YAML fails twice over.

**1. `metadata` is missing `version`.** `ComponentMetadata` (`crates/skardi/src/pipeline/types.rs:146-152`) declares `pub version: String` — no `Option`, no `#[serde(default)]`. With `metadata: { name: weekly-churn }` the server refuses to boot:

```
Pipeline loading failed: Failed to parse metadata section: missing field `version`
```

**2. With `version` added, `CAST(concat({days}, ' days') AS INTERVAL)` still fails at load.** The inferencer rewrites `{days}` to a bare `?` before planning (`inferencer.rs::convert_named_to_placeholders`). `concat` skips the unbound placeholder, so `simplify_expressions` folds the argument down to `' days'` and tries to parse that as an interval:

```
Pipeline loading failed: Failed to create logical plan: Optimizer rule 'simplify_expressions' failed
Arrow error: Parser error: Invalid input syntax for type interval: " days"
```

Neither is a query-time error — `skardi-server` exits during config load, so the reader who follows the four steps in order loses the whole server, not just this pipeline. This is also the regression path from the last round: the old `{days} * interval '1 day'` failed when the query ran; this form fails earlier and harder.

**Suggested fix** — parameterise the whole interval rather than splicing into `concat`:

```yaml
kind: pipeline
metadata:
  name: weekly-churn
  version: "1.0.0"
spec:
  query: |                                   # the varying window, parameterized
    SELECT plan, COUNT(*) AS cancels FROM warehouse.public.subs
    WHERE cancelled_at > now() - CAST({window} AS INTERVAL)
    GROUP BY plan ORDER BY cancels DESC
```

with step 4 becoming `skardi run weekly-churn -p window='7 days'` and `-d '{"window": "7 days"}'`. Verified end to end: the pipeline loads and `execute` returns the expected rows.

If keeping `days=7` in step 4 is worth more than the readability, `to_timestamp(extract(epoch from now()) - {days} * 86400)` also loads and returns the same rows, and leaves both step-4 commands exactly as they are.

**How this was checked.** Run against the `v0.5.0` server binary with a two-row CSV source standing in for `warehouse.public.subs`. That binary is valid evidence for this branch: `git diff v0.5.0 d39febe -- crates/skardi/src/pipeline/inferencer.rs` is empty, `ComponentMetadata` is unchanged, and both sides pin DataFusion 52.5.0 — the load path is identical. Step 2's `interval '7 days'` literal is fine; that one runs.

PR #206, opened 2026-08-12
feat(query-audit): extract the ledger crate + nullable identity columns

powerfulZ:
Extracts `query_audit` into `crates/query-audit` (`skardi-query-audit`, re-exported under the old path — zero call-site changes) and adds five nullable identity columns (`request_id`, `org_id`, `workspace_id`, `user_id`, `run_id`) with `record_started_for(..., Option<&QueryIdentity>)`.

**Why:** the cloud engine (skardi-cloud) is required by its design to adopt this ledger *wholesale* — same file format, same `--query-audit-db` flags, same fail-closed semantics — with its caller identity in nullable columns rather than a forked store ("upstreamed, not forked"). The extraction is what makes that possible without dragging the whole server crate into its dependency tree; the columns are what its envelope fills. OSS behavior is byte-identical: `record_started` passes `None`, rows read NULL.

Existing databases migrate on open (`PRAGMA table_info` + additive `ALTER`s, idempotent — `CREATE TABLE IF NOT EXISTS` alone cannot retrofit a live table); covered by an old-schema fixture test.

Tests: skardi-query-audit 9/9 (migration, double-open idempotence, NULL-for-OSS-rows); skardi-server 168/168 single-threaded (the parallel run trips a pre-existing env-race flake in `auth::layer::tests::build_better_auth_respects_auth_base_url` — unrelated, it reads process env that neighbouring tests mutate).

Consumed by skardi-cloud's E5 (`/query` endpoint) PR, which pins this branch's rev and bumps to main's after this merges.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9552,6 +9552,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "skardi-query-audit"
+version = "0.5.0"
+dependencies = [
+ "anyhow",
+ "chrono",
+ "serde_json",
+ "tempfile",
+ "tokio",
+ "tokio-rusqlite",
+]
+
+[[package]]
 name = "skardi-server"
 version = "0.5.0"
 dependencies = [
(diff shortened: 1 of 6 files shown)


abbccdda (approved):
Approved, although is this a refactor or a feat as a PR?

PR #203, opened 2026-08-11
feat(graph): milestone 1 — Apache AGE read-only cypher_query + graph_schema

bakey:
Milestone 1 of the [graph engine bypass design](https://github.com/SkardiLabs/skardi/pull/199): **read-only openCypher against Apache AGE (Postgres) as SQL tables** — `cypher_query(connection, cypher, params, columns)` with caller-declared columns and a planning-time-stable schema, `graph_schema(connection)` for agent discovery, backend-enforced read-only, and bounded queries. Skardi parses no Cypher and stores no graphs; AGE owns storage and traversal.

## What shipped (crates/skardi/src/sources/providers/graph/)

| Module | Contents |
|---|---|
| `error.rs` | Typed taxonomy: `MutationRejected{keyword, offset}` (never query text), `TypeMismatch` carrying JSON **kinds** never values, `RowCapExceeded`/`Timeout`, `Backend` with verbatim code + byte-capped snippet |
| `guard.rs` | Fast-path keyword guard (word-boundary `CREATE/SET/DELETE/DETACH/REMOVE/MERGE/DROP` + `CALL` + `LOAD`), caller Cypher only, at **plan time** — UX, not the boundary |
| `config.rs` | `graph:` block; pure validation: age-only, `postgres://` scheme allowlist, identifier-shaped `graph_name`, env-var-NAME-shaped credentials, bounds defaults (30s / 10 000 rows) |
| `value.rs` | Backend-shared conversion: agtype annotation-stripping parser (string-literal-aware), the friendly lowercase type vocabulary (`string/int/float/bool/json/node/relationship/path`), type-directed **batch-atomic** Arrow conversion, canonical STRUCT shapes (path = parallel typed lists, nodes one longer) |
| `client.rs` | `GraphClient` trait + `AgeClient`: every call in a `READ ONLY` transaction, `SET LOCAL statement_timeout` server-side + client wrap, row cap enforced while streaming |
| `udtf.rs` | The two UDTFs, registered once, connection resolved by name (the `open_connector_query` shape); `datafusion-functions-json` 0.52.0 getters registered alongside (verified against pinned DF 52 — the design's preferred route, no in-tree fallback needed) |

## Deliberate deviations / decisions, argued in module docs

- **sqlx, not tokio-postgres**: the workspace already ships sqlx-postgres; one Postgres stack beats two. Recorded at the top of `client.rs`.
[...]

PR #207, opened 2026-08-12
feat(graph): milestone 2 — Neo4j backend over Bolt, read-mode proven at registration

bakey:
Implements **Milestone 2** of the [graph engine bypass design](../blob/feature/graph-neo4j-m2/docs/superpowers/specs/2026-08-08-graph-engine-bypass-design.md): the Neo4j backend over Bolt, behind the same `GraphClient` trait as M1's AGE client.

> **Stacked on #203** (`feature/graph-age-m1`) — draft until the base merges. The diff shown here is M2 only.

## The access-mode spike (the milestone's hard precondition)

The design gates M2 on the pinned driver being able to express Bolt's READ access mode. Findings, verified in `neo4rs`'s source at the pinned version and recorded in `neo4j.rs`'s module doc:

- **0.8.0 (latest stable) has no read-mode channel at all** — no `execute_read`, no `Operation`, and its `BEGIN` carries no mode. M2 pins **0.9.0-rc.10** (default features).
- **The RC sends `mode: "r"` in the RUN extra** via `Graph::execute_read` — the spelling the Bolt spec honors for **auto-commit** transactions. Every M2 query is therefore one auto-commit READ transaction.
- **Explicit transactions are a dead end on this driver**: its `Begin` builder supports `with_mode`, but `Txn::new` never calls it (BEGIN always goes out `"w"`), and Bolt ignores RUN-extra mode inside an explicit transaction — so the `unstable-bolt-protocol-impl-v2` feature would add unstable surface without adding enforcement. It stays off.

## What shipped

- **`Neo4jClient`** (`graph/neo4j.rs`): auto-commit READ transactions; registration runs a reachability probe plus the **read-mode proof** — a self-erasing `CREATE (n:…) DELETE n` through the exact read channel every later query uses, passing ONLY on the server's access-mode refusal (live-pinned: `Neo.ClientError.Statement.AccessMode`). An executed write fails registration; so does any *other* error (a transient failure proves nothing about enforcement — fail closed).
- **By-NAME column binding**: Bolt records carry field names, so declared columns bind to RETURN aliases by name — AGE's positional-order footgun does not exist here, and a declared name the query never returns is a typed error naming it. Demonstrated live with out-of-order declarations (the exact shape that silently swaps on AGE).
- **JSON-`record` fallback**: omitting `columns` yields one `record: Utf8` column carrying each whole record as canonical JSON (keys sorted — the driver's rows are hash maps). On AGE the omission stays a targeted error, now worded by the backend itself (`declared_columns_requirement()` on the trait).
- **`graph_schema` grows property discovery**: output is now `(label, kind, property, property_type)` — Neo4j fills the property columns from `db.schema.nodeTypeProperties()` / `relTypeProperties()` (one row per label × property; property-less labels and label-less node types keep their rows), AGE serves them as null, structurally. The two catalog queries run concurrently; bounds semantics match the AGE client's verbatim (limit-under-cap is a clean stop; overflow is loud) and are pinned live.
- **Bolt decoding into the shared canonical shapes** (`value.rs` unchanged as the converter): multi-label nodes (`labels` array accepted alongside AGE's single `label`), paths reconstructed in traversal order with directions resolved from the Bolt index list (reversed hops pinned by unit test), temporals as ISO-8601 text (negative durations decoded through the driver's signed serde view — its `std::time::Duration` conversion wraps negatives to ~1.8e19), points as JSON objects, bytes as hex, non-finite floats as null (the agtype decoder's decision, honored across backends).
- **Bounds**: `tx_timeout` rides the RUN extra and the server kills at the bound — live verification surfaced the actual code (`Transaction.TransactionTimedOutClientConfiguration`; a suffix match would have missed it) → typed `Timeout`. Row cap and LIMIT semantics identical to AGE. Params are driver-bound (never interpolated — the design's normal road, which AGE alone cannot take); an integer beyond i64::MAX is a typed refusal, never a silent f64 narrowing.
- **Config**: `backend: neo4j` with the design's scheme allowlist (`bolt`/`bolt+s`/`neo4j`/`neo4j+s`; `+ssc` deliberately excluded); `graph_name` becomes per-backend (required on age, optional database selector on neo4j); URL-embedded-password rejection covers bolt URLs unchanged. Validation rules are one `BackendRules` row per backend so milestone 3 is one new row, and the registration dispatch has no wildcard arm (`unreachable!` instead of silently dialing the wrong driver).

## Self-review round (second commit)

An 8-angle review ran over the M2 diff before submission; every finding was verified before acting. Highlights fixed: the read-mode proof was fail-open (any error passed it); `graph_schema` had an off-by-one cap and tripped `RowCapExceeded` on small LIMITs where AGE returns cleanly; its tx_timeout mapped to a generic backend error; negative durations wrapped; u64 params narrowed silently; plus a dedup pass (shared `read_env`/`bounded`, one `rel_json` spelling, shared live-test helpers). Consciously rebutted, with reasons in the commit message: full-result buffering (explicitly allowed by the design text, same as AGE), by-value `bolt_to_json`, and a config `Backend` enum reshape (M3 work if kuzu diverges structurally).

## Verification

- **63 graph unit tests** (`sources::providers::graph`) — new coverage: Bolt decode incl. reversed-hop and zero-hop paths, malformed-index null-not-panic, negative/mixed-sign durations, params conversion + overflow refusal, multi-label and label-less node shapes, record fallback at the UDTF seam, per-backend `graph_schema` shape, config allowlists both directions.
- **Neo4j live suite (new, 7 tests) — 7/7 green** against dockerized `neo4j:5`: by-name binding on out-of-order declarations; canonical shapes incl. a real multi-label node and a temporal property; record fallback with sorted keys queried in place via `json_get_str`; **read-only enforced by the server past a bypassed keyword guard, with the write proven absent afterwards**; `graph_schema` property discovery (`KNOWS since:Long`, `Person name:String`); row cap + LIMIT + typed timeout (the live proof that RUN-extra `tx_timeout` lands) + schema cap semantics; credentials never in error text.
- **AGE live suite re-run green (9/9)** after the shared-trait changes; full lib suite **940 passed**; doctests green; fmt clean; the touched files are clippy-clean (the branch carries pre-existing findings elsewhere that main has since addressed).
- **CI**: the coverage job gains a `neo4j:5` container step mirroring the AGE one (`-E 'binary(graph_neo4j_live)' --no-tests=fail -- --ignored`).

## Design doc updates

The M2 section records the spike outcome and both live-verified error codes; the §Backend abstraction node-identity passage is corrected — the pinned driver's stable protocol implementation negotiates Bolt 4.4, which predates `elementId()`, so ids are the stringified numeric ids (same stability contract; `elementId()` stays available explicitly via `RETURN elementId(n)`).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -695,6 +695,42 @@ jobs:
         if: always()
         run: docker rm -f skardi-age || true
 
+      # Neo4j gives the graph provider's Bolt path the same real
+      # coverage — the milestone-2 backend, incl. the registration
+      # read-mode proof, which only a real server can answer.
+      - name: Start Neo4j for graph live tests
+        run: |
+          docker run -d --name skardi-neo4j \
+            -p 127.0.0.1:17687:7687 -p 127.0.0.1:17474:7474 \
+            -e NEO4J_AUTH=neo4j/skardineo4j \
+            neo4j:5
+          for i in $(seq 1 90); do
+            if curl -sf http://127.0.0.1:17474 >/dev/null 2>&1; then
+              echo "neo4j ready after ${i}s"
+              break
+            fi
+            if [ "$i" = "90" ]; then
+              echo "neo4j failed to become ready"
+              docker logs skardi-neo4j
+              exit 1
+            fi
+            sleep 1
+          done
+          # The HTTP port answers before Bolt finishes auth setup; give
+          # the server a moment to open Bolt too.
+          sleep 5
+
+      - name: Execute graph Neo4j live tests
+        env:
+          SKARDI_NEO4J_LIVE_URL: bolt://neo4j:skardineo4j@127.0.0.1:17687
+        run: |
+          cargo llvm-cov --no-report nextest --all-features \
+            -E 'binary(graph_neo4j_live)' --no-tests=fail -- --ignored
+
+      - name: Stop Neo4j
+        if: always()
+        run: docker rm -f skardi-neo4j || true
+
       - name: Generate coverage report (lcov)
         run: cargo llvm-cov report --lcov --output-path lcov.info
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -661,7 +661,13 @@ jobs:
       # the AgeClient/register_graph_source code is exercised ONLY by the
       # graph_age_live suite, which self-skips without this backend (the
       # PR #203 codecov report showed client.rs at 11% for exactly that
-      # reason). Same docker-run pattern as MinIO above.
+      # reason). Docker-run like MinIO above, but the readiness poll must
+      # go over TCP: the postgres entrypoint runs a TEMPORARY
+      # unix-socket-only server for initdb, so a socket-side pg_isready
+      # can answer "ready" during bootstrap while the real server is
+      # still seconds away — an intermittent red on the whole coverage
+      # job. The bootstrap server never listens on TCP, so -h 127.0.0.1
+      # can only succeed against the real one.
       - name: Start Apache AGE for graph live tests
         run: |
           docker run -d --name skardi-age \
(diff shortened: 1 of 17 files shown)


abbccdda:
@bakey is this ready for a review?

PR #208, opened 2026-08-13
docs(mcp): design doc for the stdio MCP binding (skardi mcp)

gracexmatin:
## What

Design doc for roadmap item `5`'s MCP binding, stdio transport first: `docs/superpowers/specs/2026-08-13-mcp-stdio-binding-design.md`. Docs only — no code changes in this PR.

## Why

Hosts with a Bash tool already use the `skardi` CLI as a tool with zero MCP config; the uncovered population is hosts without a shell (Claude Desktop, ChatGPT desktop, IDE MCP integrations). stdio ships first because the CLI is already a thin HTTP client for skardi-server, so the binding is mostly a protocol translator between two existing, stable interfaces.

## Design in one paragraph

A new `skardi mcp` subcommand speaks MCP over stdio to the host and proxies to a running skardi-server over REST via the existing `ApiClient` (URL/token config inherited unchanged). Tool surface aligns with the shipped REST/shell bindings: one tool per pipeline (name/description from YAML metadata, `inputSchema` assembled from server-emitted per-parameter JSON Schema fragments), plus built-in `query` (ad-hoc SQL, same `access_mode` governance as `/query`, optional `purpose` → `ai_context` audit passthrough) and `list_data_sources` (semantics-backed discovery). SDK is the official Rust `rmcp` (`default-features = false`, `server` + `transport-io`), implementing `ServerHandler` manually since tools are dynamic. Jobs tools, streamable HTTP, and OAuth are explicit non-goals for v1.

## Decided (was an open question in earlier revisions)

**Server-side inventory enrichment — Option A.** This design requires one small, additive change in `crates/server`: `GET /pipelines` gains `description` and `parameters`, each parameter carrying a complete `json_schema` fragment computed server-side (type mapping incl. date/timestamp formats, unconditional null unions, and `VALUES {name}` multi-row detection — the server is the only party holding the SQL template). The zero-server-change alternative is documented as rejected. See the "Server-side change" section.

## Review provenance

The spec went through an adversarially-verified multi-lens review (internal consistency, factual accuracy against the code, MCP/rmcp protocol correctness, implementer completeness): 21 findings, 20 confirmed and folded back in — including an unimplementable-as-written VALUES override, a name-sanitization algorithm that violated its own 64-char constraint, and the fact that `max_rows` never bounded pipeline results. Details in the commit messages.

## Still open (last section of the doc)

1. `purpose` parameter on `query` — optional as specced, or omit from v1.
2. Built-in tool naming — bare `query`/`list_data_sources` vs `skardi_`-prefixed.
3. Bridge lifecycle knobs — concurrent dispatch, stdin-close behavior, and the fact that `ApiClient` sets no request timeout today.

### DIFF
--- a/docs/superpowers/specs/2026-08-13-mcp-stdio-binding-design.md
+++ b/docs/superpowers/specs/2026-08-13-mcp-stdio-binding-design.md
@@ -4,16 +4,29 @@
 **Date:** 2026-08-13
 **Branch:** `mcp-stdio-binding-design`
 
+**Scope — touches two crates, not one.** This is not a CLI-only feature.
+Most of the work lives in `crates/cli` (the new `skardi mcp` subcommand),
+but shipping it also requires a small, additive, **decided** change in
+`crates/server`: `GET /pipelines` gains `description` and `parameters`
+fields it does not return today (Option A — see "Server-side change" for
+the full rationale and the rejected zero-server-change alternative). A
+server-side PR is a prerequisite for this design, not an optional
+enhancement layered on top of it.
+
 ## Summary
 
 Skardi gains its third agent-facing binding: an MCP (Model Context Protocol)
 server. v1 ships as a new CLI subcommand — `skardi mcp` — that speaks MCP over
 **stdio** to a local host (Claude Desktop, Cursor, any MCP client) and proxies
-every operation to a running `skardi-server` over the existing REST API, using
-the CLI's existing `ApiClient`. The server's execution path is untouched — the
-subcommand is a protocol translator; the only proposed server edit is a small
-additive enrichment of the pipeline inventory endpoint (see "Server-side
-change").
+every operation to a running `skardi-server` over REST, using the CLI's
+existing `ApiClient`. The server's *execution* path is untouched — no
+pipeline runs, no query executes, any differently than it does today. But
+this design does require one small, additive, **decided** server-side
+change: `GET /pipelines` gains `description` and `parameters` fields it
+doesn't return today (Option A — see "Server-side change"). That change is
+a prerequisite for the bridge, not an optional enhancement — without it,
+pipeline tools would ship with no LLM-facing description and parameter
+types rendered as raw Rust Debug strings.
 
 The tool surface mirrors what REST and shell already expose:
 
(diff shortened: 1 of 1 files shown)


docs/superpowers/specs/2026-08-13-mcp-stdio-binding-design.md
```diff
@@ -0,0 +1,479 @@
...
+`crates/server`: `GET /pipelines` gains `description` and `parameters`
+fields it does not return today (Option A — see "Server-side change" for
+the full rationale and the rejected zero-server-change alternative). A
+server-side PR is a prerequisite for this design, not an optional
+enhancement layered on top of it.
+
+## Summary
+
+Skardi gains its third agent-facing binding: an MCP (Model Context Protocol)
+server. v1 ships as a new CLI subcommand — `skardi mcp` — that speaks MCP over
```
BtXin, line 19: If we already expose mcp endpoint, does agent still need to use cli command here?

gracexmatin: No — the agent never invokes a CLI command here. `skardi mcp` isn't a tool the model calls; it's the process the **MCP host** spawns once at startup (the `command`/`args` entry in `claude_desktop_config.json`, see the Architecture section). From the agent's side the only visible surface is MCP tools: the pipeline tools, `query`, and `list_data_sources`. The CLI is just the packaging and process entry point, not the call surface.

On "if we already expose an mcp endpoint": we don't yet. There is no `/mcp` route on skardi-server today (`crates/server/src/server.rs` routes: `/pipelines`, `/pipeline/:name`, `/data_source`, `/query`, `/:name/execute`, plus jobs) — streamable HTTP is an explicit v1 non-goal. v1 deliberately adds **no MCP transport to the server at all**: no `/mcp` route, no new port, no OAuth. The host spawns the bridge over stdio and the bridge proxies to an already-running skardi-server through the REST endpoints that exist today, via the CLI's existing `ApiClient` (base URL + Bearer resolution inherited unchanged). To be precise, "no server" means no new MCP transport — skardi-server is still what executes everything. That's the cheapest, fastest path to a working MCP binding, and it's why stdio ships first.

When `/mcp` does land it won't retire this: local hosts still have uneven support for remote MCP servers and their auth flow, and pointing `--server` at a remote deployment already makes `skardi mcp` a local gateway to it. The tool-projection logic is transport-independent and shared between the two.

I've pushed a commit making both points explicit in the Summary (they were only in Architecture before, which is what made line 19 ambiguous) and noting in Non-goals that streamable HTTP won't replace the stdio bridge.

BtXin (approved):

PR #212, opened 2026-08-14
feat(sources): Outlook source pack (milestone 5.7) — messages + mail_folders over nextLink cursors, live-verified

gracexmatin:
## Summary

Milestone 5.7 — the **`outlook`** pack: `messages` and `mail_folders` as stable SQL tables over Open Connector's `outlook` service (Microsoft Graph mail, OAuth, cursor pagination over `@odata.nextLink`). Everything is reconciled against a live gateway (**v1.3.4**, `open-connector` at `2410fbe`) — never Microsoft's own docs.

**Live-verified end to end against a real MSA mailbox (2026-08-19)** — see Verification below.

All five phases are done — 1/2 (contract reconciliation + table design, recorded in `docs/superpowers/specs/2026-08-14-open-connector-m365-packs-design.md`), 3 (implementation), 4 (live verification), 5 (self-review — see Review status below).

Scope note from the phase-1/2 reconciliation: there is no `microsoft365` service upstream — the gateway splits Graph into `outlook` (mail only; no calendar, no contacts), `one_drive`, and `excel`, one OAuth connection each, and a Skardi binding carries exactly one `connection_alias`. So Microsoft 365 ships **one pack per service**: `one_drive` follows as its own milestone and PR, and the whole `excel` service is **deferred at the admission gate** (its list actions emit `nextLink` but accept no `nextLink` input, so pagination cannot be completed).

## What this adds

`packs/outlook.yaml` + `packs/outlook.rs` (module doc carries every design decision with rationale), registry entry, fixtures, `docs/open-connector-microsoft-365.md`.

| Table | Action | Resources (optional) | Pagination | Filter pushdown |
|---|---|---|---|---|
| `messages` | `outlook.list_messages` (pinned `select`) | `mailFolderId` | cursor (`top` 100) | none |
| `mail_folders` | `outlook.list_mail_folders` (pinned `includeHiddenFolders: true`) | — | cursor (`top` 1000) | none |

- **`messages`** — 24 columns: identity (`id`, `web_link`, `internet_message_id`), envelope (`subject`, `body_preview`, `from_address`/`from_name`, `sender_address`/`sender_name`, recipient lists as JSON), state (`is_read`, `is_draft`, `importance`, `flag`, `categories`), threading (`conversation_id`, `parent_folder_id`, `has_attachments`), and the four Graph timestamps. Whole mailbox when unbound; one folder's listing when the binding pins the `mailFolderId` resource.
- **`mail_folders`** — 8 columns (`id`, `display_name`, `well_known_name`, `parent_folder_id`, `child_folder_count`, `unread_item_count`, `total_item_count`, `is_hidden`): the complete root-level folder set. Folder ids from this table are what a `mailFolderId` resource wants.

```yaml
bindings:
  - name: m365                   # schema name in SQL
    source_pack: outlook
    connection_alias: my-msft    # the gateway's outlook OAuth connection
    resource:                    # optional; omit for the whole mailbox
      mailFolderId: "AQMkAGE1..."
    tables: [messages, mail_folders]
```

Excluded actions, accounting for all eleven: `get_message`/`get_profile`/`get_mailbox_settings` are single-object reads; the six draft/send/settings writes are out by the read-only allowlist.

## Key design decisions

- **`select` is pinned to exactly the mapped fields on `messages`, with `page_size: 100`** (not the schema's 1000 ceiling). Unpinned, Graph returns its default full representation — `body.content` (HTML mail) on every row — which at the declared `top` ceiling would blow the client's 16 MiB response cap. Pinned, payloads shrink by orders of magnitude, rows become deterministic, and a misspelled passthrough column turns into a loud Graph 400 instead of an always-NULL column. Cost: the pin and the column set must move together — `select_pin_mirrors_the_mapped_columns` enforces that mechanically. `body` is deliberately outside both (full message content belongs behind a future content-oriented surface, as on the Gmail pack).
- **The cursor is a URL.** The gateway validates `nextLink` as `format: uri` before credentials, and the executor pins host `graph.microsoft.com` plus an allowlisted path set — so every fixture and mock cursor is URI-shaped: an opaque `"cursor-2"` token would pass the mocks and 400 against the real gateway. Termination is the executor's explicit `null` (absent/empty also terminate — one e2e per spelling). The executor ignores all other inputs once `nextLink` is present (the link embeds its own `$top`/`$select`), pinned in the cursor e2e.
- **Raw Graph passthrough rows, heavily under-declared upstream**: `list_messages` declares 15 properties and no date field of any kind — `receivedDateTime` and twelve other `messages` columns ride `additionalProperties` passthrough outside the fingerprint gate, kept an explicitly reviewed set by the coverage-gap pin (thirteen on `messages`, one on `mail_folders`).
- **`well_known_name` on `mail_folders`** — promoted from an unmapped wire extra after the live pass (Owen-approved 2026-08-19): the live mailbox's `displayName`s were all CJK, so cross-account folder semantics need Graph's locale-independent discriminator (`inbox`/`sentitems`/…, explicit null on custom folders). The other wire extras (`@odata.etag`, `sizeInBytes`) stay deliberately unmapped.
- **Zero filter pushdown, structurally.** Outlook exposes filtering only as `filter`, a raw OData *expression* string that a per-column `FilterMapping` cannot compose (Notion precedent). Predicates re-apply locally after the bounded fetch; the scoping tools are the `mailFolderId` resource and `LIMIT` early-stop. `filter`/`orderby`/`bodyContentType` never reach the wire — negative-space guarded by exact-key-set assertions in every e2e.
- **`mail_folders` pins `includeHiddenFolders: true`** (the `state=all` move) and keeps the distinction queryable via `is_hidden`; root-level only (the executor does not recurse; `child_folder_count` reveals nesting).
- **The OAuth consent is read-write for a read-only pack.** The service's scope union requests `Mail.ReadWrite` + `Mail.Send` + `MailboxSettings.ReadWrite`, and the read actions themselves declare `requiredScopes: [User.Read, Mail.ReadWrite]` although `Mail.Read` suffices at the Graph level — the gateway would refuse a correctly-scoped read-only token. Nothing Skardi-side can narrow the consent screen; the pack doc says so plainly, the tables stay read-only by construction, and the misdeclaration is an upstream issue candidate (same class as open-connector#268).
- **Input contracts are committed artifacts**: the registration gate is output-only, so the pack ships the input half of its own capture (`fixtures/outlook/contracts/inputs/`, cross-checked byte-identical against per-action discovery) plus `generated_inputs_are_accepted_by_the_captured_input_contracts` — drift is caught on re-capture; a registration-time input fingerprint remains tracked engine work.

## Verification (phase 4, live pass 2026-08-19)

Against the pinned gateway (v1.3.4, `2410fbe`) and a real personal (MSA) mailbox over OAuth, end to end through a skardi-server SQL scan:

- **Every mapped column carried real non-NULL values** across a seeded mailbox (9 messages, 9 root folders): all 22 selected message fields arrived with the pinned spellings — including all four passthrough timestamps, `hasAttachments`, `conversationId`, `categories`, and the `emailAddress` nesting — and all folder fields. Under the select pin no field was ever absent or null; emptiness is spelled `""`/`[]`.
- **Live discovery schemas were byte-identical** to the committed contract captures (both actions, both halves); registration passed the fingerprint gate against live discovery.
- **Pagination**: a forced `top=2` walk followed real cursors five pages to a genuinely-null terminal `nextLink`. `top=1000` is accepted on the wire; `top=1001` is a schema 400 before credentials — unlike Feishu, the declared bound IS the wire bound. A foreign-host cursor is refused with "nextLink must target graph.microsoft.com".
- **Select misspellings fail loudly**: a bad field name returns a Graph 400 naming the property, surfaced through the gateway failure envelope — the behavior the select pin's design leans on.
- **`mailFolderId` forwards verbatim** (an inbox-scoped binding returned exactly that folder's rows), with the multi-page upstream caveat below.
- **Fixtures are redacted live captures** from this pass (`messages.json`/`mail_folders.json`: deterministic id maps, placeholder identities, minute-coarsened timestamps), re-audited on every test run by the `fixtures_stay_redacted` tripwire (ASCII/identity/URL allowlists, nested-JSON decode); explicit-null and hidden-folder arms stay inline-synthetic and say so.
- **Caveat recorded honestly**: this mailbox had no hidden folders, so `includeHiddenFolders: true` was accepted but its on/off responses were identical — the pin's effect is unobservable against this account.

## Known upstream defect (found in the live pass — since fixed upstream)

Graph's continuation URL for a *folder-scoped* listing uses the OData parenthesized form `/v1.0/me/mailFolders('{id}')/messages`, which the executor's path allowlist rejected — a `mailFolderId`-bound scan whose folder exceeds one page (`page_size: 100`) failed loudly with a gateway 400 on its second page ("nextLink must target Outlook message pagination endpoints"), refusing the very cursor its previous response returned. Loud, never silent; whole-mailbox scans (`/v1.0/me/messages`) paginate cleanly either way. **Fixed upstream in [oomol-lab/open-connector#372](https://github.com/oomol-lab/open-connector/pull/372)** (merged 2026-08-19, live-verified in that PR); gateways predating it — every tagged release through v1.3.5 — still carry the defect.

## Review status

- **Phase 5 (self-review) is done** — sixteen findings, fifteen fixed: vacuous test defenses (cardinality assertions arrow satisfies with `""` on null slots, mapped columns with no positive witness), a redaction audit that admitted an as-captured cursor or `webLink` on a host-prefix match alone, doc claims the pack's own fixtures contradict, and two hardcoded builtin-asset rosters lagging two shipped packs. Deferred: consolidating the pack-test helper quartet into `testutil` — seven-way duplication across the packs, so it belongs in its own cross-pack PR.
- Test counts pinned from CI on `ef5125f` (`cargo llvm-cov nextest --all-features`: 1864 tests plus the 237-test ignored suite, all green): 368 `sources::providers::open_connector` tests, 24 of them pack-scoped `packs::outlook`, inside a 1716-test skardi library binary.
- Deliberate deferrals: message bodies/attachments (future content surface), the `one_drive` pack (own milestone + PR), the entire `excel` service (blocked on an upstream `nextLink` input).


gracexmatin:
> [!WARNING]
> **Folder-scoped pagination does not work until the upstream gateway fix lands.**

A `mailFolderId`-scoped `messages` scan whose folder exceeds one page (`page_size: 100`) fails with a gateway 400 on its **second** page: Microsoft Graph returns the scoped continuation URL in the OData parenthesized form `/v1.0/me/mailFolders('{id}')/messages`, while the outlook executor's own path allowlist (`isAllowedOutlookMessageNextLinkPath`, which expects the five-segment slash form) rejects the very cursor its previous response returned.

- Found in the phase-4 live pass (2026-08-19, real MSA mailbox, pinned gateway v1.3.4). Reproduced with `top=2` inside a seeded inbox: page 1 returns rows + `nextLink`; feeding that `nextLink` back → `400 invalid_input: "nextLink must target Outlook message pagination endpoints"`.
- **Whole-mailbox scans are unaffected** — `/v1.0/me/messages` paginates cleanly (verified five pages to a genuinely-null terminal cursor). Folder-scoped bindings whose folders fit in one page also work.
- The failure is loud, never silent data corruption — but there is **no Skardi-side workaround**: the cursor is contractually opaque ("Opaque pagination URL"), and the allowlist runs inside the gateway. The fix has been submitted upstream: oomol-lab/open-connector#372 (allowlist accepts the parenthesized form; adds the first tests for this allowlist).
- Recorded in the pack's module doc, `outlook.yaml`, and the pack doc's Pagination section.

gracexmatin:
> [!NOTE]
> **Superseded** — instead of filing an issue, the fix itself went upstream as oomol-lab/open-connector#372 (live-verified before/after, plus the first tests for this allowlist). The draft below is kept for reference only.

Draft for the upstream issue referenced in the warning above — ready to file on **oomol-lab/open-connector** verbatim (title + body below). Once filed, the "issue pending" references in this PR and the pack docs get replaced with the real link.

**Suggested title:** `outlook: folder-scoped nextLink is rejected by the executor's own path allowlist (OData parenthesized segment)`

<details>
<summary>Issue body (copy-paste ready)</summary>

```markdown
## Summary

`outlook.list_messages` with a `mailFolderId` returns a `nextLink` that
the same action then refuses to consume: any folder-scoped listing that
spans more than one page fails with HTTP 400 on the second request.

## Reproduction (gateway v1.3.4, 2410fbe)

1. `POST /v1/actions/outlook.list_messages` with
   `{"input":{"select":["id"],"top":2,"mailFolderId":"<folder id>"}}`
   against a folder holding more than 2 messages.
2. The response's `nextLink` is (real capture, id elided):

   https://graph.microsoft.com/v1.0/me/mailFolders('AQMkA…')/messages?...

   Microsoft Graph emits the OData parenthesized-key form for
   folder-scoped collections — not the slash form.
3. `POST /v1/actions/outlook.list_messages` with
   `{"input":{"select":["id"],"top":2,"nextLink":"<that URL>"}}` →

   {"success":false,"message":"nextLink must target Outlook message pagination endpoints","data":{"status":400},"errorCode":"invalid_input"}

## Cause

`isAllowedOutlookMessageNextLinkPath` (src/providers/outlook/executors.ts)
allowlists exactly five path segments
(`v1.0/me/mailFolders/<id>/messages`), but Graph's real continuation URL
packs the folder id into the third segment as `mailFolders('<id>')`,
yielding four segments, so the check fails. The unscoped
`/v1.0/me/messages` form is unaffected.

## Expected

The allowlist should accept the parenthesized form Graph actually
returns for `/me/mailFolders('{id}')/messages` (or normalize it),
keeping the host/path pinning intact.

## Impact

Any consumer that pages a single mail folder past one page is broken;
the failure is loud (400) but unavoidable, since the rejected URL is the
one the previous response produced. Verified against a live personal
(MSA) mailbox through the v1.3.4 gateway.
```

</details>

yiyi-wang-skardi:
@yiyi-wang-skardi is now reviewing this PR using Claude Code — opus (reasoning effort max). Target HEAD commit: https://github.com/SkardiLabs/skardi/commit/171f881cbba2e4d8f13567776143571a54637f9e (`171f881`).

yiyi-wang-skardi:
@yiyi-wang-skardi is now reviewing this PR using Claude Code — opus (reasoning effort max). Target HEAD commit: https://github.com/SkardiLabs/skardi/commit/ef5125f2510648bda5bf5677ecff7f3b56d2b640 (`ef5125f`).

crates/skardi/src/sources/providers/open_connector/packs/outlook.rs
```diff
@@ -0,0 +1,1629 @@
...
+                        assert!(key.is_ascii(), "non-ASCII key at {path}: {key:?}");
+                        audit(child, &format!("{path}.{key}"));
+                    }
+                }
+                Value::Array(items) => {
+                    for (i, child) in items.iter().enumerate() {
+                        audit(child, &format!("{path}[{i}]"));
+                    }
+                }
+                Value::String(s) => {
```
yiyi-wang-skardi, line 604: **The audit doesn't enforce the id-remap rule it documents — and `mail_folders` has nothing else covering it.**

The redaction convention above says ids "keep their real prefix and length and carry the `Synthetic` marker", and the comment at L244 promises `fixtures_stay_redacted` "enforces the mechanical half". But this arm has exactly three string rules — ASCII, `@`-identity, URL shape — and none of them can see `id`, `parentFolderId`, `conversationId` or `@odata.etag`. Those values are ASCII, contain no `@`, and contain no `http`, so they sail through untouched.

On `messages` the gap is masked: `webLink` embeds the row's own id and *is* checked, so a forgotten message-id remap trips the URL rule. `mail_folders` has no URL column at all — a re-capture that remaps identities and folder display names but forgets the folder-id map ships nine real, persistent Graph folder identifiers (`AQMkA…`, which are stable mailbox-scoped handles) and this test stays green. That is precisely the case the comment says it guards: "what this guards is the NEXT re-capture".

Cheap close, in the same spirit as `is_redacted_url` requiring a marker on an allowed host — key-aware rather than value-only:

```rust
const ID_KEYS: [&str; 4] = ["id", "parentFolderId", "conversationId", "@odata.etag"];
// in the Object arm, before recursing:
if ID_KEYS.contains(&key.as_str()) {
    if let Value::String(v) = child {
        assert!(
            v.contains("Synthetic") || v.contains("redacted-"),
            "id-bearing value at {path}.{key} carries no redaction marker"
        );
    }
}
```

Today's fixtures already satisfy this (I checked — every id carries `Synthetic`), so it lands green and only bites the next re-capture.

docs/superpowers/specs/2026-08-14-open-connector-m365-packs-design.md
```diff
@@ -0,0 +1,437 @@
...
+predicate pushdown, following the Notion pack's precedent.
+
+## What Open Connector actually exposes for Microsoft 365
+
+Probed live against a local gateway (v1.3.4, `open-connector` at
+`2410fbe`). No assumption below is taken from Microsoft's own docs.
+
+| Service | Actions | List-shaped read actions | Gate verdict |
+|---|---|---|---|
+| `outlook` | 11 — **mail only**, no calendar or contacts | `list_messages`, `list_mail_folders` | ✅ both paginate completely |
```
yiyi-wang-skardi: **The design record still says `✅ both paginate completely`, which phase 4 disproved for the scoped case.**

Every other surface carries the folder-scoped `nextLink` defect — `outlook.yaml` L75-78, the module doc L18-27, the pack doc L143-150, the tasks doc, the PR body. This document does not: this gate-verdict row is unchanged, and the "Upstream issue candidates" list at L414 still has three entries without the allowlist defect, even though it is the most consequential finding of the live pass and the one most likely to need an upstream link later.

That matters more here than in prose, because this table *is* the admission-gate record. A future reader deciding whether `mailFolderId` should have shipped gets a verdict written before the evidence existed.

I did weigh whether the resource should be withheld the way the whole `excel` service was. I don't think it should: the gate's stated rule is "complete terminating pagination", `excel` fails it by *silently* presenting page one as the collection, and a `mailFolderId`-bound scan instead fails loudly on page two — a different class, and the unbound table is complete. Shipping it is the right call. The design doc should just say so, rather than reading as though the question never came up. Suggest annotating this row ("✅ whole-mailbox; folder-scoped continuation broken upstream — see phase 4") and adding a fourth upstream-issue candidate.

docs/open-connector-microsoft-365.md
```diff
@@ -0,0 +1,228 @@
...
+            resource:                    # optional; omit for the whole mailbox
+              mailFolderId: "AQMkAGE1..."    # scope messages to one folder
+            tables: [messages, mail_folders]
+```
+
+```sql
+SELECT from_address, subject, received_date_time
+FROM saas.m365.messages
+WHERE is_read = false
+ORDER BY received_date_time DESC
```
yiyi-wang-skardi, line 76: **The flagship example is the one query shape this pack cannot serve on a real mailbox.**

```sql
WHERE is_read = false
ORDER BY received_date_time DESC
LIMIT 20;
```

Neither bound narrows the fetch. `is_read` has no `FilterMapping` (by design), so DataFusion keeps a `FilterExec` above the scan; the `ORDER BY` forces a full sort, so the `LIMIT` becomes a TopK above that and never reaches the table provider's `fetch`. The scan therefore reads *every* page until it hits `max_pages` — 100 pages × `page_size: 100` = **10 000 rows**, then fails with `ScanBoundsExceeded` (`max_rows`' 100 000 never binds). Any ordinary mailbox is larger than that.

So the doc's first, most-copied example reads like a cheap "20 newest unread" query and is in fact the most expensive possible one — and on a real account it errors rather than returning rows. L118-120 is technically accurate ("Every predicate runs in DataFusion after the bounded fetch; `LIMIT` stops pagination early"), but the example teaches the misreading.

Two suggestions, both cheap:
1. Lead with a bare `SELECT … LIMIT 20` — that shape *does* early-stop (it's what `limit_stops_cursor_pagination_early` pins), and Graph already returns newest-first, as the capture's own `"wire order is newest-first"` assertion shows. Keep the filtered/sorted variant as the explicit "this reads the whole table" case.
2. State the ceiling in numbers next to L120-122 instead of the formula: at defaults this table tops out at 10 000 messages per scan.

crates/skardi/src/sources/providers/open_connector/packs/outlook.yaml
```diff
@@ -0,0 +1,177 @@
...
+      cursor_input: nextLink
+      next_cursor_path: "$.nextLink"
+      page_size_input: top
+      # 100, not the schema's 1000 ceiling: without `$select` Graph
+      # returns full message bodies, and even with the select pin below
+      # a conservative page keeps every response far from the client's
+      # 16 MiB cap. `top` also rides continuation requests (the engine
+      # sends the page size on every request); the executor ignores it
+      # there because the nextLink URL embeds its own $top.
+      page_size: 100
```
yiyi-wang-skardi, line 44: **Worth re-deriving `page_size: 100` now that the `select` pin exists — it is what sets the 10 000-row scan ceiling.**

The comment justifies 100 as keeping responses "far from the client's 16 MiB cap", but that reasoning was formed against the *unpinned* representation, where `body.content` made per-row size unbounded. With the pin, `body` is gone and the largest remaining field is `bodyPreview`, which Graph caps at 255 characters — a 1000-row page is on the order of 1-2 MB, an order of magnitude under the cap. The pathological case is a message with a very large recipient list, which is bounded in practice.

The cost of 100 is not throughput, it's reach: with default `max_pages: 100` this table can never scan past 10 000 messages (see the pack-doc finding), on the highest-cardinality table in the pack. Raising `page_size` is the only lever that moves that without asking every operator to retune `max_pages`.

Not asking for 1000 — the conservatism is defensible and `mail_folders` already uses the ceiling. But if 100 stays, the comment should give the real reason (the recipient-list tail, not `body`), and the pack doc should name the resulting row ceiling out loud.

crates/skardi/src/sources/providers/open_connector/packs/outlook.rs
```diff
@@ -0,0 +1,1629 @@
...
+                    page_size_param,
+                    ..
+                } => {
+                    generated.push(cursor_param);
+                    generated.extend(page_size_param);
+                }
+                _ => panic!("{short}: both tables declare the cursor strategy"),
+            }
+            for key in &generated {
+                assert!(
```
yiyi-wang-skardi, line 854: **This test checks that keys are *declared*, but not that the pinned values are the declared *type* — and type is the other half of what a strict schema 400s on.**

The test's stated job is catching "an undeclared key is a hard 400 on every scan" before it reaches the wire, and it's the pack's only guard on the input side (the registration gate is output-only). Key presence is covered; JSON type is not. `includeHiddenFolders: "true"`, or `select` pinned as a comma-joined `Str` instead of a `StrList`, would pass every assertion here and then 400 on every live scan — the exact failure mode the test exists to pre-empt, and one a reviewer wouldn't catch by eye either.

The input contracts already carry what's needed (`"type": "boolean"`, `"type": "array"` with string items, `"type": "integer"`), and `FixedValue` maps onto them one-to-one:

```rust
for (key, value) in t.fixed_inputs {
    let want = properties[*key]["type"].as_str();
    let got = match value {
        FixedValue::Bool(_) => "boolean",
        FixedValue::Int(_) => "integer",
        FixedValue::Float(_) => "number",
        FixedValue::Str(_) => "string",
        FixedValue::StrList(_) => "array",
        FixedValue::Json(_) => "object",
    };
    assert_eq!(want, Some(got), "{short}: `{key}` pins a {got} where the schema declares {want:?}");
}
```

Worth doing here rather than in a follow-up, since this is the pack that pins the most fixed inputs of any so far, and the same block would benefit gmail/slack if it later moves to `testutil`.

crates/skardi/src/sources/providers/open_connector/packs/loader.rs
```diff
@@ -762,6 +762,8 @@ mod tests {
             ("slack.yaml", include_str!("slack.yaml")),
             ("notion.yaml", include_str!("notion.yaml")),
             ("feishu.yaml", include_str!("feishu.yaml")),
+            ("discord.yaml", include_str!("discord.yaml")),
+            ("outlook.yaml", include_str!("outlook.yaml")),
```
yiyi-wang-skardi: **This roster had already drifted for a full milestone, and nothing stops the next omission.**

Adding `discord.yaml` here alongside `outlook.yaml` is a good catch — but it means `discord.yaml` shipped in 5.6 with no coverage from the test whose doc comment claims "Every shipped asset parses AND passes the same structural checks binding performs". A hand-listed array next to an `include_str!` is exactly the construct that silently falls behind, and it did.

It's also close to redundant: `SourcePackRegistry::builtins()` already calls `loader::builtin` — hence `parse_pack` — for every embedded asset, and its own test asserts the full name list. The only thing this test adds over that is the non-empty-tables check.

Either drive it off the registry:

```rust
for pack in SourcePackRegistry::builtins().expect("embedded assets parse").names() {
    assert!(!registry.require(pack).unwrap().tables.is_empty(), "{pack}: no tables");
}
```

or, if the direct `parse_pack` path is the point, assert the roster's length against the registry's pack count so an unlisted asset fails loudly instead of just going unchecked.

crates/skardi/src/sources/providers/open_connector/packs/outlook.rs
```diff
@@ -0,0 +1,1629 @@
...
+                            "@-bearing value at {path} is not a redacted identity: {s:?}"
+                        );
+                    }
+                    // Every occurrence, not just a leading one: the
+                    // capture carries `bodyPreview` (255 chars of real
+                    // mail body), where a link sits mid-sentence — ASCII
+                    // and @-free, so nothing else here would catch it.
+                    let lowered = s.to_ascii_lowercase();
+                    for (offset, _) in lowered.match_indices("http") {
+                        let url = &s[offset..];
```
yiyi-wang-skardi, line 622: **`is_redacted_url` is handed the whole tail of the string, not the URL — so the marker can be vouched for by unrelated text.**

`&s[offset..]` runs to the end of the string, so `url.contains("Synthetic")` at L560 is satisfied by any marker appearing *anywhere later* in the same value. The comment at L586-588 explains that the loop exists because `bodyPreview` carries real mail body where "a link sits mid-sentence" — which is precisely where an un-redacted `graph.microsoft.com` cursor could sit ahead of a `Synthetic…` id mentioned later in the same 255-char preview, and pass.

The host allowlist still holds, so this isn't wide open; it weakens only the marker half — the half the comment says exists so that "one that skipped the id remap fails". Cutting the token at the first character no URL can hold restores it:

```rust
let url = &s[offset..];
let url = &url[..url.find(|c: char| c.is_whitespace() || c == '"' || c == '<' || c == '>')
    .unwrap_or(url.len())];
```

crates/skardi/src/sources/providers/open_connector/packs/outlook.rs
```diff
@@ -0,0 +1,1629 @@
+//! Outlook (Microsoft 365) source pack: `messages` and `mail_folders`
+//! over Open Connector's `outlook` service, reconciled against a live
+//! gateway (v1.3.4, open-connector at `2410fbe`) on 2026-08-14. Design
+//! record: `docs/superpowers/specs/2026-08-14-open-connector-m365-packs-design.md`.
+//!
+//! **Status: phases 1–4 complete.** The phase-4 live pass ran on
+//! 2026-08-19 against the same gateway pin (v1.3.4, `2410fbe`) and a
+//! real personal (MSA) mailbox over OAuth: all 22 selected message
+//! fields and all 7 folder fields arrived with the pinned spellings
```
yiyi-wang-skardi: **"all 7 folder fields" is stale by one — it predates the `well_known_name` promotion.**

`mail_folders` maps eight columns, and the live pass verified all eight, including `wellKnownName` (which the very next paragraph, L32-36, describes as promoted *because* of that pass). Seven is the count of *declared* fields, but the parallel clause is "all 22 selected message fields", and 22 is the pinned-select count including passthrough — so the two halves of the sentence are counting different things.

`docs/open-connector-microsoft-365.md:197` already says "all 8 folder columns". One character here to match.

docs/superpowers/specs/2026-07-11-open-connector-integration-tasks.md
```diff
@@ -425,6 +425,64 @@ event carrying the scan identity and error.
...
+      upstream gateway fix (issue to be filed on
+      oomol-lab/open-connector); the live mailbox had no hidden folders,
+      so the `includeHiddenFolders` pin's effect was unobservable
+      (recorded as caveat); wire extras `@odata.etag` and `sizeInBytes`
+      remain deliberately unmapped, while `wellKnownName` was promoted
+      to a `well_known_name` column post-pass (Owen-approved
+      2026-08-19: live display names were all CJK, so cross-account
+      folder semantics need Graph's locale-independent discriminator;
+      mail_folders' coverage gap goes empty → one pinned column). Test
+      counts get pinned from CI on the next push; phase 5 review then
```
yiyi-wang-skardi: **This entry is the only 5.x item without a test count.**

Every merged milestone above closes with concrete numbers — 5.6 ends "311 tests (`cargo test -p skardi --lib sources::providers::open_connector`; 23 new — 9 keyset engine, 1 loader, 13 pack-scoped …), 868 full library suite" — and 5.1-5.5 follow the same shape. This one has a promissory note instead.

That convention is load-bearing for exactly this pack: it is the record of how much of the 1629-line test module actually runs, and "CI is the verdict" is a reason to read the numbers *off* CI, not to omit them. Worth filling in from the CI run for this PR before merge, along with flipping "phase 5 review then precedes ready" now that the review has happened.

crates/skardi/src/sources/providers/open_connector/packs/outlook.rs
```diff
@@ -0,0 +1,1629 @@
...
+    // stays honest: ids keep their real prefix and length and carry the
+    // `Synthetic` marker — messages `AQMkADAwATM3SyntheticMessage0001`
+    // padded to 144, folders and `parentFolderId` to 112,
+    // `conversationId` to 72, `@odata.etag` inside its `W/"…"` wrapper;
+    // identities become `Person N` / `personN@example.com`;
+    // `internetMessageId` becomes `<synthetic-NNNN@mail.example.com>`
+    // (brackets and subdomain kept — real ones have both); free text
+    // becomes `Synthetic subject N` / `Synthetic preview N` /
+    // `Category A`; `webLink` keeps its host and embeds the row's OWN
+    // synthetic id; timestamps are coarsened to the minute with their
```
yiyi-wang-skardi, line 272: **Question on the timestamp coarsening: did the live wire carry sub-second precision?**

All nine captured timestamps end `:00Z`. If Graph spelled `receivedDateTime` with fractional seconds on the wire (it does emit `.0000000Z` on several Graph resource types), then coarsening to the minute removed the one part of the string that `parse_timestamp` → `DateTime::parse_from_rfc3339` actually has to cope with, and the capture now pins a shape the wire doesn't produce.

That's the failure class this module is otherwise careful about — the same reasoning that made every mock cursor URI-shaped rather than `"cursor-2"`, because "an opaque token would pass these mocks and 400 live". Four passthrough timestamp columns sit outside the fingerprint gate, so the fixture is a large share of their coverage.

Sub-second digits carry no identity, so if they were there, keeping them costs nothing: coarsen to the minute and re-append the wire's own fractional spelling. If the wire really did emit whole seconds, worth saying so here — it's a fact about the provider that the next re-capture shouldn't have to rediscover.

yiyi-wang-skardi (approved):
## Review: `outlook` source pack (milestone 5.7)

Reviewed the 17 files that belong to this PR (`git diff 095d124..HEAD` — the branch is stacked on #206/#213, which I left alone): `packs/outlook.yaml`, `packs/outlook.rs` (1629 lines, all tests), the seven fixtures, the registry/loader/mod wiring, and the four docs.

**Verdict: no correctness, safety or security defect found. Nothing blocking.** This is unusually disciplined work — the design rationale is recorded where the code lives, every load-bearing assumption has a mechanical pin, and the fixtures are honestly labelled as capture vs. synthetic. It follows the existing pack idiom (gmail/discord/feishu) closely enough that reviewing it is mostly reading, not archaeology.

### What I verified mechanically

No Rust toolchain exists in this environment, so nothing was compiled — **CI remains the verdict on compilation and on the test run.** To avoid signing off on unchecked pins (the PR body notes test counts are still to come from CI), I re-implemented the relevant logic outside Rust and replayed it:

- **Both pinned fingerprints are correct.** I wrote BLAKE3 + `canonical_json` from scratch, validated it against the official BLAKE3 vectors *and* against ten already-merged pins (github.issues/commits, gmail.threads/messages/drafts, notion ×4, slack.conversations/users), then hashed the two committed contracts: `ff5dea4c…` and `c67baab7…` both reproduce exactly.
- **`select_pin_mirrors_the_mapped_columns`**: the 22 pinned keys are unique and exactly equal the 24 columns' top-level wire keys. No strays, no gaps.
- **`columns_the_coverage_walker_cannot_resolve_are_pinned`**: reimplementing the walker over both contracts yields precisely the pinned 13-column and 1-column gaps, in the pinned order.
- **`generated_inputs_are_accepted_by_the_captured_input_contracts`**: every generated key is declared, neither schema has unmet `required`, and both page sizes sit inside `top`'s declared `[1, 1000]`.
- **Loader cross-field rules**: no fixed-input / resource / pagination-key collisions, no duplicate columns, `cursor_input != page_size_input` on either table.
- **`fixtures_stay_redacted`**: reimplemented the audit (ASCII, `@`-identity both halves, URL allowlist, one nested-JSON decode level) and ran it over all three fixtures — zero failures. I also dumped every distinct string in the two live captures and read them: no real identity, address, subject, folder name or id survives.
- **Every assertion in the two conversion tests** replayed against the fixture JSON — row 8's `""`/`[]` draft, the row 1/7 conversation pair, 8 conversations across 9 rows, 4 parent folders, `categories` 1-vs-0, `webLink`-embeds-own-id, newest-first ordering, `well_known_name` explicit null on the custom folder, `total_item_count[7] == 5`, the 112-char root parent id: all hold. The mismatch fixture's `row: 1` is right (the converter is 0-based; the first row does convert).

### Notes

The live-gateway claims (real Graph spellings, `top=1001` → 400, the foreign-host cursor refusal) are unverifiable from here and I took them as stated.

The two fixtures are internally consistent in a way worth keeping: message row 2 sits in folder `…Folder0002`, which is deliberately **absent** from the root-level folder capture — exactly the nested-folder case the pack doc's `JOIN` warning is about. No test pins that relationship; a one-line cross-fixture assertion would turn the doc's caveat into a guarantee.

Housekeeping: `isDraft` is now false, so the body's "Still a draft: phase 5 (review) is pending" and "this PR stays draft until it lands" are stale — worth a body edit when the review lands.

Ten non-blocking findings follow, roughly most- to least-useful. The first is the one I'd actually act on before the next fixture re-capture.

docs/superpowers/specs/2026-08-14-open-connector-m365-packs-design.md
gracexmatin: Agreed on all the core facts — this document was the one surface with no record of the defect, and as the admission-gate record it is the surface where that gap mattered most. Both suggested edits land in the next push (the gate-verdict row annotated, a fourth upstream entry added), with two adjustments to the suggested wording:

**It is no longer an issue candidate.** The fix went upstream directly instead of through an issue: oomol-lab/open-connector#372, merged 2026-08-19 (`dd3aeea`), live-verified there with a before/after folder-scoped walk. So entry 4 records it as *found in the phase-4 live pass, since fixed*, with the caveat that matters operationally: every tagged release through v1.3.5 predates the fix, so only gateways built from upstream main at or after that commit paginate scoped scans.

**The row keeps its verdict and gains an asterisk, rather than being rewritten.** `✅ both paginate completely` is a contract-level statement — both actions declare `nextLink` on both halves, which is what the gate probes — and phase 4 found a runtime bug in the executor's allowlist, which does not retract the contract verdict. The suggested "✅ whole-mailbox" would also understate `mail_folders`, which has no scoping at all and was never affected; the defect only ever hit `messages`' folder-scoped continuation. And the annotation points at "item 4 below" rather than "see phase 4", because this document has no phase-4 section to anchor to — the phase record lives in the tasks doc.

On ship-vs-withhold: agreed, and that reasoning now lives in the document rather than in this thread — entry 4 states why the verdict stood (a loud mid-scan failure the engine turns into a failed scan, versus Excel's silent page-one-as-the-collection; and the unbound table paginates completely), so the next reader sees the question was weighed rather than missed.

crates/skardi/src/sources/providers/open_connector/packs/outlook.rs
gracexmatin: Confirmed and applied essentially as proposed — `ID_KEYS` with the four keys, checked key-aware in the Object arm before recursing, values must carry `Synthetic` or `redacted-`. The gap was real: all three value rules are blind to a marker-less id (ASCII, `@`-free, scheme-free), and `mail_folders` has no `webLink` to trip the URL rule by proxy, so a re-capture that forgot the folder-id map would have shipped nine persistent mailbox-scoped handles green.

One correction to the premise, though: "today's fixtures already satisfy this" holds for the two capture fixtures (54 id-bearing values, every one carrying `Synthetic`) but not for the third file the audit walks. The hand-written `messages_type_mismatch.json` had six id-bearing values with no marker at all (`id`/`conversationId`/`parentFolderId` on both rows — e.g. `AAQkAGE1M2IyNGNmconv-val`), so landing the check verbatim would have failed CI on its first run. Those six are renamed into the `redacted-` family the fixture already uses everywhere else (`redacted-message-val`, `redacted-conv-bad`, `redacted-folder-inbox`, …). Checked before renaming that nothing pins the old spellings: the mismatch test asserts `column`/`page`/`row`/`found` only, and the identical-looking folder id in `optional_resource_forwards_verbatim_and_only_where_declared` is an independent inline constant the audit never walks, left untouched.

Verified per house convention (no local test runs) by mirroring all four rules in Python over the three fixtures: 0 violations. Lands in the next push together with the audit-rule change.

docs/open-connector-microsoft-365.md
```diff
@@ -0,0 +1,243 @@
...
+Graph timestamps (`received_date_time`, `sent_date_time`,
+`created_date_time`, `last_modified_date_time`).
+
+The pack **pins `select` to exactly the mapped fields**. Unpinned,
+Graph's default representation carries the full (usually HTML) message
+body on every row — far past the client's 16 MiB response cap at any
+realistic page size — and the row shape becomes "whatever the default
+carries". The pin shrinks payloads by orders of magnitude and turns a
+misspelled column path into a loud Graph 400 instead of a silently
+always-NULL column. Consequence: **`body` is not a column**, by design;
```
BtXin, line 113: **`body_preview` is body content, and this guide never mentions it.**

The Columns paragraph above enumerates 23 of the 24 `messages` columns — `body_preview` is the one it leaves out — and this paragraph then states that **`body` is not a column, by design; full message content belongs behind a future content-oriented surface**. An operator reading the guide to size content/PII exposure concludes no message body is queryable. But `body_preview` is Graph's ~255-character excerpt of every message body, it is in the select pin, and it lands in every `SELECT *`.

The Gmail pack's guide sets the precedent for exactly this column class — `docs/open-connector-gmail.md:114-118`:

> **`threads.snippet` is body content**: Gmail's own excerpt of the latest message's text (roughly its first hundred characters) […] The "no bodies" shape above is `messages`' posture, not the pack's — size content/PII exposure with this column included.

Worth adding `body_preview` to the column enumeration and giving it the same explicit call-out here, so the "no bodies" claim reads as scoped to `body` rather than to message content generally.

docs/open-connector-microsoft-365.md
```diff
@@ -0,0 +1,243 @@
...
+-- The 20 newest messages. Graph returns newest-first and a bare LIMIT
+-- stops pagination early, so this reads one page, not the mailbox.
+SELECT from_address, subject, received_date_time
+FROM saas.m365.messages
+LIMIT 20;
+
+-- The same shape with a predicate or ORDER BY reads the WHOLE table:
+-- nothing pushes down (see "No filter pushdown" below), the sort turns
+-- LIMIT into a top-k above the scan, and at the default bounds the
+-- scan fails past 10 000 messages. Scope it with a mailFolderId
```
BtXin, line 69: **This points users at the one configuration the guide later documents as broken.**

`mailFolderId` is offered here (and again under "No filter pushdown", lines 130-131) as *the* remediation for the 10 000-row scan ceiling — but the Pagination section 90 lines below records that a `mailFolderId`-bound scan whose folder exceeds `page_size: 100` fails on its second page against **every tagged gateway release through v1.3.5**, because Graph's scoped continuation URL uses the parenthesized `mailFolders('{id}')/messages` form the executor's allowlist rejected.

An inbox is the folder a user will scope to first, and inboxes routinely hold more than 100 messages — so the recommended escape hatch fails for the most likely case, on every currently-released gateway. The defect is documented honestly; the problem is that the *recommendation* carries no caveat or cross-reference. A parenthetical here (and at 130-131) pointing to the upstream-defect note, or naming the gateway version the advice assumes, would close it.

docs/superpowers/specs/2026-07-11-open-connector-integration-tasks.md
```diff
@@ -425,6 +425,77 @@ event carrying the scan identity and error.
       per-table wire-declaration e2e for all three tables, UDTF parity, and
       empty-page schema stability), 868 full library suite.
 - [ ] 5.6 Later waves per the design rollout (Google Workspace, HubSpot, Jira, …) through the source-pack admission gate
+- [ ] 5.7 Outlook pack (OAuth user token, cursor pagination over Graph
```
BtXin: **5.7 ships unchecked, and "5.6" now names three different things.**

Every prior pack milestone was marked `[x]` by its own shipping PR — `git log -L386,386:<this file>` shows the Discord PR (afc154e) added `- [x] 5.6 Discord pack` in that PR, as 5.1-5.5 did. This PR states all five phases are complete but leaves 5.7 as `- [ ]`, so after merge the checklist reads as "Outlook not done".

Separately, the numbering has collided: line 385 is `- [ ] 5.6 Later waves per the design rollout (Google Calendar, Google Drive, Discord, HubSpot, Jira, …)`, line 386 is `- [x] 5.6 Discord pack`, and line 427 (just above this hunk) is a second `- [ ] 5.6 Later waves per the design rollout (Google Workspace, HubSpot, Jira, …)`. Three bullets numbered 5.6, two of them near-duplicate catch-alls. The duplicate catch-all is pre-existing, but this PR inserts 5.7 below it rather than folding it into line 385 — a good moment to collapse the two "Later waves" bullets and tick 5.7.

docs/open-connector-microsoft-365.md
```diff
@@ -0,0 +1,243 @@
...
+          - name: m365                   # schema name in SQL
+            source_pack: outlook
+            connection_alias: my-msft    # the gateway's outlook OAuth connection
+            resource:                    # optional; omit for the whole mailbox
+              mailFolderId: "AQMkAGE1..."    # scope messages to one folder
+            tables: [messages, mail_folders]
+```
+
+```sql
+-- The 20 newest messages. Graph returns newest-first and a bare LIMIT
```
BtXin: **"The 20 newest messages" is an unenforced ordering claim.**

This query has no `ORDER BY`, and the pack deliberately never sends `orderby` to the wire (lines 125-131) — so the 20 rows are whatever order Graph happened to return. The only evidence for newest-first is one live capture, pinned in `outlook.rs`'s `messages_fixture_converts_the_live_row_shapes` as `"wire order is newest-first"`, and the pack's own stated doctrine is to trust the live gateway rather than Microsoft's documentation.

If Graph's default order differs for a folder-scoped listing, or changes, this returns an arbitrary 20 rows and reports them as the newest — the silent-wrong-answer class the rest of this pack works hard to convert into loud failures. The next example already shows the honest form; consider softening this one to "a bounded first page (Graph returned newest-first in the live pass; nothing enforces it — add `ORDER BY` when the ordering matters)".

docs/open-connector-microsoft-365.md
gracexmatin: The upstream defect has been merged and resolved. It will work as expected once a new release is published, and a clear note has already been added to the documentation.

Issue #218, opened 2026-08-17
CLI: skardi query --session-id/--purpose so ad-hoc rows join ledger sessions


BtXin:
The ledger's headline property (#213) is that one agent session's ad-hoc queries and pipeline calls interleave under a single `session_id`, ordered by `created_at` — the shape intention-mining skills read. Review of #213 (@abbccdda) pointed out this isn't reachable from the shipped CLI: `skardi run --session-id` covers the pipeline half, but `skardi query` has no way to send `ai_context`, so the ad-hoc half of every CLI session lands with `session_id = NULL`. Only a direct HTTP caller can currently produce the interleaving.

Proposed: `skardi query --session-id <ID> --purpose <TEXT>`, building the `ai_context` object in the request body. Notes:

- The server requires `ai_context` to carry BOTH `purpose` and `session_id` (non-empty); the CLI should require the two flags together and error client-side if only one is given, mirroring how `run.rs` validates `--session-id` before any request is sent (accurate error, no `ApiError::Connect` misattribution, exit code preserved).
- Reuse the client-side session-id rules from `run.rs` (`MAX_SESSION_ID_CHARS = 200`, visible ASCII, no commas); `purpose` cap is 2000 chars server-side.
- `docs/server.md` documents the current limitation and should drop that sentence when this lands.

Refs: #213.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #217, opened 2026-08-17
Close the remaining parameter-value egress points on the pipeline execute path


BtXin:
#213 established the stance that caller-supplied parameter values must never reach the audit ledger, and pins it with an end-to-end canary test. Review of that PR (by @abbccdda) inventoried the remaining places on the pipeline-execute path where parameter values still egress. None of these were introduced by #213 — they predate it — but they all contradict the stance it establishes, and `docs/server.md` now explicitly scopes the ledger guarantee around them.

## The four egress points (all in `crates/server/src/pipeline_handlers.rs`)

1. **Substituted SQL at DEBUG.** `tracing::debug!("Failed SQL query: {}", sql)` on the engine-failure branch logs the fully substituted SQL — parameter literals included. DEBUG is off by default, but any operator raising `RUST_LOG` for diagnosis starts shipping param values to their trace sink. (`/query` solved this class of leak in #173 by pinning plan-printing targets; this line is the server's own, not DataFusion's, so the filter doesn't cover it.)
2. **Raw parameter value at ERROR.** In `substitute_sql_params`, the unsupported-type branch logs `tracing::error!("Unsupported parameter type for {}: {:?}", param_name, param_value)` — ERROR is on by default and fans out to every configured OTLP collector. The empty-array / mixed-shape / ragged-rows branches log names only, so this one site is the outlier.
3. **Parameter values in HTTP 400 bodies.** Two `unsupported_params.push(format!("{}: {:?}", param_name, param_value))` sites put the raw value into the `error_details.unsupported_parameters` array returned to the caller.
4. **Engine error text + record-batch schema in HTTP 500 bodies.** The engine-failure and conversion-failure responses include `engine_error: e.to_string()` (which can quote parameter literals *and* underlying data values) and the full record-batch schema. `/query` deliberately redacts both to "see server logs for details" — the pipeline endpoint should match.

## Suggested resolution

- (1) and (2): make the logs value-free — log parameter *names* and statement/pipeline identity only, matching the audit ledger's convention.
- (3) and (4): redact pipeline error responses to `/query`'s standard: fixed error kinds + a pointer to server logs. This is a breaking change to error-body contents, so it deserves its own PR and a changelog note.
- Extend #213's canary test pattern to the log stream: drive the failure modes with a planted canary value and assert a captured subscriber never sees it (the harness in `crates/server/tests/query_plan_logging.rs` already does exactly this for `/query`).

[...]

PR #213, opened 2026-08-14
feat(server): audit pipeline executions — one ledger for ad-hoc and pipeline calls

BtXin:
## Why

The query audit ledger (`--query-audit-db`, #173) records every ad-hoc `POST /query`, but pipeline executions were recorded nowhere. That makes the self-improving loop degrade its own input as it succeeds: the moment [`skardi-query-log`](https://github.com/SkardiLabs/skardi-skills/pull/25) promotes a recurring query into a pipeline, its recurrence signal leaves the ledger, and the Learn stage goes blind to exactly the tools the Act stage created.

## What

**One ledger, both kinds of executions.** `POST /:pipeline/execute` now writes to the same `query_audit` store, mirroring `/query`'s semantics:

| | ad-hoc row (existing) | pipeline row (new) |
|---|---|---|
| `statement_kind` | `Query` / `Other` (the classifier's `Debug` form — *not* SQL verbs; corrected per review) | `pipeline` |
| `sql` | raw SQL (operator opt-in) | **`name@version`** (rows outlive template revisions, so the bare name stops answering "what ran"; per round-3 review) — the versioned template lives on disk; `description` carries the purpose |
| `session_id` | from `ai_context` | from the optional `X-Skardi-Session-Id` header |
| `ai_context` | caller-sent object | always NULL |
| `error` on failure | engine error text | **fixed value-free kind** (`query_execution_error` / `result_conversion_error`) |

- **Record-before-execute, fail-closed**: the `started` row is committed after param validation and before the engine runs; if the write fails, 503 and the pipeline does not run. No store configured → behavior unchanged (pinned by test).
- **Params never reach the ledger** — including through error text: engine errors can echo substituted parameter literals, so pipeline rows persist a fixed error kind while the HTTP response keeps the full error.
- **Header, not body field**: the execute body IS the flattened params map, so a reserved key could collide with a real SQL parameter. The header is optional; malformed (empty, >200 chars, outside visible ASCII, containing a comma or tab, or duplicated) → 400 regardless of audit state — validated only after the pipeline lookup, so unknown pipelines 404 and the reject metric's label stays bounded to configured names.
- **Bounded fail-closed**: every ledger write carries a 5s timeout inside the store, so a hung fsync becomes a failed write (503) instead of wedging the single serialized writer thread all audited requests share.
- **CLI**: `skardi run --session-id <ID>` sends the header; without the flag the request is byte-identical to before.
[...]

PR #219, opened 2026-08-18
feat(server): audit job submissions — session attribution + run_id bridge to the jobs ledger

BtXin:
> **Targets `main` directly**, rebased onto #206 (the ledger-crate extraction), which merged as 095d124. Carries its own 7 commits. Review round 1 is addressed — see *Review round 1* and *Rebasing onto #206* below.
>
> Two follow-ups are stacked on this branch: #221 (auth) and #222 (bridge reconstruction), each closing a non-goal this PR records.

## Why

Job *runs* were never the blind spot pipelines were — the jobs ledger records parameters, status, and run ids. What they lacked is **attribution**: no session linkage, so the Learn stage can't stitch "this session ran four ad-hoc queries, called two pipelines, then submitted a backfill" into one intention, and an agent that manually submits the same job on a cadence (a promotion signal — it wants to be a routine) is invisible to mining. Skardi has no built-in cron, so every submission arrives through `POST /jobs/:name/run`: attribution has exactly one seam, and "no session header = not agent behavior" falls out naturally. Design discussion in `docs/superpowers/specs/2026-08-18-jobs-audit-design.md`; supersedes the jobs non-goal in #213's spec.

## What

**Record the submission event, not the run.** One new row kind in `query_audit`; run detail stays in `job_runs`, bridged by a new nullable `run_id` column. No double-logging: the audit row's lifecycle is the submission's, and the jobs ledger remains the authority on the run.

| | job row |
|---|---|
| `statement_kind` | `job` (value set is now `Query` / `Other` / `pipeline` / `job`) |
| `sql` | `name@version`, same convention as pipeline rows |
| `session_id` | from the shared `X-Skardi-Session-Id` header — same validation, same cap |
| `job_run_id` | **new nullable column** = `job_runs.id` once accepted; NULL on rejection and on all non-job rows. Added through #206's additive-column reconciler, now generalized and duplicate-column tolerant (pre-column DBs open and migrate — tests pin both fresh and migrated files) |
| `status` | submission-scoped: `started` → `succeeded` + `run_id` (accepted/enqueued) or `failed` + the executor's **static error category** (`missing_parameters`, `schema_mismatch`, …) — never message text |

- **Precedence, each step test-pinned**: jobs-disabled 503 → unknown job 404 (existence pre-check *before* header validation — #213 round-3's 404-wins/bounded-label lesson applied from the start) → malformed header 400 → audit-write 503 (`query_audit_error`, and the test proves the job was **not submitted** by asserting the jobs ledger is empty) → submit.
- **One honest asymmetry, documented**: job parameter validation happens *inside* `executor.submit` — after the audit write — so a parameter rejection leaves a `failed` row, unlike pipelines where param rejects record nothing. Structural (the server can't pre-validate job params), stated explicitly in `docs/server.md`.
- **`session_id_from_headers` moved to `query_audit`** (third consumer) — ledger-attribution plumbing now lives with the ledger, shrinking the #206 collision surface further.
- **CLI**: `skardi job run --session-id`, with the client-side validation extracted to a shared `cli/src/session.rs` used by both commands (fail-fast, no `ApiError::Connect` misattribution).
- `list_by_session` now returns an agent session's ad-hoc queries, pipeline calls, and job submissions in one totally-ordered read — the shape intention-mining skills consume.

## Tests

5 new store unit tests (job row round-trip with `run_id`, rejected-submission shape, three-kind interleaving in order, orphan reconcile, old-schema migration) and 10 integration tests in `tests/jobs_audit_http.rs` — including the fail-closed not-submitted proof, the post-audit rejection path (mutation-checked: removing the `Err`-arm stamp fails it), a param-value canary against the audit row, 404-before-400 precedence, and audit-off header validation. CLI: header present/absent + fail-fast tests over the shared validator. Full workspace suite: 1332 passed, 0 failed.

## Out of scope (deliberate)

- Auditing run *outcomes* in `query_audit` — the jobs ledger owns them; `run_id` bridges.
- Jobs handlers perform **no auth check at all** today (unlike `/query` and pipelines, which call `require_session`) — pre-existing, surfaced here as an observation; its own scope.
- `job_runs.parameters` stores raw parameter values — exactly what `query_audit` refuses to store. Flagged in the spec for the same threat-model review as #217; changing it would alter `GET /jobs/runs` behavior.

Refs: #213 (merged — this extends its ledger), #206 (ledger crate extraction — the helper moves here reduce the conflict), #217 (param egress), #218 (`skardi query` passthrough).

---

## Review round 1 — addressed

All eight comments, in `fab5857` plus the rebase below. Replies are in each thread.

| | change |
|---|---|
| Migration race | check-then-`ALTER` was not atomic; the loser's `duplicate column name` aborted startup over a condition that already held. Now counted as success, and because the reconciler is shared it covers **all six** added columns — #206 introduced the same race five more times |
| Missing index | partial `idx_query_audit_job_run_id ... WHERE job_run_id IS NOT NULL`. It cannot go in `INIT_SCHEMA_SQL` at all — that batch runs *before* the column exists, against pre-column DBs — so it lives in the reconcile step; a test pins it on fresh **and** migrated files |
| Outcome-write guard | `WHERE status = 'started'` on `record_job_outcome` *and* `record_outcome`, so a late write cannot resurrect a row reconciliation already settled. Both tests fail with the guard removed |
| `statement_kind` casing | normalized — ad-hoc rows record `query` / `other` from named `pub` constants instead of a leaked `Debug` form, so all four values share one casing |
| Unauthenticated ledger write path | documented in `docs/server.md` and the spec non-goals with the two concrete consequences; **fixed in #221** |
| One-directional `job_run_id` bridge | stated as a non-goal with the reason it is structural; **fixed in #222** via a durable reverse pointer |
| Lock-poisoning divergence | kept the recovery, commented why. `.claude/CLAUDE.md` mandates `unwrap_or_else(\|p\| p.into_inner())` for `std::sync` poisoning, so this handler follows policy and `pipeline_handlers.rs` is the outlier |
| `session_id` log field | recorded as an `Option`; `unwrap_or_default()` rendered "absent" as `""`, the one value the validator rejects |

## Rebasing onto #206

#206 moved the ledger into `crates/query-audit` and added five nullable identity columns. That produced one semantic conflict and two structural ones.

**The bridge column is now `job_run_id`.** #206's identity envelope already claims `run_id`, meaning *which of the caller's runs issued this statement* — written at insert time by a distribution that authenticates its callers. Ours means *which job run this submission produced*, stamped at outcome time. Same name, two meanings, distinguishable only by `statement_kind` — exactly the overloading review pushed back on for `statement_kind` itself. Renaming ours was the only resolution that does not redefine a just-merged field the cloud engine adopts wholesale. A test asserts the bridge stamp leaves the identity `run_id` NULL.

**One additive-column mechanism, not two.** #206's `ensure_identity_columns` already reconciles the live schema against a column list on every open — which is what this PR's standalone guarded `ALTER` was doing for one column. Generalized to `ensure_added_columns(conn, columns)`, called with two named lists: `IDENTITY_COLUMNS` (who asked) and `BRIDGE_COLUMNS` (what the work became). Kept separate because they answer different questions; sharing the mechanism is what spreads the duplicate-column fix across all six.

**`session_id_from_headers` lands in the server, not the ledger.** This PR originally relocated it into `query_audit` as its third consumer. No longer possible: it needs `axum::http::HeaderMap`, and `skardi-query-audit` depends on neither `axum` nor `skardi` — the whole point of the extraction. It now lives in `crates/server/src/session_header.rs`, shared by both handlers, still validating against the ledger's `MAX_SESSION_ID_CHARS` so the three attribution paths cannot drift. Same reasoning splits `statement_kind`: the marker strings stay `pub` in the ledger, the `StatementKind` → marker mapping sits in `query_handlers`. Vocabulary owned by the ledger, translation owned by the server.

Full workspace after the rebase: **1467 passed, 0 failed.** Clippy clean on every touched file.

### DIFF
--- a/crates/query-audit/src/lib.rs
+++ b/crates/query-audit/src/lib.rs
@@ -12,8 +12,9 @@
 //! template lives on disk; parameter values are never recorded — the version
 //! is what keeps "what ran" answerable after the promotion loop edits a
 //! template, since rows are kept forever by default). Ad-hoc rows carry
-//! `statement_kind` values from `StatementKind`'s `Debug` form — `Query` /
-//! `Other` — not SQL verbs like `select`.
+//! `statement_kind` values naming the server's statement *classification* —
+//! `query` / `other` (see [`QUERY_STATEMENT_KIND`]) — not SQL verbs like
+//! `select`.
 //!
 //! Design notes:
 //!
@@ -202,6 +203,30 @@ const IDENTITY_COLUMNS: [&str; 5] = ["request_id", "org_id", "workspace_id", "us
 /// own run — one column, one meaning.
 const BRIDGE_COLUMNS: [&str; 1] = ["job_run_id"];
 
+/// Index over the `job_run_id` bridge, applied after [`ensure_added_columns`]
+/// has guaranteed the column exists — so it cannot live in
+/// [`INIT_SCHEMA_SQL`], which also runs against pre-column databases, where
+/// indexing a missing column would fail.
+///
+/// The forward direction — session to its rows — is served by
+/// `idx_query_audit_session_created`. This covers the reverse: given a run id
+/// from `GET /jobs/runs`, which session submitted it. Partial, because
+/// `job_run_id` is NULL on every non-job row, so the index stays proportional
+/// to job submissions rather than to a ledger that is append-only and has
+/// retention off by default.
+const JOB_RUN_ID_INDEX_SQL: &str = "CREATE INDEX IF NOT EXISTS idx_query_audit_job_run_id
+    ON query_audit (job_run_id) WHERE job_run_id IS NOT NULL;";
+
+/// True for the `duplicate column name` error SQLite raises when an
+/// `ALTER TABLE ... ADD COLUMN` loses a race with another connection that
+/// already added it.
+///
+/// Matched on message text because `rusqlite` surfaces it as a generic
+/// `SqliteFailure` with `SQLITE_ERROR`, carrying no distinguishing code.
+fn is_duplicate_column(e: &rusqlite::Error) -> bool {
+    e.to_string().contains("duplicate column name")
+}
+
 /// Add any of `columns` the live table is missing, as nullable `TEXT`.
 fn ensure_added_columns(conn: &rusqlite::Connection, columns: &[&str]) -> SqlResult<()> {
     let mut stmt = conn.prepare("PRAGMA table_info(query_audit)")?;
@@ -210,19 +235,32 @@ fn ensure_added_columns(conn: &rusqlite::Connection, columns: &[&str]) -> SqlRes
         .collect::<std::result::Result<_, _>>()?;
     for col in columns {
         if !existing.contains(*col) {
-            conn.execute(
+            // The read above is not atomic with this ALTER: two processes
+            // opening the same file concurrently (a rolling restart, an
+            // operator tool, a second server pointed at the same path) can
+            // both observe the column missing and both issue it. The loser
+            // gets `duplicate column name` — which is exactly the
+            // post-condition this step wanted, so it counts as success rather
+            // than aborting startup over a benign race.
+            match conn.execute(
                 &format!("ALTER TABLE query_audit ADD COLUMN {col} TEXT"),
                 [],
-            )?;
+            ) {
+                Ok(_) => {}
+                Err(e) if is_duplicate_column(&e) => {}
+                Err(e) => return Err(e),
+            }
         }
     }
     Ok(())
 }
 
-/// Every column added to `query_audit` after its original DDL.
+/// Every column added to `query_audit` after its original DDL, plus the
+/// indexes that can only be built once those columns exist.
 fn ensure_schema_additions(conn: &rusqlite::Connection) -> SqlResult<()> {
     ensure_added_columns(conn, &IDENTITY_COLUMNS)?;
     ensure_added_columns(conn, &BRIDGE_COLUMNS)?;
+    conn.execute_batch(JOB_RUN_ID_INDEX_SQL)?;
     Ok(())
 }
 
@@ -582,6 +620,18 @@ impl QueryAuditStore {
     /// Distinct from the identity envelope's `run_id` ([`QueryIdentity`]),
     /// which names the *caller's* run rather than the job run this submission
     /// produced. One column, one meaning.
+    ///
+    /// Only applies to a row still in `started`, the same guard (and the same
+    /// reasoning) as [`Self::spawn_timeout_correction`]: the write is
+    /// monotonic, so a row that already reached a terminal state stays there.
+    /// Within one process the transition is unreachable — the 503 path returns
+    /// before any outcome stamp — but the ledger is a *file*, and nothing stops
+    /// two servers being pointed at one `--query-audit-db`. In that
+    /// configuration server B's startup [`Self::reconcile_orphaned`] can stamp
+    /// A's in-flight submission `unknown` before A's outcome write lands;
+    /// without this guard that late write would resurrect a terminal row and
+    /// defeat reconciliation. 0 rows updated is therefore not an error: it
+    /// means something else already settled the row.
     pub async fn record_job_outcome(
         &self,
         id: &str,
@@ -599,8 +649,15 @@ impl QueryAuditStore {
                 conn.execute(
                     "UPDATE query_audit
                         SET status = ?2, finished_at = ?3, job_run_id = ?4, error = ?5
-                      WHERE id = ?1",
-                    params![id, status, finished_at, job_run_id, error],
+                      WHERE id = ?1 AND status = ?6",
+                    params![
+                        id,
+                        status,
+                        finished_at,
+                        job_run_id,
+                        error,
+                        QueryAuditStatus::Started.as_str(),
+                    ],
                 )?;
                 Ok(())
             }),
@@ -612,6 +669,9 @@ impl QueryAuditStore {
     }
 
     /// Update a record with its terminal outcome.
+    ///
+    /// Shares [`Self::record_job_outcome`]'s `started`-only guard, for the
+    /// same reason.
     pub async fn record_outcome(
         &self,
         id: &str,
@@ -630,8 +690,15 @@ impl QueryAuditStore {
                 conn.execute(
                     "UPDATE query_audit
                         SET status = ?2, finished_at = ?3, row_count = ?4, error = ?5
-                      WHERE id = ?1",
-                    params![id, status, finished_at, row_count, error],
+                      WHERE id = ?1 AND status = ?6",
+                    params![
+                        id,
+                        status,
+                        finished_at,
+                        row_count,
+                        error,
+                        QueryAuditStatus::Started.as_str(),
+                    ],
                 )?;
                 Ok(())
             }),
@@ -912,6 +979,24 @@ fn restrict_permissions(path: &Path) -> Result<()> {
 /// Statement-kind marker distinguishing pipeline rows from ad-hoc SQL rows.
 const PIPELINE_STATEMENT_KIND: &str = "pipeline";
 
+/// Statement-kind markers for the two ad-hoc row kinds.
+///
+/// `pub` and named rather than derived from the server's statement-classifier
+/// `Debug` form. The ledger's `statement_kind` vocabulary is a consumer-facing
+/// contract ("consumers filtering the ledger must match these exact strings"),
+/// and leaking `Debug` both bound that contract to a type in another crate
+/// whose variant names could be renamed freely, and mixed `PascalCase` ad-hoc
+/// values with the `lowercase` `pipeline` / `job` markers. All four values now
+/// live here, in one casing.
+///
+/// The classifier itself stays in the server — this crate deliberately does
+/// not depend on `skardi` (#206) — so the mapping from its variants onto these
+/// constants lives in `server::query_handlers::adhoc_statement_kind`. The
+/// vocabulary is owned here; only the translation is over there.
+pub const QUERY_STATEMENT_KIND: &str = "query";
+/// See [`QUERY_STATEMENT_KIND`].
+pub const OTHER_STATEMENT_KIND: &str = "other";
+
 /// Statement-kind marker for job-submission rows. Nothing outside this
 /// module reads it directly (Task 2 goes through `record_job_submitted`),
 /// so it stays private, mirroring `PIPELINE_STATEMENT_KIND`'s narrowing.
@@ -1040,7 +1125,7 @@ mod tests {
                 "SELECT * FROM t WHERE ssn = '123-45-6789'",
                 Some(&ctx),
                 100,
-                "Query",
+                "query",
             )
             .await
             .unwrap();
@@ -1053,7 +1138,7 @@ mod tests {
         assert_eq!(record["ai_context"]["purpose"], json!("kyc"));
         assert_eq!(record["session_id"], json!("sess-1"));
         assert_eq!(record["max_rows"], json!(100));
-        assert_eq!(record["statement_kind"], json!("Query"));
+        assert_eq!(record["statement_kind"], json!("query"));
         assert_eq!(record["status"], json!("started"));
         assert!(record["finished_at"].is_null());
 
@@ -1071,7 +1156,7 @@ mod tests {
     async fn failed_outcome_carries_the_error() {
         let store = QueryAuditStore::open_in_memory().await.unwrap();
         let id = store
-            .record_started("SELECT 1", None, 1, "Query")
+            .record_started("SELECT 1", None, 1, "query")
             .await
             .unwrap();
         store
@@ -1089,15 +1174,15 @@ mod tests {
         let ctx = json!({"purpose": "audit", "session_id": "sess-42"});
         let other = json!({"purpose": "audit", "session_id": "sess-other"});
         store
-            .record_started("SELECT 1", Some(&ctx), 1, "Query")
+            .record_started("SELECT 1", Some(&ctx), 1, "query")
             .await
             .unwrap();
         store
-            .record_started("SELECT 2", Some(&other), 1, "Query")
+            .record_started("SELECT 2", Some(&other), 1, "query")
             .await
             .unwrap();
         store
-            .record_started("SELECT 3", Some(&ctx), 1, "Query")
+            .record_started("SELECT 3", Some(&ctx), 1, "query")
             .await
             .unwrap();
 
@@ -1111,11 +1196,11 @@ mod tests {
     async fn reconcile_marks_orphans_unknown() {
         let store = QueryAuditStore::open_in_memory().await.unwrap();
         let orphan = store
-            .record_started("SELECT 1", None, 1, "Query")
+            .record_started("SELECT 1", None, 1, "query")
             .await
             .unwrap();
         let done = store
-            .record_started("SELECT 2", None, 1, "Query")
+            .record_started("SELECT 2", None, 1, "query")
             .await
             .unwrap();
         store
@@ -1141,7 +1226,7 @@ mod tests {
     async fn prune_deletes_only_older_records() {
         let store = QueryAuditStore::open_in_memory().await.unwrap();
         store
-            .record_started("SELECT 1", None, 1, "Query")
+            .record_started("SELECT 1", None, 1, "query")
             .await
             .unwrap();
         assert_eq!(store.count().await.unwrap(), 1);
@@ -1170,7 +1255,7 @@ mod tests {
         let total_rows = (PRUNE_BATCH_SIZE * 2) + 500;
         for _ in 0..total_rows {
             store
-                .record_started("SELECT 1", None, 1, "Query")
+                .record_started("SELECT 1", None, 1, "query")
                 .await
                 .unwrap();
         }
@@ -1280,7 +1365,7 @@ mod tests {
         let path = dir.path().join("nested").join("audit.db");
         let store = QueryAuditStore::open(&path).await.unwrap();
         store
-            .record_started("SELECT 1", None, 1, "Query")
+            .record_started("SELECT 1", None, 1, "query")
             .await
             .unwrap();
 
@@ -1311,7 +1396,7 @@ mod tests {
         let id = {
             let store = QueryAuditStore::open(&path).await.unwrap();
             store
-                .record_started("SELECT 'durable'", None, 1, "Query")
+                .record_started("SELECT 'durable'", None, 1, "query")
                 .await
                 .unwrap()
         };
@@ -1357,7 +1442,7 @@ mod tests {
         let store = QueryAuditStore::open_in_memory().await.unwrap();
         let ctx = serde_json::json!({"purpose": "p", "session_id": "sess-mix"});
         store
-            .record_started("SELECT 1", Some(&ctx), 10, "Query")
+            .record_started("SELECT 1", Some(&ctx), 10, "query")
             .await
             .unwrap();
         store
@@ -1369,7 +1454,7 @@ mod tests {
         // The promised property is the *ordering* (insertion order via
         // created_at with id as tie-break), plus the sql-column overload —
         // a bare count passes with ORDER BY dropped or reversed.
-        assert_eq!(rows[0]["statement_kind"], json!("Query"));
+        assert_eq!(rows[0]["statement_kind"], json!("query"));
         assert_eq!(rows[0]["sql"], json!("SELECT 1"));
         assert_eq!(rows[1]["statement_kind"], json!("pipeline"));
         assert_eq!(rows[1]["sql"], json!("weekly-churn@1.0.0"));
@@ -1435,7 +1520,7 @@ mod tests {
         let store = QueryAuditStore::open_in_memory().await.unwrap();
         let ctx = serde_json::json!({"purpose": "p", "session_id": "sess-all"});
         store
-            .record_started("SELECT 1", Some(&ctx), 10, "Query")
+            .record_started("SELECT 1", Some(&ctx), 10, "query")
             .await
             .unwrap();
         store
@@ -1448,7 +1533,7 @@ mod tests {
             .unwrap();
         let rows = store.list_by_session("sess-all").await.unwrap();
         assert_eq!(rows.len(), 3);
-        assert_eq!(rows[0]["statement_kind"], json!("Query"));
+        assert_eq!(rows[0]["statement_kind"], json!("query"));
         assert_eq!(rows[1]["statement_kind"], json!("pipeline"));
         assert_eq!(rows[2]["statement_kind"], json!("job"));
         assert_eq!(rows[2]["sql"], json!("nightly-backfill@2.1.0"));
@@ -1497,5 +1582,126 @@ mod tests {
         let row = store.get(&id).await.unwrap().unwrap();
         assert_eq!(row["statement_kind"], json!("job"));
         assert!(row["job_run_id"].is_null());
+    }
+
+    #[tokio::test]
+    async fn open_creates_job_run_id_index_on_migrated_and_fresh_databases() {
+        // The index cannot live in INIT_SCHEMA_SQL — that batch also runs
+        // against pre-column databases, where indexing job_run_id would fail
+        // — so it is applied in `ensure_schema_additions`, after the column is
+        // guaranteed. Pin that it lands on both a fresh database and a
+        // migrated one.
+        async fn index_exists(store: &QueryAuditStore) -> bool {
+            store
+                .conn
+                .call(|conn| -> SqlResult<bool> {
+                    conn.prepare(
+                        "SELECT 1 FROM sqlite_master
+                          WHERE type = 'index' AND name = 'idx_query_audit_job_run_id'",
+                    )?
+                    .exists([])
+                })
+                .await
+                .unwrap()
+        }
+
+        let tmp = tempfile::TempDir::new().unwrap();
+
+        let fresh = QueryAuditStore::open(tmp.path().join("fresh.db"))
+            .await
+            .unwrap();
+        assert!(
+            index_exists(&fresh).await,
+            "fresh database is missing the job_run_id index"
+        );
+
+        let old_path = tmp.path().join("old.db");
+        {
+            let conn = rusqlite::Connection::open(&old_path).unwrap();
+            conn.execute_batch(
+                "CREATE TABLE query_audit (
+                    id TEXT PRIMARY KEY, created_at TEXT NOT NULL,
+                    finished_at TEXT, sql TEXT NOT NULL, ai_context TEXT,
+                    session_id TEXT, max_rows INTEGER NOT NULL,
+                    statement_kind TEXT NOT NULL, status TEXT NOT NULL,
+                    row_count INTEGER, error TEXT);",
+            )
+            .unwrap();
+        }
+        let migrated = QueryAuditStore::open(&old_path).await.unwrap();
+        assert!(
+            index_exists(&migrated).await,
+            "migrated database is missing the job_run_id index"
+        );
+    }
+
+    #[test]
+    fn duplicate_column_error_is_recognised() {
+        // The migration's check-then-ALTER is not atomic, so a second process
+        // can win the race and leave this connection's ALTER failing with
+        // `duplicate column name`. That is the post-condition the step wanted,
+        // so `open()` treats it as success — but only if it can tell that
+        // error apart from a real one. rusqlite reports it as a generic
+        // SQLITE_ERROR with no distinguishing code, hence the text match;
+        // this test is what keeps the match honest against a rusqlite bump.
+        let conn = rusqlite::Connection::open_in_memory().unwrap();
+        conn.execute_batch("CREATE TABLE query_audit (id TEXT PRIMARY KEY);")
+            .unwrap();
+        conn.execute("ALTER TABLE query_audit ADD COLUMN job_run_id TEXT", [])
+            .unwrap();
+        let err = conn
+            .execute("ALTER TABLE query_audit ADD COLUMN job_run_id TEXT", [])
+            .expect_err("adding an existing column must fail");
+        assert!(is_duplicate_column(&err), "unrecognised: {err}");
+
+        // A genuinely different failure must not be swallowed.
+        let other = conn
+            .execute("ALTER TABLE nonexistent ADD COLUMN run_id TEXT", [])
+            .expect_err("altering a missing table must fail");
+        assert!(!is_duplicate_column(&other), "over-matched: {other}");
+    }
+
+    #[tokio::test]
+    async fn job_outcome_does_not_resurrect_a_reconciled_row() {
+        // Two servers sharing one --query-audit-db file: B's startup
+        // reconciliation settles A's in-flight submission to `unknown`, then
+        // A's outcome write lands. Without the started-only guard that late
+        // write would flip a terminal row back to `succeeded` and defeat
+        // reconciliation.
+        let store = QueryAuditStore::open_in_memory().await.unwrap();
+        let id = store
+            .record_job_submitted("nightly-backfill", "2.1.0", Some("sess-j"))
+            .await
+            .unwrap();
+        store.reconcile_orphaned("server B restart").await.unwrap();
+
+        // Not an error: 0 rows updated means something else already settled it.
+        store
+            .record_job_outcome(&id, Some("run-late"), QueryAuditStatus::Succeeded, None)
+            .await
+            .unwrap();
+
+        let row = store.get(&id).await.unwrap().unwrap();
+        assert_eq!(row["status"], json!("unknown"));
+        assert!(row["run_id"].is_null(), "terminal row was resurrected");
+    }
+
+    #[tokio::test]
+    async fn query_outcome_does_not_resurrect_a_reconciled_row() {
+        let store = QueryAuditStore::open_in_memory().await.unwrap();
+        let id = store
+            .record_started("SELECT 1", None, 10, "query")
+            .await
+            .unwrap();
+        store.reconcile_orphaned("server B restart").await.unwrap();
+
+        store
+            .record_outcome(&id, QueryAuditStatus::Succeeded, Some(7), None)
+            .await
+            .unwrap();
+
+        let row = store.get(&id).await.unwrap().unwrap();
+        assert_eq!(row["status"], json!("unknown"));
+        assert!(row["row_count"].is_null(), "terminal row was resurrected");
     }
 }
--- a/crates/query-audit/src/lib.rs
+++ b/crates/query-audit/src/lib.rs
@@ -35,7 +35,6 @@
 //! forever, and pruning/rotation is the operator's call.
 
 use anyhow::{Context, Result, anyhow};
-use axum::http::HeaderMap;
 use serde_json::Value;
 use std::future::Future;
 use std::path::{Path, PathBuf};
@@ -64,77 +63,18 @@ pub(crate) const AUDIT_WRITE_TIMEOUT: Duration = Duration::from_secs(5);
 /// Maximum length of a session id, in characters — shared by `/query`'s
 /// `ai_context.session_id` and the `x-skardi-session-id` header consumed by
 /// both the pipeline execute endpoint and the jobs submit endpoint (see
-/// `query_handlers::validate_ai_context` and [`session_id_from_headers`]),
+/// `query_handlers::validate_ai_context` and
+/// `session_header::session_id_from_headers` in the server crate),
 /// so all three paths can't drift apart. It is an opaque grouping key, not a
 /// payload.
 ///
-/// `skardi-cli` restates this cap under the same name (`run.rs`) because the
-/// CLI crate does not depend on this one; keep them in sync.
+/// `skardi-cli` restates this cap under the same name
+/// (`crates/cli/src/session.rs`) because the CLI crate does not depend on this
+/// one; keep them in sync.
 /// `pub` because the handlers that enforce this bound now live in a
 /// different crate: `pub(crate)` was the right scope while this module was
 /// inside the server, and the extraction is what widened it.
 pub const MAX_SESSION_ID_CHARS: usize = 200;
-
-/// Optional caller-supplied session header. A header (not a body field)
-/// because the pipeline-execute body IS the flattened parameter map — a
-/// reserved key could collide with a legitimate SQL parameter of the same
-/// name. The jobs submit endpoint shares the header for the same reason:
-/// its body is the job's parameter map.
-///
-/// Not an auth credential: unrelated to `require_session`. The value is
-/// caller-asserted, so ledger session attribution is self-reported rather
-/// than derived from an authenticated principal — the same property as
-/// `/query`'s `ai_context.session_id`.
-pub(crate) const SESSION_ID_HEADER: &str = "x-skardi-session-id";
-
-/// Extract and validate the session header. `Ok(None)` when absent; `Err`
-/// when present but malformed — silently dropping a malformed value would
-/// corrupt session stitching, the one job this field has.
-///
-/// Allowed characters are visible ASCII graphic characters, excluding comma —
-/// no space, no tab, no other whitespace or control character. Space is
-/// rejected (not just trimmed) because `httparse` strips leading/trailing
-/// whitespace per RFC 9110 §5.5 before this code ever sees the value: if
-/// spaces were merely tolerated, `"  sess-1  "` would be recorded under the
-/// different key `sess-1` (silently breaking the session stitching this
-/// field exists for) while `"   "` would 400 here — the exact fail-late
-/// round trip a client-side check exists to prevent. A grouping key has no
-/// legitimate use for spaces, so this rejects them outright instead of
-/// defining trimming semantics. Commas are rejected because RFC 9110 §5.3
-/// lets any intermediary merge repeated header lines into one comma-joined
-/// value, which would otherwise slip a duplicate past the check below as a
-/// single merged `"id1, id2"`. This predicate is mirrored exactly by the
-/// CLI's own check in `crates/cli/src/session.rs`.
-pub(crate) fn session_id_from_headers(headers: &HeaderMap) -> Result<Option<String>, String> {
-    let mut values = headers.get_all(SESSION_ID_HEADER).iter();
-    let Some(value) = values.next() else {
-        return Ok(None);
-    };
-    if values.next().is_some() {
-        return Err(format!(
-            "{SESSION_ID_HEADER} must not be sent more than once"
-        ));
-    }
-    let s = value
-        .to_str()
-        .map_err(|_| format!("{SESSION_ID_HEADER} must contain only visible ASCII characters"))?;
-    if !s.chars().all(|c| c.is_ascii_graphic() && c != ',') {
-        return Err(format!(
-            "{SESSION_ID_HEADER} must contain only visible ASCII characters, \
-             with no spaces and no commas (proxies may merge repeated header \
-             lines into one comma-separated value)"
-        ));
-    }
-    if s.is_empty() {
-        return Err(format!("{SESSION_ID_HEADER} must not be empty"));
-    }
-    if s.chars().count() > MAX_SESSION_ID_CHARS {
-        return Err(format!(
-            "{SESSION_ID_HEADER} must be at most {MAX_SESSION_ID_CHARS} characters"
-        ));
-    }
-    Ok(Some(s.to_string()))
-}
 
 /// Error from [`bounded`] that keeps "the write timed out" distinguishable
 /// from "the write ran and failed" (e.g. `ConnectionClosed`). Only the
--- a/crates/query-audit/src/lib.rs
+++ b/crates/query-audit/src/lib.rs
@@ -35,6 +35,7 @@
 //! forever, and pruning/rotation is the operator's call.
 
 use anyhow::{Context, Result, anyhow};
+use axum::http::HeaderMap;
 use serde_json::Value;
 use std::future::Future;
 use std::path::{Path, PathBuf};
@@ -61,10 +62,11 @@ use tokio_rusqlite::{Connection, params, rusqlite};
 pub(crate) const AUDIT_WRITE_TIMEOUT: Duration = Duration::from_secs(5);
 
 /// Maximum length of a session id, in characters — shared by `/query`'s
-/// `ai_context.session_id` and the pipeline execute endpoint's
-/// `x-skardi-session-id` header (see `query_handlers::validate_ai_context`
-/// and `pipeline_handlers::session_id_from_headers`), so the two paths can't
-/// drift apart. It is an opaque grouping key, not a payload.
+/// `ai_context.session_id` and the `x-skardi-session-id` header consumed by
+/// both the pipeline execute endpoint and the jobs submit endpoint (see
+/// `query_handlers::validate_ai_context` and [`session_id_from_headers`]),
+/// so all three paths can't drift apart. It is an opaque grouping key, not a
+/// payload.
 ///
 /// `skardi-cli` restates this cap under the same name (`run.rs`) because the
 /// CLI crate does not depend on this one; keep them in sync.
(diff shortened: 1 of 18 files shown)


crates/server/src/jobs_handlers.rs
```diff
@@ -106,11 +131,96 @@ pub async fn submit_job_run(
...
+                        &format!("Job '{name}' not found"),
+                        "unknown_job",
+                        Some(serde_json::json!({ "job": name })),
+                    ),
+                ));
+            }
+        }
+    };
+
+    let session_id = session_id_from_headers(&headers).map_err(|msg| {
```
abbccdda, line 184: **Security — the ledger's session attribution is now writable by unauthenticated callers, and this PR is what makes that matter.**

`execute_pipeline_by_name` calls `require_session(&app_state, &headers).await?` as its very first statement (`pipeline_handlers.rs:879`), and `/query` does the same. `submit_job_run` calls it nowhere. The PR body and spec both flag "jobs handlers perform no auth check at all" as pre-existing and out of scope — and as a *submission* concern that's fair. But the consequence changes here:

1. **Attribution forgery.** Before this commit, `X-Skardi-Session-Id` could only reach the ledger through two authenticated seams. After it, any unauthenticated caller who can reach `POST /jobs/:name/run` can mint arbitrary `session_id` values into `query_audit` — including ones belonging to real agent sessions. The Learn stage stitches on exactly this key, so a forged row lands inside a legitimate session's `list_by_session` read and is indistinguishable from a real submission. "Self-reported rather than derived from an authenticated principal" (the doc comment on `SESSION_ID_HEADER`) is a defensible property when the caller is at least *authenticated*; on this endpoint the caller is neither.

2. **Cross-endpoint availability coupling.** Every submission — including ones `executor.submit` will reject — now issues writes on `QueryAuditStore`'s **single serialized writer thread**, which the module docs describe as shared with "every audited request's pre-execution INSERT." `prune_before`'s comment spells out the failure mode: work that exceeds `AUDIT_WRITE_TIMEOUT` on that thread "would make every *other* concurrent `/query` and pipeline-execute request time out and 503 even though nothing is wrong with them." This PR hands an unauthenticated actor a cheap, unrate-limited way to queue work onto that thread. Two writes per request, plus unbounded append-only growth of a DB whose retention is off by default.

I'm not asking for `require_session` to be backfilled in this PR — that genuinely is its own scope. But "pre-existing, surfaced as an observation" undersells it now: **the jobs endpoint has become the only unauthenticated write path into the audit ledger.** At minimum this should be called out in `docs/server.md`'s ledger section (so an operator enabling `--query-audit-db` behind auth knows one seam is not gated), and ideally tracked as a follow-up issue rather than a spec bullet.

crates/server/src/jobs_handlers.rs
```diff
@@ -106,11 +131,96 @@ pub async fn submit_job_run(
...
-        })),
+        Ok(run_id) => {
+            finish_job_audit(
+                app_state.query_audit.as_deref(),
+                audit_id.as_deref(),
+                Some(&run_id),
+                QueryAuditStatus::Succeeded,
+                None,
+            )
+            .await;
```
abbccdda, line 246: **Reliability — the `run_id` bridge is lossy and one-directional, with no recovery path.** This is my main design concern against the spec.

The spec's rationale for `run_id` is exactness: "name + timestamp correlation is ambiguous under concurrent submissions of the same job." Agreed. But the stamp is written *after* `executor.submit` has already returned `Ok` and created the run, on a best-effort basis:

- If `record_job_outcome` fails or times out, `finish_job_audit` logs and moves on (documented, and correct — you can't un-submit). The row keeps `status = started`, `run_id = NULL`.
- If the process dies in that window, `reconcile_orphaned` rewrites the row to `status = unknown`, `run_id = NULL`.

In both cases the run exists in `job_runs` and really did execute, but the correlation is **permanently unrecoverable** — because `job_runs` carries no `session_id` and no audit-row id, so there is no reverse pointer to rebuild from. Precisely the ambiguity `run_id` was introduced to eliminate returns, silently, and the row that survives is the *least* informative one (`unknown` means "may have run after a crash" for query rows; for a job row it means "definitely submitted, linkage lost" — a different fact wearing the same label).

This is structurally weaker than the pipeline case, where a lost outcome stamp costs you only the result, not the identity of the thing that ran.

Cheapest fix that closes it: persist the audit row id (or the session id) alongside the run at submit time, e.g. pass it into `executor.submit` and store it on `job_runs`, making the bridge bidirectional and reconstructable. If that's too invasive for this PR, an alternative is to have `record_job_submitted` mint the audit id and thread it through so the *jobs ledger* owns the durable half. Either way I'd want the current gap stated in the spec's non-goals rather than left implicit, since "exact correlation" is the stated point of the column.

crates/server/src/query_audit.rs
```diff
@@ -160,7 +223,8 @@ const INIT_SCHEMA_SQL: &str = "CREATE TABLE IF NOT EXISTS query_audit (
     statement_kind TEXT NOT NULL,
     status         TEXT NOT NULL,
     row_count      INTEGER,
-    error          TEXT
+    error          TEXT,
+    run_id         TEXT
```
abbccdda: **Missing index on `run_id`.** `INIT_SCHEMA_SQL` gains four indexes (`session_id, created_at`, `created_at`, `status`, `statement_kind`) but none on the new column.

The bridge's whole purpose is correlation, and the interesting direction is usually run → session ("I'm looking at run `r-abc` in `GET /jobs/runs`; which agent session submitted it?"). That query is a full table scan on a table that is append-only, has retention **off by default**, and is explicitly documented as growing "forever." The forward direction (session → runs) is already served by `idx_query_audit_session_created`; the reverse has nothing.

One line, and it costs almost nothing since the column is NULL on every non-job row:

```sql
CREATE INDEX IF NOT EXISTS idx_query_audit_run_id
    ON query_audit (run_id) WHERE run_id IS NOT NULL;
```

Note this also needs to be added to the migration block below, not just to `INIT_SCHEMA_SQL` — `CREATE INDEX IF NOT EXISTS` is idempotent, so it can simply live in the same guarded step.

crates/server/src/query_audit.rs
```diff
@@ -250,6 +314,22 @@ impl QueryAuditStore {
...
+        // Databases created before the run_id column (the 2026-08-18
+        // jobs-audit change) lack it, and CREATE TABLE IF NOT EXISTS will not
+        // add columns. Idempotent:
+        // checked against pragma table_info on every open.
+        conn.call(|conn| -> SqlResult<()> {
+            let has_run_id = conn
+                .prepare("SELECT 1 FROM pragma_table_info('query_audit') WHERE name = 'run_id'")?
+                .exists([])?;
+            if !has_run_id {
+                conn.execute("ALTER TABLE query_audit ADD COLUMN run_id TEXT", [])?;
```
abbccdda: **Migration race: concurrent `open()` on the same file aborts server startup.**

The check-then-act is not atomic. Two processes opening the same `--query-audit-db` concurrently (rolling restart, a second server pointed at the same path, an operator tool) both observe `has_run_id == false`, both issue the `ALTER`, and the loser gets `duplicate column name: run_id`. That error propagates out of `open()` → `.context("Failed to migrate query-audit schema (run_id column)")` → startup failure. The failure mode is "server won't boot" for a condition that is actually benign (the column exists, which is all we wanted).

The module already assumes a shared-file world elsewhere (WAL mode, sidecar permission handling), so this isn't hypothetical. Two options, both small:

- Wrap the read + ALTER in a single `BEGIN IMMEDIATE` transaction, or
- Treat the duplicate-column error as success:

```rust
if !has_run_id {
    match conn.execute("ALTER TABLE query_audit ADD COLUMN run_id TEXT", []) {
        Ok(_) => {}
        // Another process won the race; the column exists, which is all
        // this step needed to guarantee.
        Err(e) if e.to_string().contains("duplicate column name") => {}
        Err(e) => return Err(e),
    }
}
```

The existing `open_migrates_old_schema_without_run_id_column` test is good and pins the happy path; this is the concurrent one it can't see.

crates/query-audit/src/lib.rs
```diff
@@ -423,6 +503,97 @@ impl QueryAuditStore {
...
+    ) -> Result<()> {
+        let id = id.to_string();
+        let finished_at = chrono::Utc::now().to_rfc3339();
+        let status = status.as_str();
+        let run_id = run_id.map(str::to_string);
+        let error = error.map(str::to_string);
+        bounded(
+            self.conn.call(move |conn| -> SqlResult<()> {
+                conn.execute(
+                    "UPDATE query_audit
```
abbccdda, line 696: **`record_job_outcome` has no `WHERE status = 'started'` guard**, unlike `spawn_timeout_correction` a few hundred lines down, which deliberately adds `AND status = ?5` and documents why ("0 rows means either the INSERT never landed, or it landed and something else already moved it off `started`").

Within one process this is unreachable — the 503 path returns before any outcome stamp. But this store is a *file*, and combined with the migration race above it's clear multiple processes on one audit DB is a real configuration. In that world: server B starts, `reconcile_orphaned` stamps A's in-flight submission `unknown`, then A's outcome write lands and flips it to `succeeded` — resurrecting a terminal row and defeating the point of reconciliation.

I'd rather see the guard added for the same reason `spawn_timeout_correction` has it: it makes the write monotonic and self-documenting, and it costs nothing. (`record_outcome` shares the gap — consistency argues for fixing both or explicitly documenting that the ledger assumes a single writing process. The latter is a reasonable call, but it should be written down somewhere, since nothing currently prevents the configuration.)

crates/server/src/jobs_handlers.rs
```diff
@@ -106,11 +131,96 @@ pub async fn submit_job_run(
...
+    // the same job set, and today nothing writes to either map after
+    // startup, so the two lookups can never actually disagree. If a future
+    // hot-reload updates one map without the other, this pre-check and
+    // `executor.submit` could resolve different versions of the same job —
+    // the ledger might record a version the run didn't use, or this could
+    // 404 a job the executor would still accept (or vice versa). That
+    // divergence is that future feature's problem to solve, not something
+    // this handler guards against today.
+    let version = {
+        let config = app_state.config.read().unwrap_or_else(|p| p.into_inner());
```
abbccdda, line 168: Nit — **lock-poisoning policy diverges from the sibling handler for the same lock.** Here: `.read().unwrap_or_else(|p| p.into_inner())` (recover and continue). In `execute_pipeline_by_name`: `.read().map_err(...)` → `500 internal_error` (`pipeline_handlers.rs:886`). Same `Arc<RwLock<ServerConfig>>`, opposite conclusions about whether a poisoned config is usable.

Yours is arguably the better choice — `ServerConfig` is effectively immutable after startup, so poison carries no torn state — but two policies for one lock is the kind of thing that reads as an accident later. Worth a one-line comment justifying the divergence, or aligning them.

crates/server/src/jobs_handlers.rs
```diff
@@ -106,11 +131,96 @@ pub async fn submit_job_run(
...
+    };
+
+    let session_id = session_id_from_headers(&headers).map_err(|msg| {
+        (
+            StatusCode::BAD_REQUEST,
+            error_json(&msg, "parameter_validation_error", None),
+        )
+    })?;
+
+    tracing::info!(
```
abbccdda, line 206: Nit — `session_id.as_deref().unwrap_or_default()` emits `session_id=""` on every unattributed submission. The empty string is the one value `session_id_from_headers` explicitly rejects, so the log field's "absent" rendering collides with a value the validator treats as malformed. `tracing` handles `Option` natively; recording `session_id = ?session_id` (or skipping the field when `None`) keeps "no header sent" distinguishable in log-based analysis, which is the same distinction the NULL-vs-value column preserves in the ledger.

docs/server.md
```diff
@@ -178,8 +178,9 @@ execution and updated with its outcome afterwards:
 | `sql` | raw statement text |
 | `ai_context` | the caller's object, verbatim JSON |
 | `session_id` | denormalised from `ai_context` for indexed session lookup |
-| `max_rows` | requested row cap (`0` on pipeline rows — not applicable) |
-| `statement_kind` | `Query` or `Other` for ad-hoc rows (the `Debug` form of the server's statement classifier — not SQL verbs like `select`/`dml`), `pipeline` for pipeline rows. Consumers filtering the ledger must match these exact strings. |
+| `max_rows` | requested row cap (`0` on pipeline and job rows — not applicable) |
+| `run_id` | job_runs.id bridge, job rows only |
+| `statement_kind` | `Query` or `Other` for ad-hoc rows (the `Debug` form of the server's statement classifier — not SQL verbs like `select`/`dml`), `pipeline` for pipeline rows, `job` for job rows. Consumers filtering the ledger must match these exact strings. |
```
abbccdda: Nit — the `statement_kind` value set is now `Query` / `Other` / `pipeline` / `job`: two `PascalCase` values (the `Debug` form of the statement classifier) and two `lowercase` ones. The line already warns "Consumers filtering the ledger must match these exact strings," which is the right mitigation for today, but this PR is the second one to widen the set and every addition makes the inconsistency more expensive to unwind. Since `--query-audit-db` is still unreleased, normalizing to one casing (with the classifier's `Debug` output mapped explicitly rather than leaked) is cheaper now than it will ever be again. Not blocking — but worth a decision rather than a third widening.

abbccdda (commented):
Thorough pass focused on correctness/security/maintainability first, then the implementation against `docs/superpowers/specs/2026-08-18-jobs-audit-design.md`.

The spec is faithfully implemented — every clause I checked (row shape, `run_id` bridge, guarded migration, precedence ladder, record-before-submit fail-closed, helper relocation, CLI extraction, the documented parameter-validation asymmetry) has a matching test. The comments are unusually load-bearing and the TOCTOU note is *verified accurate*: `JobExecutor` is built from `config.jobs.clone()` (`server.rs:253`) and the only `config.write()` in the tree is test-only (`pipeline_handlers.rs:1988`).

What I'd want addressed before merge is concentrated on **end-to-end reliability of the bridge itself** and one **security seam the spec explicitly deferred but that this PR materially changes**. Details inline.

crates/server/src/query_audit.rs
BtXin: Added, as a partial index so it stays proportional to job submissions rather than to the whole ledger:

```sql
CREATE INDEX IF NOT EXISTS idx_query_audit_run_id
    ON query_audit (run_id) WHERE run_id IS NOT NULL;
```

One correction to the suggested placement: it cannot go in `INIT_SCHEMA_SQL` at all, not even alongside. That batch also runs against pre-column databases — it executes *before* the migration step — so indexing `run_id` there would fail on exactly the databases the migration exists to rescue. It lives in the guarded step instead (after the ALTER, unconditionally, since `IF NOT EXISTS` is idempotent), plus `open_in_memory`.

`open_creates_run_id_index_on_migrated_and_fresh_databases` pins it on both paths against `sqlite_master`. Docs updated with the reverse-lookup rationale.

crates/server/src/query_audit.rs
BtXin: Fixed with the second option — treating duplicate-column as success, since it *is* the post-condition the step wanted:

```rust
match conn.execute("ALTER TABLE query_audit ADD COLUMN run_id TEXT", []) {
    Ok(_) => {}
    Err(e) if is_duplicate_column(&e) => {}
    Err(e) => return Err(e),
}
```

Preferred over `BEGIN IMMEDIATE` because it also covers the case where the other writer is not a concurrent `open()` at all.

The predicate matches on message text — `rusqlite` surfaces this as a generic `SqliteFailure`/`SQLITE_ERROR` with no distinguishing code — so `duplicate_column_error_is_recognised` produces a real one from a real double-ALTER and asserts on it, plus a missing-table error to prove it does not over-match and swallow genuine failures. That is what keeps the text match honest across a rusqlite bump.

crates/query-audit/src/lib.rs
BtXin: Guard added to both, for the reason you give: the ledger is a file, nothing prevents two servers on one `--query-audit-db`, and reconciliation is defeated if a late write can resurrect a terminal row.

Both now carry `AND status = ?N` bound to `Started`, and the doc comment on `record_job_outcome` spells out the B-reconciles-then-A-stamps sequence and that 0 rows updated is not an error. `record_outcome` points at it rather than restating.

Mutation-checked — dropping the guard from both fails `job_outcome_does_not_resurrect_a_reconciled_row` and `query_outcome_does_not_resurrect_a_reconciled_row`.

crates/server/src/jobs_handlers.rs
BtXin: Kept the recovery and documented why, rather than aligning — because the divergence points the other way from how it reads.

The repo error-handling policy in `.claude/CLAUDE.md` mandates `.unwrap_or_else(|p| p.into_inner())` for `std::sync` lock poisoning specifically so a poisoned lock does not turn into a panic or an outage. So this handler is following policy and `execute_pipeline_by_name` is the outlier; aligning to it would mean converting a correct answer into a 500 for a condition that carries no torn state (`ServerConfig` is a startup snapshot — the only `config.write()` in the tree is test-only, as you verified for the TOCTOU note).

Added a comment naming the sibling explicitly so the next reader sees the divergence is deliberate. Bringing `pipeline_handlers.rs` into line is a separate, safe cleanup — outside this diff.

crates/server/src/jobs_handlers.rs
BtXin: Fixed — now `session_id = ?session_id`, recording the `Option` directly.

Comment on the line makes the collision explicit, since it is the non-obvious half: `""` is precisely the shape `session_id_from_headers` rejects, so the old rendering made "no header sent" indistinguishable from "header the validator called malformed" in log analysis — the same distinction NULL-vs-value preserves in the ledger.

docs/server.md
BtXin: Normalized. Ad-hoc rows now record `query` / `other`, so all four values share one casing.

The `Debug` form is no longer leaked at all — the mapping is explicit in `query_audit`, next to `PIPELINE_STATEMENT_KIND` and `JOB_STATEMENT_KIND`, so the whole vocabulary lives in one module:

```rust
pub(crate) fn adhoc_statement_kind(kind: StatementKind) -> &'static str {
    match kind {
        StatementKind::Query => QUERY_STATEMENT_KIND,
        StatementKind::Other => OTHER_STATEMENT_KIND,
    }
}
```

That also unbinds the ledger contract from a type in another crate whose variant names could be renamed freely — a rename would have silently changed a documented consumer-facing string.

Updated `docs/server.md`, the module docs, and test assertions. `2026-08-14-pipeline-audit-design.md`'s `statement_kind` row is marked superseded and this spec's scope line records the decision, so the reasoning survives past the diff.

crates/server/src/jobs_handlers.rs
BtXin: Agreed that "pre-existing, surfaced as an observation" undersold it — the second edge you name (this PR is what makes an unauthenticated caller able to write into the ledger at all) is the part the spec bullet failed to say.

Both asks done, and then the fix itself:

**In this PR** — `docs/server.md`'s ledger section now states the seam, and the spec non-goal names the two concrete consequences rather than gesturing at "no auth check": forgeable `session_id` values landing inside a real session's `list_by_session` read, and unrate-limited writes queued onto the shared serialized writer thread whose stalls 503 every concurrent `/query` and pipeline execute.

**Fixed in #221**, stacked on this branch. `require_session` on all five `/jobs/*` handlers, not only the write path — `GET /jobs/runs` returns `job_runs.parameters`, i.e. raw parameter *values*, so gating submissions alone would have closed the smaller hole. Placed as the first statement, ahead of the jobs-disabled 503 and the unknown-job 404 and deliberately outside the precedence ladder this PR establishes among them, so an unauthenticated caller learns neither whether jobs are enabled here nor which job names exist. `verify_session` short-circuits with no auth layer configured, so no-auth deployments are unaffected — pinned by a test rather than left to inspection.

The 401 tests are table-driven over the route list, so a route added without a gate surfaces as a missing entry rather than as silently uncovered surface. Authentication still does not make the header attestable — it bounds who can self-report — and `docs/server.md` now says exactly that instead of the older "not gated" warning.

crates/server/src/jobs_handlers.rs
BtXin: You are right that this was the design gap, and that `unknown` on a job row is a different fact wearing the same label. **Fixed in #222**, stacked on #221.

Took the first option you suggested, with one change in framing. `job_runs` gains a nullable `submission_id`: an *opaque correlation token supplied by the submitter, stored verbatim and never interpreted by the jobs subsystem*. That keeps the core `skardi` crate ignorant of the server's audit ledger — it stores a token whose meaning lives with whoever supplied it, and adds the one thing that makes it useful, `JobStore::get_run_by_submission_id`. Without that framing the change would have put a server-layer audit concern into the core jobs subsystem, which is what made me hesitate to do it inside #219.

The key property is *when* it is written: in the same INSERT that creates the run, not stamped afterwards. So the halves are asymmetric on purpose —

| direction | column | when written | reliability |
| --- | --- | --- | --- |
| audit row → run | `query_audit.run_id` | after `submit` returns | best-effort |
| run → audit row | `job_runs.submission_id` | with the run's INSERT | durable |

`run_id` stays the fast lookup; `submission_id` is what you fall back to when it is NULL. Both directions indexed.

The test that carries the argument is `correlation_survives_a_lost_forward_stamp`: submit with a token, **never** record an outcome, `reconcile_orphaned` to reproduce the crash, assert the audit row really is `unknown` with `run_id = NULL`, then recover the run from the token alone. Driven at the executor seam rather than over HTTP precisely because the handler would race to write the forward stamp the test needs absent.

Also took the "state it in the non-goals" half for what remains genuinely out of reach: `run_id` cannot itself be made reliable, since the run id does not exist until `submit` returns. The fix is a second pointer, not a better first one — and that is now written down rather than implicit.

BtXin:
Rebased onto `main` — #206 (the ledger-crate extraction) had landed, and the conflict turned out to be semantic rather than textual, so flagging the resolution rather than leaving it to be discovered in the diff.

**#206 already added a `run_id` column**, as one of five nullable caller-identity columns:

| | #206's `run_id` | this PR's bridge |
|---|---|---|
| means | which of the *caller's* runs issued this statement | `job_runs.id` — the run this submission produced |
| written | at insert, by `record_started_for` | at outcome, by `record_job_outcome` |
| consumer | skardi-cloud's identity envelope | Learn-stage correlation |

Merging as-is would have left one column meaning two different things depending on `statement_kind` — the same overloading the `statement_kind` casing thread pushed back on, and worse than `sql`'s overloading, which at least varies along one documented axis. Renaming #206's field was not an option: it is merged, and the cloud engine is designed to adopt that ledger wholesale. **So the bridge column is now `job_run_id`**, and a test asserts the bridge stamp leaves the identity `run_id` NULL.

Two structural consequences, both of which made the round-1 fixes land better than they did before:

- **One additive-column mechanism instead of two.** #206's `ensure_identity_columns` already reconciles the live schema against a column list on every open — which is what this PR's standalone guarded `ALTER` was doing for a single column. Generalized to `ensure_added_columns(conn, columns)` and called with two named lists, `IDENTITY_COLUMNS` (who asked) and `BRIDGE_COLUMNS` (what the work became), kept separate because they answer different questions. The upshot for the migration-race fix: the duplicate-column tolerance now covers **all six** added columns, not just ours — #206 introduced the same check-then-act race five more times. Likewise the `job_run_id` index moved into that step, which is also where it has to be: `INIT_SCHEMA_SQL` runs *before* the column exists.

- **`session_id_from_headers` lands in the server, not the ledger.** This PR originally moved it into `query_audit` as its third consumer. That is no longer possible — it needs `axum::http::HeaderMap`, and `skardi-query-audit` depends on neither `axum` nor `skardi`, which is the entire point of the extraction. It now lives in `crates/server/src/session_header.rs`, shared by both handlers, still validating against the ledger's `MAX_SESSION_ID_CHARS` so the three attribution paths cannot drift. The same dependency direction splits `statement_kind`: the marker strings stay `pub` in the ledger, the `StatementKind` → marker mapping sits in `query_handlers`. Vocabulary owned by the ledger, translation owned by the server — which still satisfies the round-1 ask that all four values live in one place.

The two follow-ups are rebased on top and green: #221 (auth) needed no content change beyond the helper's new path; #222 (bridge reconstruction) picked up the rename.

Verified after the rebase — full workspace **1467 passed, 0 failed** on this branch (1471 on #221, 1479 on #222), clippy clean on every touched file. Details in `docs/superpowers/specs/2026-08-18-jobs-audit-design.md` under *Rebase decisions*.

crates/query-audit/src/lib.rs
```diff
@@ -793,8 +979,32 @@ fn restrict_permissions(path: &Path) -> Result<()> {
...
+/// and leaking `Debug` both bound that contract to a type in another crate
+/// whose variant names could be renamed freely, and mixed `PascalCase` ad-hoc
+/// values with the `lowercase` `pipeline` / `job` markers. All four values now
+/// live here, in one casing.
+///
+/// The classifier itself stays in the server — this crate deliberately does
+/// not depend on `skardi` (#206) — so the mapping from its variants onto these
+/// constants lives in `server::query_handlers::adhoc_statement_kind`. The
+/// vocabulary is owned here; only the translation is over there.
+pub const QUERY_STATEMENT_KIND: &str = "query";
```
abbccdda, line 1042: **`statement_kind` casing change has no backfill for existing ledger files.**

Ad-hoc rows previously recorded `Query` / `Other` (the `Debug` form) and this PR switches them to `query` / `other`, but nothing normalizes rows already on disk. `query_audit` is durable, append-only, and retention is off by default, so any server that already ran with `--query-audit-db` ends up with a file containing both casings for the same concept.

Concrete misbehavior: after upgrading, an operator (or the intention-mining consumer this ledger exists for) runs exactly what `docs/server.md` now documents — `SELECT ... WHERE statement_kind = 'query'` — and silently gets only post-upgrade rows; every ad-hoc statement recorded before the upgrade is invisible. The reverse query (`= 'Query'`) silently loses everything after. `idx_query_audit_statement_kind` makes both look fast and correct, so there is no symptom other than wrong counts.

`ensure_schema_additions` is already the one place that reconciles the live file on every open, and is the natural home for a one-shot `UPDATE query_audit SET statement_kind = 'query' WHERE statement_kind = 'Query'` (and the same for `Other`) — idempotent and a no-op on fresh databases, exactly like the column reconciler next to it.

crates/query-audit/src/lib.rs
```diff
@@ -292,7 +352,7 @@ impl QueryAuditStore {
             conn.pragma_update(None, "journal_mode", "WAL")?;
             conn.pragma_update(None, "synchronous", "FULL")?;
             conn.execute_batch(INIT_SCHEMA_SQL)?;
-            ensure_identity_columns(conn)?;
+            ensure_schema_additions(conn)?;
```
abbccdda, line 401: **The concurrent-open scenario this migration now caters to will usually fail with `SQLITE_BUSY`, not `duplicate column name`.**

`ensure_added_columns` now tolerates a lost `ALTER TABLE` race, and `record_job_outcome`'s `started`-only guard is justified by "nothing stops two servers being pointed at one `--query-audit-db`". But this connection never sets `busy_timeout` (only `journal_mode = WAL` and `synchronous = FULL` on line 352-353), so SQLite installs no busy handler and returns `SQLITE_BUSY` immediately whenever another connection holds the write lock.

Concrete scenario: during a rolling restart, server B calls `open()` while server A's writer thread is mid-INSERT. B's `ALTER TABLE ... ADD COLUMN` (or the `CREATE INDEX` right after it) returns `database is locked`, which is not `duplicate column name`, so it propagates out of `ensure_schema_additions` -> `open()` and aborts B's startup — over a transient lock, not a schema problem. The same gap makes every audited request's INSERT fail with `database is locked` under two-writer contention, which the fail-closed policy turns into a 503 for a request that was otherwise fine.

A `conn.busy_timeout(AUDIT_WRITE_TIMEOUT)` (or `pragma_update(None, "busy_timeout", ...)`) alongside the other pragmas would make the duplicate-column tolerance the case it is documented to be, rather than the rare one.

crates/server/src/jobs_handlers.rs
```diff
@@ -106,11 +135,110 @@ pub async fn submit_job_run(
...
+            }
+        }
+    };
+
+    let session_id = session_id_from_headers(&headers).map_err(|msg| {
+        (
+            StatusCode::BAD_REQUEST,
+            error_json(&msg, "parameter_validation_error", None),
+        )
+    })?;
```
abbccdda, line 199: **The jobs 400 drops the `details.header` field the pipeline path supplies, so the two `parameter_validation_error` shapes are not interchangeable.**

`pipeline_handlers.rs:942-956` rejects the identical malformed header with `create_error_response(&msg, "parameter_validation_error", Some(json!({"header": SESSION_ID_HEADER})))`. Here the same `error_type` is returned with `details: None`.

Concrete scenario: a client (or the CLI's error mapping) that branches on `error_type == "parameter_validation_error"` and reads `details.header` to decide "this was the session header, not a job parameter" works against `POST /:pipeline/execute` and silently falls into the *job parameter was invalid* branch against `POST /jobs/:name/run` — the only machine-readable discriminator is absent, and `error_type` is shared with real parameter rejections on this endpoint.

```suggestion
    let session_id = session_id_from_headers(&headers).map_err(|msg| {
        (
            StatusCode::BAD_REQUEST,
            error_json(
                &msg,
                "parameter_validation_error",
                Some(serde_json::json!({ "header": SESSION_ID_HEADER })),
            ),
        )
    })?;
```
(needs `SESSION_ID_HEADER` added to the `crate::session_header` import.)

docs/jobs.md
```diff
@@ -178,6 +178,19 @@ On startup the server:
...
+### Session attribution
+
+`POST /jobs/:name/run` accepts an optional `X-Skardi-Session-Id` request
+header (non-empty, ≤ 200 characters) that attributes the submission to an
+agent session in the query audit ledger. A malformed header is rejected
+with `400 parameter_validation_error` — the header is always validated once
+the job exists, regardless of audit configuration. When `--query-audit-db`
+is configured, the submission is recorded as a `job` row with the session
+id and `name@version` in the ledger, with the submission's outcome
+(`succeeded` or `failed`) and any bridge `run_id`. When the server has
```
abbccdda: **Names the bridge column `run_id`, but the column that ships is `job_run_id`.**

The rebase onto #206 renamed the bridge to `job_run_id` precisely because `run_id` already means something else in the ledger (the caller's own run, from the identity envelope) — `docs/server.md` documents both correctly. This line sends a reader looking for `run_id` on a `job` row, where `run_id` is always NULL and the value they want is in `job_run_id`. The two-column distinction is exactly the thing this PR argues must not be blurred.

```suggestion
(`succeeded` or `failed`) and the `job_run_id` bridge to the run. When the server has
```

Same drift in the newly-added `docs/superpowers/plans/2026-08-18-jobs-audit.md` (line 7 says "a `run_id` bridge" / "One new nullable ledger column (`run_id`)", and line 96 still asserts `statement_kind == "Query"` — both contradicted by what shipped).

crates/query-audit/src/lib.rs
```diff
@@ -1262,4 +1472,236 @@ mod tests {
...
+
+        // Not an error: 0 rows updated means something else already settled it.
+        store
+            .record_job_outcome(&id, Some("run-late"), QueryAuditStatus::Succeeded, None)
+            .await
+            .unwrap();
+
+        let row = store.get(&id).await.unwrap().unwrap();
+        assert_eq!(row["status"], json!("unknown"));
+        assert!(row["run_id"].is_null(), "terminal row was resurrected");
```
abbccdda: **The guard-regression test asserts the wrong column, so half of it is vacuous.**

`record_job_outcome` writes `job_run_id = ?4`; it never touches `run_id` (that is #206's identity envelope, always NULL on this server). So `row["run_id"].is_null()` holds whether or not the `WHERE status = 'started'` guard exists — the assertion labelled *"terminal row was resurrected"* cannot ever detect a resurrection.

The test still fails on guard removal via the `status == "unknown"` assertion two lines up, so the coverage claim in the PR body survives. But the assertion that is supposed to prove the *stamp* did not land is checking a column the stamp never writes:

```rust
assert!(row["job_run_id"].is_null(), "terminal row was resurrected");
```

Worth fixing precisely because this is the `run_id` / `job_run_id` conflation the rebase section spends three paragraphs preventing — it slipped into the one test asserting that very distinction.

crates/server/src/query_handlers.rs
```diff
@@ -126,6 +129,21 @@ fn validate_context_string(
 // is the ledger's, not `/query`'s, and it is shared with `pipeline_handlers`.
 
 /// Execute ad-hoc SQL endpoint - POST /query
+/// The ledger's `statement_kind` marker for an ad-hoc statement of the given
```
abbccdda, line 131: **Misplaced doc comment — `execute_query` lost its documentation to `adhoc_statement_kind`.**

The new function was inserted *between* `/// Execute ad-hoc SQL endpoint - POST /query` (line 131) and the `pub async fn execute_query` it documented. Rustdoc now reads the whole block as one comment on `adhoc_statement_kind`:

> Execute ad-hoc SQL endpoint - POST /query
> The ledger's `statement_kind` marker for an ad-hoc statement of the given kind. …

and the endpoint handler is left undocumented. Move line 131 back down onto `execute_query`.

docs/superpowers/plans/2026-08-18-jobs-audit.md
```diff
@@ -0,0 +1,439 @@
+# Job-Submission Attribution Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Record every job submission in the `query_audit` ledger with session attribution and a `run_id` bridge to the jobs ledger, so agent-submitted jobs join session mining.
+
+**Architecture:** `POST /jobs/:name/run` writes a `started` row (statement_kind=`job`, `name@version`, optional session header) before `executor.submit`, fail-closed like #213; the outcome stamp carries `run_id` on success or a fixed error kind on rejection. Run detail stays in `job_runs`. One new nullable ledger column (`run_id`) via a guarded migration. `session_id_from_headers` moves to `query_audit` (third consumer). CLI gains `skardi job run --session-id` with the validation extracted to a shared module.
```
abbccdda, line 21: **This plan document is added by this PR and is already stale against the code in the same PR.**

The rebase onto #206 changed three decisions, all recorded in the spec's *Rebase decisions* section — but the plan (439 new lines, `docs/superpowers/plans/` being the repo's design record) still describes the pre-rebase design:

| line | says | code does |
| --- | --- | --- |
| 7 | "One new nullable ledger column (`run_id`)" | `job_run_id` |
| 7 | "`session_id_from_headers` moves to `query_audit`" | `crates/server/src/session_header.rs` |
| 96 | `assert_eq!(rows[0]["statement_kind"], json!("Query"))` | `"query"` |

The companion spec was diligent about this — `2026-08-14-pipeline-audit-design.md` got explicit **Superseded** notes in the same PR. Landing a *new* plan that contradicts its own implementation on day one is the one case where a stale doc is strictly worse than no doc: a later reader has no way to tell which of the two documents is current.

Either sync the three lines or add the same one-line supersession pointer to the spec's *Rebase decisions*.

abbccdda (commented):
## Review — jobs-audit (round 2, post-rebase)

Read against `docs/superpowers/specs/2026-08-18-jobs-audit-design.md`, focused on correctness/security first and then on whether the implementation actually delivers what the spec promises end-to-end.

**The core mechanism holds up.** I traced the five properties the spec stakes its reliability claim on and all five are real, not asserted:

- **Fail-closed is genuinely fail-closed.** `executor.submit` awaits `store.create_run` before returning `Ok` (`crates/skardi/src/jobs/executor.rs`), so `audit_write_failure_is_503_and_job_is_not_submitted`'s "jobs ledger is empty" assertion is a sound proof rather than a race. This is stronger evidence than #213 could produce.
- **Precedence is exactly the documented chain** — 503 -> 404 -> 400 -> 503 -> submit — and the 404 pre-check reads `config.jobs`, which `create_server` guarantees is non-empty whenever `app_state.jobs` is `Some` (the executor is built from `config.jobs.clone()`). No deployment can 404 a job the executor would accept.
- **Value-free by construction.** `JobSubmitError::category()` is `&'static str` per variant, so the `failed` row cannot carry parameter text even though `SqlPlanFailure`'s `Display` embeds the rendered SQL. The HTTP body keeps the message; the ledger never sees it. The canary test pins it.
- **The identity/bridge separation is enforced, not just documented** — `record_job_outcome` writes only `job_run_id`, and the round-trip test asserts `run_id` stays NULL.
- **Round 1's four correctness fixes are all real.** Duplicate-column tolerance now covers all six added columns; the `started`-only guard is on both outcome writers; the partial index is correctly placed *after* the column reconcile (it genuinely cannot live in `INIT_SCHEMA_SQL`) and pinned on fresh and migrated files.

Nothing below changes the design. Five findings, all complementary to the three already posted: one test that is weaker than it reads, and four documentation inconsistencies that all orbit the same `run_id` / `job_run_id` collision this PR exists to disentangle.

I did not run the suite locally, per instruction — CI owns that.

---

**On the two deferred items** (unauthenticated `POST /jobs/:name/run`, one-directional bridge): both were raised in round 1 and answered with stacked PRs (#221, #222) rather than argument, and #222's `submission_id`-written-in-the-same-INSERT design is the right shape — the durable half belongs to the ledger that owns the run. I'd note only that until #221 lands, this PR *newly* couples an unauthenticated endpoint to the ledger's single serialized writer: a flood of anonymous submissions can stall that thread, and because `/query` and pipeline execute are fail-closed on it, they 503 in response. That is a new cross-endpoint availability coupling, not merely a new forgeable field, and `docs/server.md` states it plainly. Fine to merge ahead of #221 if the ordering is deliberate; worth being explicit that it is.

**One supporting note on the `busy_timeout` finding** posted at `lib.rs:355`: it is not just an omission relative to SQLite's default (no busy handler — `SQLITE_BUSY` returns immediately). This repo already treats that default as wrong everywhere else it opens SQLite — `crates/skardi/src/sources/providers/sqlite/mod.rs` takes a `busy_timeout_ms` option and *defaults it to 5000*. The audit ledger is the one SQLite connection in the tree that sets `journal_mode` and `synchronous` but not `busy_timeout`, which makes the two-writer scenario this PR now explicitly designs for (duplicate-column tolerance, the `started`-only guard) fail on the lock before it ever reaches the races being guarded.

docs/superpowers/specs/2026-08-18-jobs-audit-design.md
```diff
@@ -0,0 +1,204 @@
...
+| `max_rows` | 0 sentinel (same invariant note) |
+| `job_run_id` | **new nullable column** — `job_runs.id` once the submission is accepted; NULL on non-job rows and on rejected submissions |
+| `status` | submission-scoped: `started` before submit → `succeeded` + `job_run_id` once enqueued → `failed` + fixed error kind if the executor rejects it. For job rows, `succeeded` means "submission accepted", documented on the method; run outcome lives in `job_runs`. (Alternative considered: a new `submitted` terminal status — rejected to keep the status enum and every existing consumer's filter unchanged.) |
+| `error` on rejection | `JobSubmitError::category()` — static per-variant strings (`unknown_job`, `schema_mismatch`, …), value-free by construction; the HTTP response keeps the full message |
+
+### Schema change: `job_run_id`
+
+First actual schema change to the ledger. `--query-audit-db` remains
+unreleased, but dev databases exist, so `open()` gains an idempotent guarded
+migration: `ALTER TABLE query_audit ADD COLUMN run_id TEXT` when
```
abbccdda: **The spec contradicts itself on the column name.**

> `open()` gains an idempotent guarded migration: `ALTER TABLE query_audit ADD COLUMN run_id TEXT` when `pragma table_info` lacks it.

Line 100 of this same file then says the bridge column is `job_run_id`, "not `run_id`", and explains at length why the rename was the only safe resolution. The *Schema change* section above was not updated with it — and it is the section a reader goes to for the actual DDL.

Same one-word fix as the `docs/jobs.md` line, and worth doing for the same reason: the whole point of the rename was that these two names are not interchangeable.

abbccdda (approved):
Approving. The design is sound and the round-1 response was substantive — the two deferrals were answered with stacked PRs (#221, #222) rather than argument, and #222's write-`submission_id`-in-the-same-INSERT shape is the right fix for the bridge.

What I verified holds up end-to-end:

- **Fail-closed is a real proof, not an assertion.** `executor.submit` awaits `store.create_run` before returning `Ok`, so `audit_write_failure_is_503_and_job_is_not_submitted`'s "jobs ledger is empty" is sound rather than racy.
- **Value-free by construction.** `JobSubmitError::category()` is `&'static str` per variant, so `SqlPlanFailure`'s rendered-SQL message cannot reach the ledger even though it reaches the HTTP body.
- **No deployment can 404 a job the executor would accept** — `create_server` builds the executor from `config.jobs.clone()` and only when it is non-empty, so the new pre-check reads the same set.
- **The identity/bridge separation is enforced, not just documented** — `record_job_outcome` writes only `job_run_id`.

None of my seven comments block: the two `low`s are polish, the docs drift is mechanical, and the vacuous test assertion still fails on guard removal via its `status` check. I'd like the two `medium`s in before merge — the missing `busy_timeout` and the `statement_kind` backfill — but neither changes the design and both are one commit.

---

**The caveat, and the reason I want it acknowledged rather than just noted:**

Until #221 lands, this PR newly couples an **unauthenticated** endpoint to the ledger's single serialized writer thread. `docs/server.md` states the two consequences accurately, but the framing there — "one seam is not gated" — reads as an *attribution* problem, and the second edge is an *availability* one:

`/query` and `POST /:pipeline/execute` are fail-closed on that writer. Before this PR, `POST /jobs/:name/run` did not touch it at all. After it, anyone who can reach the jobs endpoint can queue unrate-limited writes onto the thread those two authenticated endpoints block on — and when it stalls, they 503 by design. That is not "an unauthenticated caller can forge a session id"; it is "an unauthenticated caller can deny service to authenticated ones, through a mechanism this PR introduces."

I am fine merging ahead of #221 — the ordering is defensible, and holding a correct change hostage to its own follow-up is worse. But it should be a decision someone made, not a property that fell out of the stacking order. Concretely, I would want either:

- #221 to land before this is deployed anywhere with `--query-audit-db` enabled behind auth, or
- the `docs/server.md` bullet to say plainly that the exposure is to *availability of the other two endpoints*, so an operator reading it weighs the right risk.

Everything else is in the inline threads.

crates/query-audit/src/lib.rs
BtXin: Fixed in `a5f1d73`. `conn.busy_timeout(AUDIT_WRITE_TIMEOUT)` alongside the other two pragmas.

The framing that decided it was the second half of the finding rather than the first: the failure is not only startup aborting on `ALTER TABLE`, it is that the duplicate-column tolerance and the `started`-only guard are both written for a two-writer world the connection could not actually survive. Without a busy handler, contention resolves as `database is locked` before either guard is reached — so the two round-1 fixes were guarding races that could not be reached in the configuration that produces them.

Bounded by `AUDIT_WRITE_TIMEOUT` deliberately: every write already passes through `bounded` with that same value, so waiting for the lock cannot outlast the bound the caller is already waiting out — it converts an immediate 503 into a wait the request was budgeted for anyway. The sqlite-provider precedent you cite is what settled the value; matching it keeps the two SQLite openings in the tree on one story.

crates/query-audit/src/lib.rs
BtXin: Fixed in `a5f1d73` — `normalise_statement_kind_casing`, called from `ensure_schema_additions` right after the column reconcile, as suggested.

The detail that made this worth a test rather than a one-liner is the one you named: there is no symptom. `idx_query_audit_statement_kind` serves both spellings equally fast, so a post-upgrade `WHERE statement_kind = 'query'` returns a plausible number that is simply missing every pre-upgrade row — and since the ledger is append-only with retention off by default, those rows never age out of the wrong answer.

`open_normalises_legacy_statement_kind_casing` seeds a pre-column file with `Query`, `Other`, and a `pipeline` row, opens it, adds a post-upgrade row, and asserts the documented filter now sees both eras, that `pipeline` was untouched, and that a second open does not disturb what it normalised. It fails (`left: 1, right: 2`) with the call removed.

`statement_kind` carries no `COLLATE NOCASE`, so the match is case-sensitive and normalised rows cannot be rewritten again — which is what makes the repeat-open half of the test meaningful rather than decorative.

crates/server/src/jobs_handlers.rs
BtXin: Fixed in `a5f1d73`, with your suggestion and the `SESSION_ID_HEADER` import.

The reason this is worse than plain inconsistency, and worth recording next to the change: on `POST /:pipeline/execute`, `parameter_validation_error` without `details.header` is unambiguous, because that handler validates the header *before* it ever looks at pipeline parameters. On this endpoint the same `error_type` genuinely covers both — job parameter validation happens inside `executor.submit`, which is the documented asymmetry this PR already carries. So `details.header` is not a nicety here, it is the only discriminator that exists; without it a client cannot distinguish the two rejections at all.

The existing 400 test now asserts both `error_type` and `details.header`.

crates/query-audit/src/lib.rs
BtXin: Fixed in `a5f1d73`. Now asserts `job_run_id`.

Verified it is actually load-bearing rather than just correct: with the `WHERE status = 'started'` guard removed *and* the `status` assertion neutralised, the test fails on this line alone (`terminal row was resurrected`). With `run_id` it could not have — you are right that half the test was inert.

Worth noting where it came from: the assertion predates the rebase, when the column genuinely was `run_id`. The rename updated the implementation and the round-trip test but not this one, which is exactly the failure mode the rebase section argues against — so the fix carries a comment saying why the column choice is the load-bearing part, not incidental.

crates/server/src/query_handlers.rs
BtXin: Fixed in `a5f1d73` — `/// Execute ad-hoc SQL endpoint - POST /query` moved back onto `execute_query`.

docs/jobs.md
BtXin: Fixed in `a5f1d73`, your suggestion applied verbatim.

Both of the drifts you flag in the same comment are also fixed: the spec's *Schema change* section (thread below) and the plan document (thread below that).

docs/superpowers/specs/2026-08-18-jobs-audit-design.md
BtXin: Fixed in `a5f1d73`. The *Schema change* section now names `job_run_id` and carries a forward pointer to *Rebase decisions*, so a reader who lands on the DDL section — which, as you say, is where they go for the actual DDL — does not have to reach line 100 to learn the two names are not interchangeable.

docs/superpowers/plans/2026-08-18-jobs-audit.md
BtXin: Took the second of your two options in `a5f1d73` — a supersession banner rather than syncing the three lines — and the reason is that three understates it. Grepping the file, `run_id` appears on 20 lines, and several are code snippets of the *pre-rebase* guarded `ALTER TABLE` that no longer exists in any form: the plan does not merely name the column wrongly, it specifies a mechanism `ensure_added_columns` replaced. Editing those to match what shipped would turn a record of what was planned into a fabricated record of a plan nobody wrote — and `docs/superpowers/plans/` is the repo's record of the former.

So the banner sits directly under the title and names all four divergences in a table — the column, the migration mechanism, the helper's home, and line 96's `"Query"` — each with what shipped and why. It states that the spec's *Rebase decisions* and the code are the authority wherever the two disagree, which is the property your objection actually turns on: not that the plan is stale, but that a later reader has no way to tell which document is current. They now do, before reading a line of it.

If you would rather have the three lines edited in place regardless, say so and I'll switch — the disagreement is about which document lies, not about whether the drift matters.

docs/server.md
BtXin: Reopening this one after your approval note, since the reframe belongs here.

Rewritten in `a5f1d73`. The bullet now splits the seam into two numbered consequences with the availability one second in order but first in weight, and says plainly what you asked for: `/query` and `POST /:pipeline/execute` are fail-closed on the ledger's single serialized writer; before this change `POST /jobs/:name/run` did not touch that thread at all; after it, an unauthenticated caller can stall writes that two *authenticated* endpoints block on, and they 503 by design.

Your framing is what changed the text. The old bullet described the writer-thread contact as a trailing clause of the forgery sentence, which reads as one attribution problem with a side effect. It is two exposures with different victims — the forgery corrupts the Learn stage's view of one session, the contention denies service to callers who did authenticate. An operator weighing "should I enable `--query-audit-db` here" needs the second, and the old text did not give it to them.

BtXin:
## Review round 2 — addressed

All seven inline findings plus the approval caveat, in `a5f1d73`. Replies are in each thread.

| | change |
|---|---|
| Missing `busy_timeout` | `conn.busy_timeout(AUDIT_WRITE_TIMEOUT)` alongside the other pragmas. The sharp half of this was the second one: without a busy handler, contention resolves as `database is locked` *before* either round-1 guard is reached — so the duplicate-column tolerance and the `started`-only guard were protecting races unreachable in the configuration that produces them |
| `statement_kind` casing backfill | one-shot idempotent `UPDATE` in `ensure_schema_additions`, beside the column reconcile. New test seeds a pre-column file with `Query` / `Other` / `pipeline`, opens it, and pins that the documented `WHERE statement_kind = 'query'` sees both eras, that `pipeline` is untouched, and that a second open is a no-op. Fails with the call removed |
| Jobs 400 drops `details.header` | added, your suggestion verbatim. Worse than inconsistency here specifically: the pipeline path validates the header *before* it looks at parameters, so its `error_type` is unambiguous without the field. This endpoint returns the same `error_type` for genuine parameter rejections — the documented asymmetry — so `details.header` is the only discriminator that exists |
| Vacuous guard assertion | now asserts `job_run_id`. Verified load-bearing, not merely correct: with the guard removed *and* the `status` assertion neutralised, it fails on this line alone |
| Misplaced doc comment | `/// Execute ad-hoc SQL endpoint - POST /query` moved back onto `execute_query` |
| `docs/jobs.md` + spec *Schema change* | both name `job_run_id`; the spec section gained a forward pointer to *Rebase decisions*, so the DDL section carries the distinction where a reader actually looks for it |
| Stale plan document | took your second option — supersession banner, not a three-line sync. Reasoning in the thread: `run_id` appears on 20 lines and several are snippets of the pre-rebase `ALTER TABLE` mechanism `ensure_added_columns` replaced, so editing them would fabricate a plan nobody wrote. The banner names all four divergences and points at the spec and the code as authority |

## On the ordering caveat

Taking your second option, and recording it as a decision rather than letting it fall out of the stack: **#219 may merge ahead of #221**, because `docs/server.md` now carries the real risk rather than an understatement of it.

The bullet is rewritten. It splits the seam into two numbered consequences instead of describing the writer-thread contact as a trailing clause of the forgery sentence, and says plainly what you asked for — `/query` and `POST /:pipeline/execute` are fail-closed on the ledger's single serialized writer; before this change `POST /jobs/:name/run` did not touch that thread at all; after it, an anonymous caller can stall writes that two *authenticated* endpoints block on, and they 503 by design. It closes by telling an operator to weigh that when deciding where to enable `--query-audit-db`.

Your framing is what changed the text. The two exposures have different victims — forgery corrupts the Learn stage's view of one session, contention denies service to callers who did authenticate — and the old wording gave an operator only the first.

## Verification

Full workspace green. Two pre-existing flakes, neither touched by this change and both confirmed failing on the unmodified branch:

- `auth::layer` tests race on process-global env vars (1–3 of 6 fail per run). Already fixed in #221 by `cb3591b`.
- `unknown_pipeline_request_is_logged` uses a thread-local `with_default` subscriber that loses to concurrent tests under load; passes in isolation and on rerun.

Clippy clean on every touched file. (`gui.rs:216`'s `approx_constant` error is pre-existing on `main` and untouched here.)

The stack is rebased on top and green: **#221** at `261f1fc`, **#222** at `d31a7a3`. #221's one conflict was `docs/server.md` — it replaces this bullet wholesale, since it closes the gap the bullet describes.

PR #221, opened 2026-08-19
feat(server): require an authenticated session on every /jobs/* endpoint

BtXin:
> **Stacked on #219** (`worktree-jobs-audit`). Review that one first — this branch contains only its own 2 commits. Retarget to `main` once #219 lands.
>
> Rebased after #219 was rebased onto #206 (the ledger-crate extraction). No content change here — the only knock-on was `session_id_from_headers` moving to `crates/server/src/session_header.rs`, since the extracted ledger crate depends on neither `axum` nor `skardi`. Full workspace on this tip: **1471 passed, 0 failed.**

Closes the non-goal #219 recorded rather than fixed, raised in [review](https://github.com/SkardiLabs/skardi/pull/219#discussion_r3802536379).

## Why

The jobs handlers have never called `require_session`, while `/query` ([query_handlers.rs:134](https://github.com/SkardiLabs/skardi/blob/main/crates/server/src/query_handlers.rs#L134)) and `POST /:pipeline/execute` ([pipeline_handlers.rs:879](https://github.com/SkardiLabs/skardi/blob/main/crates/server/src/pipeline_handlers.rs#L879)) both call it as their first statement. On its own that is a submission-authorization gap, and arguably pre-existing. **#219 changed what it costs**, in two ways:

1. **Attribution forgery.** Job submissions now write `session_id` into `query_audit` from `X-Skardi-Session-Id`, which made `POST /jobs/:name/run` the ledger's *only* unauthenticated write path. Anyone who could reach it could mint rows carrying a real agent session's id — and the Learn stage stitches intentions on exactly that key, so a forged row lands inside a legitimate session's `list_by_session` read indistinguishable from a real submission.
2. **Cross-endpoint availability coupling.** Every submission — including ones the executor rejects — issues writes on `QueryAuditStore`'s single serialized writer thread, shared with every audited request. Work exceeding `AUDIT_WRITE_TIMEOUT` there 503s *other* concurrent `/query` and pipeline executions. The endpoint handed an unauthenticated actor an unrate-limited way to queue onto it.

Authentication does not make the header trustworthy — it stays self-reported by design, here as on the other two endpoints. It bounds *who* can self-report to callers the operator admitted.

## What

**All five handlers, not only the write path.** `GET /jobs` exposes job definitions and parameter names; `GET /jobs/runs` returns `job_runs.parameters` — raw parameter *values*, exactly what `query_audit` refuses to store. Gating submissions alone would close the smaller hole.

**First statement, ahead of the jobs-disabled 503 and the unknown-job 404** — deliberately outside the precedence ladder #219 established among those checks. An unauthenticated caller learns neither whether jobs are enabled on this server nor which job names exist; both are otherwise readable straight off the status code.

**One envelope.** `require_session` reports in the shared `ErrorResponse` shape while the jobs endpoints answer in `JobErrorResponse`. A thin `require_job_session` adapter keeps the status and re-renders the body, so one endpoint family does not answer in two shapes. `unauthorized` is already the crate's `error_type` for the condition.

**No behaviour change without auth.** `verify_session` returns `Ok` when `auth_layer.as_better_auth()` is `None`, so servers that never configured auth are unaffected — pinned by a test rather than left to inspection.

## Tests

New `crates/server/tests/jobs_auth_http.rs`, 4 tests. The 401 cases are **table-driven over the route list**, so a route added to `configure_routes` without a gate shows up as a missing entry rather than as silently uncovered surface — that seemed worth more than five near-identical tests, given the failure mode this PR exists to fix.

- every route 401s with no session, and with a bogus bearer token
- the rejection arrives in `JobErrorResponse`'s envelope, not `/query`'s
- auth precedes both the unknown-job 404 and the jobs-disabled 503
- a no-auth server still serves `GET /jobs`, `POST /jobs/:name/run` (real run created) and `GET /jobs/runs`

Mutation-checked: removing the gate from any single handler fails the table-driven tests.

## Also here: a pre-existing test flake, split into its own commit

`auth::layer::tests::build_better_auth_missing_secret_errors` removes `AUTH_SECRET` and asserts the build fails; four sibling tests set it. They share one test binary that libtest runs in parallel, and env vars are process-global — so the removal test intermittently sees a secret another test just set. It reproduces under `cargo test --workspace` (concurrent test binaries change the interleaving) and **not** under `cargo test -p skardi-server --lib` alone, which is why it has read as random.

Unrelated to the auth gate; separated as `72ee9b3` so this change is not blamed for it. Serialized with a `tokio::sync::Mutex` rather than `std::sync::Mutex` because the guard must be held across `AuthLayer::build(..).await` — locking only the mutation would still let another test set the variable before the build reads it.

## Not fixed here

`crates/server/tests/pipeline_audit_http.rs`'s `unknown_pipeline_request_is_logged` is also flaky under full-workspace load (thread-local `tracing` dispatcher). Pre-existing on `main`, in a file neither this PR nor #219 touches — flagging rather than folding in.

## Non-goals

- Authorization beyond "has a valid session" — no per-job ACLs, no submit/cancel distinction. Same posture as `/query` and pipelines today.
- Making the session header attestable. It stays self-reported; see the ledger section of `docs/server.md`.
- `job_runs.parameters` storing raw values. Gating the read narrows exposure but does not resolve the threat-model question #219 flagged alongside #217.

Design notes in `docs/superpowers/specs/2026-08-19-jobs-auth-design.md`.

### DIFF
--- a/crates/server/src/jobs_handlers.rs
+++ b/crates/server/src/jobs_handlers.rs
@@ -21,6 +21,7 @@ use serde_json::Value;
 use skardi::jobs::{JobRun, JobSubmitError};
 use std::collections::HashMap;
 
+use crate::auth::routes::require_session;
 use crate::query_audit::{QueryAuditStatus, QueryAuditStore};
 use crate::server::AppState;
 use crate::session_header::{SESSION_ID_HEADER, session_id_from_headers};
@@ -59,6 +60,26 @@ fn error_json(msg: &str, kind: &str, details: Option<Value>) -> Json<JobErrorRes
         details,
         timestamp: chrono::Utc::now().to_rfc3339(),
     })
+}
+
+/// Reject the request unless it carries a valid session.
+///
+/// `require_session` reports through the shared `ErrorResponse` shape; the
+/// jobs endpoints answer in `JobErrorResponse`. Rather than let one endpoint
+/// family answer in two shapes, the status is kept and the body re-rendered —
+/// `unauthorized` is already this crate's `error_type` for the condition, so
+/// the only thing a client sees differ from `/query` is the envelope it
+/// already expects from `/jobs/*`.
+///
+/// A no-auth server short-circuits inside `verify_session`, so this is a
+/// no-op for deployments that never configured auth.
+async fn require_job_session(
+    app_state: &AppState,
+    headers: &HeaderMap,
+) -> Result<(), (StatusCode, Json<JobErrorResponse>)> {
+    require_session(app_state, headers)
+        .await
+        .map_err(|(status, body)| (status, error_json(&body.error, "unauthorized", None)))
 }
 
 fn submit_error_status(err: &JobSubmitError) -> StatusCode {
@@ -124,6 +145,13 @@ pub async fn submit_job_run(
     Path(name): Path<String>,
     Json(req): Json<SubmitRunRequest>,
 ) -> Result<Json<SubmitRunResponse>, (StatusCode, Json<JobErrorResponse>)> {
+    // First statement, ahead of the jobs-disabled probe and the job-existence
+    // lookup, matching `execute_pipeline_by_name` and `/query`. Deliberately
+    // *outside* the status-precedence ladder those two checks establish: an
+    // unauthenticated caller learns neither whether jobs are enabled on this
+    // server nor which job names exist.
+    require_job_session(&app_state, &headers).await?;
+
     let Some(executor) = app_state.jobs.clone() else {
         return Err((
             StatusCode::SERVICE_UNAVAILABLE,
@@ -277,8 +305,10 @@ pub async fn submit_job_run(
 /// `GET /jobs/runs/:run_id`
 pub async fn get_job_run(
     State(app_state): State<AppState>,
+    headers: HeaderMap,
     Path(run_id): Path<String>,
 ) -> Result<Json<Value>, (StatusCode, Json<JobErrorResponse>)> {
+    require_job_session(&app_state, &headers).await?;
     let Some(executor) = app_state.jobs.clone() else {
         return Err((
             StatusCode::SERVICE_UNAVAILABLE,
@@ -307,8 +337,10 @@ pub struct ListRunsQuery {
 /// `GET /jobs/runs?job=...&limit=...`
 pub async fn list_job_runs(
     State(app_state): State<AppState>,
+    headers: HeaderMap,
     Query(q): Query<ListRunsQuery>,
 ) -> Result<Json<Value>, (StatusCode, Json<JobErrorResponse>)> {
+    require_job_session(&app_state, &headers).await?;
     let Some(executor) = app_state.jobs.clone() else {
         return Err((
             StatusCode::SERVICE_UNAVAILABLE,
@@ -337,8 +369,10 @@ pub async fn list_job_runs(
 /// the executor.
 pub async fn cancel_job_run(
     State(app_state): State<AppState>,
+    headers: HeaderMap,
     Path(run_id): Path<String>,
 ) -> Result<Json<Value>, (StatusCode, Json<JobErrorResponse>)> {
+    require_job_session(&app_state, &headers).await?;
     let Some(executor) = app_state.jobs.clone() else {
         return Err((
             StatusCode::SERVICE_UNAVAILABLE,
@@ -362,7 +396,9 @@ pub async fn cancel_job_run(
 /// for CLI discovery.
 pub async fn list_jobs(
     State(app_state): State<AppState>,
+    headers: HeaderMap,
 ) -> Result<Json<Value>, (StatusCode, Json<JobErrorResponse>)> {
+    require_job_session(&app_state, &headers).await?;
     let Some(executor) = app_state.jobs.clone() else {
         return Err((
             StatusCode::SERVICE_UNAVAILABLE,
--- a/crates/server/src/jobs_handlers.rs
+++ b/crates/server/src/jobs_handlers.rs
@@ -66,10 +66,15 @@ fn error_json(msg: &str, kind: &str, details: Option<Value>) -> Json<JobErrorRes
 ///
 /// `require_session` reports through the shared `ErrorResponse` shape; the
 /// jobs endpoints answer in `JobErrorResponse`. Rather than let one endpoint
-/// family answer in two shapes, the status is kept and the body re-rendered —
-/// `unauthorized` is already this crate's `error_type` for the condition, so
-/// the only thing a client sees differ from `/query` is the envelope it
-/// already expects from `/jobs/*`.
+/// family answer in two shapes, the status is kept and the body re-rendered
+/// into this crate's envelope — so the only thing a client sees differ from
+/// `/query` is the wrapper it already expects from `/jobs/*`.
+///
+/// `error_type` is propagated rather than pinned to `unauthorized`, which is
+/// all `require_session` classifies today: `verify_session` already
+/// distinguishes "Authentication required" from "Invalid or expired session"
+/// internally, so a second classification growing upstream would otherwise
+/// change the message while silently leaving the type behind.
 ///
 /// A no-auth server short-circuits inside `verify_session`, so this is a
 /// no-op for deployments that never configured auth.
@@ -79,7 +84,7 @@ async fn require_job_session(
 ) -> Result<(), (StatusCode, Json<JobErrorResponse>)> {
     require_session(app_state, headers)
         .await
-        .map_err(|(status, body)| (status, error_json(&body.error, "unauthorized", None)))
+        .map_err(|(status, body)| (status, error_json(&body.error, &body.error_type, None)))
 }
 
 fn submit_error_status(err: &JobSubmitError) -> StatusCode {
@@ -145,11 +150,17 @@ pub async fn submit_job_run(
     Path(name): Path<String>,
     Json(req): Json<SubmitRunRequest>,
 ) -> Result<Json<SubmitRunResponse>, (StatusCode, Json<JobErrorResponse>)> {
-    // First statement, ahead of the jobs-disabled probe and the job-existence
-    // lookup, matching `execute_pipeline_by_name` and `/query`. Deliberately
-    // *outside* the status-precedence ladder those two checks establish: an
-    // unauthenticated caller learns neither whether jobs are enabled on this
-    // server nor which job names exist.
+    // First statement in the handler, ahead of the jobs-disabled probe and the
+    // job-existence lookup, matching `execute_pipeline_by_name` and `/query`.
+    // Deliberately *outside* the status-precedence ladder those two checks
+    // establish, so an unauthenticated caller cannot read either off the
+    // status code.
+    //
+    // "First" is about the handler body: axum runs extractors ahead of it, so
+    // a malformed `Query`/`Json` still answers 400/422 pre-gate. That leak is
+    // schema-shaped rather than data-shaped, identical on `/query` and the
+    // pipeline endpoints, and does not reach the audit writes this gate exists
+    // to protect — those all live below.
     require_job_session(&app_state, &headers).await?;
 
     let Some(executor) = app_state.jobs.clone() else {
--- a/crates/server/tests/jobs_auth_http.rs
+++ b/crates/server/tests/jobs_auth_http.rs
@@ -136,7 +136,24 @@ spec:
     (state, tmp)
 }
 
+/// Serializes the `AUTH_*` mutations below.
+///
+/// Environment variables are process-global and libtest runs this binary's
+/// tests on parallel threads, so concurrent `set_var` calls are a data race —
+/// which is why edition 2024 makes `set_var` `unsafe`. Identical values do not
+/// rescue it, and the file would become order-dependently flaky the moment one
+/// test wanted a different `AUTH_DB_PATH`.
+///
+/// A `tokio::sync::Mutex` rather than `std::sync::Mutex` because the guard has
+/// to be held across `AuthLayer::build(..).await`: `build` is what reads these
+/// variables, so releasing after the mutation would still let another test
+/// overwrite them before the read. Mirrors `auth::test_env::lock`, which
+/// cannot be reused here — it is `#[cfg(test)]` on the lib and so invisible to
+/// an integration-test binary.
+static ENV_LOCK: tokio::sync::Mutex<()> = tokio::sync::Mutex::const_new(());
+
 async fn better_auth_layer() -> AuthLayer {
+    let _env = ENV_LOCK.lock().await;
     unsafe {
         std::env::set_var("AUTH_SECRET", "test-secret-that-is-at-least-32-characters!");
         std::env::set_var("AUTH_DB_PATH", ":memory:");
--- a/crates/server/tests/jobs_auth_http.rs
+++ b/crates/server/tests/jobs_auth_http.rs
@@ -1,0 +1,249 @@
+//! HTTP integration tests for authentication on the `/jobs/*` endpoints.
+//!
+//! Until this change the jobs handlers performed no auth check at all, while
+//! `/query` and `POST /:pipeline/execute` both called `require_session` as
+//! their first statement. That gap became load-bearing once job submissions
+//! started writing into the query-audit ledger (#219): it made
+//! `POST /jobs/:name/run` the ledger's only unauthenticated write path, so an
+//! unauthenticated caller could mint `session_id` values into rows the Learn
+//! stage stitches on.
+//!
+//! These tests pin both halves: every endpoint 401s without a session when
+//! auth is configured, and every endpoint is unaffected when it is not.
+
+use arrow::array::Int64Array;
+use arrow::datatypes::{DataType, Field, Schema};
+use arrow::record_batch::RecordBatch;
+use axum::body::Body;
+use axum::http::{Request, StatusCode};
+use datafusion::datasource::MemTable;
+use datafusion::prelude::SessionContext;
+use http_body_util::BodyExt;
+use serde_json::{Value, json};
+use skardi::jobs::{JobDefinition, JobExecutor, JobStore, SqliteJobStore};
+use skardi::sources::DataSourceType;
+use std::collections::HashMap;
+use std::io::Write;
+use std::sync::Arc;
+use tempfile::TempDir;
+use tower::ServiceExt;
+
+use skardi_server::auth::layer::AuthLayer;
+use skardi_server::auth::mode::AuthMode;
+use skardi_server::config::{CliArgs, ServerConfig};
+use skardi_server::semantics::SemanticsRegistry;
+use skardi_server::server::{AppState, configure_routes};
+
+const TEST_JOB_NAME: &str = "ingest-all";
+
+/// Every route under `/jobs`, as (method, uri). Kept as one list so a route
+/// added to `configure_routes` without an auth gate shows up as a missing
+/// entry here rather than as silently uncovered surface.
+fn all_job_routes() -> Vec<(&'static str, String)> {
+    vec![
+        ("GET", "/jobs".to_string()),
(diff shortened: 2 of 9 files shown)


crates/server/tests/jobs_auth_http.rs
```diff
@@ -0,0 +1,249 @@
...
+        Arc::clone(&ctx),
+        auth_layer,
+        executor,
+        None,
+        Default::default(),
+    );
+    (state, tmp)
+}
+
+async fn better_auth_layer() -> AuthLayer {
```
abbccdda, line 155: **Test hygiene / latent flake — the one item I'd fix before merge.**

`better_auth_layer()` mutates three process-global env vars and is called from four sites across three `#[tokio::test]`s in this binary (lines 176, 195, 212, 221), which libtest runs on parallel threads. Nothing serializes them.

What makes this worth raising rather than shrugging at: commit `f7f44b91` **in this same PR** adds `auth::test_env::lock()` specifically to fix this hazard for the unit tests, with a docstring explaining that the guard must be held across `AuthLayer::build(..).await` because locking only around the mutation still lets another test change the variable between the write and the build that reads it. That reasoning applies verbatim here, but this file can't reuse the fix — `test_env` is `#[cfg(test)]` inside the lib, so it isn't visible to an integration-test binary.

Two distinct problems:

- **Soundness.** The crate is edition 2024, where `std::env::set_var` is `unsafe` precisely because a concurrent set/get is a data race and UB. Concurrent writes are a race even when the written value is identical, so "all three set the same string" doesn't rescue it. It's UB today that happens to produce the expected result.
- **Latent flake.** It's stable only as long as every caller writes identical values. The moment one test wants a different `AUTH_BASE_URL` or a non-`:memory:` `AUTH_DB_PATH`, this becomes an order-dependent failure that reproduces on CI and not locally — the exact failure mode `f7f44b91` was written to eliminate.

Note `crates/server/tests/query_http.rs` is *not* precedent for leaving this: it mutates env from exactly one test, so it has no intra-binary race.

Cheapest fix that keeps the tests independent — a file-local guard held across the build, mirroring `test_env`:

```rust
static ENV_LOCK: tokio::sync::Mutex<()> = tokio::sync::Mutex::const_new(());

async fn better_auth_layer() -> AuthLayer {
    // Held across `build`, not just the mutation: `build` is what reads these.
    let _env = ENV_LOCK.lock().await;
    unsafe {
        std::env::set_var("AUTH_SECRET", "test-secret-that-is-at-least-32-characters!");
        std::env::set_var("AUTH_DB_PATH", ":memory:");
        std::env::remove_var("AUTH_BASE_URL");
    }
    AuthLayer::build(&AuthMode::BetterAuthDieselSqlite).await.unwrap()
}
```

If you'd rather not duplicate the lock, promoting `test_env` behind a `test-support` feature would let both binaries share one guard — but that's a bigger change than this PR needs, and the duplicate is fine given the docstring next door explains why.

crates/server/src/jobs_handlers.rs
```diff
@@ -124,6 +145,13 @@ pub async fn submit_job_run(
     Path(name): Path<String>,
     Json(req): Json<SubmitRunRequest>,
 ) -> Result<Json<SubmitRunResponse>, (StatusCode, Json<JobErrorResponse>)> {
+    // First statement, ahead of the jobs-disabled probe and the job-existence
```
abbccdda: **Low severity — the stated invariant is slightly stronger than what's enforced.**

"First statement" is accurate about the handler body, but axum runs extractors *before* the body, so the gate isn't actually the first thing an unauthenticated request meets. The spec's justification for hoisting the gate outside the precedence ladder is that a caller must not read anything off the status code, and the extractor layer sits ahead of that.

Most visible on `list_job_runs`, where `Query<ListRunsQuery>` has `limit: Option<usize>`:

```
GET /jobs/runs?limit=abc     (no session)  ->  400, from the Query extractor
GET /jobs/runs               (no session)  ->  401, from the gate
```

So an unauthenticated caller can still distinguish "this route exists and takes a `limit` I got wrong" from "this route exists." Same class on `submit_job_run`: `Json<SubmitRunRequest>` deserializes pre-gate, so a non-object body (`[]`, `"x"`) answers 422 rather than 401. `SubmitRunRequest` is `#[serde(default, flatten)] HashMap<String, Value>`, so any JSON *object* parses — which is why the table test's `json!({})` body never trips this and the tests pass.

This is not a regression and not worth restructuring for: `execute_query` and `execute_pipeline_by_name` have the identical shape, the leak is schema-shaped rather than data-shaped, and axum's 2MB default body cap bounds the pre-auth work. The audit-writer DoS that motivated the PR is genuinely fixed, since that write lives after the gate.

Suggest just narrowing the comment — say the gate precedes every check *inside the handler*, and drop or qualify the absolute "learns neither ... nor ...", since the extractor layer is outside its reach. Worth the same edit in the spec's Placement paragraph. If you do want the invariant to hold literally, the move is `Query<HashMap<String, String>>` (or a middleware gate on the `/jobs` subtree) — I'd call that out of scope here.

crates/server/src/jobs_handlers.rs
```diff
@@ -362,7 +396,9 @@ pub async fn cancel_job_run(
 /// for CLI discovery.
 pub async fn list_jobs(
     State(app_state): State<AppState>,
+    headers: HeaderMap,
 ) -> Result<Json<Value>, (StatusCode, Json<JobErrorResponse>)> {
+    require_job_session(&app_state, &headers).await?;
```
abbccdda, line 412: **Spec-vs-implementation gap: the confidentiality rationale for gating the read paths is defeated by an ungated sibling route.**

The spec argues the read paths must be gated too, because `GET /jobs` "exposes job definitions and parameter names," and that an unauthenticated caller should learn "neither whether jobs are enabled on this server nor which job names exist."

`GET /` (`serve_dashboard`, `crates/server/src/gui.rs:415`, wired at `crates/server/src/server.rs:398`) is not gated and renders a **superset** of what this handler returns. `render_job_card` (`gui.rs:275`) emits, for every registered job: name, version, destination, the full parameter list with types, the `/jobs/{name}/run` endpoint, and a ready-to-send example request body. `serve_dashboard` also branches on `jobs_enabled` (`JOBS_DISABLED_EMPTY` vs. job cards) and interpolates `{{JOB_COUNT}}`.

So every fact the spec lists as protected by this gate is readable with one unauthenticated `curl /`:

```
curl -s http://host/ | grep -o '/jobs/[^/]*/run'    # every job name
```

To be clear about what this does and doesn't undermine: the security-critical half of the PR is untouched. Ledger write attribution and the audit-thread DoS depend only on gating `POST /jobs/:name/run`, and that is fully closed. This is specifically about the *additional* justification given for gating the three read paths — that argument doesn't currently buy what it claims.

Gating the reads is still the right call regardless (`GET /jobs/runs` returns `job_runs.parameters`, i.e. raw parameter *values*, which the dashboard does **not** expose and which is genuinely the bigger exposure — worth making the load-bearing argument, since it stands on its own). So no code change needed here. What I'd change is one of:

- narrow the spec's claim to the run-history reads and drop the "learns ... which job names exist" assertion, or
- add gating the dashboard to the non-goals list with a sentence on why it's deferred, or
- gate `GET /` too, if the intent really is that job names are non-public.

Option 1 or 2 seems right for this PR's scope; option 3 changes dashboard UX and deserves its own change.

docs/server.md
```diff
@@ -328,34 +328,15 @@ ad-hoc queries, pipeline calls, and job submissions in one ordered read.
...
-     Before job auditing existed, `POST /jobs/:name/run` did not touch that
-     thread at all. It now does, unrate-limited and unauthenticated — so an
-     anonymous caller can stall writes that two *authenticated* endpoints
-     block on, plus grow a ledger whose retention is off by default.
-
-  Weigh the second one when deciding where to enable `--query-audit-db`: if
-  the deployment is behind auth, this seam is not gated and the exposure is
-  to the availability of the endpoints that are. Backfilling the check is
-  tracked separately.
+- **Attribution is authenticated but still self-reported.** Every `/jobs/*`
```
abbccdda, line 331: **Docs gap: `docs/jobs.md` wasn't updated, and it's the doc an operator actually reads for this feature.**

This revision is good — "read `session_id` as a correlation key, never as evidence of who ran what" is exactly the right framing, and it correctly stops short of claiming the header became trustworthy. I verified the stronger claim in it too: `pipeline_health_check` only probes table registration and never executes or audits, so there is genuinely no other unauthenticated ledger write path. Accurate as written.

But `docs/jobs.md` still describes the pre-PR world:

- The **HTTP endpoints** table (`docs/jobs.md:168`) lists all five routes with no indication any of them requires a session.
- Line 179: *"When the server was started without `--jobs`, every endpoint above returns `503 Service Unavailable` with `error_type: jobs_disabled`."* This PR deliberately inverts that for authenticated deployments — the gate now sits ahead of the 503, so an unauthenticated caller gets `401 unauthorized` and never sees `jobs_disabled`. As written this sentence will send someone debugging a 401 down the wrong path.
- The **Session attribution** section (`docs/jobs.md:181`) explains `X-Skardi-Session-Id` without mentioning that a session is now also *required*. Those two header concerns are easy to conflate — one is attribution, one is authentication — and this is the spot where the distinction most needs stating.
- The CLI section below it doesn't mention that `skardi job` needs a configured token against an auth-enabled server.

On that last point, the end-to-end path does work: `ApiClient::with_auth` (`crates/cli/src/client.rs:178`) attaches `Authorization: Bearer <token>` on every request, and all five `skardi job` subcommands go through `client.get`/`client.post`/`client.post_with_headers`, so they all inherit it. No CLI code change needed — the gap is purely that nothing tells the user a token is now required, so the first symptom is an opaque 401 from a command that worked yesterday.

Suggest an "Authentication" line under the endpoints table: every `/jobs/*` route requires a valid session when the server is started with auth, the rejection is `401` with `error_type: unauthorized`, the check precedes both the `jobs_disabled` 503 and the unknown-job 404, and servers without auth configured are unaffected. Then qualify line 179 with "for an authenticated caller."

crates/server/src/jobs_handlers.rs
```diff
@@ -61,6 +62,26 @@ fn error_json(msg: &str, kind: &str, details: Option<Value>) -> Json<JobErrorRes
...
+/// `require_session` reports through the shared `ErrorResponse` shape; the
+/// jobs endpoints answer in `JobErrorResponse`. Rather than let one endpoint
+/// family answer in two shapes, the status is kept and the body re-rendered —
+/// `unauthorized` is already this crate's `error_type` for the condition, so
+/// the only thing a client sees differ from `/query` is the envelope it
+/// already expects from `/jobs/*`.
+///
+/// A no-auth server short-circuits inside `verify_session`, so this is a
+/// no-op for deployments that never configured auth.
+async fn require_job_session(
```
abbccdda, line 81: **Nit, no action strictly needed.** The adapter is the right call — one endpoint family answering in two envelopes would be worse than this thin shim.

Two small things:

1. `error_type` is hardcoded to `"unauthorized"` while the message is propagated from `body.error`. That's correct today because `require_session` always passes `"unauthorized"` to `create_error_response`, so the constant and the source agree. But it means if `require_session` ever grows a second classification (say `session_expired`, which `verify_session` already distinguishes internally — it returns different bodies for "Authentication required" vs. "Invalid or expired session"), the message will change and the type silently won't. `.map_err(|(status, body)| (status, error_json(&body.error, &body.error_type, None)))` propagates both and reads no worse. Worth noting the docstring's "`unauthorized` is already this crate's `error_type` for the condition" is the justification for the constant — if you keep it, that's fine, but the coupling is what a future reader will trip on.

2. The rejection body is serialized and reparsed on the way through: `verify_session` builds a JSON string, `require_session` calls `to_bytes` and `serde_json::from_slice` to pull `.error` back out, then this re-renders it into `JobErrorResponse`. Three passes over a static string on an unauthenticated request. Irrelevant for throughput, but it does mean the 401 path has an `unwrap_or_default()` fallback that silently degrades the message to "Authentication required" if the shape ever drifts. Pre-existing in `require_session`, inherited rather than introduced here — flagging only so it's not mistaken for new.

abbccdda (approved):
## Review: `require_session` on `/jobs/*`

**Score: 8.0 / 10 for mergeability.** Approving — the core change is correct, minimal, and closes a real hole. Nothing below blocks the merge; the items are small and can land as fixups on this branch or as an immediate follow-up.

### What's good

The change does exactly what the spec says and no more. `require_job_session` as the first statement in all five handlers matches how `execute_query` and `execute_pipeline_by_name` already gate, so there's one convention rather than two. Placing the gate outside the 503/404 precedence ladder is the right call and is pinned by a test rather than asserted in a comment. The `AuthLayer::None` short-circuit means no-auth deployments are untouched, and that's pinned too — which is the property most likely to break operators and the easiest to regress silently.

The test design is the strongest part. Driving all five routes off one `all_job_routes()` list means a sixth route added to `configure_routes` without a gate surfaces as a missing table entry, which is a structurally different (and better) failure mode than five copy-pasted tests. Asserting the envelope shape, not just the status, catches the realistic regression where someone swaps the adapter for a bare `require_session` and the family starts answering in two shapes. And the claimed mutation check — remove the gate from any one handler and the table test fails — is the right bar for a security gate.

The security reasoning also holds up where it counts. The genuinely serious half of #219's exposure was the write path: `POST /jobs/:name/run` was the audit ledger's only unauthenticated write, so anyone reachable could forge `session_id` rows the Learn stage stitches on, and could queue unrate-limited work onto `QueryAuditStore`'s single serialized writer thread and 503 the two endpoints that *were* authenticated. Both are closed here. I checked the other ungated routes for a second write path and there isn't one — `pipeline_health_check` only probes table registration, it never executes or audits. So the new `docs/server.md` claim that the ledger has no unauthenticated write path is accurate as written.

### What needs doing

Ranked. Only the first is worth doing before merge, and even that is a test-hygiene issue rather than a defect in shipped behavior.

1. **Serialize the env mutation in `jobs_auth_http.rs`** (inline comment). Three tests in the new binary mutate `AUTH_SECRET`/`AUTH_DB_PATH`/`AUTH_BASE_URL` in parallel with no lock. Notable because commit `f7f44b91` in this very PR adds `auth::test_env::lock()` for precisely this hazard — the new file reintroduces it because `test_env` is `#[cfg(test)]` on the lib and integration tests can't reach it.
2. **Update `docs/jobs.md`** (inline comment). `docs/server.md` was revised but the operator-facing jobs doc still documents the endpoints with no auth requirement at all, and its `503`-without-`--jobs` sentence is now only true for authenticated callers.
3. **Reconcile the `GET /jobs` gating rationale with the ungated dashboard** (inline comment). The spec gates the read paths because they expose job definitions and parameter names, but `GET /` renders a superset of that ungated. Either gate the dashboard or narrow the spec's claim — the ledger-integrity goal is unaffected either way.
4. Two low-severity nits inline: extractor-before-gate ordering slightly weakens the "learns nothing from the status code" invariant, and the adapter hardcodes `error_type` instead of propagating it.

### Non-goals I agree with

Leaving the header self-reported, skipping per-job ACLs, and not touching `job_runs.parameters` are all correctly scoped out and honestly documented. The revised `docs/server.md` wording — read `session_id` as a correlation key, never as evidence — is the right thing to tell an operator, and it's an improvement on what it replaced.

crates/server/tests/jobs_auth_http.rs
BtXin: Fixed in a7c3fbf. Added a file-local `tokio::sync::Mutex` held across `AuthLayer::build(..).await`, as suggested, with a docstring pointing at `auth::test_env::lock` and noting why it cannot be reused (`#[cfg(test)]` on the lib, invisible to an integration-test binary).

Took the duplicate rather than the `test-support` feature — the feature would need to be plumbed through `[dev-dependencies]` self-referencing the crate, which is more moving parts than this PR should add for one guard.

Agreed on the soundness framing specifically: this was not "stable because the values match," it was UB that happened to land on the expected result. The write is a race regardless of the value written.

`cargo test -p skardi-server --test jobs_auth_http` → 4 passed.

crates/server/src/jobs_handlers.rs
BtXin: Verified the ordering empirically before editing — with no session, `GET /jobs/runs?limit=abc` → `400 Bad Request` from the `Query` extractor, `GET /jobs/runs` → `401 Unauthorized` from the gate. Exactly as described.

Narrowed rather than restructured, per your recommendation. The comment now says "first statement in the handler," states that axum runs extractors ahead of the body so a malformed `Query`/`Json` still answers 400/422 pre-gate, and characterises the residue as schema-shaped rather than data-shaped. The spec's Placement paragraph carries the same qualification plus the note that a middleware gate on the `/jobs` subtree is the move if the invariant ever needs to hold literally.

Did not add a test pinning the 400-pre-gate behaviour — asserting it would enshrine the leak as intended rather than tolerated, and the probe above was enough to confirm the claim.

crates/server/src/jobs_handlers.rs
BtXin: Confirmed — `serve_dashboard` (`gui.rs:415`) is wired ungated at `server.rs:398`, and `render_job_card` emits name, version, destination, the typed parameter list, the `/jobs/{name}/run` endpoint and an example body for every job. The `curl -s host/ | grep -o "/jobs/[^/]*/run"` path works. The spec claim did not buy what it asserted.

Took options 1 and 2 together in a7c3fbf, since neither is complete alone:

- The `What` section now makes the run history the load-bearing argument — `GET /jobs/runs` and `GET /jobs/runs/:run_id` return `job_runs.parameters`, raw parameter *values*, which nothing else on the server exposes. `GET /jobs` is described as gated for consistency across the family, explicitly not for confidentiality, with the dashboard superset named.
- Non-goals gained an entry for gating `GET /` — why it is deferred (a dashboard UX decision, not a ledger-integrity one) and that the write-attribution and audit-thread fixes do not depend on it.

Left option 3 alone; agreed it changes dashboard UX and deserves its own change. The `Placement` paragraph keeps the status-code claim, which is still true within the handler — that is a different assertion from job names being non-public, and only the latter was defeated.

docs/server.md
BtXin: Fixed in a7c3fbf. `docs/jobs.md` gained an `### Authentication` section under the endpoints table covering all four points: every `/jobs/*` route requires a valid session under auth, the rejection is `401` with `error_type: unauthorized`, the check precedes both the `jobs_disabled` 503 and the unknown-job 404, and no-auth servers are unaffected.

Also:

- Line 179 rewritten — the `503` now reads as holding "for an authenticated caller," with the note that an unauthenticated one sees the 401 instead and never learns whether jobs are enabled.
- Added the distinction you flagged as easiest to conflate, at the seam between the two sections: authentication answers *may this caller use the endpoint at all*, `X-Skardi-Session-Id` answers *which session should this be attributed to* and stays self-reported. Passing the header does not authenticate; authenticating does not make the header trustworthy.
- CLI section notes that `skardi job` needs a token against an auth-enabled server, via `--token` / `$SKARDI_API_TOKEN` / `~/.skardi/config.yaml`, and that the symptom of omitting it is an opaque 401. No code change, matching your read of `ApiClient::with_auth` — the gap was purely that nothing told the user.

crates/server/src/jobs_handlers.rs
BtXin: Took (1). `error_type` now propagates: `.map_err(|(status, body)| (status, error_json(&body.error, &body.error_type, None)))`. Equivalent today, since `require_session` only ever passes `"unauthorized"` to `create_error_response` — but the coupling was the problem, not the current value, and the propagating form reads no worse. Docstring updated to explain that rather than to justify the constant.

On (2): agreed it is inherited from `require_session`, not introduced here, and left alone. The serialize/reparse is three passes over a static string on a rejected request, and the `unwrap_or_default()` degradation only bites if `verify_session`'s body shape drifts — a fix belongs in `require_session` itself (returning the classification rather than a rendered `Response`), which would touch `/query` and the pipeline endpoints too. Noting it here rather than folding it in.

PR #222, opened 2026-08-19
feat(jobs): durable reverse pointer so a lost run_id stamp cannot lose the correlation

BtXin:
> **Stacked on #221**, which has landed. Rebased onto `main`; this branch carries its own 2 commits (the change, then the review fixes).

Closes the second non-goal #219 recorded rather than fixed, raised in [review](https://github.com/SkardiLabs/skardi/pull/219#discussion_r3802536385).

## Why

#219 added `query_audit.job_run_id` because "name + timestamp correlation is ambiguous under concurrent submissions of the same job." True — but the column only removes that ambiguity on the happy path, because the stamp is written **after** `executor.submit` has already returned and created the run:

- `record_job_outcome` fails or times out → `finish_job_audit` logs and moves on (correct — you can't un-submit). Row keeps `status = started`, `job_run_id = NULL`.
- process dies in that window → `reconcile_orphaned` rewrites it to `status = unknown`, `job_run_id = NULL`.

In both cases the run exists in `job_runs` and really did execute, but the correlation was **permanently unrecoverable** — `job_runs` carried neither a session id nor an audit-row id, so there was nothing to rebuild from. The ambiguity `job_run_id` exists to eliminate came back silently, and the surviving row was the *least* informative one: `unknown` means "may have run after a crash" for a query row, but "definitely submitted, linkage lost" for a job row — a different fact wearing the same label.

Structurally weaker than the pipeline case, where a lost outcome stamp costs the result, not the identity of the thing that ran.

## What

`job_runs` gains a nullable `submission_id`: **an opaque correlation token supplied by the submitter, stored verbatim, never interpreted by the jobs subsystem.**

That framing is the layering answer. The core `skardi` crate has no notion of the server's query-audit ledger and does not acquire one here — it stores a token whose meaning lives entirely with whoever supplied it. What it *does* add is the lookup that makes the token useful: `JobStore::get_run_by_submission_id`.

**Written in the same INSERT that creates the run**, not stamped afterwards. That is the whole point — the reverse pointer is durable the moment the run exists, so it cannot be lost to the same window that loses the forward one. The two halves are asymmetric on purpose:

| direction | column | when written | reliability |
| --- | --- | --- | --- |
| audit row → run | `query_audit.job_run_id` | after `submit` returns | best-effort |
| run → audit row | `job_runs.submission_id` | with the run's INSERT | **durable** |

Read the pair as one bridge with a fast half and a reliable half: `job_run_id` for the common lookup, `submission_id` when it is NULL. Both directions are indexed (`job_run_id`'s partial index came with #219; `submission_id` gets its own here).

**`submit` takes the token as a required argument**, not an optional one behind a defaulted overload. It's an attribution seam, so a new call site should have to decide rather than inherit a silent `None`. Seven existing test call sites pass `None` explicitly.

**Migration** reuses #219's pattern: `CREATE TABLE IF NOT EXISTS` no-ops on an existing `jobs.db` and won't add columns, so a guarded idempotent `ALTER` bolts it on — and a process that loses the check-then-act race gets exactly the post-condition it wanted instead of a startup failure.

**Repaired into the ledger at startup, not left to an operator.** A durable pointer only makes the correlation *recoverable* — on its own nothing recovers it, and `record_job_outcome` guards on `status = 'started'`, so once `reconcile_orphaned` has run no later well-behaved write can stamp the row. `setup_app_state` holds both ledgers, so after each has reconciled its own orphans it passes over job rows left `unknown` with `job_run_id IS NULL` and re-links each through `get_run_by_submission_id`. `status` stays `unknown` — the outcome was never observed, only the linkage is recovered. Idempotent, cannot overwrite a correct pointer, cannot touch a row still `started`, logs rather than fails startup.

**Exposure: a filter on the way in, not a field on the way out.** `GET /jobs/runs?submission_id=<token>` resolves the single matching run; run payloads deliberately omit it. A pointer readable only by opening the SQLite file is no recovery path — but neither is a broadcast one: `submission_id` is a `query_audit` primary key, that ledger is chmod 0600, and `GET /jobs/runs` returns every run to any authenticated session (#221 added authentication, not authorization). The filter also serves the operator strictly better, since `list_runs` clamps to 500 with no offset and the run wanted during an incident is exactly the one off the window.

**Uniqueness is enforced** by a partial unique index. `get_run_by_submission_id` resolves `ORDER BY created_at DESC LIMIT 1`, so an unconstrained duplicate returned a *confidently wrong* run — no error, same shape as a right answer. Constrains nothing today (one writer, fresh audit id per submission); NULLs stay free.

**`jobs.db` gets the audit ledger's file posture** — owner-only create, chmod on every open, WAL sidecars included. `job_runs.parameters` holds the raw values `query_audit` refuses to store, and the reverse pointer now links the file to a protected row; two halves of one audit record cannot have two permission decisions.

**WAL + `busy_timeout`**, which is what makes the migration's duplicate-column tolerance complete rather than half: without them the *other* interleaving returns `database is locked`, which is not tolerated and took startup down over the same benign race.

## Tests

**Store units** (5): token round-trips and resolves in reverse; a miss returns `None`; unattributed runs keep NULL without colliding; a pre-column `jobs.db` migrates on open, keeps its old rows readable, accepts attributed writes, and ends up with the index; duplicate-column detection exercised against a real rusqlite error *and* a non-duplicate one so it can't over-match.

**HTTP integration** (4): the bridge points both ways after a real submission; `submission_id` is NULL on an unaudited server; the runs API exposes it.

The load-bearing one is **`correlation_survives_a_lost_forward_stamp`** — it submits with a token, **never** records an outcome, runs `reconcile_orphaned` to reproduce the crash, asserts the audit row really is `unknown` with `job_run_id = NULL`, then recovers the run from the token alone. Driven at the executor seam rather than over HTTP precisely because the handler would race to write the forward stamp the test needs absent.

Mutation-checked: passing `None` instead of the audit id in the handler fails the bridge tests; disabling the startup repair call site fails `jobs_bridge_startup.rs`. Full workspace: **1517 passed, 0 failed**.\n\n### Review round (9dfd447)\n\nAll ten findings from the review are addressed — see the inline replies. Three changed shipped behaviour: the startup repair pass, the token becoming a filter rather than a response field, and `jobs.db` permissions. One sub-point pushed back on with reasoning (a length cap on the token) and recorded as a non-goal.\n\n`pipeline_audit_http::unknown_pipeline_request_is_logged` fails intermittently under full-workspace load and passes in isolation — pre-existing on `main`, in a file this PR does not touch.

## Non-goals

- Making `job_run_id` itself reliable. It can't be — the run id doesn't exist until `submit` returns. The fix is a second pointer, not a better first one.
- A length cap on the token. `submit` takes an unbounded `&str`, matching the column's opaque-TEXT contract; the only production caller supplies a server-minted audit id, and a large string in a TEXT column is not the class of failure the unique index addresses.
- Periodic repair. The pass runs at startup only; its guards make a scheduled variant additive.
- Putting `session_id` on `job_runs` too. The audit row holds it and is reachable from the token, so a second copy would be denormalized state that can disagree with the ledger.
- `job_runs.parameters` storing raw parameter values — still the open threat-model question from #219 and #217.

Design notes in `docs/superpowers/specs/2026-08-19-jobs-bridge-design.md`.

### DIFF
--- a/crates/server/src/jobs_handlers.rs
+++ b/crates/server/src/jobs_handlers.rs
@@ -140,7 +140,6 @@ fn job_run_to_json(run: &JobRun) -> Value {
         "rows_written": run.rows_written,
         "snapshot_id": run.snapshot_id,
         "error": run.error,
-        "submission_id": run.submission_id,
     })
 }
 
@@ -351,6 +350,13 @@ pub async fn get_job_run(
 pub struct ListRunsQuery {
     pub job: Option<String>,
     pub limit: Option<usize>,
+    /// Resolve the single run carrying this correlation token.
+    ///
+    /// The recovery direction of the audit bridge: an operator holding a
+    /// `query_audit` row id whose `job_run_id` was lost asks for it here. A
+    /// filter on the way in rather than the token on the way out — see
+    /// [`job_run_to_json`], which deliberately does not emit it.
+    pub submission_id: Option<String>,
 }
 
 /// `GET /jobs/runs?job=...&limit=...`
@@ -366,6 +372,27 @@ pub async fn list_job_runs(
             error_json("Jobs subsystem is not enabled", "jobs_disabled", None),
         ));
     };
+    // An exact-token lookup, not a filter over the recent window: the run an
+    // operator needs during an incident is the one that has already fallen off
+    // it, which is exactly when concurrent submissions made `job_run_id` worth
+    // having. `job` and `limit` are ignored — one token names at most one run.
+    if let Some(token) = q.submission_id.as_deref() {
+        return match executor.store().get_run_by_submission_id(token).await {
+            Ok(run) => {
+                let body: Vec<Value> = run.iter().map(job_run_to_json).collect();
+                Ok(Json(serde_json::json!({
+                    "success": true,
+                    "runs": body,
+                    "count": body.len(),
+                })))
+            }
+            Err(e) => Err((
+                StatusCode::INTERNAL_SERVER_ERROR,
+                error_json(&e.to_string(), "internal_error", None),
+            )),
+        };
+    }
+
     let limit = q.limit.unwrap_or(50).clamp(1, 500);
     match executor.store().list_runs(q.job.as_deref(), limit).await {
         Ok(runs) => {
--- a/crates/server/src/jobs_handlers.rs
+++ b/crates/server/src/jobs_handlers.rs
@@ -140,6 +140,7 @@ fn job_run_to_json(run: &JobRun) -> Value {
         "rows_written": run.rows_written,
         "snapshot_id": run.snapshot_id,
         "error": run.error,
+        "submission_id": run.submission_id,
     })
 }
 
@@ -273,7 +274,14 @@ pub async fn submit_job_run(
         None => None,
     };
 
-    match executor.submit(&name, req.parameters).await {
+    // The audit row id doubles as the run's submission token, so the two
+    // ledgers point at each other. This half is durable at run-creation time;
+    // the `run_id` stamp below is best-effort, and if it is lost the
+    // correlation is still recoverable through `get_run_by_submission_id`.
+    match executor
+        .submit(&name, req.parameters, audit_id.as_deref())
+        .await
+    {
         Ok(run_id) => {
             finish_job_audit(
                 app_state.query_audit.as_deref(),
--- a/crates/skardi/src/jobs/store.rs
+++ b/crates/skardi/src/jobs/store.rs
@@ -30,12 +30,31 @@ const INIT_SCHEMA_SQL: &str = "CREATE TABLE IF NOT EXISTS job_runs (
     finished_at   TEXT,
     rows_written  INTEGER,
     snapshot_id   TEXT,
-    error         TEXT
+    error         TEXT,
+    submission_id TEXT
 );
 CREATE INDEX IF NOT EXISTS idx_job_runs_name_created
     ON job_runs (job_name, created_at DESC);
 CREATE INDEX IF NOT EXISTS idx_job_runs_status
     ON job_runs (status);";
+
+/// Index over [`JobRun::submission_id`], applied after the migration below has
+/// guaranteed the column exists — so it cannot live in [`INIT_SCHEMA_SQL`],
+/// which also runs against pre-column ledgers.
+///
+/// Partial, because the column is NULL for every run not submitted through an
+/// audited server, and because the only query against it is an equality
+/// lookup for one submission.
+const SUBMISSION_ID_INDEX_SQL: &str = "CREATE INDEX IF NOT EXISTS idx_job_runs_submission_id
+    ON job_runs (submission_id) WHERE submission_id IS NOT NULL;";
+
+/// True for the `duplicate column name` error SQLite raises when an
+/// `ALTER TABLE ... ADD COLUMN` loses a race with another connection that
+/// already added it. Matched on message text because `rusqlite` reports it as
+/// a generic `SqliteFailure` with no distinguishing code.
+fn is_duplicate_column(e: &rusqlite::Error) -> bool {
+    e.to_string().contains("duplicate column name")
+}
 
 /// Lifecycle status of a single job run.
 ///
@@ -102,6 +121,16 @@ pub struct JobRun {
     /// string). For DB destinations this is typically null.
     pub snapshot_id: Option<String>,
     pub error: Option<String>,
+    /// Opaque correlation token supplied by whoever submitted the run, stored
+    /// verbatim and never interpreted here.
+    ///
+    /// The jobs subsystem has no notion of the query-audit ledger; this is the
+    /// seam that lets a caller which *does* — the server — make the
+    /// correlation reconstructable from either side. It writes its audit row
+    /// id here in the same INSERT that creates the run, so the pointer is
+    /// durable at run-creation time rather than stamped afterwards. See the
+    /// ledger section of `docs/server.md`.
+    pub submission_id: Option<String>,
 }
 
 /// Trait over the run ledger. All methods are async because the default
@@ -110,6 +139,16 @@ pub struct JobRun {
 pub trait JobStore: Send + Sync {
     async fn create_run(&self, run: &JobRun) -> Result<()>;
     async fn get_run(&self, run_id: &str) -> Result<Option<JobRun>>;
+    /// Look a run up by the opaque token its submitter stored on it.
+    ///
+    /// This is the direction that makes the correlation *reconstructable*:
+    /// the submitter already knows its own token, so it can recover the run
+    /// even when its own forward pointer was never written.
+    ///
+    /// Returns the most recently created match. Duplicate tokens are a
+    /// caller error rather than something this ledger enforces — nothing
+    /// here can know whether a given token is meant to be unique.
+    async fn get_run_by_submission_id(&self, submission_id: &str) -> Result<Option<JobRun>>;
     async fn list_runs(&self, job_name: Option<&str>, limit: usize) -> Result<Vec<JobRun>>;
     async fn update_status(
         &self,
@@ -187,6 +226,38 @@ impl SqliteJobStore {
             })
             .await
             .map_err(|e| anyhow::anyhow!("Failed to initialize jobs.db schema: {e}"))?;
+
+        // `CREATE TABLE IF NOT EXISTS` above no-ops on an existing ledger and
+        // will not add columns, so ledgers written before `submission_id`
+        // need the column bolted on. Idempotent: re-checked against
+        // pragma table_info on every open.
+        self.conn
+            .call(|conn| -> std::result::Result<(), rusqlite::Error> {
+                let has_column = conn
+                    .prepare(
+                        "SELECT 1 FROM pragma_table_info('job_runs') \
+                         WHERE name = 'submission_id'",
+                    )?
+                    .exists([])?;
+                if !has_column {
+                    // The check is not atomic with the ALTER, so a second
+                    // process opening the same file can win the race. Its
+                    // `duplicate column name` is exactly the post-condition
+                    // wanted here — treat it as success rather than failing
+                    // startup over a benign race.
+                    match conn.execute("ALTER TABLE job_runs ADD COLUMN submission_id TEXT", []) {
+                        Ok(_) => {}
+                        Err(e) if is_duplicate_column(&e) => {}
+                        Err(e) => return Err(e),
+                    }
+                }
+                conn.execute_batch(SUBMISSION_ID_INDEX_SQL)?;
+                Ok(())
+            })
+            .await
+            .map_err(|e| {
+                anyhow::anyhow!("Failed to migrate jobs.db schema (submission_id): {e}")
+            })?;
         Ok(())
     }
 }
@@ -211,6 +282,7 @@ fn row_to_job_run(row: &Row<'_>) -> rusqlite::Result<JobRun> {
         rows_written: row.get::<_, Option<i64>>("rows_written")?.map(|v| v as u64),
         snapshot_id: row.get("snapshot_id")?,
         error: row.get("error")?,
+        submission_id: row.get("submission_id")?,
     })
 }
 
@@ -223,8 +295,8 @@ impl JobStore for SqliteJobStore {
                 conn.execute(
                     "INSERT INTO job_runs
                          (id, job_name, parameters, status, created_at, started_at,
-                          finished_at, rows_written, snapshot_id, error)
-                     VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, ?9, ?10)",
+                          finished_at, rows_written, snapshot_id, error, submission_id)
+                     VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, ?9, ?10, ?11)",
                     rusqlite::params![
                         run.id,
                         run.job_name,
@@ -236,6 +308,7 @@ impl JobStore for SqliteJobStore {
                         run.rows_written.map(|v| v as i64),
                         run.snapshot_id,
                         run.error,
+                        run.submission_id,
                     ],
                 )?;
                 Ok(())
@@ -253,7 +326,7 @@ impl JobStore for SqliteJobStore {
                 move |conn| -> std::result::Result<Option<JobRun>, rusqlite::Error> {
                     let mut stmt = conn.prepare(
                         "SELECT id, job_name, parameters, status, created_at, started_at,
-                            finished_at, rows_written, snapshot_id, error
+                            finished_at, rows_written, snapshot_id, error, submission_id
                      FROM job_runs
                      WHERE id = ?1",
                     )?;
@@ -269,6 +342,32 @@ impl JobStore for SqliteJobStore {
         Ok(row)
     }
 
+    async fn get_run_by_submission_id(&self, submission_id: &str) -> Result<Option<JobRun>> {
+        let submission_id = submission_id.to_string();
+        let row = self
+            .conn
+            .call(
+                move |conn| -> std::result::Result<Option<JobRun>, rusqlite::Error> {
+                    let mut stmt = conn.prepare(
+                        "SELECT id, job_name, parameters, status, created_at, started_at,
+                            finished_at, rows_written, snapshot_id, error, submission_id
+                     FROM job_runs
+                     WHERE submission_id = ?1
+                     ORDER BY created_at DESC
+                     LIMIT 1",
+                    )?;
+                    let mut rows = stmt.query(rusqlite::params![submission_id])?;
+                    match rows.next()? {
+                        Some(row) => Ok(Some(row_to_job_run(row)?)),
+                        None => Ok(None),
+                    }
+                },
+            )
+            .await
+            .map_err(|e| anyhow::anyhow!("get_run_by_submission_id failed: {e}"))?;
+        Ok(row)
+    }
+
     async fn list_runs(&self, job_name: Option<&str>, limit: usize) -> Result<Vec<JobRun>> {
         let job_name = job_name.map(|s| s.to_string());
         let limit = limit as i64;
@@ -279,7 +378,7 @@ impl JobStore for SqliteJobStore {
                     let (sql, params): (&str, Vec<rusqlite::types::Value>) = match &job_name {
                         Some(name) => (
                             "SELECT id, job_name, parameters, status, created_at, started_at,
-                                    finished_at, rows_written, snapshot_id, error
+                                    finished_at, rows_written, snapshot_id, error, submission_id
                              FROM job_runs
                              WHERE job_name = ?1
                              ORDER BY created_at DESC
@@ -291,7 +390,7 @@ impl JobStore for SqliteJobStore {
                         ),
                         None => (
                             "SELECT id, job_name, parameters, status, created_at, started_at,
-                                    finished_at, rows_written, snapshot_id, error
+                                    finished_at, rows_written, snapshot_id, error, submission_id
                              FROM job_runs
                              ORDER BY created_at DESC
                              LIMIT ?1",
@@ -379,6 +478,10 @@ mod tests {
     use super::*;
 
     fn sample_run(id: &str, job_name: &str) -> JobRun {
+        sample_run_with_submission(id, job_name, None)
+    }
+
+    fn sample_run_with_submission(id: &str, job_name: &str, submission_id: Option<&str>) -> JobRun {
         JobRun {
             id: id.to_string(),
             job_name: job_name.to_string(),
@@ -390,6 +493,7 @@ mod tests {
             rows_written: None,
             snapshot_id: None,
             error: None,
+            submission_id: submission_id.map(str::to_string),
         }
     }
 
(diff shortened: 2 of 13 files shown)


crates/server/src/jobs_handlers.rs
```diff
@@ -273,7 +274,14 @@ pub async fn submit_job_run(
         None => None,
     };
 
-    match executor.submit(&name, req.parameters).await {
+    // The audit row id doubles as the run's submission token, so the two
+    // ledgers point at each other. This half is durable at run-creation time;
+    // the `run_id` stamp below is best-effort, and if it is lost the
+    // correlation is still recoverable through `get_run_by_submission_id`.
```
abbccdda, line 279: **Nothing ever repairs the ledger, so the artifact an auditor reads is unchanged.**

This is the finding I'd want addressed before merge. The change makes the correlation *recoverable*, but no code path in the tree ever recovers it. After the failure this PR exists for, `query_audit` permanently says `status = unknown, job_run_id = NULL` — and `record_job_outcome` guards on `WHERE status = 'started'`, so once `reconcile_orphaned` has run the forward pointer can never be stamped again, not even by a later well-behaved write. The only repair is an operator with `sqlite3` and both files open.

That leaves the PR's own diagnosis in place: *"the surviving row was the least informative one — `unknown` means 'may have run after a crash' for a query row but 'definitely submitted, linkage lost' for a job row."* A reader of the ledger still can't tell those apart; they just now have a way to find out by hand.

The repair is already in reach at startup. `setup_app_state` (`crates/server/src/server.rs:229` and `:276`) opens `SqliteJobStore` *before* `QueryAuditStore` and holds both, so after the audit `reconcile_orphaned` you can pass over job rows left `unknown` with `job_run_id IS NULL` and back-fill each from `job_runs.submission_id`. Needs one new ledger method (`record_job_outcome`'s status guard rules out reuse) and turns "linkage lost, ask an operator" into "linkage restored on the next boot" — which is what "the correlation is not best-effort" would actually mean end to end.

If that's deliberately a follow-up, it should be a stated non-goal here rather than absent: the current docs and spec both read as if the correlation is now whole, when what shipped is a durable pointer plus a manual procedure.

crates/skardi/src/jobs/store.rs
```diff
@@ -110,6 +139,16 @@ pub struct JobRun {
...
+    /// Look a run up by the opaque token its submitter stored on it.
+    ///
+    /// This is the direction that makes the correlation *reconstructable*:
+    /// the submitter already knows its own token, so it can recover the run
+    /// even when its own forward pointer was never written.
+    ///
+    /// Returns the most recently created match. Duplicate tokens are a
+    /// caller error rather than something this ledger enforces — nothing
+    /// here can know whether a given token is meant to be unique.
+    async fn get_run_by_submission_id(&self, submission_id: &str) -> Result<Option<JobRun>>;
```
abbccdda, line 273: **No production caller — the lookup this change is built around is only reachable from a test.**

`get_run_by_submission_id` is called exactly once in the tree, from `crates/server/tests/jobs_audit_http.rs:535`. The spec calls it "the lookup that makes the token useful", but nothing in `crates/server` or `crates/cli` can invoke it, so the recovery path it implements has no entry point. Combined with the exposure story (see my note on `jobs_handlers.rs:143`), the shipped operator procedure is still "open the SQLite file" — the thing the spec's Exposure section says isn't much of a recovery path.

Smallest change that closes it: `?submission_id=` on `GET /jobs/runs` (or `GET /jobs/runs/by-submission/:token`), which is a handful of lines on top of the method that already exists. Without a caller, this is a public trait method that CI can only prove compiles.

crates/server/src/jobs_handlers.rs
```diff
@@ -140,6 +140,7 @@ fn job_run_to_json(run: &JobRun) -> Value {
         "rows_written": run.rows_written,
         "snapshot_id": run.snapshot_id,
         "error": run.error,
+        "submission_id": run.submission_id,
```
abbccdda: **The exposure choice doesn't reach the operator, and it does reach every tenant.**

Two problems with surfacing the token only through the run payloads:

1. **It can't be searched.** `list_job_runs` clamps `limit` to 500 with no offset and no `submission_id` filter (`jobs_handlers.rs:368`). An operator holding an audit-row id has to fetch up to 500 recent runs and grep. On a server busy enough for the concurrent-submission ambiguity that motivated `job_run_id` in the first place, the run they need is exactly the one that has fallen off the window — so the API path fails in precisely the conditions the bridge exists for, and they are back to the ledger file.

2. **It's unscoped.** `GET /jobs/runs` returns every run to any authenticated session (#221 added authentication, not authorization), so each caller now also reads the internal `query_audit` row id of every *other* caller's submission. No endpoint resolves an audit id today, so it isn't directly exploitable — but the ledger is a deliberately access-restricted artifact (`query-audit` chmods its file 0600), and this publishes one of its primary keys into a per-tenant response body. A filtered lookup on the way in would serve the operator better than a broadcast on the way out.

crates/skardi/src/jobs/store.rs
```diff
@@ -187,6 +226,38 @@ impl SqliteJobStore {
...
+                         WHERE name = 'submission_id'",
+                    )?
+                    .exists([])?;
+                if !has_column {
+                    // The check is not atomic with the ALTER, so a second
+                    // process opening the same file can win the race. Its
+                    // `duplicate column name` is exactly the post-condition
+                    // wanted here — treat it as success rather than failing
+                    // startup over a benign race.
+                    match conn.execute("ALTER TABLE job_runs ADD COLUMN submission_id TEXT", []) {
```
abbccdda: **The concurrent-startup race this tolerates is not the one that actually happens.**

The comment claims a process losing the check-then-ACT race "gets exactly the post-condition it wanted instead of a startup failure". That holds only for the narrow interleaving where the other process's `ALTER` has already *committed* by the time this one runs. In the other interleaving — the other process is mid-write when this `ALTER` fires — SQLite returns `SQLITE_BUSY` / `database is locked`, `is_duplicate_column` correctly declines to match it, and `ensure_schema` fails the open, taking startup down over the same benign race the comment says is handled.

`jobs.db` sets neither `busy_timeout` nor WAL, so the default busy handler is "fail immediately". Compare the precedent this PR says it reuses: `crates/query-audit/src/lib.rs:380-401` sets `journal_mode = WAL` and `conn.busy_timeout(AUDIT_WRITE_TIMEOUT)` *before* `ensure_schema_additions`, which is what makes its duplicate-column tolerance a complete story rather than half of one. Adding the same two pragmas here would cost two lines and also fix the same exposure on every other `job_runs` write, which today has neither a busy timeout nor a bounded-write wrapper.

crates/skardi/src/jobs/store.rs
```diff
@@ -30,13 +30,32 @@ const INIT_SCHEMA_SQL: &str = "CREATE TABLE IF NOT EXISTS job_runs (
...
     ON job_runs (status);";
 
+/// Index over [`JobRun::submission_id`], applied after the migration below has
+/// guaranteed the column exists — so it cannot live in [`INIT_SCHEMA_SQL`],
+/// which also runs against pre-column ledgers.
+///
+/// Partial, because the column is NULL for every run not submitted through an
+/// audited server, and because the only query against it is an equality
+/// lookup for one submission.
+const SUBMISSION_ID_INDEX_SQL: &str = "CREATE INDEX IF NOT EXISTS idx_job_runs_submission_id
```
abbccdda: **A duplicate token silently shadows a run; a partial UNIQUE index would make it loud.**

Non-goal #2 says uniqueness can't be known here, so duplicates are a caller error resolved by `ORDER BY created_at DESC LIMIT 1`. That's a defensible read of the layering, but note what it costs: a duplicate makes `get_run_by_submission_id` return a *confidently wrong* run — no error, no signal, the same shape as a correct answer. For a column whose entire purpose is disambiguating which run a submission produced, silently picking one is the failure mode the column was introduced to remove.

`CREATE UNIQUE INDEX ... WHERE submission_id IS NOT NULL` gives the same lookup performance, keeps NULLs unconstrained (so unattributed runs are unaffected), and turns a caller error into a `create_run` failure at the moment it is made rather than a wrong answer during an incident. The one production writer already guarantees uniqueness (a fresh `new_id()` per submission), so this constrains nothing that exists today — it only prevents a future call site from reusing a token undetected.

If uniqueness really must stay unenforced, the ledger can at least stop hiding it: count the matches and log a warning when >1. Related: `submit` takes an unbounded `&str`, so a caller-supplied token has no length cap (contrast the ≤200-char validation on `X-Skardi-Session-Id`).

crates/skardi/src/jobs/store.rs
```diff
@@ -102,6 +121,16 @@ pub struct JobRun {
...
+    /// Opaque correlation token supplied by whoever submitted the run, stored
+    /// verbatim and never interpreted here.
+    ///
+    /// The jobs subsystem has no notion of the query-audit ledger; this is the
+    /// seam that lets a caller which *does* — the server — make the
+    /// correlation reconstructable from either side. It writes its audit row
+    /// id here in the same INSERT that creates the run, so the pointer is
+    /// durable at run-creation time rather than stamped afterwards. See the
+    /// ledger section of `docs/server.md`.
+    pub submission_id: Option<String>,
```
abbccdda, line 255: **The two files are now one audit record, but only one of them is protected.**

`QueryAuditStore::open` goes out of its way to keep the ledger private: `create_private_file` before open so it is never briefly world-readable, then `restrict_permissions` (0600) on the db and its `-wal`/`-shm` sidecars (`crates/query-audit/src/lib.rs:371-417`). `SqliteJobStore::open` uses a plain `Connection::open`, so `jobs.db` lands at whatever the umask allows — typically 0644.

That asymmetry was already awkward, since `job_runs.parameters` holds exactly the raw values `query_audit` refuses to store (the open threat-model question this PR re-flags). This change makes it load-bearing: the reverse pointer means a world-readable `jobs.db` now also reveals which protected audit row each run belongs to, so the file with the weaker posture links into the one with the stronger. If the two halves are "one bridge", they should be one permission decision — reuse the same private-create/chmod path here, or state explicitly in `docs/server.md` that `jobs.db` is not treated as a protected artifact and why.

docs/server.md
```diff
@@ -338,15 +338,20 @@ Two properties of this seam an operator should know before relying on it:
...
-  keeps `job_run_id = NULL` (and reconciles to `unknown` on restart). `job_runs`
-  carries no `session_id` and no audit-row id, so there is no reverse pointer
-  to rebuild the correlation from — it is lost, not merely delayed. Exact
-  correlation is the column's purpose, so treat it as *usually* exact rather
-  than guaranteed; making it reconstructable needs a durable half in the
-  jobs ledger, which is out of scope here.
+- **The `job_run_id` bridge is best-effort, but the correlation is not.**
+  `query_audit.job_run_id` is stamped *after* `executor.submit` returns, so if
+  that write fails, times out, or the process dies in the window, the audit
+  row keeps `job_run_id = NULL` and reconciles to `unknown`. The correlation
```
abbccdda, line 344: **"Durable" needs the retention caveat, and the useful inference from a miss should be documented.**

`query_audit` has a retention knob (`--query-audit-retention-days`, pruning at startup and hourly — `server.rs:316`); `job_runs` has no pruning at all. So under retention the halves expire on different clocks: an old run keeps a `submission_id` that resolves to a row that has been deleted, indistinguishable from a token that was never valid. The pointer stays durable; its target does not. Worth one sentence here, since this is the section an operator reads before relying on the bridge.

The flip side is worth stating too, because it's the part that makes the bridge genuinely stronger than the PR claims: since `job_runs` is never pruned, *no match for a token* is positive evidence that `submit` never created a run — which cleanly separates "definitely submitted, linkage lost" from "never ran" for an `unknown` job row, the exact ambiguity the Why section opens with. That's the operator procedure, and it currently exists only as an implication of two facts documented in different files. It also stops being sound the moment anyone adds pruning to `job_runs`, so writing it down here is what protects it.

abbccdda (commented):
Review of the durable reverse pointer. The core mechanism is right and the load-bearing test is well chosen; my findings are about how far the reliability claim actually reaches end to end (nothing repairs the ledger, no caller for the lookup, and the migration's concurrency story is half of #219's).

crates/skardi/src/jobs/store.rs
```diff
@@ -30,13 +30,32 @@ const INIT_SCHEMA_SQL: &str = "CREATE TABLE IF NOT EXISTS job_runs (
     finished_at   TEXT,
     rows_written  INTEGER,
     snapshot_id   TEXT,
-    error         TEXT
+    error         TEXT,
+    submission_id TEXT
```
abbccdda, line 36: **Two other docs describe this schema and neither was updated.**

- `docs/agent_data_plane.md:127` embeds the `job_runs` DDL verbatim and says "The schema is fixed by `crates/skardi/src/jobs/store.rs` — the `INIT_SCHEMA_SQL` constant". It now omits `submission_id`, so a doc that names its source of truth has silently diverged from it.
- `docs/jobs.md:352` has the run-row field table, introduced as "Row fields, matching the CLI `status` response" — that response now carries `submission_id`, so the table is incomplete on a claim it makes explicitly.
- `docs/jobs.md:212` still describes "the `job_run_id` bridge to the run" with no mention of the reverse half, while `docs/server.md` was updated to say the bridge has two halves. `docs/jobs.md` is the jobs-facing doc, so an operator following the jobs docs alone gets the pre-PR story.

crates/skardi/src/jobs/store.rs
```diff
@@ -30,13 +30,32 @@ const INIT_SCHEMA_SQL: &str = "CREATE TABLE IF NOT EXISTS job_runs (
...
+/// audited server, and because the only query against it is an equality
+/// lookup for one submission.
+const SUBMISSION_ID_INDEX_SQL: &str = "CREATE INDEX IF NOT EXISTS idx_job_runs_submission_id
+    ON job_runs (submission_id) WHERE submission_id IS NOT NULL;";
+
+/// True for the `duplicate column name` error SQLite raises when an
+/// `ALTER TABLE ... ADD COLUMN` loses a race with another connection that
+/// already added it. Matched on message text because `rusqlite` reports it as
+/// a generic `SqliteFailure` with no distinguishing code.
+fn is_duplicate_column(e: &rusqlite::Error) -> bool {
```
abbccdda, line 94: **Third copy of this migration pattern; #219 already generalized it.**

`is_duplicate_column` here is byte-for-byte the helper at `crates/query-audit/src/lib.rs:226`, doc comment included. The duplication itself is defensible — `crates/skardi` must not depend on `skardi-query-audit`, and a two-line predicate isn't worth a shared crate — but the *shape* around it needn't be: `query-audit` factored this into `ensure_added_columns(conn, &[cols])`, which handles N columns idempotently, while this hand-rolls the single-column case inline in `ensure_schema`. The next column on `job_runs` copy-pastes this block a fourth time.

Worth either lifting the loop into a local `ensure_added_columns(conn, "job_runs", &["submission_id"])` now, or a comment pointing at the query-audit original so the two stay in sync when the predicate needs widening (a future SQLite wording change breaks both, silently, in different crates).

crates/skardi/src/jobs/store.rs
```diff
@@ -517,4 +621,137 @@ mod tests {
...
+        s.create_run(&sample_run("r2", "ingest")).await.unwrap();
+        assert!(
+            s.get_run("r1")
+                .await
+                .unwrap()
+                .unwrap()
+                .submission_id
+                .is_none()
+        );
+        assert!(s.list_runs(None, 10).await.unwrap().len() == 2);
```
abbccdda: Nit: `assert!(x.len() == 2)` prints only `assertion failed` — `assert_eq!(s.list_runs(None, 10).await.unwrap().len(), 2)` reports the actual count, which is the fact you'd want on failure. The surrounding tests use `assert_eq!` for exactly this.

crates/server/src/jobs_handlers.rs
BtXin: Fixed in 9dfd447 — this was the right call to make blocking. Verified the mechanism first: `record_job_outcome` guards `WHERE id = ?1 AND status = ?6` with `?6 = started`, so once `reconcile_orphaned` has run the row is permanently unstampable. The docs and spec did read as if the correlation were whole when what shipped was a pointer plus a manual procedure.

Took your suggested shape. `setup_app_state` now runs a repair pass after **both** stores have reconciled their own orphans — the audit reconcile is what creates the candidates, so repairing earlier would miss exactly the rows the crash just made. Two new ledger methods rather than reusing `record_job_outcome`, as you noted its guard rules that out:

- `job_rows_missing_run_id()` — `statement_kind = job AND status = unknown AND job_run_id IS NULL`, oldest first.
- `backfill_job_run_id(id, run_id)` — guarded on `unknown` + NULL, the mirror image of the guard that excluded reuse. So it can neither overwrite a correctly written pointer nor touch a row still `started`.

`status` stays `unknown`: the outcome genuinely was never observed, only the linkage is recovered. A candidate with no matching run is left alone — per your `docs/server.md` note, that miss is positive evidence `submit` never created a run, which is a fact worth keeping rather than papering over. Failure is logged, not fatal; a ledger that cannot be repaired should not stop a server from serving.

Tests: `jobs_audit_http.rs` drives the same `repair_lost_job_correlations` the server calls (re-links the row, idempotent on a second boot, leaves a never-ran submission alone, declines a row in flight), plus one ledger unit test that pins the premise the pass rests on — that `record_job_outcome` cannot stamp a reconciled row — so if that guard ever changes, the pass is flagged as dead code rather than quietly kept.

Those pin what the pass does but not that anything calls it, which is the failure mode your finding describes, so there is also a new `jobs_bridge_startup.rs` that boots `setup_app_state` twice over the same two on-disk ledgers. Mutation-checked: disabling the call site fails it.

Startup-only rather than periodic is now a stated non-goal; the guards make a scheduled variant purely additive.

crates/skardi/src/jobs/store.rs
BtXin: Fixed. It now has two production callers, both from this review: the startup repair pass (your `jobs_handlers.rs:280` finding) resolves every candidate through it, and `GET /jobs/runs?submission_id=` exposes it over HTTP.

Took the `?submission_id=` form on the existing route rather than a new `by-submission/:token` path — it reuses the handler's auth gate and error envelope, and one token names at most one run so `job`/`limit` simply do not apply (pinned by a test). Confirmed your count before changing anything: the only caller was `jobs_audit_http.rs:535`.

crates/server/src/jobs_handlers.rs
BtXin: Both halves fixed, by the same change — a filter on the way in, and the field dropped from the way out. Your framing in the last line is what the implementation follows.

1. **Searchability.** `GET /jobs/runs?submission_id=<token>` is an exact lookup through `get_run_by_submission_id`, not a filter over the recent window, so the 500-clamp is irrelevant to it. There is a test that buries the wanted run under later ones, confirms the list route only returns a bounded window, and then resolves it by token anyway — the incident case you describe.

2. **Scoping.** `job_run_to_json` no longer emits `submission_id`, from either the list or the detail route. Verified the exposure was as you describe: `GET /jobs/runs` returns every run to any authenticated session, since #221 added authentication and explicitly not authorization, and `query-audit` does chmod its file 0600. Publishing one of its primary keys into a per-tenant body for a field nothing needed to read was not a trade worth keeping once the lookup exists.

A test asserts the absence on both routes, so re-adding the field fails rather than passing silently. `docs/server.md` and the spec now carry the rationale, so it does not read as an oversight to a future reader.

crates/skardi/src/jobs/store.rs
BtXin: Confirmed and fixed. You are right that the tolerance only covered the already-committed interleaving; in the other one `is_duplicate_column` correctly declines `database is locked` and `ensure_schema` fails the open.

`SqliteJobStore::open` now sets `journal_mode = WAL` and `busy_timeout(JOBS_BUSY_TIMEOUT)` (5s) before `ensure_schema`, matching the `query-audit` precedent you cite. As you note, this also covers every other `job_runs` write, which had neither. Applied in `open` only, not `open_in_memory` — WAL is meaningless on a memory database and there is no second connection to contend with.

Pinned by a test that reads both pragmas back rather than trusting the calls, and the comment records why the timeout is what makes the duplicate-column tolerance a complete story instead of half of one.

crates/skardi/src/jobs/store.rs
BtXin: Took the unique index. The argument that decided it is yours: a duplicate made `get_run_by_submission_id` return a confidently wrong run, with no error and the same shape as a right answer, for the one column whose purpose is disambiguation. Non-goal #2 traded a caller error for a wrong answer during an incident, which is the wrong direction. It constrains nothing today, exactly as you say — one writer, a fresh `new_id()` per submission.

One wrinkle worth flagging, since it would have made the change a no-op in practice: `CREATE UNIQUE INDEX IF NOT EXISTS` matches on **name**, so against a `jobs.db` written by an earlier build of this branch — which carries the non-unique index under the same name — it silently no-ops and leaves a ledger that looks constrained and is not. So it is `DROP INDEX IF EXISTS idx_job_runs_submission_id` followed by a create under `idx_job_runs_submission_id_unique`, which makes the outcome identical whichever build wrote the file. Three tests: a duplicate fails `create_run`, five NULL rows still insert freely under the partial clause, and a ledger carrying the old index name converges (opened twice, to pin re-runnability).

**Not taken: the length cap.** Pushing back on this one sub-point. `submit` stays an unbounded `&str`, matching the column's opaque-TEXT contract. The only production caller supplies a server-minted audit id, and the failure a cap prevents — a large string in a TEXT column — is not in the class the unique index addresses, which is a silently *wrong* answer. The `X-Skardi-Session-Id` cap earns its keep because that value arrives from outside the process; this one does not. Adding a `JobSubmitError` variant for a token the server generates itself would add a failure mode no caller can currently reach. Recorded as an explicit non-goal rather than left unsaid — happy to add it if you read the layering differently, since `submit` is public API and a future caller need not be the server.

crates/skardi/src/jobs/store.rs
BtXin: Fixed — the asymmetry was real and this change did make it load-bearing. `SqliteJobStore::open` now does what `QueryAuditStore::open` does: `create_private_file` before open so the file is never briefly world-readable, then `restrict_permissions` (0600) on the db and its `-wal`/`-shm` sidecars, which the new WAL setting makes relevant.

Both helpers are duplicated into `crates/skardi/src/jobs/store.rs` rather than shared, for the same no-dependency reason as `is_duplicate_column`, with the doc comment saying so. Went with the code change over the docs escape hatch: `job_runs.parameters` holds exactly the raw values `query_audit` refuses to store, so the file with the weaker posture was already the one holding the more sensitive data — the reverse pointer just made that impossible to argue around. `docs/server.md` states the posture too, since it changes what an operator can expect of the file.

Pinned by a test asserting the mode on the db and both sidecars after a write. Off Unix the chmod is a no-op, as with the audit ledger.

docs/server.md
BtXin: Both added. Confirmed the asymmetry — `--query-audit-retention-days` prunes at startup and hourly (`server.rs:316`), `job_runs` has no pruning anywhere — so `docs/server.md` now says the pointer stays durable while its target does not, and that an expired target is indistinguishable from a token that was never valid.

The flip side is written down as its own paragraph, and you are right that it is the stronger half: because `job_runs` is never pruned, no match for a token is positive evidence that `submit` never created a run, which is what separates "definitely submitted, linkage lost" from "never ran" for an `unknown` job row — the ambiguity the Why section opens with. It was previously only an implication of two facts in different files. The note that it stops being sound the moment anything prunes `job_runs` is included, since that is what protects it.

It also became load-bearing in code, which is worth recording: the repair pass deliberately leaves a candidate with no matching run alone, and that inference is the reason it is correct to do so rather than a gap. The behaviour has its own test.

crates/skardi/src/jobs/store.rs
BtXin: All three fixed.

- `docs/agent_data_plane.md` — the DDL block now carries `submission_id`, with the same inline comment style as its neighbours. A doc that names `INIT_SCHEMA_SQL` as its source of truth diverging from it was the worst of the three.
- `docs/jobs.md:352` — the row-field table stays honest in a way that needed a decision, because of the exposure change in your `jobs_handlers.rs:143` finding: `submission_id` is now deliberately **not** in the CLI `status` response, so adding a table row would have made the table wrong in the other direction. Instead there is a note under it saying the row carries one field the response does not, why it is withheld, and how to reach it (`?submission_id=`).
- `docs/jobs.md:212` — now describes both halves and the startup repair, so following the jobs docs alone no longer gives the pre-PR story.

Also documented the new query parameter in the endpoints table there, since that is where an operator would look for it.

crates/skardi/src/jobs/store.rs
BtXin: Fixed — `assert_eq!(s.list_runs(None, 10).await.unwrap().len(), 2)`. It was the only `assert!(… == …)` in the file.

crates/skardi/src/jobs/store.rs
BtXin: Took the loop. `ensure_added_columns(conn, &ADDED_COLUMNS)` in `store.rs` now mirrors the `query-audit` shape, so the next column on `job_runs` is a list entry rather than a fourth copy of the block — and the `ALTER`-race comment lives in one place instead of being restated inline.

Kept the duplicated `is_duplicate_column` for the reason you give, and took the comment suggestion as well as the refactor rather than instead of it: the predicate now names its `crates/query-audit/src/lib.rs` original and says explicitly that a future SQLite rewording of that message breaks both silently in different crates, so change them together. That is the failure mode the cross-reference is actually for; the shared crate would be the heavier fix for a two-line string match.

The same commit adds the two pragmas from your `store.rs:248` finding, which is what makes this migration path match the precedent in behaviour and not just in shape.

abbccdda (approved):

PR #223, opened 2026-08-19
feat(sources): OneDrive source pack (milestone 5.8) — drive_items + drive_item_search

gracexmatin:
Milestone **5.8** — the `one_drive` source pack: `drive_items` and `drive_item_search` over Open Connector's `one_drive` service.

**Live-verified.** The contract is reconciled against a live gateway (v1.3.4, open-connector at `2410fbe`), and the real-row pass ran 2026-08-21 against a personal (MSA) drive: raw probes plus end-to-end skardi-server scans of both tables through the live fingerprint gate. The row fixtures are redacted live captures, not synthetic.

## Why its own pack

There is no `microsoft365` service upstream: the gateway splits Graph into `outlook` (mail), `one_drive` (files) and `excel`, each with its **own OAuth connection** — and a Skardi binding carries exactly one `connection_alias`, so a cross-service pack would silently span two grants and fail half its tables at scan time when only one is connected. M365 therefore ships one pack per service. `outlook` is milestone 5.7 (#212, merged); `excel` is deferred whole at the admission gate, because its list actions emit a `nextLink` continuation but accept no `nextLink` input, so their pagination cannot be completed.

One engine addition, `exclusive_resources` (below); otherwise `packs/` files plus two roster lines.

## What shipped

**`packs/one_drive.yaml`** — 2 tables, cursor pagination over Graph's `@odata.nextLink` (`page_size: 999`, the declared ceiling; 1000 and 0 both 400 live, and 999 was confirmed on the wire). `drive_items` → `one_drive.list_folder_children`, all three resources optional (`driveId`, `folderItemId`, `folderPath`); with none bound the table is the drive root's children. `drive_item_search` → `one_drive.search_items`, `query` required, `driveId` optional. No `fixed_inputs`, no `filters`, no `error_path` on either table.

**`packs/one_drive.rs`** — module doc recording every decision with its rationale, the `OnceLock` pack accessor, and **30 tests** (17 contract/conversion, 13 end-to-end through `MockGateway`).

**Fixtures** — the declared output schemas (the fingerprint input) and the input schemas, both re-fetched from live discovery; three row fixtures, two of them redacted live captures.

**Docs** — `docs/open-connector-one-drive.md`, the 5.8 milestone entry, the design record, the README capability row, and the status note in `docs/open-connector.md`.

## The live pass changed the pack: 16 columns, then 14

The declared contract says the two actions return the same `driveItem` collection schema — byte-identical captures, one shared fingerprint. The wire disagrees. Real search hits are a reduced Substrate projection that **never carries `eTag`/`cTag`** (zero occurrences across 1800+ live hits, files and folders alike), so `drive_item_search` maps **fourteen** columns to `drive_items`' sixteen. Mapping the pair there would have shipped two always-NULL columns — precisely the defect a live pass exists to catch, and invisible to every contract-level gate. A test pins the search column list as `drive_items`' minus exactly those two, since a retype, rename, re-path or reorder on one table alone would slip past both the width pins and the coverage gate.

Other live findings, all recorded in the pack doc:

- **Search continuations fail server-side on personal drives.** Following a real search cursor answers Graph's own "Error Calling Substrate Search" — deterministic across retries and across the `/me/drive` and `/drives/{id}` forms, cursor forwarded byte-identically. Upstream, not a gateway or pack defect. The scan **fails loudly** through the failure envelope rather than truncating; queries whose hits fit one page terminate on a clean null.
- **The one truncation risk is closed.** A null cursor is the only end-of-collection signal (Graph publishes no has-more flag), so a gateway that answered an allowlist-rejected link with null instead of an error would make a partial scan look complete. It errors.
- **`folderItemId` silently wins** when a binding sets it alongside `folderPath` (structural: the executor checks id → path → root), so the path becomes dead configuration.
- **Search rows spell `parent_drive_id` lowercase without the leading zero** while children rows carry the `0FAB…` form of the same drive, so a naive join on that column misses.
- `description` was never witnessed non-NULL across 2800+ rows. It stays mapped (declared in-schema, so drift stays loud) with the caveat recorded; an all-NULL column here is expected, not the always-NULL bug.

## Design decisions

**No `select` pin, and the reason is coverage.** Every mapped column of both tables resolves *inside* the declared item schema, so the coverage-gap pin is **empty** — the sharpest contrast with `outlook.messages`, whose thirteen undeclared columns are exactly why it needs one. Two distinct mechanisms deliver loudness and the docs keep them apart: an **upstream** rename or retype changes the schema hash and fails registration, while a **mapping typo inside the pack** leaves the hash intact and is caught by the coverage test in CI.

**`query` is required, and not where it looks.** The input schema's `required` array is empty and `query` is merely `minLength: 1`, which splits three ways live: an empty string 400s as `invalid_input` at the schema layer, while a **missing** and a **whitespace-only** query both pass validation and die in the executor's own trim check. Declaring it a required resource closes the missing case — the check is presence plus non-null, so `query: " "` still registers cleanly and fails at scan time. A registration test pins that a binding without one is refused before any HTTP.

**Zero filter pushdown, structurally.** Neither action exposes a filter input at all. The exact input key set of every request is asserted per table, absence included, which is what discharges the claim. `LIMIT` bounds requests, not bytes: `top` is sent verbatim, so `LIMIT 10` costs one request but still transfers a full page.

**Item type is facet presence, not a field**, so `file_mime_type` and `folder_child_count` double as the discriminator — with `childCount: 0` kept distinct from NULL, or an empty folder would be indistinguishable from a non-folder. A live `remoteItem` row (the Personal Vault stub) carries neither facet, so that gap is witnessed, not hypothetical.

**Input contracts are pinned too.** The fingerprint gate is output-only, so a renamed input key would register cleanly and then 400 every scan. Both input halves are committed with a test locking every key the pack can send. Gating inputs at registration remains engine work.

## Verification

```
cargo test -p skardi --lib sources::providers::open_connector::packs::one_drive
```

**30 tests** pack-scoped, plus four loader rejection arms and two loader round-trip tests for `exclusive_resources`; suite-wide counts come from CI, per this repo's convention of not running tests locally. `cargo fmt` clean.

Coverage: per-table wire-declaration scans asserting the exact input key set of every request (including absence); null-cursor termination, the only spelling either action can emit; pagination-loop and failure-envelope arms; LIMIT early-stop asserting `top` is not narrowed; resource forwarding and withholding; a search binding with no `query` refused before any HTTP; a contract-legal item with no `id` failing the page rather than yielding a null key; wire nulls and empty strings converting without failing the page; the search/children column relation; no-pushdown row identity; UDTF parity; drift-refusal at registration through the public entry point; and a key-scoped default-deny fixture audit with thirteen self-trip probes.

## What review found and fixed

Worth calling out, because these are the class of defect that passes CI-by-mocks:

- A **mock that contradicted the committed contract** — the search table's terminal page omitted `nextLink`, but both output contracts are `required: ["items", "nextLink"]` with `additionalProperties: false`, so that is a shape the gateway cannot produce.
- **Gates with no failing arm**: `query`'s requirement, `id`'s non-nullability, and the fixture audit — whose "carries no real drive data" half rested on an assertion that could never fail.
- **The audit's own doctrine, violated by three of its arms.** It holds that a shape check enforces nothing, yet `path` checked only a prefix (so a real folder name passed where the same string under `name` would fail its allowlist) and the two URL arms checked host plus the synthetic cid, leaving the per-row `resid`/`UniqueId` GUIDs and the human-readable path segments unexamined. Every probe tested a value wrong all the way through, so none covered the **partial** redaction a re-capture actually produces. Now one shared name allowlist spans `name`, path segments and URL segments; every URL path segment and query parameter is default-deny, an unknown parameter being a failure; and four probes cover the partial class.
- **A hand-listed fixture roster**, now enumerated from the directory — `loader.rs` records what the listed form costs, discord.yaml having shipped unlisted for a full milestone.
- **Coverage the live-capture swap dropped**: twelve explicit wire nulls and an empty-string row went to zero, leaving only NULL-via-missing-key assertions. Both shapes return as an inline synthetic page, covering the nested nulls that traverse *through* a null and `""` staying distinguishable from absent.
- **A pairwise invariant that lived only in a comment.** The yaml derives `drive_item_search`'s columns from `drive_items`' as "MINUS the two concurrency tags", and the width pins catch an added or dropped column — but a retype, a rename, a re-path to another declared key, or a reorder on ONE table slipped past every gate and quietly made that comment false. `search_columns_are_drive_items_minus_the_two_concurrency_tags` now pins the relation field by field, name/path/type/nullability and order included.
- **The partial-redaction hole the audit's own doctrine forbids, still open on `id`.** The arm accepted `FAB1234CD567890!` plus anything, but that suffix is the per-item half of a personal-drive id and varies per row — the exact shape a re-capture that scrubs the one global cid produces, and the reason the URL arms default-deny `resid`/`UniqueId`. Both call sites now share one predicate pinning the two suffix families whole, and two new probes cover a real suffix under a synthetic cid (nothing among the thirteen existing ones did).
- **An operator trap the pack could close itself.** `folderItemId` and `folderPath` are alternatives; upstream a binding carrying both is legal and the id wins, so the scan succeeds against a folder nobody named while the path becomes dead configuration — the one misconfiguration here that yields confidently wrong rows instead of an error. Documented in three places, which is the weakest available mitigation. It is now a registration failure: `SourcePackTable` gained `exclusive_resources` (additive, defaulted empty), the loader validates the declaration and registration rejects the binding by name. Deliberately stricter than the gateway — no precedence is the right answer when the operator stated two scopes.
- **Doc claims that were not true**: `LIMIT` does not narrow `top`; three of the seven "verified to 400 live" input keys are declared-but-deliberately-unsent; "a misspelled path fails at registration" holds only for upstream renames; a required resource closes the missing query, not the whitespace-only one; the design record still described the search table as sixteen columns; and the README capability row never gained the pack.

## Reviewer notes

- The OAuth consent for `one_drive` is **read-write** even though this pack is read-only: the gateway's scope union is `User.Read`, `Files.ReadWrite`, `offline_access`, because the same connection serves the service's upload and delete actions. Skardi's tables stay read-only by construction (the registrar rejects `read_write` for source packs), but the pack doc says plainly that you consent to more than Skardi uses. A narrower `Files.Read` union is an upstream ask.
- `drive_items` is **non-recursive** — one binding sees one folder level. That is what makes `drive_item_search` worth shipping.
- The live pass ran on a **personal (MSA)** drive. The search-continuation failure is scoped to that in the docs; a business/SharePoint-backed drive is unprobed.


crates/skardi/src/sources/providers/open_connector/packs/one_drive.rs
```diff
@@ -0,0 +1,2001 @@
...
+                Value::String(s) => {
+                    let allowed = match key {
+                        // One key, several id families: driveItem ids
+                        // (`FAB…!103`, `FAB…!s<hex32>`), the user id (the
+                        // cid), application ids (repeated-digit GUIDs plus
+                        // Office's well-known first-party constant), the
+                        // remoteItem's business-shaped id, and the
+                        // type-mismatch fixture's marked synthetic ids.
+                        "id" => {
+                            s.starts_with("FAB1234CD567890!")
```
yiyi-wang-skardi: **The partial-redaction class this commit closes for URLs is still open on `id`.**

`s.starts_with("FAB1234CD567890!")` accepts *any* suffix, but the suffix is the per-item half of a personal-drive id (`!103`, `!s<hex32>`) and it varies per row — precisely the shape a re-capture that scrubs the one global cid and misses the per-row identifiers produces. That is the reason `url_parts_are_redacted` now default-denies `resid`/`UniqueId`, and the same id family rides in the URL query arm at line 726 with the same loose check, so a real `FAB1234CD567890!s1a2b3c4d…` sails through both.

Every committed id is already either an ordinal or `s` + 32 repeated hex digits (I checked all ten distinct values), so tightening costs nothing:

```rust
"id" => match s.strip_prefix("FAB1234CD567890!") {
    Some(rest) => {
        rest.bytes().all(|b| b.is_ascii_digit())
            || (rest.len() == 33
                && rest.starts_with('s')
                && rest[1..].bytes().all(|b| b == rest.as_bytes()[1]))
    }
    None => s == "0FAB1234CD567890" || repeated_digit_guid(s) || …
}
```

Worth a fourteenth tripwire too — none of the current probes covers a real id suffix under a synthetic cid, which is the one leak shape this arm would wave through.

crates/skardi/src/sources/providers/open_connector/packs/one_drive.yaml
```diff
@@ -0,0 +1,229 @@
...
+    # /" but enforced only as `minLength: 1`, so a slash-less path passes
+    # validation and fails in the executor.
+    # `select`/`expand`/`orderBy` are deliberately
+    # never sent (negative-space guarded): `orderBy` would only reorder a
+    # set the scan reads in full, and `select` would narrow rows the
+    # declared coverage above already protects. Note the camelCase — the
+    # live gateway 400s lower-case `orderby`, which is the spelling the
+    # sibling `outlook` service uses.
+    resources:
+      optional: [driveId, folderItemId, folderPath]
```
yiyi-wang-skardi, line 115: **`folderItemId` silently winning over `folderPath` is a hazard the pack can close itself, not just document.**

The design record's own framing of this (open item 7) is "that binding scans a folder the operator did not name", and phase 4 confirmed exactly that. The resolution was to record it in three places — but a binding that carries both still scans the wrong folder, with a green status and no diagnostic anywhere, and prose is the weakest available mitigation for that.

Two pack-level options, either of which turns it loud:

1. Declare only `folderItemId` here. `folderPath` is pure convenience (an id is what the executor prefers anyway), and dropping it makes "one way to scope a folder" structural — the KISS reading. `mod.rs`'s `UnknownResourceKey` check then rejects a `folderPath` binding by name at registration.
2. Keep both and add mutually-exclusive resource groups to the loader's validation, alongside the existing required/optional overlap check at `loader.rs:219`. More work, but it generalises — this cannot be the last pack with alternative scoping resources.

Not blocking: the behaviour is correct-by-upstream and disclosed. But it is the one place in this pack where a plausible misconfiguration produces confidently wrong rows instead of an error, which is the opposite of the doctrine the rest of the pack holds to.

crates/skardi/src/sources/providers/open_connector/packs/one_drive.yaml
```diff
@@ -0,0 +1,229 @@
...
+    # resource (the binding pins the term, and the table is "the drive
+    # items matching this binding's query" — same semantics as
+    # notion.block_children requiring a blockId) rather than a fixed
+    # input. `list_folder_children` being non-recursive is what makes
+    # this table worth shipping: it is the only way to see a whole drive.
+    resources:
+      required: [query]
+      optional: [driveId]
+    # drive_items' column set MINUS the two concurrency tags: the
+    # declared schema is shared, but real search rows are a reduced
```
yiyi-wang-skardi, line 218: **The 16→14 decision rests on one personal drive, and nothing tracks re-probing it.**

Dropping `e_tag`/`c_tag` is the right call on the evidence, and the doc is honest that a business/SharePoint-backed drive is unprobed. But the failure mode of being wrong is asymmetric: if a business drive *does* return the concurrency tags, the pack silently discards data it declared it could not get — invisible to the fingerprint gate (the tags are in the shared declared schema), invisible to the coverage gate (an unmapped declared key is not a gap), and invisible to every test here.

The caveat currently lives only in `docs/open-connector-one-drive.md:201`. The tasks spec ticks 5.8 with no residual item and the design record's Phase 4 results section closes item 4 without one, so this observation has no home outside user-facing prose. Suggest a tracked open item — "re-probe `search_items` on a business/SharePoint drive; if it carries `eTag`/`cTag`, the search table grows back to 16" — in the design record next to the finding it qualifies. That also tells whoever hits it that widening the table back is the expected outcome rather than a contract break.

docs/open-connector-one-drive.md
```diff
@@ -0,0 +1,373 @@
...
+Microsoft 365 reaches Skardi as **one source pack per Open Connector
+service**. There is no `microsoft365` service upstream: the gateway
+splits Microsoft Graph into `outlook` (mail only), `one_drive` (files)
+and `excel` (spreadsheets), each with its **own OAuth connection** — and
+a Skardi binding carries exactly one connection alias, so a
+cross-service pack would silently span two OAuth grants and fail half
+its tables at scan time when only one service is connected.
+
+This document covers the built-in **`one_drive` pack** — `drive_items`
+and `drive_item_search` over a drive. The `outlook` pack ships as its own
```
yiyi-wang-skardi, line 16: **The two Microsoft 365 guides are now disconnected, and this reverses a decision recorded in the merged M365 design record.**

`docs/superpowers/specs/2026-08-14-open-connector-m365-packs-design.md:399` decided explicitly: "A single `docs/open-connector-microsoft-365.md` holds the Microsoft 365 story for both packs — it ships covering `outlook` alone and **gains the `one_drive` tables with that milestone** — so the product-level story survives the split."

This PR instead ships a separate guide and leaves `docs/open-connector-microsoft-365.md` untouched, where line 12 still reads "The `one_drive` pack ships separately" with no pointer, under the title "Microsoft 365 Source Packs (Outlook)". So a reader who lands on the M365 page — which is where the README's *Outlook* link goes, and the obvious landing page for "Microsoft 365" — has no path to the OneDrive guide, and this page has no path back.

A separate guide is defensible (both docs are long, and the OAuth grants really are independent), so the ask is not to merge them: add the missing link in each direction, and amend that line of the design record so the reversal is recorded rather than silent. Given that this branch spent a whole commit correcting stale claims in that same family of documents, this one belongs with them.

crates/skardi/src/sources/providers/open_connector/packs/one_drive.yaml
```diff
@@ -0,0 +1,229 @@
...
+    row_path: "$.items"
+    fingerprint: 6fc6a6b319766064e084865122b06db8e5562b35193ae9dadc4d64b8f9ac207b
+    pagination:
+      strategy: cursor
+      cursor_input: nextLink
+      next_cursor_path: "$.nextLink"
+      page_size_input: top
+      # The declared ceiling (`top`: 1–999; 1000 and 0 both 400 against
+      # the live gateway). driveItem rows are metadata-only, so a full
+      # page stays orders of magnitude under the 16 MiB response cap.
```
yiyi-wang-skardi: **"orders of magnitude under the 16 MiB response cap" is about one order of magnitude, and the committed fixtures let you say the real number.**

The captured children rows average 1 397 bytes serialized (max 1 708), so a full 999-row page is ≈ **1.3 MiB** — roughly 8 % of `DEFAULT_MAX_RESPONSE_BYTES` (`client.rs:77`), i.e. ~12× headroom, not orders of magnitude. Search rows are lighter (≈ 0.76 MiB/page).

That is still comfortable, so the `page_size: 999` decision is unaffected. But 12× is a number an operator might actually care about — a drive whose rows carry more facets and hash sets than these captures is not implausibly far from the cap, whereas "orders of magnitude" reads as "never think about this". The claim repeats verbatim in `one_drive.rs:62-63` and the design record's Table-design section; a measured figure derived from the committed fixtures would be both more useful and more in keeping with the rest of this pack's write-up.

docs/open-connector-one-drive.md
```diff
@@ -0,0 +1,373 @@
...
+        runtime_token_env: OPEN_CONNECTOR_TOKEN
+        bindings:
+          - name: drive                    # schema name in SQL
+            source_pack: one_drive
+            connection_alias: my-msft-drive  # the gateway's one_drive OAuth connection
+            resource:                      # omit entirely for the drive root
+              folderItemId: "01ABCDEF..."      # scope drive_items to one folder
+            tables: [drive_items]
+
+          # drive_item_search needs its own binding: `query` is required,
```
yiyi-wang-skardi: **"drive_item_search needs its own binding" overstates the constraint.**

A single binding *can* serve both tables. `mod.rs:200` requires each supplied resource key to be declared by **at least one** bound table, and `table.rs:94` filters each table's request inputs to the keys it declares — so `resource: {folderItemId: …, query: budget}` with `tables: [drive_items, drive_item_search]` registers cleanly, and `drive_items` simply withholds `query`.

The real constraint is one search *term* per binding, which is what the following sentence and the `bind one search per term you care about` line further down already say correctly. Separate bindings are still the better example (distinct schema names read better in SQL) — it is just the word "needs" that is not true, and this doc is otherwise careful enough about exactly this distinction that the exception stands out.

crates/skardi/src/sources/providers/open_connector/packs/one_drive.rs
```diff
@@ -0,0 +1,2001 @@
...
+                "示例报告.docx",
+                "copilotUploads",
+                "documents",
+                "wrong-types.bin",
+            ];
+            // Search rows percent-encode CJK path segments where children
+            // rows send them literally, so one allowlist covers both
+            // spellings only after decoding. Invalid UTF-8 decodes to the
+            // empty string, which is on no allowlist — default-deny holds.
+            fn percent_decode(seg: &str) -> String {
```
yiyi-wang-skardi, line 681: Nit (DRY): `percent-encoding = "2.3"` and `url = { workspace = true }` are both already direct dependencies of this crate, so the hand-rolled decoder is re-implementing `percent_encoding::percent_decode_str(seg).decode_utf8().unwrap_or_default()` — same semantics including the empty-string-on-invalid-UTF-8 behaviour the comment above relies on for default-deny.

`url::Url::parse` would fold in more of `url_parts_are_redacted` as well: `path_segments()` and `query_pairs()` (already decoded) replace the `split('/').skip(3)` and `split('&')` + `split_once('=')` hand-parsing, and drop the implicit assumption that the string always begins with `scheme://host` — which holds today only because each caller prefix-checks the host first. Test-only code, so purely a simplification; worth it if this audit gets copied to the next pack, which is the pattern so far.

yiyi-wang-skardi (approved):
## Audit of the `one_drive` source pack (milestone 5.8)

Reviewed the full diff against `main` (`97859f8`), i.e. the seven OneDrive commits: `packs/one_drive.yaml`, `packs/one_drive.rs` (27 tests), the eight fixtures, the two roster lines, and the docs. Read the surrounding engine to check the claims rather than take them on trust.

**What I verified independently (all correct):**

- **Column coverage** — all 30 mapped paths (16 + 14) resolve inside the declared item schema, so `no_column_escapes_the_fingerprint_gate` really is empty; declared JSON types line up with every pack type (`string`→utf8/timestamp, `integer`→int64).
- **The shared fingerprint** — the two committed output contracts are byte-identical (same md5), so the duplicate pin is the upstream fact the test says it is.
- **Input contracts** — `top` really is `1–999` (so the pinned 999 is the ceiling exactly), `nextLink` is `format: uri`, `query` is `minLength: 1` with no `required` array, and `filter`/`orderby`/`skip`/`page`/`perPage`/`pageSize` are all absent while `orderBy` is present. The negative-space claims hold.
- **The `query` enforcement boundary** — `mod.rs:186` is `contains_key` and `config.rs:324` rejects only nulls, so the corrected "closes the *missing* case only" wording is accurate.
- **Scan bounds** — `max_pages` defaults to 100 and `max_rows` to 100 000, so the doc's 99 900-vs-100 000 arithmetic and "`max_pages` binds first" are right.
- **Redaction** — walked every string leaf of all three row fixtures against the audit's arms by hand: they pass, and all thirteen tripwire probes do trip. An independent sweep of the whole diff for GUIDs, emails, cids and bearer tokens found nothing that is not obviously synthetic (`00000000-0000-0000-0000-0000480728c5` is Microsoft's public first-party Office app id).
- **Rosters** — `outlook` appears in exactly two Rust files (`packs/mod.rs`, `source_pack.rs`), both updated; the pack guide is linked from the README and `docs/open-connector.md`. `builtin_assets_parse_and_validate` is registry-driven, so no third list was missed.
- **Compilability of the new test code** — `FieldType` is `Copy + PartialEq + Debug` (so `shape()` works), `MockResponse::new` takes `impl Into<Vec<u8>>` (so the `format!` bodies compile), and the `catch_unwind` + `json!({ key: leak })` probe pattern is the one `discord.rs` already ships.

The pack's genuinely load-bearing values (the two fingerprints) are checked by CI both directly and via eight e2e registrations, so a bad pin cannot ship. I found no correctness, safety or security defect that blocks merge, and the phase-4 write-up matches what the fixtures and the engine actually say.

**Findings are all non-blocking.** The sharpest is the redaction audit's `id` arm, which still admits the partial-redaction class this PR set out to close everywhere else. The rest are an operator trap the pack could close itself (`folderItemId` vs `folderPath`), one decision resting on a single drive with no tracked follow-up, three doc claims that are slightly stronger than the evidence, and one small DRY nit.

One thing worth stating plainly since the PR asks for it: the 16→14 reduction is the right call on the evidence, and pinning it with `search_columns_are_drive_items_minus_the_two_concurrency_tags` rather than prose is the correct instinct — a reorder or retype on one table only is exactly what the width and coverage gates miss.

BtXin (approved):

crates/skardi/src/sources/providers/open_connector/packs/one_drive.yaml
gracexmatin: Done — option 2, with one correction to where it lands.

`loader.rs:219` can only do half of it. `validate_table` takes a `&SourcePackTable`, so it sees the pack's *declaration*; "this binding carries both keys" is only visible at binding resolution in `mod.rs`. A check added purely alongside the overlap check would never fire. So enforcement is split:

- **loader** validates the declaration: a group needs ≥2 members to be a choice, every member must be a declared resource, no member may be `required` (it is always present, so every other member would be unreachable — a group that silently forbids configuration the pack still advertises), and no key may sit in two groups (which would make the reported pair arbitrary).
- **registration** validates the binding, in the per-table loop next to the required-resource check and before discovery. A conflict fails with a new `ConflictingResourceInputs` naming the table and both keys.

`SourcePackTable` gained `exclusive_resources: &'static [&'static [&'static str]]` — additive, defaulted empty, so the other seven packs are untouched — plus a `conflicting_resources` predicate. `one_drive.yaml` declares `exclusive: [[folderItemId, folderPath]]` on `drive_items`; `driveId` is deliberately *not* in the group, since it selects the drive and composes with either scope.

On option 1: I did not take it. "`folderPath` is pure convenience" holds for the executor's precedence but not for the operator — anyone writing a binding by hand knows `/Documents/Finance` and would need an extra API call to get the corresponding id. It is also live-verified working capability (phase 4 forwarded a real `folderPath`), so removing it to close a diagnostic gap trades away more than it buys.

Framing worth adding to the record: this isn't really about *which* key should win. A binding carrying both has stated two different scopes, so any precedence — id or path — silently discards half of what the operator asked for. Refusing the ambiguity is the only answer that doesn't guess, and it turns the yaml's "ALTERNATIVES" into a checked fact rather than prose. This is the one place the pack is deliberately stricter than the gateway, and it says so in the code and the docs.

Pack-scoped tests 27 → 30: the group's declaration (and its *absence* on `drive_item_search`, so it cannot be acquired by copy-paste), a both-scopes binding refused before any HTTP (asserting the gateway sees only the health probe), and a single-scope-plus-`driveId` binding still registering — without that arm the group could quietly forbid the configuration the pack advertises. Loader gains four rejection arms and two round-trip tests.

Design record open item 7 is now closed rather than "recorded", the pack doc states that Skardi refuses the binding and why, and the tasks-spec entry carries the engine addition. `3684d9a`.

PR #224, opened 2026-08-21
feat(cli): named contexts, a current context, and `skardi config` (contexts/login M1)

bakey:
**M1 of [the CLI contexts and login design](https://github.com/SkardiLabs/skardi-cloud/blob/main/docs/superpowers/specs/2026-08-20-skardi-cli-contexts-and-login-design.md)** (the design doc lives in `skardi-cloud`; this milestone is assigned to `skardi` by its §4 and §11): the config file shape, per-field resolution, context selection, the `skardi config` surface, atomic `0600` writes, mutation safety, and legacy back-compat. **No network code** — `login`/`logout` are M2.

## Scope note

The design sequences one PR per milestone. Of the six, **M3 and M6 are blocked**: §7.0 sequences them explicitly after `skardi-cloud#288`, which is still open. #301 has merged, so M5 is unblocked, and M4 has no dependency. This PR is M1, which everything else rests on.

## The load-bearing decision

A **`mode: cloud` context is authoritative** for its `server` and `token`, inverting the precedence a `mode: server` context has always had:

```
mode: server   server: --server > $SKARDI_SERVER_URL > context > default
mode: cloud    server: --server > context            (env is a hard error)
```

A cloud context's `{server, token}` are a matched pair — the PAT is scoped to one workspace at one role, and the gateway that honours it is the one `login` wrote beside it. A `SKARDI_SERVER_URL` left exported from the single-server era would otherwise send a workspace-scoped credential to whatever listens there. So the conflict is **refused by name** rather than silently applied or silently ignored. Flags still win, because `--server` is a deliberate act at the point of use.

## Everything ambiguous is a hard error, not a fallback

| Situation | Behaviour |
|---|---|
| Unknown `--context` / `$SKARDI_CONTEXT` / `current-context` | lists what exists; never falls through to `http://127.0.0.1:8080` — the one failure that would quietly send a cloud query to a local server |
| `mode: cloud` with no `workspace` | refuses at resolution, naming the repair, instead of issuing a request the gateway answers `workspace_required` |
| Env var conflicting with a cloud context | names the variable |

## Reads tolerate a broken file; mutations refuse it (§5.4)

Deliberately asymmetric. A read warns, ignores, falls through — so `--server` keeps working while the file is repaired. A mutation refuses to open it: a rewrite built from an empty parse tree would atomically replace a malformed-but-credential-bearing file, and the rename is torn-write-safe, not data-loss-safe. There is no `--force`, because that file may hold the only copy of a PAT nobody can re-mint without re-authenticating.

## Write path

Temp file in the **same** directory (a cross-filesystem rename is not atomic), created `0600` **before any bytes land** so a token never sits in a world-readable file even briefly, then renamed. Unknown top-level and per-context keys are preserved, so an older CLI's `use-context` cannot downgrade a file a newer one wrote. `config view` redacts tokens but keeps a recognizable prefix, the way git shows short hashes.

## Back-compat

A legacy `spec:` file resolves as one context named `default`, and its **first edit promotes it** into `contexts:` rather than dropping the credential it holds. A file carrying both prefers `contexts:` and warns once.

## Verification — all run locally

- **34 new unit tests** — 24 in `config.rs` over the precedence/selection matrix and the file surface, 10 in `commands/config.rs` covering every `config` subcommand in-process. (93 unit total in the crate, including 59 pre-existing.)
- **7 integration tests driving the real binary** with `HOME` redirected at a temp dir: create/list/switch/delete, redaction, unknown-key preservation, legacy promotion, the serverless-cloud refusal, and each resolution failure asserted at **exit 1 rather than the connect code `2`** — which is how "no request was issued" is demonstrated rather than assumed. One asserts a broken file survives a refused mutation **byte for byte**; another asserts `--server` *does* let the request through by expecting exit 2 against a `.invalid` host.
- **Live against a real `skardi-server` on :8080**: a server-mode context, `--context`, `$SKARDI_CONTEXT`, and a legacy `spec:` file each return rows.

`cargo test -p skardi-cli`: **93 unit + 7 integration + 4 e2e (2 ignored, server-gated)**. Clippy clean, `cargo fmt` clean.

## Review rounds

Three P1s and four P2s from review are fixed in this branch. The ones worth knowing about as a reviewer:

- **A cloud context with no `server` fell through to `http://127.0.0.1:8080` carrying the workspace PAT** — a hole in this PR's own central claim, reproduced against the real binary. Now a typed refusal at resolution *and* at `set-context`.
- **`redact()` printed short tokens in full** while labelling them `(redacted)`; legacy `spec:` tokens are arbitrary strings, so `abc123` rendered as `abc123…(redacted)`. Prefix dropped entirely — the context name already identifies the credential.
- **The env-conflict error's advice was false**: it said to pass `--server`, but the check ran before flags applied. Now per field, and only where the env would actually win.
- **Parse errors quoted the file**, so a token on a broken line hit stderr on every command. All complaints now report line/column only.
- **`docs/cli.md`** documented the pre-contexts precedence and no contexts at all; rewritten, with every command and flag checked against `--help`.
- **The lone-context selection step** is a documented deviation from §5.1 rather than an implicit one — it is load-bearing for §5.2's legacy files, and §5.1 is amended in skardi-cloud to say so.

## Reviewer notes

- `ClientConfig` gained a `context` field, so the seven test helpers that build it literally each gained `context: None`. No behaviour change to `client.rs` or any existing command.
- Resolution is deliberately **lazy, inside `dispatch`** after the `config` arm: `set-context` is how an operator repairs a config whose cloud context is missing its workspace, so it must not be gated on that resolution succeeding.
- M2 will add `ClientConfig::workspace()` and the `Skardi-Workspace` header together; the accessor is deliberately absent here rather than shipped unused.
- **Coverage note for anyone reading the Codecov diff:** the integration suite contributes none, because CI instruments test binaries and those tests spawn `skardi` as a subprocess. That is why `commands/config.rs` is covered in-process as well as through the binary — the two layers prove different things (logic vs. wiring and exit codes).

### DIFF
--- a/crates/cli/src/commands/config.rs
+++ b/crates/cli/src/commands/config.rs
@@ -58,8 +58,9 @@ pub enum ConfigCmd {
         current: bool,
     },
 
-    /// Delete one context. Does not revoke its credential — see `skardi logout
-    /// --revoke` for that.
+    /// Delete one context. Does NOT revoke its credential: a PAT stays valid
+    /// until it is revoked in the console (or by a future `skardi logout
+    /// --revoke`, which does not exist yet).
     DeleteContext {
         /// context name
         name: String,
@@ -136,8 +137,14 @@ fn current_context(path: &Path) -> Result<()> {
 
 fn use_context(path: &Path, name: &str) -> Result<()> {
     let mut file = config::load_for_mutation(path)?;
-    if file.find(name).is_none() {
-        bail!("no context named '{name}'. Available: {}", available(&file));
+    // One call: `effective_contexts` warns when the file carries both shapes,
+    // and asking twice printed that "warns once" twice.
+    let names = file.context_names();
+    if !names.iter().any(|n| n == name) {
+        bail!(
+            "no context named '{name}'. Available: {}",
+            render_names(&names)
+        );
     }
     file.current_context = Some(name.to_string());
     config::save(path, &file)?;
@@ -162,6 +169,22 @@ fn set_context(
         Some("cloud") => Some(ContextMode::Cloud),
         Some(other) => bail!("--mode must be 'server' or 'cloud', not '{other}'"),
     };
+
+    // Trim, and treat whitespace-only as unset — exactly as resolution does.
+    // Without this `--workspace "  "` passed the cloud guard below and wrote
+    // the dud the guard exists to prevent: resolution then refused it, so the
+    // context could be created but never used.
+    let trimmed = |value: Option<String>| -> Option<String> {
+        value
+            .map(|v| v.trim().to_string())
+            .filter(|v| !v.is_empty())
+    };
+    let (server, workspace, token, user) = (
+        trimmed(server),
+        trimmed(workspace),
+        trimmed(token),
+        trimmed(user),
+    );
 
     let mut file = config::load_for_mutation(path)?;
     // A legacy `spec:`-only file gains its `default` context on first edit,
@@ -240,7 +263,11 @@ fn delete_context(path: &Path, name: &str) -> Result<()> {
     let before = file.contexts.len();
     file.contexts.retain(|c| c.name != name);
     if file.contexts.len() == before {
-        bail!("no context named '{name}'. Available: {}", available(&file));
+        let names: Vec<String> = file.contexts.iter().map(|c| c.name.clone()).collect();
+        bail!(
+            "no context named '{name}'. Available: {}",
+            render_names(&names)
+        );
     }
     // A dangling current-context would make every later command a hard error
     // (§5.1's unknown-context rule), so clear the pointer with the context.
@@ -254,6 +281,14 @@ fn delete_context(path: &Path, name: &str) -> Result<()> {
 }
 
 fn view(path: &Path, show_tokens: bool) -> Result<()> {
+    print!("{}", render_view(path, show_tokens)?);
+    Ok(())
+}
+
+/// Render what `view` prints. Separate from the printing so tests can assert
+/// the TEXT — the earlier test only checked that `view` returned `Ok`, which
+/// is why it did not notice that short tokens were being printed in full.
+fn render_view(path: &Path, show_tokens: bool) -> Result<String> {
     let content = match std::fs::read_to_string(path) {
         Ok(content) => content,
         Err(err) => bail!("cannot read {}: {err}", path.display()),
@@ -279,11 +314,7 @@ fn view(path: &Path, show_tokens: bool) -> Result<()> {
             spec.token = spec.token.as_deref().map(redact);
         }
     }
-    print!(
-        "{}",
-        serde_yaml::to_string(&file).context("serialize config")?
-    );
-    Ok(())
+    serde_yaml::to_string(&file).context("serialize config")
 }
 
 /// Replace a token entirely. No prefix survives, deliberately.
@@ -300,8 +331,7 @@ fn redact(_token: &str) -> String {
     "(redacted)".to_string()
 }
 
-fn available(file: &ContextsFile) -> String {
-    let names = file.context_names();
+fn render_names(names: &[String]) -> String {
     if names.is_empty() {
         "(none)".to_string()
     } else {
@@ -427,6 +457,23 @@ mod tests {
             !path.exists(),
             "a rejected cloud context must not be written"
         );
+
+        // Whitespace is unset on the WRITE path too, as it is at resolution.
+        // Otherwise `--workspace "  "` passed this guard and wrote the dud the
+        // guard exists to prevent — creatable, then unusable.
+        let err = set_context(
+            &path,
+            "c",
+            Some("https://gw".to_string()),
+            Some("cloud".to_string()),
+            Some("   ".to_string()),
+            None,
+            None,
+            false,
+        )
+        .unwrap_err();
+        assert!(err.to_string().contains("needs a workspace"), "{err}");
+        assert!(!path.exists());
 
         // With both it lands.
         set_context(
@@ -573,13 +620,25 @@ mod tests {
         assert!(err.to_string().contains("cannot read"), "{err}");
 
         // A legacy `spec:` token is redacted too — it is just as live as a
-        // context's, and only this path reaches it.
+        // context's, and only this path reaches it. Asserted on the RENDERED
+        // text: the earlier version of this test only checked that `view`
+        // returned Ok, which is why it did not notice short tokens being
+        // printed in full.
         write(
             &path,
             "kind: client\nspec:\n  server: http://legacy:8080\n  token: legacy-secret\n",
         );
-        view(&path, false).unwrap();
-        view(&path, true).unwrap();
+        let redacted = render_view(&path, false).unwrap();
+        assert!(
+            !redacted.contains("legacy-secret"),
+            "the legacy token leaked:\n{redacted}"
+        );
+        assert!(redacted.contains("(redacted)"), "{redacted}");
+        // The rest of the file still renders, so `view` stays useful.
+        assert!(redacted.contains("http://legacy:8080"), "{redacted}");
+
+        let shown = render_view(&path, true).unwrap();
+        assert!(shown.contains("legacy-secret"), "--show-tokens prints it");
 
         // And an unparsable file is the one case `view` exists to diagnose.
         write(&path, "contexts: [broken\n");
@@ -601,10 +660,13 @@ mod tests {
     }
 
     #[test]
-    fn available_lists_names_or_says_none() {
-        let empty = ContextsFile::default();
-        assert_eq!(available(&empty), "(none)");
+    fn render_names_lists_names_or_says_none() {
+        assert_eq!(render_names(&[]), "(none)");
+        assert_eq!(render_names(&["a".to_string()]), "a");
+        assert_eq!(render_names(&["a".to_string(), "b".to_string()]), "a, b");
 
+        // A legacy `spec:` file reports its promoted name, so the listing in
+        // an error message is never empty just because `contexts:` is.
         let legacy = ContextsFile {
             spec: Some(LegacySpec {
                 server: Some("http://x:1".to_string()),
@@ -612,6 +674,9 @@ mod tests {
             }),
             ..ContextsFile::default()
         };
-        assert_eq!(available(&legacy), config::LEGACY_CONTEXT_NAME);
+        assert_eq!(
+            render_names(&legacy.context_names()),
+            config::LEGACY_CONTEXT_NAME
+        );
     }
 }
--- a/crates/cli/src/commands/config.rs
+++ b/crates/cli/src/commands/config.rs
@@ -1,0 +1,285 @@
+//! `skardi config …` — read and edit `~/.skardi/config.yaml`.
+//!
+//! Every subcommand here is a pure file edit: no network, no server contact.
+//! That is a deliberate boundary — `login` is the only module that knows a
+//! control plane exists, so these commands keep working (and keep being
+//! testable) with no cloud reachable at all.
+//!
+//! Mutations go through [`config::load_for_mutation`], which refuses a file it
+//! cannot parse. Reads use the tolerant loader, so `view` can still show the
+//! operator what is wrong with a broken file rather than only failing.
+
+use crate::config::{self, Context, ContextMode, ContextsFile};
+use anyhow::{Context as _, Result, bail};
+use clap::Subcommand;
+use std::path::Path;
+
+#[derive(Subcommand, Debug)]
+pub enum ConfigCmd {
+    /// List every context, marking the current one with `*`.
+    GetContexts,
+
+    /// Print the current context's name.
+    CurrentContext,
+
+    /// Switch the current context.
+    UseContext {
+        /// context name (see `skardi config get-contexts`)
+        name: String,
+    },
+
+    /// Create or update one context, then optionally make it current.
+    SetContext {
+        /// context name; created if it does not exist
+        name: String,
+
+        #[arg(long, value_name = "URL")]
+        server: Option<String>,
+
+        /// `cloud` reaches a skardi-cloud gateway with a workspace-scoped PAT;
+        /// `server` reaches a skardi-server directly
+        #[arg(long, value_name = "MODE")]
+        mode: Option<String>,
+
+        /// workspace slug — required by a cloud context, sent per request
+        #[arg(long, value_name = "SLUG")]
+        workspace: Option<String>,
+
+        #[arg(long, value_name = "TOKEN")]
+        token: Option<String>,
+
+        #[arg(long, value_name = "EMAIL")]
+        user: Option<String>,
+
+        /// also set this context as current
+        #[arg(long)]
+        current: bool,
+    },
+
+    /// Delete one context. Does not revoke its credential — see `skardi logout
+    /// --revoke` for that.
+    DeleteContext {
+        /// context name
+        name: String,
+    },
+
+    /// Print the config file, with tokens redacted.
+    View {
+        /// print tokens in full instead of redacting them
+        #[arg(long)]
+        show_tokens: bool,
+    },
+}
+
+/// Dispatch one `skardi config` subcommand.
+pub fn run(cmd: ConfigCmd) -> Result<()> {
+    let path = config::default_config_path()
+        .context("cannot determine the home directory for ~/.skardi/config.yaml")?;
+
+    match cmd {
+        ConfigCmd::GetContexts => get_contexts(&path),
+        ConfigCmd::CurrentContext => current_context(&path),
+        ConfigCmd::UseContext { name } => use_context(&path, &name),
+        ConfigCmd::SetContext {
+            name,
+            server,
+            mode,
+            workspace,
+            token,
+            user,
+            current,
+        } => set_context(&path, &name, server, mode, workspace, token, user, current),
+        ConfigCmd::DeleteContext { name } => delete_context(&path, &name),
+        ConfigCmd::View { show_tokens } => view(&path, show_tokens),
+    }
+}
+
+fn get_contexts(path: &Path) -> Result<()> {
+    let file = config::load(path).unwrap_or_default();
+    let contexts = file.effective_contexts();
+    if contexts.is_empty() {
+        println!("no contexts configured ({})", path.display());
+        return Ok(());
+    }
+    let current = file.current_context.as_deref().unwrap_or("");
+    println!(
+        "{:<2} {:<24} {:<8} {:<20} SERVER",
+        "", "NAME", "MODE", "WORKSPACE"
+    );
+    for context in contexts {
+        println!(
+            "{:<2} {:<24} {:<8} {:<20} {}",
+            if context.name == current { "*" } else { "" },
+            context.name,
+            context.mode,
+            context.workspace.as_deref().unwrap_or("-"),
+            context.server.as_deref().unwrap_or("-"),
+        );
+    }
+    Ok(())
+}
+
+fn current_context(path: &Path) -> Result<()> {
+    let file = config::load(path).unwrap_or_default();
+    match file.current_context.as_deref().map(str::trim) {
+        Some(name) if !name.is_empty() => {
+            println!("{name}");
+            Ok(())
+        }
+        // Exit non-zero: a script asking "which context" and getting silence
+        // plus success would proceed against the wrong one.
+        _ => bail!("no current context is set ({})", path.display()),
+    }
+}
+
+fn use_context(path: &Path, name: &str) -> Result<()> {
+    let mut file = config::load_for_mutation(path)?;
+    if file.find(name).is_none() {
+        bail!("no context named '{name}'. Available: {}", available(&file));
+    }
+    file.current_context = Some(name.to_string());
+    config::save(path, &file)?;
+    println!("switched to context '{name}'");
+    Ok(())
+}
+
+#[allow(clippy::too_many_arguments)]
+fn set_context(
+    path: &Path,
+    name: &str,
+    server: Option<String>,
+    mode: Option<String>,
+    workspace: Option<String>,
+    token: Option<String>,
+    user: Option<String>,
+    current: bool,
+) -> Result<()> {
+    let mode = match mode.as_deref() {
+        None => None,
+        Some("server") => Some(ContextMode::Server),
+        Some("cloud") => Some(ContextMode::Cloud),
+        Some(other) => bail!("--mode must be 'server' or 'cloud', not '{other}'"),
+    };
+
+    let mut file = config::load_for_mutation(path)?;
+    // A legacy `spec:`-only file gains its `default` context on first edit,
+    // so the mutation does not silently drop the credential it holds.
+    if file.contexts.is_empty() {
+        file.contexts = file.effective_contexts();
+        file.spec = None;
+    }
+
+    let existing = file.contexts.iter().position(|c| c.name == name);
+    let created = existing.is_none();
+    let index = existing.unwrap_or_else(|| {
+        file.contexts.push(Context {
+            name: name.to_string(),
+            ..Context::default()
+        });
+        file.contexts.len() - 1
+    });
+    let context = &mut file.contexts[index];
+    // Only fields the caller named are touched: `set-context --server X` must
+    // not clear a token the same context already holds.
+    if let Some(server) = server {
+        context.server = Some(server);
+    }
+    if let Some(mode) = mode {
+        context.mode = mode;
+    }
+    if let Some(workspace) = workspace {
+        context.workspace = Some(workspace);
+    }
+    if let Some(token) = token {
+        context.token = Some(token);
+    }
+    if let Some(user) = user {
+        context.user = Some(user);
+    }
+
+    // Refused here rather than at the next command: a cloud context with no
+    // workspace cannot issue a request, so writing one is writing a dud.
+    if context.mode == ContextMode::Cloud && context.workspace.is_none() {
+        bail!(
+            "context '{name}' is mode: cloud, so it needs a workspace: pass \
+             --workspace SLUG"
+        );
+    }
+
+    if current {
+        file.current_context = Some(name.to_string());
+    }
+    config::save(path, &file)?;
+    println!(
+        "{} context '{name}'{}",
+        if created { "created" } else { "updated" },
+        if current { " (now current)" } else { "" }
+    );
+    Ok(())
+}
+
+fn delete_context(path: &Path, name: &str) -> Result<()> {
+    let mut file = config::load_for_mutation(path)?;
+    if file.contexts.is_empty() {
+        file.contexts = file.effective_contexts();
+        file.spec = None;
+    }
+    let before = file.contexts.len();
+    file.contexts.retain(|c| c.name != name);
+    if file.contexts.len() == before {
+        bail!("no context named '{name}'. Available: {}", available(&file));
+    }
+    // A dangling current-context would make every later command a hard error
+    // (§5.1's unknown-context rule), so clear the pointer with the context.
+    if file.current_context.as_deref() == Some(name) {
+        file.current_context = None;
+        println!("note: '{name}' was the current context; no context is current now");
+    }
+    config::save(path, &file)?;
+    println!("deleted context '{name}'");
+    Ok(())
+}
+
+fn view(path: &Path, show_tokens: bool) -> Result<()> {
+    let content = match std::fs::read_to_string(path) {
+        Ok(content) => content,
+        Err(err) => bail!("cannot read {}: {err}", path.display()),
+    };
+    let mut file: ContextsFile = match serde_yaml::from_str(&content) {
+        Ok(file) => file,
+        // `view` is the diagnostic for a broken file, so it prints the parse
+        // error and the path rather than the tolerant loader's warning.
+        Err(err) => bail!("{} does not parse: {err}", path.display()),
+    };
+    if !show_tokens {
+        for context in &mut file.contexts {
+            if let Some(token) = &context.token {
+                context.token = Some(redact(token));
+            }
+        }
+        if let Some(spec) = file.spec.as_mut().filter(|s| s.token.is_some()) {
+            spec.token = spec.token.as_deref().map(redact);
+        }
+    }
+    print!(
+        "{}",
+        serde_yaml::to_string(&file).context("serialize config")?
+    );
+    Ok(())
+}
+
+/// Keep enough of a token to recognize WHICH credential it is, without
+/// printing anything usable — the same reason `git` shows short hashes.
+fn redact(token: &str) -> String {
+    let visible: String = token.chars().take(12).collect();
+    format!("{visible}…(redacted)")
+}
+
+fn available(file: &ContextsFile) -> String {
+    let names = file.context_names();
+    if names.is_empty() {
+        "(none)".to_string()
+    } else {
+        names.join(", ")
+    }
+}
--- a/crates/cli/src/commands/config.rs
+++ b/crates/cli/src/commands/config.rs
@@ -260,9 +260,14 @@ fn view(path: &Path, show_tokens: bool) -> Result<()> {
     };
     let mut file: ContextsFile = match serde_yaml::from_str(&content) {
         Ok(file) => file,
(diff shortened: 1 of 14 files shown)


crates/cli/src/config.rs
```diff
@@ -1,307 +1,1288 @@
...
 
-    match serde_yaml::from_str::<ConfigManifest>(&content) {
-        Ok(manifest) => manifest.spec,
+/// Load `path` for READING. Tolerant by design (§5.4): a missing file is
+/// `None`, and an unparsable one warns and resolves to `None` so that
+/// `--server`/`--token` still work while the operator fixes it. Never fatal.
+///
+/// Mutations use [`load_for_mutation`] instead, which refuses the same file.
+pub fn load(path: &Path) -> Option<ContextsFile> {
+    let content = std::fs::read_to_string(path).ok()?;
```
yiyi-wang-skardi: `read_to_string(path).ok()?` collapses *every* read failure into "no config", including the ones that are not absence. `load_for_mutation` right below is careful to separate `NotFound` from everything else, for exactly the reason that matters here — and the read path lost that distinction.

Concrete case: `~/.skardi/config.yaml` ends up root-owned (a `sudo skardi login`, a restored backup, a container volume with a different uid). `skardi query` then gets `EACCES`, silently resolves to `http://127.0.0.1:8080` with no token, and prints nothing about it. That is the same shape as the failure the unknown-context and `CloudContextWithoutServer` guards exist to prevent — the operator believes they are talking to their cloud gateway and are in fact talking to whatever is on :8080 — just reached through a route where nobody is watching.

A warning is enough; the read path should stay non-fatal per §5.4. Something like:

```rust
let content = match std::fs::read_to_string(path) {
    Ok(content) => content,
    Err(err) if err.kind() == std::io::ErrorKind::NotFound => return None,
    Err(err) => {
        eprintln!("warning: ignoring unreadable config file {}: {err}", path.display());
        return None;
    }
};
```

Pre-existing behaviour (the old `load_file_config` did the same), so not a regression — but this PR is what turns a missing warning into a credential-routing question.

crates/cli/src/config.rs
```diff
@@ -1,307 +1,1288 @@
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
     pub token: Option<String>,
+    /// The PAT's id, needed by `logout --revoke` — the token itself cannot
+    /// name itself to the revoke endpoint.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub token_id: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub token_expires_at: Option<String>,
+    #[serde(flatten)]
+    pub extra: BTreeMap<String, serde_yaml::Value>,
```
yiyi-wang-skardi, line 123: `extra` absorbing unknown per-context keys is the right call for forward compatibility, but it also means a **typo in a security-relevant key is silently accepted, and it fails in the unsafe direction**.

`mode` is the case that matters. Hand-edit `Mode: cloud` (capital M — or `modee`, or a stray trailing space in the key) and:

- the key lands in `extra` and is faithfully preserved,
- `mode` takes its `Default` → `ContextMode::Server`,
- the `token` and `server` are still read from the context,
- so the cloud-authoritative rule no longer applies, and `$SKARDI_SERVER_URL` can once again send the workspace-scoped PAT to whatever listens there.

That is the module doc's opening scenario, reachable by one character. The value-side typo is safe by contrast (`mode: Cloud` is an unknown variant → parse error → tolerant read ignores the whole file → default server *with no token*), which is a nice illustration that the unsafe direction here is specifically the *key* side.

`deny_unknown_fields` is not available alongside `flatten`, and you would not want it anyway. But a warning on unknown per-context keys — or just on ones that match a known key case-insensitively — would keep the preservation guarantee while making the typo visible:

```rust
for key in context.extra.keys() {
    if KNOWN_CONTEXT_KEYS.iter().any(|k| k.eq_ignore_ascii_case(key)) {
        eprintln!("warning: context '{}' has key '{key}', which differs only in case from a key this CLI reads; it is being ignored", context.name);
    }
}
```

Low likelihood while the file is machine-written — but the PR's own integration test hand-edits it, and §5.4 exists precisely because operators do.

crates/cli/src/config.rs
```diff
@@ -1,307 +1,1288 @@
...
+
+    let yaml = serde_yaml::to_string(file).context("serialize config")?;
+    let stem = path
+        .file_name()
+        .map(|n| n.to_string_lossy().to_string())
+        .unwrap_or_else(|| "config.yaml".to_string());
+    // A crash between create and rename leaves a temp file holding a full copy
+    // of every token, and nothing else ever removed it. Sweep the strays we
+    // can see before adding another.
+    remove_stale_temp_files(parent, &stem, path);
```
yiyi-wang-skardi, line 715: The sweep has no age or liveness guard, so it cannot tell a stray from **a temp file another process is still writing**. That converts the concurrency gap the doc comment above documents ("the later rename wins, so one edit is silently lost") into a spurious hard failure:

1. A: sweeps (clean), creates `.config.yaml.tmp-100-500`, writing…
2. B: sweeps → sees `.config.yaml.tmp-100-500` (regular file, not the target) → **deletes it** → writes its own temp → renames → exits 0.
3. A: `rename(".config.yaml.tmp-100-500", "config.yaml")` → `ENOENT` → `error: replace /home/u/.skardi/config.yaml atomically: No such file or directory`.

So A's write is still lost, but now with an error message that points at the wrong thing entirely, and the cleanup `remove_file` on the following line silently fails too. `skardi login` (M2) racing a `skardi config use-context` in a shell is the realistic trigger; a wrapper script that fires two `set-context` calls in parallel is another.

The doc comment on `save` reasons carefully about concurrency but predates this sweep. An mtime guard keeps it a pure garbage collector and costs one `metadata` call:

```rust
// Only sweep what no live writer could still own.
let stale = entry.metadata().ok()
    .and_then(|m| m.modified().ok())
    .and_then(|t| t.elapsed().ok())
    .is_some_and(|age| age > std::time::Duration::from_secs(60));
if !stale { continue; }
```

Sweeping only the current pid's own leftovers would also work and needs no clock.

crates/cli/src/commands/config.rs
```diff
@@ -0,0 +1,682 @@
...
+        // `view` is the diagnostic for a broken file — position and path, not
+        // the raw serde message, which quotes the offending scalar and would
+        // print a token from the very command that promises to redact them.
+        Err(err) => bail!(
+            "{} does not parse ({}). Open it to see the offending line",
+            path.display(),
+            config::describe_parse_error(&err)
+        ),
+    };
+    if !show_tokens {
```
yiyi-wang-skardi, line 369: Redaction is an **allowlist over the fields this binary models**, while the file model is deliberately open — which means the two mechanisms disagree about the same file.

`render_view` re-serializes the parsed tree, so anything sitting in `Context::extra` or `ContextsFile::extra` is printed verbatim. A newer CLI that adds, say, `refresh-token:` or `client-secret:` under a context gets its value preserved by `save` (correct, that is the whole point of `extra`) and then printed in full by an older CLI's `config view` — the command whose doc comment calls it "the command people run while screen-sharing", and which prints `(redacted)` two lines above, so the output actively reads as safe.

This is forward-looking today (no such key exists yet), but the fix is cheap and belongs with the mechanism that creates the exposure rather than with the field that eventually trips it:

```rust
const SENSITIVE: [&str; 3] = ["token", "secret", "password"];
let looks_sensitive = |k: &str| {
    let k = k.to_ascii_lowercase();
    SENSITIVE.iter().any(|needle| k.contains(needle))
};
```

and run it over both `extra` maps under the same `if !show_tokens`.

Unrelated nit two lines below: `file.spec.as_mut().filter(|s| s.token.is_some())` — the `filter` is redundant, since `spec.token.as_deref().map(redact)` already leaves `None` alone. `if let Some(spec) = file.spec.as_mut() { spec.token = spec.token.as_deref().map(redact); }` says the same thing.

crates/cli/src/commands/config.rs
```diff
@@ -0,0 +1,682 @@
...
+}
+
+fn get_contexts(path: &Path) -> Result<()> {
+    let file = config::load(path).unwrap_or_default();
+    let contexts = file.effective_contexts();
+    if contexts.is_empty() {
+        println!("no contexts configured ({})", path.display());
+        return Ok(());
+    }
+    let current = file.current_context.as_deref().unwrap_or("");
```
yiyi-wang-skardi: The `*` marker re-implements "which context is current" instead of asking resolution, and the two answers differ in three ways:

1. **No lone-context auto-selection.** The single most common case — an existing pre-contexts `spec:` install — shows `default` with *no* `*`, and `skardi config current-context` (line 127) exits 1 with "no current context is set", while every actual command is happily using `default`. `resolve_from`'s comment calls that auto-selection step load-bearing; the surface that reports it doesn't know about it.
2. **No trim.** `current-context: " local "` is selected by `resolve_from` (via `non_empty`) but does not match here, so nothing gets marked.
3. **`--context` / `$SKARDI_CONTEXT` are ignored.** They are `global = true` flags, so `skardi --context acme/prod config current-context` parses fine and prints the file's pointer instead. A script doing `ctx=$(skardi config current-context)` to find out what it is about to talk to gets the wrong answer under a harness that exports `SKARDI_CONTEXT` — and the test suite notes that env var is "the one an agent harness sets".

All three fall out for free if the marker comes from resolution rather than from the raw field — `ClientConfig::resolve(None, None, flag_context)` already returns `Some(SelectedContext { name, .. })` and already handles trim, auto-selection, and the env var. Same for `current_context`, which could print the resolved name and only fail when resolution really selects nothing.

The underlying reason this is worth a fix rather than a shrug: this module's own comments justify a hard error on a both-shapes file because "silently choosing either way would leave the operator guessing which credential their commands used". `get-contexts` showing no `*` on a file that definitely has an effective context is the same guessing game.

crates/cli/src/commands/config.rs
```diff
@@ -0,0 +1,682 @@
...
+        /// `cloud` reaches a skardi-cloud gateway with a workspace-scoped PAT;
+        /// `server` reaches a skardi-server directly
+        #[arg(long, value_name = "MODE")]
+        mode: Option<String>,
+
+        /// workspace slug — required by a cloud context, sent per request
+        #[arg(long, value_name = "SLUG")]
+        workspace: Option<String>,
+
+        #[arg(long, value_name = "TOKEN")]
```
yiyi-wang-skardi: `--token` puts a live PAT on `argv`, which is the weakest link in an otherwise careful credential path.

On Linux `/proc/<pid>/cmdline` is world-readable by default (no `hidepid`), so any local user can read the token for the lifetime of the process; it also lands in `~/.bash_history` / `~/.zsh_history` unmodified, and in the shell's own process listing. This PR goes to real lengths to keep a token out of a world-readable *file* even briefly (`create_new` + `mode(0o600)` before any bytes land, the stale-temp sweep because a leftover "holds a full copy of every token") — and `docs/cli.md:111` then documents `set-context --token skardi_pat_…` as the way to get it in there.

`--token` as a global connection flag predates this PR, so this is not new surface. What *is* new is that `set-context` is the first command that **persists** a credential taken from `argv`, and `docs/cli.md` now teaches it as the normal way to configure a cloud context.

The conventional shape is a stdin path — `docker login --password-stdin`, `gh auth login --with-token`, `kubectl create secret --from-file`. `--token-stdin` (or honouring a `SKARDI_PAT` env var, which at least is not world-readable) would cost a few lines here and is worth settling before M2's `login` becomes the primary way tokens reach this file. Keeping `--token` for scripts is fine; it just should not be the documented default.

crates/cli/src/config.rs
```diff
@@ -1,307 +1,1288 @@
...
+    /// Every context the file defines, with the legacy `spec:` block folded in
+    /// as `default` when there is no `contexts:` list.
+    ///
+    /// A file carrying BOTH prefers `contexts:` and warns once — silently
+    /// choosing either way would leave the operator guessing which credential
+    /// their commands used.
+    pub fn effective_contexts(&self) -> Vec<Context> {
+        if !self.contexts.is_empty() {
+            if self.spec.is_some() {
+                eprintln!(
```
yiyi-wang-skardi: `effective_contexts` is an accessor with an `eprintln!` in it, and that has already cost this PR two bugs: the double both-shapes warning fixed in `resolve_from` (the comment at line 362 records it) and the same one fixed in `use_context` (line 140). Both fixes are call-site discipline — "call this exactly once" — which is the fragile kind: `get_contexts` calls `effective_contexts`, `context_names` calls it again, and any future caller that needs the list twice re-introduces the bug with nothing to catch it.

The side effect is also in the wrong place. Whether the file carries both shapes is a property of the *file*, discovered when it is read; it does not need re-deriving on every access. Warning once in `load`/`load_for_mutation` (where the file is already in hand, and where the malformed-file warning already lives) makes `effective_contexts` pure and idempotent, and the "warns once" promise structural instead of conventional:

```rust
// in load / load_for_mutation, right after a successful parse
if !file.contexts.is_empty() && file.spec.is_some() {
    eprintln!("warning: config has both 'contexts:' and a legacy 'spec:' block; …");
}
```

Returning the condition from `effective_contexts` and letting callers decide would work too, but is more machinery than this needs.

crates/cli/src/commands/config.rs
```diff
@@ -0,0 +1,682 @@
...
+        "{} context '{name}'{}",
+        if created { "created" } else { "updated" },
+        if current { " (now current)" } else { "" }
+    );
+    Ok(())
+}
+
+fn delete_context(path: &Path, name: &str) -> Result<()> {
+    let mut file = config::load_for_mutation(path)?;
+    if file.contexts.is_empty() {
```
yiyi-wang-skardi: This legacy-promotion block is verbatim identical to the one in `set_context` (lines 192-195), including the ordering that matters (promote, *then* clear `spec`, so the credential is not dropped). Two copies of a three-line invariant that guards a credential is the kind of duplication that survives until one copy is updated and the other isn't.

A method on `ContextsFile` next to `effective_contexts` reads better than a free function here, since it is a property of the file rather than of either command:

```rust
/// Fold a legacy `spec:` block into `contexts:` so a mutation cannot drop
/// the credential it holds. No-op once `contexts:` is populated.
pub fn promote_legacy_spec(&mut self) {
    if self.contexts.is_empty() {
        self.contexts = self.effective_contexts();
        self.spec = None;
    }
}
```

Both call sites then become one line, and M2's `login` — which will need exactly this before writing a context — gets it for free rather than adding a third copy.

crates/cli/src/config.rs
```diff
@@ -1,307 +1,1288 @@
...
+        // Any read failure (absent, or unreadable) yields a fresh tree only
+        // when the file truly is absent; an unreadable-but-present file must
+        // not be clobbered either.
+        Err(err) if err.kind() == std::io::ErrorKind::NotFound => {
+            return Ok(ContextsFile {
+                kind: Some("client".to_string()),
+                ..ContextsFile::default()
+            });
+        }
+        Err(err) => {
```
yiyi-wang-skardi, line 664: Refusing on any non-`NotFound` read error is the right behaviour, but reporting it as `UnparsableForMutation` makes the message say something untrue. `Display` renders "refusing to modify {path}: **it exists but does not parse** (Permission denied)" — the operator goes looking for a YAML syntax error that isn't there.

The test at line 1243 hits this exact path (a directory standing in for an unreadable file), so the message it produces is "…does not parse (Is a directory (os error 21))".

Either a second variant, or a small reword that covers both — the actionable advice ("fix or move the file first") is the same in either case:

```rust
UnreadableForMutation { path: PathBuf, error: String },
// "refusing to modify {path}: it exists but cannot be read ({error}). …"
```

crates/cli/src/config.rs
```diff
@@ -1,307 +1,1288 @@
...
+///
+/// Temp file in the SAME directory then rename: a cross-filesystem rename is
+/// not atomic, and `~/.skardi` is where the target lives. The temp file is
+/// created `0600` before any bytes are written, so a token never exists in a
+/// world-readable file even briefly.
+pub fn save(path: &Path, file: &ContextsFile) -> anyhow::Result<()> {
+    use anyhow::Context as _;
+
+    let parent = path.parent().unwrap_or_else(|| Path::new("."));
+    std::fs::create_dir_all(parent)
```
yiyi-wang-skardi: `create_dir_all` uses the process umask, so a first `set-context`/`login` on a typical `umask 022` box creates `~/.skardi` as `0755`. The config file itself is `0600` and the temp files inherit that, so no credential is exposed — but the directory that exists to hold credentials is group- and world-listable, which lets any local user enumerate that the file exists, watch temp files appear and disappear, and (with write access to a shared parent, which is not the normal case) is one step from the usual dotfile-directory tricks.

`~/.ssh` and `~/.gnupg` are `0700` for this reason, and given `warn_if_loose_permissions` already exists for the file, the directory is the one half not covered. `std::fs::DirBuilder` with `mode(0o700)` under `#[cfg(unix)]` (recursive, so it matches `create_dir_all`'s semantics) covers it; note that `mode` only applies to directories this call actually creates, so a pre-existing loose `~/.skardi` is untouched — the same asymmetry `warn_if_loose_permissions` handles for the file, if you want to mirror it.

yiyi-wang-skardi (approved):
## Audit of M1: named contexts, `current-context`, and `skardi config`

I read every changed file in full (`config.rs`, `commands/config.rs`, `main.rs`, `tests/contexts.rs`, `docs/cli.md`, and the seven one-line test-helper touch-ups), and diffed the new resolution against the pre-PR `resolve_from` to check nothing regressed for existing installs.

**Caveat on verification:** no Rust toolchain is available in this review environment, so I could not run `cargo test`/`clippy`/`fmt` myself. Everything below is from reading the code; the 93+7 test results in the description are taken at face value.

### What holds up

- **The central safety claim checks out.** In `resolve_from`, a `mode: cloud` context can only reach `server.unwrap_or_else(DEFAULT_SERVER_URL)` through a `Some(..)`, because the `CloudContextWithoutServer` guard fires first. I traced every route into that line (flag, env, context, auto-selected lone context, legacy `spec:`) and found none that lands a workspace PAT on `http://127.0.0.1:8080`.
- **The env-conflict guard is right, and right for the right reason.** Checking `env.is_some() && flag.is_none()` per field makes the error's own advice ("pass `--server`/`--token`") actually followable, and it correctly lets an exported-but-empty `SKARDI_API_TOKEN=` fall through rather than becoming a hard error.
- **Read/write asymmetry (§5.4) is implemented as documented,** and `load_for_mutation` correctly distinguishes `NotFound` (fresh tree) from every other error (refuse) — so a present-but-unreadable file is not clobbered.
- **The write path is sound:** temp file in the same directory, `create_new` + `mode(0o600)` before any bytes land, `sync_all`, rename, then a best-effort parent-directory fsync. The `create_new` choice (vs. `create`+`truncate`) is the correct one and the comment explains why.
- **Selection is total.** Every branch either selects a context, errors, or explicitly selects nothing — there is no path where a typo'd name silently degrades to the default.
- **Tests are proportionate to the risk.** Asserting exit 1 vs. the connect code 2 is the right way to prove no request was issued, and the byte-for-byte survival assertion on a refused mutation is exactly the test that change needed.

### Findings

**Nothing here blocks the merge.** Ten non-blocking items are inline. The two I would prioritise are both instances of the same failure class this PR exists to close — a config the CLI cannot fully understand quietly resolving to the local default:

1. `load` swallows every read error, including `EACCES`, so an existing-but-unreadable config falls through to `http://127.0.0.1:8080` with no warning (pre-existing, but this PR is what makes it consequential).
2. A mistyped per-context key (`Mode:` for `mode:`) is absorbed into `extra`, the context resolves as `mode: server`, and `$SKARDI_SERVER_URL` becomes able to redirect the workspace PAT again.

The rest are a temp-file sweep that can break a concurrent writer, a redaction allowlist that misses `extra`, a PAT arriving on `argv`, `get-contexts`/`current-context` not using resolution's own selection logic, and four small KISS/DRY/message nits.

### Three things too small or too diffuse to pin to a line

- **`LegacySpec` has no `extra`,** so unknown keys inside a `spec:` block are dropped rather than preserved — and `use-context` (which does *not* promote) will re-serialize `spec:` from the two modeled fields alone. Low risk, since the documented legacy shape only ever had `server`/`token`, but it is the one place the "unknown keys survive an older CLI" promise does not apply.
- **`set-context --mode cloud` requires `--server` and `--workspace` but not `--token`,** so a tokenless cloud context is still creatable. That is not a security problem (worst case is a 401 rather than a misdirected credential), but if the "writing a dud" rationale in the guard applies to the other two fields, it is worth saying out loud why it does not apply here — presumably because M2's `login` writes the token in a second step.
- **`docs/jobs.md:226` and `:268` still describe the pre-contexts resolution** (`$SKARDI_SERVER_URL` / `--token` / `~/.skardi/config.yaml`). Still true for `mode: server`, but now incomplete; a pointer to `docs/cli.md#connecting` would keep it from drifting further.

crates/cli/src/config.rs
bakey: Fixed in 6fd4ea3, and reproduced against the real binary first — `chmod 000` on the config, then `skardi query`, and it silently POSTed to `http://127.0.0.1:8080/query` with nothing on stderr. You're right that this is the same shape as the guards this PR added, reached by a route nobody was watching.

Implemented as you suggested: `NotFound` returns silently (absence is the common case and says nothing), everything else warns and still resolves to `None`, so the read path stays non-fatal per §5.4. It now prints:

```
warning: ignoring unreadable config file /tmp/t1/.skardi/config.yaml: Permission denied (os error 13)
```

Two notes on scope. I left it a warning rather than a hard error because §5.4's rationale — `--server` must keep working while the file is repaired — applies just as much to an unreadable file as to an unparsable one. And the test uses a directory as the unreadable path rather than `chmod 000`, so it doesn't depend on the test user not being root.

crates/cli/src/config.rs
bakey: Confirmed and fixed in 6fd4ea3. This was the sharper of the two, and the reproduction is stark: `Mode: cloud` with `SKARDI_SERVER_URL=http://127.0.0.1:9` sent the request to `:9`, exactly as you traced. One capital letter turns off the whole cloud-authoritative rule.

Your observation that the unsafe direction is specifically the *key* side is the part that made this actionable — a bad *value* (`mode: Cloud`) fails closed as an unknown-variant parse error, so only the key needs guarding. Implemented close to your sketch, with a `KNOWN_CONTEXT_KEYS` list checked case-insensitively at read time:

```
warning: context 'c' has key 'Mode', which differs only in case from 'mode'; it is being preserved but NOT read
```

Kept as a warning rather than a refusal so `extra`'s forward-compatibility promise still holds — a newer CLI's genuinely new key must not become an error in an older one. The warning lives in the new `warn_about_file`, which runs once per read (see the `effective_contexts` thread), so it cannot double-print.

crates/cli/src/config.rs
bakey: You're right, and this one is mine from the round before: the sweep I added to clean up crash leftovers had no liveness guard, so it could delete a live writer's temp file between its create and its rename. Your trace is exactly the failure — A's write is still lost, but now with `replace … atomically: No such file or directory`, which points at entirely the wrong thing, and the cleanup `remove_file` on the next line fails silently too.

Fixed in 6fd4ea3 with the mtime guard, essentially as you wrote it, behind a named `STALE_TEMP_AGE` of 60s. I chose age over "sweep only this pid's leftovers" because the whole reason the sweep exists is to remove *other* runs' abandoned files — a crashed process's temp file holds a full copy of every token, and pid-scoping would leave precisely those behind.

Tested both directions: a backdated stale file is removed, and a freshly-written one with a temp-shaped name survives the save. That second assertion is the one that would have caught this.

crates/cli/src/commands/config.rs
bakey: Good catch, and the framing is the right one — the fix belongs with the mechanism that creates the exposure, not with whichever field eventually trips it. Fixed in 6fd4ea3.

`redact_sensitive_extras` runs over both `extra` maps under the same `if !show_tokens`, matching key names against `token`/`secret`/`password`/`credential` (I added the fourth). Name-based rather than value-based by necessity: `extra` exists precisely for fields this binary doesn't understand, so the name is all there is to go on.

A test pins all three behaviours that matter: `refresh-token:` and `client-secret:` are redacted, a `note:` key still renders (so `view` stays useful rather than becoming a wall of `(redacted)`), and `--show-tokens` still prints everything.

The `filter` nit is fixed too — you're right that `map` already leaves `None` alone.

crates/cli/src/commands/config.rs
bakey: All three divergences confirmed and fixed in 6fd4ea3, and the first one is the most embarrassing: a legacy `spec:` install — the most common existing case — showed no `*` at all and `current-context` exited 1, while every real command was happily using `default`.

Done the way you suggested, by routing both commands through resolution's own chain rather than the raw field. I extracted `select_context(contexts, current, flag, env)` from `resolve_from` and both surfaces now call it, so trim, the lone-context step, and `--context`/`$SKARDI_CONTEXT` all come for free and cannot diverge again. Verified live:

```
$ HOME=/tmp/t5 skardi config get-contexts      # legacy spec: install
*  default                  server   -                    http://legacy:8080
$ HOME=/tmp/t5 skardi config current-context
default
$ SKARDI_CONTEXT=b skardi config current-context
b
```

One behaviour change worth flagging, because a test had to be updated for it: after `delete-context` removes the pointer's target, the single remaining context is now selected, so `current-context` reports it instead of erroring. That's the correct answer — it's what the next command will use — but it is a visible change from "reports the raw field".

`current_context`'s failure message became "no context is selected" rather than "no current context is set", since with the lone-context step the pointer is no longer the only thing that selects.

crates/cli/src/commands/config.rs
bakey: Agreed, and the asymmetry you point at is real: this PR goes to some length to keep a token out of a world-readable file even briefly, and then the docs I wrote in the same PR teach `--token` as the normal way to put one in.

Added `--token-stdin` in 6fd4ea3, in the `docker login --password-stdin` shape:

```bash
skardi config set-context acme/prod --mode cloud --workspace acme-prod \
  --server https://gateway.skardi.ai --token-stdin < token.txt
```

`conflicts_with` on the pair so the two can't both be given; the first line of stdin is used, trailing newline stripped (so `echo … | skardi …` works); an empty read is an error rather than a context written with an empty credential. `--token` still works for scripts, and `docs/cli.md` now leads with stdin and says why in one line.

Reading happens in the dispatch layer rather than inside `set_context`, so that function stays a pure file edit and remains unit-testable without stdin.

I did not add `SKARDI_PAT`, deliberately: M2 introduces `login` as the primary path, and adding a third credential source in M1 would mean specifying its precedence against `--token`/context/`SKARDI_API_TOKEN` with no design text to point at. Worth settling in M2 rather than guessing here.

crates/cli/src/config.rs
bakey: You're right, and the evidence is in this PR's own history: the same bug twice, patched at two call sites. Fixed structurally in 6fd4ea3.

`effective_contexts` is now pure and idempotent, and the both-shapes warning moved into a `warn_about_file` called from `load` and `load_for_mutation` — where the file is already in hand and the malformed-file warning already lives. That makes "warns once" a property of reading the file rather than a rule callers have to remember, which is what you were getting at. The new key-typo warning from the `extra` thread lives there too, for the same reason.

I took the `eprintln!`-in-load option over returning the condition, agreeing that the latter is more machinery than this needs — every caller would then have to decide, which is the same distributed-discipline problem in a different shape.

crates/cli/src/commands/config.rs
bakey: Fixed in 6fd4ea3, as a method on `ContextsFile` exactly as you suggested:

```rust
pub fn promote_legacy_spec(&mut self) {
    if self.contexts.is_empty() {
        self.contexts = self.effective_contexts();
        self.spec = None;
    }
}
```

Both call sites are one line now. The ordering you flagged — promote first, *then* clear `spec` — is the reason this deserved to be one function rather than two copies of three lines, and the doc comment says so, since the failure mode if someone reverses it is a dropped credential rather than a compile error.

And you're right that M2's `login` needs exactly this before writing a context; it now gets it instead of adding a third copy.

crates/cli/src/config.rs
bakey: Fixed in 6fd4ea3 with the second variant. You're right that the advice is identical but the diagnosis isn't — "does not parse (Permission denied)" sends someone hunting for a YAML error that doesn't exist.

`UnreadableForMutation` now renders:

> refusing to modify {path}: it exists but cannot be read ({error}). Fix the permissions or move the file first — rewriting it would discard the credentials it may still hold

The test you pointed at is the one that proved the old message wrong, so it now asserts the new variant *and* that the message does **not** contain "does not parse" — otherwise a future refactor could quietly route it back through the parse path and the test would still pass.

crates/cli/src/config.rs
bakey: Fixed in 6fd4ea3. `create_dir_all` was indeed leaving `~/.skardi` at 0755 under a typical umask — verified, and it's now `drwx------`.

Used `DirBuilder::mode(0o700)` with `recursive(true)` under `#[cfg(unix)]`, as you suggested, with a `create_dir_all` fallback elsewhere. The `~/.ssh` precedent is cited in the comment.

On mirroring `warn_if_loose_permissions` for a pre-existing loose directory: I left that out, and the asymmetry is deliberate rather than an oversight. The file case rewrites as well as warns, because the file is ours to replace on every save; a pre-existing `~/.skardi` may hold anything (other tools' state, a user's own files), so silently tightening it would be reaching outside what this command owns. Warning without acting seemed like noise on a path that runs on every mutation. Happy to add the warning if you'd rather have it — the comment records the reasoning either way.

bakey:
Thanks — this was a careful read, and the two you prioritised were the right two: both were the failure class this PR exists to close, reached by routes the new guards didn't cover. I reproduced both against the real binary before touching anything, and both are fixed in 6fd4ea3 along with the other eight threads and the three diffuse notes.

**The unreadable config** silently POSTed to `127.0.0.1:8080` with nothing on stderr. `load` now separates `NotFound` (silent — absence says nothing) from everything else (warned, still non-fatal per §5.4), which is the distinction `load_for_mutation` already made.

**The `Mode:` typo** was the sharper one. With `SKARDI_SERVER_URL` set, the request went to the env-named host — one capital letter disables the entire cloud-authoritative rule. Your point that the unsafe direction is specifically the *key* side is what made it tractable: a bad value fails closed as a parse error, so only keys need guarding. Case-insensitive matches against the known keys now warn while `extra` still preserves them.

**The sweep hazard was mine from the previous round** — I added a garbage collector with no liveness guard, which could delete a live writer's temp file and make that process fail its rename with a message pointing somewhere else entirely. Now age-guarded at 60s, with a test asserting a *fresh* temp file survives; that assertion is the one that would have caught it.

**Selection had two implementations**, and the legacy `spec:` case was the embarrassing one — no `*`, `current-context` exiting 1, while every real command used `default`. Both surfaces now call one extracted `select_context` that `resolve_from` also uses. One visible consequence: after `delete-context` clears the pointer, a lone remaining context is now reported, because that is what the next command will use.

**Redaction was an allowlist over a deliberately open model**, so a future `refresh-token:` would print in full two lines under a `(redacted)`. Now name-matched wherever the key sits.

Also in: `--token-stdin` (this PR made `set-context` the first command that *persists* an argv credential, and my own docs taught `--token` as the default); `effective_contexts` made pure with the warning moved to read-time, so "warns once" is structural rather than a rule two call sites had already broken; `promote_legacy_spec` extracted; `UnreadableForMutation` so the message stops claiming a parse error; `~/.skardi` created 0700; `LegacySpec.extra`; and `docs/jobs.md` pointing at `cli.md#connecting`.

**Three places I made a call rather than complying, each explained in-thread:** age-guarding the sweep instead of pid-scoping it (pid-scoping leaves exactly the crashed-run token copies the sweep exists for); no `SKARDI_PAT` env var yet (its precedence against the other three sources needs M2's design text, not a guess); and no warning for a pre-existing loose `~/.skardi` (the file is ours to rewrite, a directory that may hold other tools' state is not — happy to add it if you disagree).

100 unit + 7 integration + 4 e2e (2 ignored) green, clippy and fmt clean. Every fix verified against the real binary, not just the suite.

PR #225, opened 2026-08-24
feat(cli): skardi login/logout, cloud capability gating, and the workspace selector (contexts/login M2)

bakey:
Milestone 2 of `docs/superpowers/specs/2026-08-20-skardi-cli-contexts-and-login-design.md`
(the design lives in skardi-cloud; §11 assigns M1 and M2 to this repo).
Stacked on #224, now rebased onto `main`.

M2's scope, per §11: `login`/`logout` including the §6.5 saga, §6.2's URL
precedence, §6.3's `--identity` dev path, the `Skardi-Workspace` header,
capability gating and §8's error messages.

### `client.rs` — the workspace selector, and a header bug it exposed

A `mode: cloud` context now sends `Skardi-Workspace: <slug>` on every request
(§7.3), keyed off `mode` rather than off the field being populated, so a stray
`workspace:` on a server-mode context cannot start selecting anything. It sits
outside the reserved `x-skardi-*` prefix deliberately: #288's invariant strips
client-supplied headers in that namespace before the gateway forwards
upstream, so a selector wearing it would be silently dropped.

Writing it surfaced a real defect in the existing client.
`RequestBuilder::header` **appends**, so a per-call header of the same name
produced *two* values on the wire and left the choice to the peer's parser.
For an authorization-relevant selector that is a spoofing surface, and the
same was already true of `Authorization`. Both names are now `insert`ed on a
built `Request`, and both values are built once in `ApiClient::new` — so an
unsendable token is a config error with one clear message instead of a failure
at every request, and the bearer is marked sensitive as `bearer_auth` did.
Pinned by a test that passes a forged pair through `post_with_headers` and
asserts exactly one value of each, with the client's own winning.

### `cloud.rs` (new) — §8, as pure functions

Three cloud-only behaviours, all pre/post-processing around an unchanged
request path (`mode` never selects a URL):

- **Capability gating.** The gateway mounts `query` and `schema`; `run`,
  `pipeline`, `job`, and `health` refuse before a request is built, because a
  404 from a route that was never served reads as a broken server rather than
  as a context that cannot do that. The "Available:" list is derived from the
  same table the gate reads, so the two cannot drift.
- **Credential expiry**, checked against `token-expires-at` with no round
  trip. An unparsable stamp is deliberately not fatal: it is an optional field
  in a hand-editable file, and refusing every command over a malformed
  annotation is worse than one authoritative request. Comparison is by
  instant, not by text, so a legal non-UTC offset resolves correctly.
- **Error translation.** 401 is *replaced* — the transport message's advice
  (set `SKARDI_API_TOKEN`) is refused by resolution for a cloud context and
  actively misleads. A workspace-boundary 403 is *annotated*, keeping the
  gateway's own sentence, which is what distinguishes "not a member" from "no
  matching workspace"; `identity_required` and `identity_unresolved` also
  mention "workspace" and are excluded by `error_type`, so a message-only rule
  cannot misclassify them. A 503 is the schema-read limiter only **with**
  `Retry-After` and only on `schema` — `ApiError::Http` now carries
  `retry_after` for exactly that distinction.

### `login/` (new) — §6

- `pkce.rs` — S256 and the `state` nonce, both straight from the OS CSPRNG.
- `loopback.rs` — a single-use listener on `127.0.0.1:0`, never `0.0.0.0` and
  never a fixed port. Single-use is about the *callback*: a favicon or a port
  probe gets a 404 and the wait continues (bounded), so it cannot consume the
  one authorization response. The page never echoes a query value.
- `oauth.rs` — authorization URL, `state` check *before* the code is read,
  code exchange. The ID token is held in memory only; no refresh token is
  requested and nothing the provider returns is written to disk (§9.1). Both
  endpoint URLs and the browser opener are parameters with production
  defaults, which is what makes the whole path testable.
- `control_plane.rs` — `GET /v1/me/workspaces`, `POST /v1/me/tokens`,
  `DELETE /v1/me/tokens/{id}`. Deliberately **not** `ApiClient`: the control
  plane's error envelope is nested (`{"error": {"code", "message", …}}`) while
  the engine's is flat, the credential is an ID token rather than the PAT, and
  `org_ambiguous` carries an org list one client cannot read without guessing.
- `mod.rs` — the flow, and the saga.

### `commands/login.rs`, `commands/logout.rs` (new)

`logout` is deliberately two things. Plain, it is a local edit that says the
PAT stays **valid** until it expires; a token cannot manage tokens, so
`--revoke` re-authenticates. The local edit happens first, so an unreachable
control plane still leaves the credential off this disk. `--all` covers cloud
contexts only: a server-mode token was configured by hand and there is no
`login` to put it back. A cleared context with no `token-id` is *named*, not
skipped — clearing it quietly would read as a revocation that happened.

## Design decisions worth reviewing

**Minting is a saga, not a loop (§6.5).** Each mint commits independently at
the control plane, so every `{token_id, workspace}` is retained in memory,
contexts are written only after all selected mints succeed, and any failure —
a later mint, a failed verification probe, the config write — revokes
everything this run created before reporting the original cause. A rollback
that cannot complete exits non-zero and prints the surviving ids, because a
live credential nobody knows about is worse than a loud failure. Control-plane
calls carry a 30s timeout for the rollback's sake specifically: a revoke that
hangs forever is a silent terminal in front of an unreported credential.

**Verification is the real path (§6.1 step 6).** The probe is a `select 1`
through the same `ApiClient` a query will use, so it carries the same
`Skardi-Workspace` header and the same bearer. A credential that cannot answer
is revoked and its context is never written — a login that reports success
while its first query would 403 is the failure this step exists to remove.

**`--identity` is loopback-only (§6.3),** not loopback-or-private: RFC1918 is
exactly where a shared internal staging cluster lives, and a `dev:` bearer
there is impersonation of any user. `--i-know-this-is-dev-auth` is the visible
opt-out, and the guard is one function shared with `logout --revoke` so the
two cannot drift. Every dev-path run warns, naming what it authenticated
against.

**Wire shapes are pinned against skardi-global's handlers, not prose:**
`{"workspaces": [MembershipView…]}` and `201 {"token": …, "pat": {"token_id":
…}}`. `gateway_url` is `Option` because §7.1 is M4's work — a control plane
that predates it omits the key, and §6.2's precedence then decides.

**The global flags mean something different here.** `--server` under `login`
is the *gateway* URL written into contexts, `--context` is the *name* to
write, and `--token` means nothing at all. Both subcommands carry long help
saying which, and the flags they cannot honour are refused by name rather than
silently dropped.

## Two deliberate deviations from the design's text

Both because the text would ship as false:

1. **§6.1 step 1 ends its chain in a "built-in default" control plane, and the
   browser path implies a built-in OAuth client id. Neither value exists.**
   The only committed control-plane URL is a commented `https://global.skardi.ai`
   example and localhost; the client id is per deployment (skardi-cloud feeds
   it from `$SKARDI_GOOGLE_CLIENT_ID`, and §7.2 makes the audience a set
   precisely so the CLI can have its own). A hostname that answers nothing
   would fail at DNS without mentioning the real inputs, so both chains end
   the way §6.2's deliberately does: a typed error naming
   `--control-plane`/`$SKARDI_CONTROL_PLANE_URL`/`control-plane:` and
   `--client-id`/`$SKARDI_OAUTH_CLIENT_ID`. One constant each replaces this
   when the hosted values exist.
2. **Wording.** The refusal names `'job'`, the subcommand this CLI has (§8's
   example wrote `'jobs'`), and the pre-flight expiry message states what was
   observed ("expired at \<stamp\>") instead of reusing the 401's "was
   rejected" — nothing rejected anything, and no request was sent.

## Verification

```
cargo test -p skardi-cli          # 211 passing, 2 ignored (need a live server)
cargo clippy -p skardi-cli --all-targets   # clean
cargo fmt -p skardi-cli --check             # clean
```

211 passing on this branch against 109 on `main` — **102 new tests** (94 unit,
8 integration). No new crate versions enter `Cargo.lock`: `chrono`, `sha2`,
`base64`, and `getrandom` are all already in the tree at the versions sibling
crates pin.

Coverage on the files this PR adds or touches, measured with the same tool CI
uses (`cargo llvm-cov -p skardi-cli --all-targets`): 96.87% lines overall
(from 94.80%), with `login/mod.rs` at 97.68%, `client.rs` 98.20%,
`control_plane.rs` 98.87%, `loopback.rs` 99.55%, `cloud.rs` 99.17%,
`commands/login.rs` 99.20%, `commands/logout.rs` 96.79%.

Three things stay uncovered because they are uncoverable, not untested:
`open_in_browser`'s `Command` invocation (a test would launch a real browser —
so the per-platform argv is split into `browser_command()` and tested
separately), `dispatch`'s per-command arms (they need a resolved config and a
live server; the integration tests do exercise them, but `cargo llvm-cov
nextest` does not instrument the spawned binary), and `logout --revoke`'s
browser re-authentication.

What the tests actually pin, beyond counts:

- **The saga was mutation-checked.** Disabling the rollback loop
  (`.iter().rev().take(0)`) failed exactly the four tests that assert it —
  second-mint failure revokes the first, a config-write failure revokes all, a
  failed verification revokes and writes nothing, and a failed rollback prints
  the surviving ids. A guard that cannot be shown to bite is a guard that
  quietly stops guarding.
- **The browser path is covered end to end**, with a closure standing in for
  the user agent: it reads the authorization URL as a browser would, follows
  the redirect with the `state` it was given, and the flow completes against a
  stub token endpoint. That opener is a parameter *because* of this test.
- **RFC 7636's own appendix B vector** pins S256 to the spec rather than to
  itself. A `state` mismatch is refused without the planted code appearing in
  the error. A timeout releases the port, proven by rebinding it.
- **§10's `login` list runs in-process** against a mock control plane and
  gateway, which is what makes "the first PAT was revoked" an assertion rather
  than an inference: minted scope and name, per-org gateway URLs, the full
  §6.2 precedence including the no-source error, non-`active` skips,
  `org_ambiguous`, re-login replacement, `--keep-old-token`.
- **Two integration tests assert absence** — a gated command and an expired
  credential record *zero* requests at the server, which no unit test can show.
- **One integration test drives the real binary** through login → query →
  gated refusal → logout, proving the context `login` writes is the one
  resolution then picks and sends the selector for, that the config lands at
  `0600`, and that the raw PAT appears on neither stream.

### Not machine-verified

§11's acceptance line for M2 is "smoke-tested on the preview compose via
`--identity`". The end-to-end test above is that same flow against a mock
control plane and gateway; running it against the live preview compose needs
an environment this branch cannot reach, and is the one check still owed
before merge.

### DIFF
--- a/crates/cli/src/cloud.rs
+++ b/crates/cli/src/cloud.rs
@@ -180,8 +180,9 @@ pub fn diagnose(
         // and `Retry-After` is what separates it from a gateway that is simply
         // down. Scoped to `schema` because that is the only route that
         // bounds concurrency this way.
-        503 if capability == Capability::Schema && retry_after.is_some() => {
-            let seconds = retry_after.unwrap_or_default();
+        503 if capability == Capability::Schema
+            && let Some(seconds) = *retry_after =>
+        {
             anyhow!("the gateway is at its schema-read limit; retry in {seconds}s")
         }
         _ => err,
--- a/crates/cli/src/cloud.rs
+++ b/crates/cli/src/cloud.rs
@@ -1,0 +1,475 @@
+//! What a `mode: cloud` context changes about a command's outcome (§8).
+//!
+//! Three things, all of them pre- or post-processing around an unchanged
+//! request path — `mode` never selects a URL, so nothing here touches routing:
+//!
+//! 1. **Capability gating.** A skardi-cloud gateway mounts `query` and
+//!    `schema`; it has no jobs, pipelines, or health surface. Those commands
+//!    fail before a request is built, because a 404 from a gateway that never
+//!    served the route reads as "the server is broken" rather than "this
+//!    context cannot do that".
+//! 2. **Credential expiry.** A context carries `token-expires-at`, so an
+//!    expired PAT is knowable without spending a round trip.
+//! 3. **Error translation.** The gateway's typed refusals name the *deployment*
+//!    ("the token's org has no workspace matching the one this gateway
+//!    serves"); the operator thinks in contexts. Translation happens here, at
+//!    the one place that knows both.
+//!
+//! Everything is a pure function of `(error, capability, config)` so the
+//! matrix is unit-testable without a server.
+
+use crate::client::ApiError;
+use crate::config::ClientConfig;
+use anyhow::anyhow;
+use chrono::{DateTime, Utc};
+
+/// One remote subcommand, for gating and for the error mapping that is
+/// specific to one route (§7.4.2's schema-read limit).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Capability {
+    Query,
+    Schema,
+    Run,
+    Pipeline,
+    Job,
+    Health,
+}
+
+impl Capability {
+    /// The subcommand name as the user typed it.
+    ///
+    /// §8's example message says `'jobs'`; this CLI's subcommand is `job`
+    /// (`skardi job list`), and naming a command that does not exist would be
+    /// a worse message than the design's illustration.
+    pub const fn command(self) -> &'static str {
+        match self {
+            Capability::Query => "query",
+            Capability::Schema => "schema",
+            Capability::Run => "run",
+            Capability::Pipeline => "pipeline",
+            Capability::Job => "job",
+            Capability::Health => "health",
+        }
+    }
+
+    /// Whether a skardi-cloud gateway serves this command.
+    ///
+    /// The gateway's route table is `POST /query` and `GET /data_source`
+    /// (§7.4); everything else is an engine-local surface that only a
+    /// `mode: server` context reaches.
+    pub const fn served_by_gateway(self) -> bool {
+        matches!(self, Capability::Query | Capability::Schema)
+    }
+}
+
+/// Every capability, so the "Available:" list is DERIVED from
+/// [`Capability::served_by_gateway`] rather than restated next to it — a
+/// hand-written list is the kind that survives the table changing under it.
+const ALL_CAPABILITIES: [Capability; 6] = [
+    Capability::Query,
+    Capability::Schema,
+    Capability::Run,
+    Capability::Pipeline,
+    Capability::Job,
+    Capability::Health,
+];
+
+/// The comma-separated commands a cloud context can run.
+fn cloud_capability_list() -> String {
+    ALL_CAPABILITIES
+        .iter()
+        .filter(|c| c.served_by_gateway())
+        .map(|c| c.command())
+        .collect::<Vec<_>>()
+        .join(", ")
+}
+
+/// Refuse a command the selected cloud context cannot serve, before any
+/// request is built (§8, row 1). Exits `1` through `main`'s mapping — `2` is
+/// reserved for "cannot reach the server", and this never reached it.
+pub fn ensure_available(capability: Capability, config: &ClientConfig) -> anyhow::Result<()> {
+    let Some(context) = config.cloud_context() else {
+        return Ok(());
+    };
+    if capability.served_by_gateway() {
+        return Ok(());
+    }
+    Err(anyhow!(
+        "'{}' is not available in a cloud context ({}). Available: {}.",
+        capability.command(),
+        context.name,
+        cloud_capability_list()
+    ))
+}
+
+/// Refuse when the stored PAT's `token-expires-at` has already passed (§8,
+/// row 3), so an expired credential costs no round trip.
+///
+/// `now` is a parameter rather than a `Utc::now()` call so the check stays a
+/// pure function, the way resolution is.
+pub fn ensure_credential_fresh(config: &ClientConfig, now: DateTime<Utc>) -> anyhow::Result<()> {
+    let Some(context) = config.cloud_context() else {
+        return Ok(());
+    };
+    let Some(raw) = context.token_expires_at.as_deref() else {
+        return Ok(());
+    };
+    // An unparsable stamp is NOT fatal. It is an optional field in a file a
+    // human may have hand-edited, and refusing every command over a malformed
+    // annotation would be a worse outcome than sending one request the gateway
+    // answers authoritatively.
+    let Ok(expires_at) = DateTime::parse_from_rfc3339(raw) else {
+        return Ok(());
+    };
+    if expires_at.with_timezone(&Utc) > now {
+        return Ok(());
+    }
+    // §8 prescribes the 401 message verbatim here ("was rejected"). Said
+    // pre-flight that is untrue — nothing rejected anything, no request was
+    // sent — so this keeps the design's shape (context named, `skardi login`
+    // as the action) and states what was actually observed.
+    Err(anyhow!(
+        "credential for context '{}' expired at {raw}. Run 'skardi login'.",
+        context.name
+    ))
+}
+
+/// Rewrite a gateway failure into the context's vocabulary (§8, rows 2, 4, 5).
+///
+/// Non-cloud contexts and unrecognised failures pass through untouched: this
+/// only ever replaces a message it can improve. `ApiError::Connect` is never
+/// touched either — its own message already names the URL and the flags, and
+/// `main` reads the same error back out to choose exit code `2`.
+pub fn diagnose(
+    err: anyhow::Error,
+    capability: Capability,
+    config: &ClientConfig,
+) -> anyhow::Error {
+    let Some(context) = config.cloud_context() else {
+        return err;
+    };
+    let Some(ApiError::Http {
+        status,
+        error_type,
+        message,
+        retry_after,
+    }) = err.downcast_ref::<ApiError>()
+    else {
+        return err;
+    };
+
+    match *status {
+        // REPLACED, not annotated: the transport-level 401 message tells the
+        // caller to set SKARDI_API_TOKEN or edit `token` in the config, and
+        // for a cloud context that advice is actively wrong — resolution
+        // refuses the env var (§5.1) and the PAT is `login`'s to write.
+        401 => anyhow!(
+            "credential for context '{}' was rejected — it may be expired or revoked. Run 'skardi login'.",
+            context.name
+        ),
+        // ANNOTATED, not replaced: the gateway's own sentence says which
+        // boundary was crossed (not a member / no matching workspace), which
+        // is the detail that distinguishes the two, and it never leaks a slug
+        // the caller didn't already send.
+        403 if names_workspace(error_type.as_deref(), message) => err.context(format!(
+            "context '{}' names workspace '{}', which this gateway does not serve",
+            context.name,
+            context.workspace.as_deref().unwrap_or("(unset)"),
+        )),
+        // §7.4.2's schema-read limiter is a LOAD signal, not a schema error,
+        // and `Retry-After` is what separates it from a gateway that is simply
+        // down. Scoped to `schema` because that is the only route that
+        // bounds concurrency this way.
+        503 if capability == Capability::Schema && retry_after.is_some() => {
+            let seconds = retry_after.unwrap_or_default();
+            anyhow!("the gateway is at its schema-read limit; retry in {seconds}s")
+        }
+        _ => err,
+    }
+}
+
+/// Whether a `403` is the workspace boundary rather than an authorization
+/// failure inside the workspace.
+///
+/// Today's gateway answers every `PatAuthError::NotAuthorized` with
+/// `error_type: "forbidden"` and a deliberately caller-safe message that names
+/// no slug ("the token's org has no workspace matching the one this gateway
+/// serves"), so the word in the message is the only discriminator available.
+/// The typed tokens are matched too: §7.3's org-admission path maps
+/// `Directory.Resolve`'s `workspace_not_found` to a typed 403, and accepting
+/// them now means the CLI reads that release without needing one of its own.
+///
+/// `identity_required` and `identity_unresolved` also mention "workspace" —
+/// they are excluded because they carry their own `error_type`, and they are
+/// genuinely a different failure (a credential with no envelope, not a slug
+/// the deployment refuses).
+fn names_workspace(error_type: Option<&str>, message: &str) -> bool {
+    match error_type {
+        Some("workspace_not_found" | "workspace_required") => true,
+        Some("forbidden") => message.to_ascii_lowercase().contains("workspace"),
+        _ => false,
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{
+        ALL_CAPABILITIES, Capability, diagnose, ensure_available, ensure_credential_fresh,
+    };
+    use crate::client::ApiError;
+    use crate::config::{ClientConfig, ContextMode, SelectedContext};
+    use chrono::{DateTime, Utc};
+
+    fn at(rfc3339: &str) -> DateTime<Utc> {
+        DateTime::parse_from_rfc3339(rfc3339)
+            .unwrap()
+            .with_timezone(&Utc)
+    }
+
+    fn config(mode: Option<ContextMode>, expires: Option<&str>) -> ClientConfig {
+        ClientConfig {
+            server: "https://gw.example".to_string(),
+            token: Some("pat".to_string()),
+            context: mode.map(|mode| SelectedContext {
+                name: "acme/prod".to_string(),
+                mode,
+                workspace: Some("acme-prod".to_string()),
+                token_expires_at: expires.map(str::to_string),
+            }),
+        }
+    }
+
+    fn http(
+        status: u16,
+        error_type: Option<&str>,
+        message: &str,
+        retry_after: Option<u64>,
+    ) -> anyhow::Error {
+        anyhow::Error::new(ApiError::Http {
+            status,
+            error_type: error_type.map(str::to_string),
+            message: message.to_string(),
+            retry_after,
+        })
+    }
+
+    /// §10's "table test over (subcommand, mode)". The mode column is what
+    /// decides, so both non-cloud shapes (server context, no context at all)
+    /// are exercised for every capability.
+    #[test]
+    fn gating_matrix_refuses_exactly_the_engine_local_commands() {
+        for capability in ALL_CAPABILITIES {
+            let cloud = ensure_available(capability, &config(Some(ContextMode::Cloud), None));
+            assert_eq!(
+                cloud.is_err(),
+                !capability.served_by_gateway(),
+                "cloud gating disagreed with served_by_gateway for '{}'",
+                capability.command()
+            );
+            for non_cloud in [Some(ContextMode::Server), None] {
+                assert!(
+                    ensure_available(capability, &config(non_cloud, None)).is_ok(),
+                    "'{}' must stay available without a cloud context",
+                    capability.command()
+                );
+            }
+        }
+    }
+
+    #[test]
+    fn refusal_names_the_command_the_context_and_what_is_left() {
+        let err = ensure_available(Capability::Job, &config(Some(ContextMode::Cloud), None))
+            .unwrap_err()
+            .to_string();
+        assert_eq!(
+            err,
+            "'job' is not available in a cloud context (acme/prod). Available: query, schema."
+        );
+    }
+
+    #[test]
+    fn expiry_is_checked_only_for_a_cloud_context_and_only_when_it_has_passed() {
+        let now = at("2026-06-01T12:00:00Z");
+
+        let expired = ensure_credential_fresh(
+            &config(Some(ContextMode::Cloud), Some("2026-05-31T12:00:00Z")),
+            now,
+        );
+        assert_eq!(
+            expired.unwrap_err().to_string(),
+            "credential for context 'acme/prod' expired at 2026-05-31T12:00:00Z. Run 'skardi login'."
+        );
+
+        // Fresh, absent, unparsable, and non-cloud all pass without a request.
+        for (mode, expires) in [
+            (Some(ContextMode::Cloud), Some("2026-06-01T12:00:01Z")),
+            (Some(ContextMode::Cloud), None),
+            (Some(ContextMode::Cloud), Some("not-a-timestamp")),
+            (Some(ContextMode::Server), Some("2020-01-01T00:00:00Z")),
+            (None, None),
+        ] {
+            assert!(
+                ensure_credential_fresh(&config(mode, expires), now).is_ok(),
+                "unexpected refusal for mode={mode:?} expires={expires:?}"
+            );
+        }
+    }
+
+    /// A non-UTC offset is a legal RFC 3339 stamp and must be compared as an
+    /// instant, not as text: this one is 23:00Z, an hour AFTER `now`, while
+    /// its digits sort before it.
+    #[test]
+    fn expiry_compares_instants_not_strings() {
+        assert!(
+            ensure_credential_fresh(
+                &config(Some(ContextMode::Cloud), Some("2026-06-01T18:00:00-05:00")),
+                at("2026-06-01T22:00:00Z"),
+            )
+            .is_ok()
+        );
+    }
+
+    #[test]
+    fn unauthorized_is_replaced_with_the_context_and_the_login_action() {
+        let err = diagnose(
+            http(401, Some("unauthorized"), "invalid token", None),
+            Capability::Query,
+            &config(Some(ContextMode::Cloud), None),
+        );
+        assert_eq!(
+            format!("{err:#}"),
+            "credential for context 'acme/prod' was rejected — it may be expired or revoked. Run 'skardi login'."
+        );
+        // Replaced outright: the transport message's advice (set
+        // SKARDI_API_TOKEN) is refused by resolution for a cloud context.
+        assert!(!format!("{err:#}").contains("SKARDI_API_TOKEN"));
+    }
+
+    #[test]
+    fn workspace_boundary_403_names_the_context_and_keeps_the_gateway_reason() {
+        let err = diagnose(
+            http(
+                403,
+                Some("forbidden"),
+                "the token's org has no workspace matching the one this gateway serves",
+                None,
+            ),
+            Capability::Query,
+            &config(Some(ContextMode::Cloud), None),
+        );
+        let rendered = format!("{err:#}");
+        assert!(
+            rendered.starts_with(
+                "context 'acme/prod' names workspace 'acme-prod', which this gateway does not serve"
+            ),
+            "rendered: {rendered}"
+        );
+        assert!(
+            rendered.contains("no workspace matching"),
+            "rendered: {rendered}"
+        );
+        // Still an ApiError underneath, so `main` keeps choosing exit 1 (not
+        // 2) from the same downcast.
+        assert!(matches!(
+            err.downcast_ref::<ApiError>(),
+            Some(ApiError::Http { status: 403, .. })
+        ));
+    }
+
+    /// The 403s that are NOT the workspace boundary pass through untouched,
+    /// including `identity_required` — whose message mentions "workspace" and
+    /// would be misclassified by a message-only rule.
+    #[test]
+    fn other_403s_are_left_alone() {
+        for (error_type, message) in [
+            (
+                Some("identity_required"),
+                "this deployment routes queries per workspace; the presented credential resolved locally",
+            ),
+            (
+                Some("identity_unresolved"),
+                "resolved identity could not be turned into a request envelope",
+            ),
+            (
+                Some("forbidden"),
+                "the caller's role does not permit this operation",
+            ),
+            (None, "forbidden"),
+        ] {
+            let err = diagnose(
+                http(403, error_type, message, None),
+                Capability::Query,
+                &config(Some(ContextMode::Cloud), None),
+            );
+            assert!(
+                !format!("{err:#}").contains("does not serve"),
+                "{error_type:?} was rewritten as a workspace boundary miss"
+            );
+        }
+    }
+
+    #[test]
+    fn schema_read_limit_is_reported_as_load_only_with_retry_after() {
+        let cloud = config(Some(ContextMode::Cloud), None);
+        let limited = diagnose(
+            http(
+                503,
+                Some("schema_read_limit"),
+                "too many concurrent schema reads",
+                Some(7),
+            ),
+            Capability::Schema,
+            &cloud,
+        );
+        assert_eq!(
+            format!("{limited:#}"),
+            "the gateway is at its schema-read limit; retry in 7s"
+        );
+
+        // No Retry-After: a plain outage, reported as the gateway phrased it.
+        let outage = diagnose(
+            http(503, Some("backend_unavailable"), "engine unreachable", None),
+            Capability::Schema,
+            &cloud,
+        );
+        assert!(format!("{outage:#}").contains("engine unreachable"));
+
+        // Same 503 on another route is not the schema limiter.
+        let query = diagnose(
+            http(
+                503,
+                Some("backend_unavailable"),
+                "engine unreachable",
+                Some(7),
+            ),
+            Capability::Query,
+            &cloud,
+        );
+        assert!(format!("{query:#}").contains("engine unreachable"));
+    }
+
+    #[test]
+    fn non_cloud_contexts_and_transport_failures_pass_through() {
+        for mode in [Some(ContextMode::Server), None] {
+            let err = diagnose(
+                http(401, Some("unauthorized"), "invalid token", None),
+                Capability::Query,
+                &config(mode, None),
+            );
+            assert!(!format!("{err:#}").contains("skardi login"));
+        }
+
+        let connect = diagnose(
+            anyhow::Error::new(ApiError::Connect {
+                url: "https://gw.example/query".to_string(),
+                message: "connection refused".to_string(),
+            }),
+            Capability::Query,
+            &config(Some(ContextMode::Cloud), None),
+        );
+        assert!(matches!(
+            connect.downcast_ref::<ApiError>(),
+            Some(ApiError::Connect { .. })
+        ));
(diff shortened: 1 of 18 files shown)


yiyi-wang-skardi:
@yiyi-wang-skardi is now reviewing this PR using Claude Code — opus (reasoning effort max). Target HEAD commit: https://github.com/SkardiLabs/skardi/commit/d36161a8a478809ffb3aeb8dbdfcfc14e6fb6045 (`d36161a`).

crates/cli/src/login/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+/// hard error. Deliberately no built-in default and no fall-through to
+/// `http://127.0.0.1:8080` — writing a context that points at a local port
+/// would fail later and further from the cause.
+fn resolve_server(options: &LoginOptions, membership: &Membership) -> Result<String> {
+    let candidate = options
+        .server_override
+        .as_deref()
+        .or(options.env_gateway_url.as_deref())
+        .or(membership.gateway_url.as_deref())
+        .map(str::trim)
```
yiyi-wang-skardi: `Option::or` short-circuits on the first `Some`, and `.map(str::trim).filter(…)` then applies **only to the winner** — so a blank value doesn't fall through, it shadows everything after it.

With `SKARDI_GATEWAY_URL` exported empty (a shell profile, or `SKARDI_GATEWAY_URL=$UNSET skardi login`), `env_gateway_url` is `Some("")`. That beats `membership.gateway_url`, gets filtered to `None`, and the login dies with:

> the control plane did not say which gateway serves workspace 'acme-prod' — pass --server \<URL\>, set $SKARDI_GATEWAY_URL, or configure gateway_url for the org

which is untrue: the control plane did say, and the message points at the very variable that caused it. `--server ""` does the same. `resolve_control_plane` in `commands/login.rs:169` already handles this correctly — `.map(trim).find(non-empty)` over the flattened list — and `blank_values_are_skipped_not_honoured` pins it there. `the_gateway_url_comes_from_the_flag_then_the_env_then_the_membership` only ever passes populated values, so this side is uncovered.

Same shape as the other one:

```rust
let candidate = [
    options.server_override.as_deref(),
    options.env_gateway_url.as_deref(),
    membership.gateway_url.as_deref(),
]
.into_iter()
.flatten()
.map(str::trim)
.find(|url| !url.is_empty());
```

crates/cli/src/login/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+    if options.context_name.is_some() && selected.len() > 1 {
+        bail!(
+            "--context names ONE context, but {} workspaces were selected — drop it, or select a single workspace with --workspace",
+            selected.len()
+        );
+    }
+
+    // §6.5: mint, then verify, retaining every id. Any failure past the first
+    // mint rolls the whole run back before reporting.
+    let expires_at = (options.now + options.expires).to_rfc3339();
```
yiyi-wang-skardi: `DateTime<Utc> + chrono::Duration` is `checked_add_signed(rhs).expect("`DateTime + TimeDelta` overflowed")` — it panics rather than erroring, and `parse_expires` does not bound the input tightly enough to prevent it.

`parse_expires` rejects a duration only when `Duration::try_days` returns `None`, i.e. past roughly 1.07e11 days (`TimeDelta`'s own range). `DateTime`'s range is much smaller: `NaiveDate::MAX` is year 262143, about 9.5e7 days from now. Anything in between builds a perfectly valid `Duration` and then blows up here:

    skardi login --expires 100000000d

the process aborts with a chrono panic and a backtrace instead of the typed refusal every other bad `--expires` gets. It lands *after* `acquire_bearer` and `cp.memberships()`, so on the browser path the user completes a full sign-in first — nothing is minted, so no credential is stranded, but it's an ugly exit for what `parse_expires`'s last line already means to catch ("longer than any usable credential").

Cheapest fix is to make the addition fallible where it happens:

```rust
let expires_at = options
    .now
    .checked_add_signed(options.expires)
    .ok_or_else(|| anyhow!("--expires is longer than any usable credential"))?
    .to_rfc3339();
```

(or clamp in `parse_expires` — but the ceiling really is a function of `now`, so here reads truer.) Worth a case in `expires_accepts_days_and_hours_and_refuses_the_rest`, which currently stops at `-5d`.

crates/cli/src/login/control_plane.rs
```diff
@@ -0,0 +1,387 @@
...
+
+/// The HTTP client every control-plane conversation uses — `login`'s and
+/// `logout --revoke`'s alike.
+///
+/// One builder, so a caller cannot acquire an UNBOUNDED client by accident:
+/// `logout --revoke` did exactly that, and a stalled `DELETE` would have hung
+/// the command after the local credential was already gone. `timeout` is a
+/// parameter only so a test can prove the bound bites without waiting
+/// [`CONTROL_PLANE_TIMEOUT`] for it.
+pub fn client(timeout: Duration) -> Result<reqwest::Client> {
```
yiyi-wang-skardi, line 38: This client has a timeout, `no_proxy`, and a body cap — but no equivalent of `client.rs`'s cleartext check, and it is the leg that carries the most sensitive traffic in the whole flow: the OIDC **ID token** goes up on every call, and `POST /v1/me/tokens` returns the **raw PAT** in the response body.

`ApiClient::new` warns when a bearer would cross plain `http://` to a non-loopback host (`is_cleartext_remote`, `client.rs:357`). Nothing does that for the control plane, so:

    skardi login --control-plane http://global.example.com --client-id …

sends an identity assertion and receives a 90-day credential entirely in the clear, silently. That's reachable from a mistyped scheme, a URL pasted out of a staging README, or a `control-plane:` line that outlived a TLS migration — and once it's in the config file, every later `login`/`logout --revoke` reuses it with no flag typed.

`check_dev_identity`'s loopback guard doesn't cover this: it fires only for the `dev:` path, and the browser path (the one with a real ID token) has no scheme check at all.

This is the same predicate, so I'd lift `is_cleartext_remote` to a shared spot and warn once in `ControlPlane::new` (or in `resolve_control_plane`, where it can be said before the browser opens). Warn rather than refuse, matching `ApiClient`'s reasoning about TLS-terminating proxies.

crates/cli/src/login/loopback.rs
```diff
@@ -0,0 +1,479 @@
...
+        ignored: &AtomicUsize,
+    ) -> Result<Callback> {
+        for _ in 0..=MAX_UNRELATED_REQUESTS {
+            let (mut stream, _peer) = self
+                .listener
+                .accept()
+                .await
+                .context("accept the login redirect")?;
+
+            let Some(target) = read_request_target(&mut stream).await? else {
```
yiyi-wang-skardi: The accept loop is strictly sequential and `read_request_target` has no read deadline, so one connection that opens and **sends nothing** holds the loop here until the 120-second callback timeout — and the real redirect, already sitting completed in the accept backlog, is never read.

The module docs promise the opposite: "a favicon or a port probe gets a 404 and the wait continues". That holds for a probe that *closes* (`Ok(0)` → break → 400 → continue), which is what `malformed_and_non_get_requests_are_answered_and_ignored` exercises with `drop(TcpStream::connect(…))`. It does not hold for a probe that stays open.

Two ways in:

- **Adversarial.** Anything on the machine can reach a loopback port — that's the premise the `state` check is built on. A single `nc 127.0.0.1 <port>` that never types opens a 120-second denial of the sign-in. Cheaper than the `MAX_UNRELATED_REQUESTS` route, which needs nine round trips.
- **Accidental.** Browsers open speculative preconnect sockets to a navigation target and hold them idle for seconds. If the preconnect socket is accepted before the one carrying the GET, the callback waits behind it.

A short per-connection deadline fixes it without leaving the single-task design:

```rust
let target = match tokio::time::timeout(
    Duration::from_secs(5),
    read_request_target(&mut stream),
).await {
    Ok(Ok(Some(target))) => target,
    // A silent or malformed connection is noise, not this sign-in's response.
    _ => { respond(&mut stream, "400 Bad Request", "Malformed request.").await; continue }
};
```

Separately, and much smaller: `read_request_target` reads one byte per `await`, up to 8192 of them. `BufReader::read_until(b'\n', …)` under a `take(MAX_REQUEST_BYTES)` is the same bound in one call.

crates/cli/src/login/pkce.rs
```diff
@@ -0,0 +1,93 @@
...
+    use super::{Pkce, URL_SAFE_NO_PAD, random_urlsafe};
+    use base64::Engine as _;
+    use sha2::{Digest, Sha256};
+
+    /// RFC 7636 appendix B's worked example, so the transform is pinned to the
+    /// spec rather than to itself.
+    #[test]
+    fn s256_matches_rfc_7636_appendix_b() {
+        let verifier = "dBjftJeZ4CVP-mB92K27uhbUJU1p1r_wW1gFWFOEjXk";
+        let challenge = URL_SAFE_NO_PAD.encode(Sha256::digest(verifier.as_bytes()));
```
yiyi-wang-skardi: This test doesn't call anything in the module — it recomputes `URL_SAFE_NO_PAD.encode(Sha256::digest(…))` inline and compares that to the appendix-B value. So what it pins is `sha2` + `base64`, not `Pkce`. If `generate` switched to `STANDARD` base64, or to SHA-1, this test would still pass.

`generated_pairs_are_unreserved_fresh_and_self_consistent` has the same shape for the same reason: `assert_eq!(first.challenge, URL_SAFE_NO_PAD.encode(Sha256::digest(first.verifier.as_bytes())))` is true by construction whatever `generate` does, since both sides are the same expression. Between them there is no assertion that the *production* challenge is S256-of-the-verifier — which matters, because `code_challenge_method=S256` is asserted in `authorization_url`, so a mismatch here would be a login that fails only against a real provider.

The verifier is random, which is what blocks the direct test — so split the transform out and let the vector run through it:

```rust
impl Pkce {
    fn from_verifier(verifier: String) -> Pkce {
        let challenge = URL_SAFE_NO_PAD.encode(Sha256::digest(verifier.as_bytes()));
        Pkce { verifier, challenge }
    }
    pub fn generate() -> Result<Pkce> { Ok(Pkce::from_verifier(random_urlsafe()?)) }
}

// then:
assert_eq!(
    Pkce::from_verifier("dBjftJeZ4CVP-mB92K27uhbUJU1p1r_wW1gFWFOEjXk".into()).challenge,
    "E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM",
);
```

The browser-path test does prove challenge and verifier agree *with each other* end to end, but a stub token endpoint accepts any pair — only the vector can say which algorithm.

crates/cli/src/commands/logout.rs
```diff
@@ -0,0 +1,636 @@
...
+        // Cloud contexts only: a server-mode context's token is one the
+        // operator configured by hand, and `logout --all` sweeping it away
+        // would be a surprise with no login to undo it.
+        file.contexts
+            .iter()
+            .filter(|c| c.mode == ContextMode::Cloud)
+            .map(|c| c.name.clone())
+            .collect()
+    } else {
+        let selected = config::select_context(
```
yiyi-wang-skardi, line 165: `--all` filters to `ContextMode::Cloud` on a stated rationale — "a server-mode token is one the operator configured by hand, and there is no `login` to put it back" — but this branch applies no mode filter at all. Everything in that sentence is still true when the context was picked by `current-context` instead of by `--all`.

So with a server-mode context current (the default shape for anyone using this CLI against a plain skardi-server), a bare `skardi logout` deletes the hand-pasted token and prints:

> the credential is gone from this machine but stays VALID until it expires — run 'skardi logout --revoke' to revoke it at the control plane

Both halves mislead: there's no control plane in play, and `--revoke` can't help either — a `set-context` context has no `token-id`, so it takes the `bail!` at line 79. Meanwhile the only copy of the token is gone, and unlike a cloud context nothing can re-mint it.

The `logout_clears_the_selected_contexts_credential_and_keeps_the_context` test seeds `current-context: acme/prod` (cloud), so this path isn't covered.

I'd either apply the same `mode == Cloud` filter here and refuse a server-mode selection by name, or — if clearing an explicitly-named server context is meant to work — say what's about to be lost before writing. The silent case is a `current-context` nobody re-checked.

crates/cli/src/login/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+    // BOUNDED here rather than on `ApiClient` itself: the same client runs
+    // `skardi query`, where a statement legitimately takes minutes, so a
+    // client-wide total timeout would cap real work. A gateway that accepts
+    // the connection and then never answers must not hang the flow — this is
+    // the one point past which a PAT exists and no context does, so the saga
+    // has to be reachable (§6.5).
+    match tokio::time::timeout(timeout, probe_request).await {
+        Ok(Ok(_)) => Ok(()),
+        Ok(Err(err)) => Err(anyhow!(
+            "the new credential for workspace '{}' could not query {server}: {err} — the context was not written (pass --no-verify to skip this check)",
```
yiyi-wang-skardi: `{err}` here is the raw `ApiError`, and its `Display` for a 401 is:

> unauthorized — set SKARDI_API_TOKEN or 'token' in ~/.skardi/config.yaml

So a gateway that rejects the freshly minted PAT produces "the new credential for workspace 'acme-prod' could not query https://gw… : unauthorized — set SKARDI_API_TOKEN or 'token' in ~/.skardi/config.yaml — the context was not written". That advice is exactly what `cloud::diagnose` exists to replace for a cloud context (`cloud.rs:161`: resolution refuses the env var, and the PAT is `login`'s to write) — and it's *more* wrong here than at query time, since the credential was minted seconds ago and there is no context yet to point `skardi login` at.

`verify` already builds a full cloud `ClientConfig` for the probe, so the translation is available for free:

```rust
Ok(Err(err)) => {
    let err = crate::cloud::diagnose(err.into(), crate::cloud::Capability::Query, &probe);
    Err(anyhow!("the new credential for workspace '{}' could not query {server}: {err:#} — …", …))
}
```

`a_credential_that_cannot_query_is_revoked_and_no_context_is_written` uses a 403 and only asserts "could not query", so it wouldn't notice either way.

crates/cli/src/cloud.rs
```diff
@@ -0,0 +1,475 @@
...
+            "context '{}' names workspace '{}', which this gateway does not serve",
+            context.name,
+            context.workspace.as_deref().unwrap_or("(unset)"),
+        )),
+        // §7.4.2's schema-read limiter is a LOAD signal, not a schema error,
+        // and `Retry-After` is what separates it from a gateway that is simply
+        // down. Scoped to `schema` because that is the only route that
+        // bounds concurrency this way.
+        503 if capability == Capability::Schema && retry_after.is_some() => {
+            let seconds = retry_after.unwrap_or_default();
```
yiyi-wang-skardi: Nit: the guard on line 183 already established `retry_after.is_some()`, so this second look can't fail — but `unwrap_or_default()` says it can, and its fallback (`retry in 0s`) is a value that would be wrong if it ever ran.

The arm can carry the binding itself, which is one read instead of two and drops the unreachable default:

```rust
503 if capability == Capability::Schema
    && let Some(seconds) = *retry_after =>
{
    anyhow!("the gateway is at its schema-read limit; retry in {seconds}s")
}
```

(let-chains in match guards are stable on edition 2024, and `login/mod.rs:196` already uses one.)

crates/cli/src/commands/logout.rs
```diff
@@ -0,0 +1,636 @@
...
+
+/// Re-authenticate for `--revoke`, reusing `login`'s acquirers so the dev path
+/// and its loopback guard behave identically here.
+async fn authenticate(args: &LogoutArgs, path: &Path, env: &LogoutEnv<'_>) -> Result<ControlPlane> {
+    // The SAME bounded builder `login` uses. Rolling a local client here is
+    // how this path ended up unbounded, so there is now one place to get one.
+    let http = control_plane::client(env.request_timeout)?;
+    let control_plane =
+        super::login::control_plane_for_revoke(args, path, env.env_control_plane.as_deref())?;
+    let bearer = match &args.identity {
```
yiyi-wang-skardi: Nit, consistency: `login` resolves the dev bearer as flag-then-env (`commands/login.rs:140-142`, `--identity` > `$SKARDI_DEV_IDENTITY`), but `--revoke` reads only `args.identity`. `--client-id` right below *does* consult its env var, so the asymmetry is within this one function too.

Effect: in a dev shell where `SKARDI_DEV_IDENTITY=dev:alice` makes plain `skardi login` work, `skardi logout --revoke` falls into the browser branch and fails with "--revoke re-authenticates, which needs --client-id …" — after the token ids have already left the disk, so it takes the `unrevoked_warning` path for a reason that isn't really a missing input.

Since both commands go through `check_dev_identity` anyway, the loopback guard still applies either way:

```rust
let identity = args.identity.clone()
    .or_else(|| std::env::var("SKARDI_DEV_IDENTITY").ok())
    .filter(|id| !id.trim().is_empty());
```

Worth reading both env names off `LogoutEnv` while you're there — `SKARDI_OAUTH_CLIENT_ID` is fetched inline here while `SKARDI_CONTROL_PLANE_URL` is threaded through the struct, and the struct is what makes this testable.

yiyi-wang-skardi (approved):
## Audit of `skardi login`/`logout`, cloud gating, and the workspace selector

I read the whole 18-file diff against `main` (6019 added lines) — `client.rs`, `cloud.rs`, `login/{mod,oauth,loopback,pkce,control_plane}.rs`, `commands/{login,logout}.rs`, `main.rs`, both integration suites, and `docs/cli.md`. `cargo` is not available in this environment, so nothing below was executed; every finding is from reading the code.

**This is careful work.** The things I went looking for are already there: `state` is checked *before* the code is read and a mismatch is ignored rather than fatal; the loopback binds `127.0.0.1:0` and the page echoes no query value; the ID token is never written to disk and no refresh token is requested; `Minted`'s `Debug` is hand-written to redact the PAT and `LoginOptions`/`ApiClient` deliberately derive none; both response readers are byte-capped and chunk-streamed; `Authorization` and `Skardi-Workspace` are `insert`ed on a built `Request` so a per-call header cannot append a second value; the config write is `create_new` + `0600` + same-dir rename; and there is **no `unwrap`/`expect`/`panic!` anywhere in the non-test code** of the new modules. The saga's rollback is mutation-checked, and the two integration tests that assert *absence* (a gated command and an expired credential record zero requests) are the right shape for what they prove.

Nothing I found blocks the merge. Nine notes below, roughly in order of how much I'd want them addressed:

1. **A blank `--server`/`$SKARDI_GATEWAY_URL` shadows the control plane's `gateway_url`** (`login/mod.rs`) — `Option::or` picks the first `Some` and only *then* trims, so an exported-empty variable wins the chain and produces an error that says the control plane stayed silent when it did not. The sibling `resolve_control_plane` gets this right and has a test for exactly this case; `resolve_server` does not.
2. **`--expires` can panic** — the guard bounds the value by `TimeDelta`'s range, not `DateTime`'s, so a large-but-parseable duration reaches `now + expires` and `expect`s.
3. **No cleartext warning on the control-plane leg** — `ApiClient` warns when a bearer would cross plain `http://` to a non-loopback host; the control-plane client, which carries the ID token up and the raw PAT back, has no equivalent. `is_cleartext_remote` already exists.
4. **An idle connection stalls the loopback wait** — the accept loop is sequential and the request read has no deadline, so a socket that connects and sends nothing blocks the real redirect behind it until the 120s timeout.
5. The RFC 7636 appendix-B vector does not go through `Pkce`, so it pins `sha2`+`base64` rather than the production transform.
6. Plain `logout` clears a server-mode context's hand-configured token, which is the case `--all` deliberately protects.
7. The verify failure interpolates the raw `ApiError`, re-surfacing the `SKARDI_API_TOKEN` advice `cloud::diagnose` exists to replace.
8. / 9. Two small consistency nits (`unwrap_or_default()` under an `is_some()` guard; `logout --revoke` ignoring `$SKARDI_DEV_IDENTITY`).

One process note, not a finding: the PR body names the preview-compose `--identity` smoke test as "the one check still owed before merge". I could not run it here either — the `login_e2e` test is a faithful stand-in but not the same thing.

crates/cli/src/login/mod.rs
bakey: Fixed in e76056a — you are right, and the asymmetry with `resolve_control_plane` was the tell.

Adopted your shape verbatim (flatten-then-find), with a comment recording *why* it is not `or`, since the two functions sitting side by side with different shapes is what let this survive review once already.

The test I added covers four blank spellings — `--server ""`, `--server "   "`, an empty `SKARDI_GATEWAY_URL`, and both blank at once — each asserting the context still points at the membership's `gateway_url`. Mutation-checked: restoring the `or` chain fails it.

crates/cli/src/login/mod.rs
bakey: Fixed in e76056a. Confirmed the range gap before changing anything — `Duration::try_days(100_000_000)` is `Some` and `Utc::now().checked_add_signed(that)` is `None`, so `parse_expires` really does hand a valid `TimeDelta` to an addition that aborts.

Took your first option (`checked_add_signed` at the addition) for the reason you gave: the ceiling is a function of `now`, so clamping in `parse_expires` would be a constant standing in for something that is not one. The message reuses the wording `parse_expires`'s last line already meant.

`expires_accepts_days_and_hours_and_refuses_the_rest` stays as the parser's test; the overflow gets its own case through the flow, which also asserts nothing was minted before the refusal.

crates/cli/src/login/control_plane.rs
bakey: Fixed in e76056a. This was the one I was most glad you caught — `ApiClient` warning for a bearer while the leg carrying the ID token *and* the raw PAT said nothing is exactly backwards.

`is_cleartext_remote` is now `pub(crate)` and the warning fires in `resolve_control_plane`, i.e. your second suggestion: before the browser opens, so it is not advice delivered after a completed sign-in, and it covers `logout --revoke` through `control_plane_for_revoke` for free. Warning rather than refusal, for the TLS-terminating-proxy reason you named.

Verified through the real binary: `--control-plane http://global.example.com` warns, `http://127.0.0.1:1` and `https://…` stay silent.

crates/cli/src/login/loopback.rs
bakey: Fixed in e76056a. Verified the distinction you drew: `malformed_and_non_get_requests_are_answered_and_ignored` uses `drop(TcpStream::connect(…))`, which is `Ok(0)` → break → 400 → continue, so it never covered a socket that stays open. The module doc promised behaviour the code only had for the hang-up case.

Took your fix (a per-connection deadline, 5s), plus the `read_until`-under-`take` rewrite — same bound, one call instead of up to 8192 awaits.

The deadline is a field on `Loopback` with a `#[cfg(test)]` setter rather than a bare const, so the regression test costs 150ms instead of 5s; `ApiClient::with_max_response_bytes` set that precedent in this crate. The test opens an idle socket, waits for the loop to accept it, then sends the real redirect and asserts it arrives in under 2s — mutation-checked by relaxing the deadline to 600s, which fails it.

crates/cli/src/login/pkce.rs
bakey: Fixed in e76056a, and you are right about both halves — the second is the sharper one. `assert_eq!(first.challenge, URL_SAFE_NO_PAD.encode(Sha256::digest(first.verifier.as_bytes())))` is the same expression on both sides; it could not fail.

Adopted `Pkce::from_verifier` as you wrote it, and the appendix-B vector now runs through it. The self-consistency assertion is deleted rather than kept — with the vector in place it adds nothing, and leaving it would keep implying coverage it never had. `generated_pairs_are_unreserved_and_fresh` now claims only what it checks: freshness, length, charset.

Mutation-checked: switching `from_verifier` to `STANDARD` base64 fails both PKCE tests. Under the old shape it would have failed neither.

crates/cli/src/commands/logout.rs
bakey: Fixed in e76056a. Confirmed the whole chain, including that `--revoke` cannot rescue it: a `set-context` context has no `token-id`, so it takes the `bail!` at line 79 — the advice printed is unusable in the one case where the token is already gone.

Went with your first option: the `mode == Cloud` rule now applies however the context was selected, and a server-mode selection is refused by name with a pointer at `config delete-context` (or editing the file). I chose refusing over warning-then-writing because the failure is unrecoverable and silent — there is no `login` to re-mint, so a warning the operator reads after the write is worth nothing.

Two tests: one via `current-context`, one via `--context local`, both asserting the file is byte-identical afterwards. `docs/cli.md` now says `logout` is cloud-only and why.

crates/cli/src/login/mod.rs
bakey: Fixed in e76056a — and you are right that it is worse here than at query time, which is the part I would have missed. At query time the advice is merely refused by resolution; at mint time there is no context yet for `skardi login` to point at either.

Routed through `cloud::diagnose` with the `probe` config, as you suggested. Your read of the existing test was correct too: it used a 403 and asserted only "could not query", so it was blind to this. The new test uses a **401** and asserts the message contains `skardi login` and does NOT contain `SKARDI_API_TOKEN`, while still asserting the saga ran (PAT revoked, no context written).

Mutation-checked by restoring the raw interpolation, which fails it.

crates/cli/src/cloud.rs
bakey: Fixed in e76056a. Agreed on the substance rather than just the style: `retry in 0s` is not a harmless default — it is a number the operator would act on, produced by a branch that cannot legitimately run.

Adopted the let-chain binding verbatim. Existing tests cover the arm both ways (with `Retry-After` and without), and they still pass, so the rewrite is behaviour-preserving.

crates/cli/src/commands/logout.rs
bakey: Fixed in e76056a. The consequence you traced is the reason this is more than a nit: the failure lands *after* the ids have left the disk, so it takes the `unrevoked_warning` path — a scary, correct-but-avoidable message for an input that was actually supplied.

`--identity` now resolves flag-then-env exactly as `login` does, and both `SKARDI_DEV_IDENTITY` and `SKARDI_OAUTH_CLIENT_ID` are read off `LogoutEnv` rather than inline, which is what let me test it: a `--revoke` with `args.identity = None` and `env_dev_identity = Some("dev:alice")` now revokes instead of demanding `--client-id`. `check_dev_identity` still runs either way, so the loopback guard is unchanged.

bakey:
Thanks — this was a genuinely useful read. All nine addressed in e76056a (docs in 023840a), each verified against the code first. Four turned out to be real defects, three were guards that were not guarding, two were nits worth taking.

**The four that could bite a user:**

- **Blank-value shadowing** (1) — `Option::or` short-circuits, so an exported-empty `SKARDI_GATEWAY_URL` beat the control plane's answer and produced an error blaming the control plane for a silence the variable itself caused. Your flatten-then-find shape, with a comment recording why it is not `or` — the two functions sitting side by side with different shapes is what let this survive review once already.
- **The `--expires` panic** (2) — confirmed the range gap directly before touching it: `try_days(100_000_000)` is `Some` and `checked_add_signed` on it is `None`. Fixed at the addition rather than in the parser, for the reason you gave: the ceiling is a function of `now`.
- **No cleartext check on the control-plane leg** (3) — the leg carrying the ID token up and the raw PAT back said nothing while `ApiClient` warned for a mere bearer. Warning now fires at resolution, so it lands before the browser opens and covers `logout --revoke` too.
- **One idle socket denied a sign-in** (4) — the best catch in the set. The existing test used `drop(TcpStream::connect(…))`, which is the *hang-up* case; a socket that stays open held the sequential loop for the full 120 seconds with the real redirect sitting unread in the backlog. Per-connection deadline plus your `read_until`-under-`take` rewrite.

**The three tests that were not testing:**

- The RFC 7636 vector (5) recomputed the transform inline, and its companion assertion was the same expression on both sides. `Pkce::from_verifier` now runs the vector through the production path; I deleted the tautology rather than keeping it, since leaving it implies coverage it never had.
- Plain `logout` (6) applied the cloud filter only under `--all`, so it destroyed hand-configured tokens — and `--revoke` cannot rescue them, since a `set-context` context has no `token-id`. Refused now, however the context was selected.
- The verify failure (7) re-surfaced the `SKARDI_API_TOKEN` advice `cloud::diagnose` exists to replace, and the existing test used a 403 so it was blind to it. Routed through `diagnose`; the new test uses a 401.

**Both nits taken** (8, 9) — the `retry in 0s` fallback was a number an operator would have acted on, and the `$SKARDI_DEV_IDENTITY` asymmetry failed *after* the token ids had left the disk, so it produced the alarming `unrevoked_warning` path for an input that had in fact been supplied. Both env names now come off `LogoutEnv`, which is what made the fix testable.

**On verification.** All seven behavioural fixes are mutation-checked — each reverted in place, each failing exactly its own test and nothing else. Worth reporting that two of those mutations silently no-op'd on the first pass (rustfmt had reflowed the text my patch was matching) and reported a false pass; re-run with an assertion that the edit actually applied, both failed as expected. 225 tests, clippy and fmt clean.

**On the process note** — agreed, and I have not claimed otherwise: `login_e2e` drives the real binary through login → query → gated refusal → logout against a mock control plane and gateway, which is the same flow minus a real skardi-global. The preview-compose run is still owed and still needs an environment neither of us has here. It is the one item I would not merge on my own judgement.

powerfulZ (approved):
Re-reviewed the current head f2835afc, including the fixes for every resolved review thread and the two follow-up gateway_url commits.

The CLI now consumes gateway_url from each membership while correctly supporting the deployment-wide gateway bridge, with distinct Skardi-Workspace routing covered for multiple workspaces. I also rechecked cloud capability gating, URL precedence, OAuth state/PKCE, bounded loopback and control-plane reads, token redaction, rollback, and atomic owner-only config writes.

Local skardi-cli tests pass (227 passed, 2 explicitly ignored), as do formatting and diff checks; all required CI checks are green. No blocking issues found.

Two non-blocking wording nits remain: one test comment still says “per-org URL”, and the CLI docs line break drops “plane’s” from “control plane’s gateway_url”.

PR #226, opened 2026-08-25
feat(sources): Google Drive source pack (milestone 5.9) — files, drives, file_permissions

gracexmatin:
Milestone 5.9: a source pack for Google Drive over Open Connector's `googledrive` service — three tables (`files`, `drives`, `file_permissions`), **live-verified end to end against a real Workspace account** on 2026-08-25. Contract reconciled against gateway v1.3.4 / open-connector `2410fbe` (the same pin as 5.7/5.8); on the provider side the executors pin Google Drive **`v3`** in every URL. No engine changes: the diff is `packs/` plus the registration rosters and docs.

Design record: `docs/superpowers/specs/2026-08-25-open-connector-google-drive-pack-design.md` (phases 1–4, risk R1 resolved). User docs: `docs/open-connector-google-drive.md`.

## What shipped

- `packs/google_drive.yaml` — the declarative pack: `files` (14 columns), `drives` (13), `file_permissions` (11), every design decision annotated inline with its rationale.
- `packs/google_drive.rs` — module doc (status, 11 design decisions, each held by a named test), OnceLock accessor, 25 tests.
- `packs/fixtures/google_drive/` — row fixtures as **redacted live captures** under a default-deny redaction audit, plus both halves of all three contracts (`contracts/` + `contracts/inputs/`, gmail-style).
- Registry entries in `packs/mod.rs` / `source_pack.rs` (+1 roster test), pack doc, status line in `docs/open-connector.md`, the README's source matrix, a cross-link from the Gmail guide — Google splits into one pack per service upstream, so a reader landing on Gmail now has a path to Drive — and the milestone entry in the tasks spec.

## Service shape

The service is spelled `googledrive` and its action IDs carry **two dots** (`googledrive.files.list`) — new to shipped packs, verified engine-safe end to end in phase 1 (`validate_action_id` rejects only `/` and empty dot segments; the URL path keeps dots verbatim; the `rsplit('.')` short-name lookups read table IDs, never action IDs).

Rows are **NORMALIZED** — the opposite of one_drive. Every executor rebuilds rows through a `normalize*` function into a closed key set (`additionalProperties: false`) and pins its own provider-side `fields` projection (`files.list` does not even declare a `fields` input; sending one 400s). So the declared contract is column truth to an unusual degree, and the wire keys are the normalizer's (`sizeBytes` as a number, not Google's string-typed `size`).

## Design decisions

- **The all-drives pin is the pack's one load-bearing fixed input.** `files.list` forwards `supportsAllDrives` verbatim with NO upstream default (unlike its siblings, which default it true), so an unpinned scan silently omits every shared-drive file — the confidently-wrong-rows failure mode. Both halves of the pair are pinned `true`.
- **`q` is an optional RESOURCE, not a filter mapping, structurally**: it is a whole query language, and a FilterMapping's column-op-value literal is never a legal `q`. Trashed files deliberately stay in (`trashed` is a column; filtering belongs in SQL — a `q` pin would collide with the resource and silently narrow the table).
- **`fileId` is a required resource on `file_permissions`** with a three-way enforcement boundary (upstream declares NO `required` array): missing → refused at registration, `""` → schema-layer 400, whitespace → executor 400 at scan time. Loud all three ways. The table means "the permissions of the file this binding names" — the `notion.block_children` shape.
- **The five `restrictions.*` columns trade fingerprint coverage for audit value, knowingly.** `restrictions` is a bare open object, so the five flags sit outside the gate; a coverage test pins exactly those five as the pack's only uncovered columns. Phase 4 supplied what no capture can: a real drive row carried all five spellings verbatim. The per-caller sibling `capabilities` stays unmapped (its values change with the OAuth identity doing the scan).
- Page sizes pinned at the declared ceilings (1000/100/100); the deprecated `teamDriveId` alias is not exposed (avoiding the silent-precedence trap 5.8 needed `exclusive_resources` for); no table declares `error_path` (executors consume in-band Google errors into the gateway failure envelope).
- **Deferred with reasons** (design record): `changes.list` (its cursor must be bootstrapped by a second action — an incremental-sync feed, not a collection), comments / replies / revisions / labels / accessproposals (second wave; the last is Workspace-approval-gated).

## Phase-4 live verification (real Workspace account, seeded corpus)

Seeded: a folder, four files spanning every shape, one shared drive, one shared-drive file, three grant types on one file.

- **Risk R1 resolved in the best direction.** The account started at zero shared drives — that clean empty scan was itself witnessed — then proved able to CREATE one, so `drives` was verified on real rows instead of deferred. All five `restrictions.*` spellings came back verbatim (present `false`s, alongside an unmapped nested `downloadRestriction`).
- **Three-table end-to-end skardi-server scans**, registration passing the fingerprint gate against LIVE discovery. Every mapped column extracted a real non-NULL value somewhere, except three structural residuals (below).
- **The all-drives pin does its job**: the seeded shared-drive file came back with its real `drive_id`. Live absence pattern captured: shared-drive rows drop exactly `owners`+`shared`; `parents`/`starred`/`trashed` stay.
- **Wire bounds confirmed**: 1000/100/100 each 200 with real credentials; 1001/101/0 each 400. Real `pageSize: 1` multi-page walks (~444-char cursors, true terminal `null` on every table). Through skardi-server with a small page size, `LIMIT` stopped a three-page scan after two requests.
- `q` verified live both as a direct input and as a bound binding resource; `fileId` forwarded verbatim on every page; a nonexistent `fileId` produced the documented gateway failure envelope.
- **Live corrections folded into the fixtures**: native Google Docs DO report `sizeBytes` (the null-size row is a folder); every grant carries `permissionDetails`, not just shared-drive ones; owner objects always carry four keys.

**Three columns remain unwitnessed non-null**, each key present-and-null on real rows, each structurally out of reach, each documented as a residual: `drives.org_unit_id` (Workspace org units only), `drives.theme_id` (null even when a drive is CREATED with an explicit theme — it materializes as `color_rgb` + `background_image_link`), `file_permissions.expiration_time` (Google 403s expirations on domain/anyone grants; a user-grant expiration needs a second real account).

## Verification

**CI green on `f763ce8`** ([run 32854955874](https://github.com/SkardiLabs/skardi/actions/runs/32854955874)): 2125 workspace tests passed via `cargo llvm-cov --no-report nextest --all-features`, plus 237 `--ignored` integration tests and 37 doc tests. CI runs the whole workspace through nextest rather than the per-module filter, so the per-pack figures are static counts of test functions in the tree (none `#[ignore]`d): 26 new — 25 in the pack module (`cargo test -p skardi --lib sources::providers::open_connector::packs::google_drive`) plus 1 registry test, taking `sources::providers::open_connector` from 402 to 428.

- 13 contract/conversion: per-table fixture conversions over the redacted live captures; boundary shapes one live tenant cannot produce (empty strings/arrays, partial `restrictions`, a non-null expiration) as inline synthetic pages, marked as such; the default-deny redaction audit over every fixture string leaf; `sizeBytes`-as-string type mismatch with row-scoped error identity; id-less row failing the page; fingerprint pins for all three contracts; the five-column coverage-gap pin; pagination / resource / fixed-input declarations; the captured-input-contract test locking every sendable key plus the declared-but-never-sent negative space per action.
- 12 e2e through MockGateway: per-table wire-declaration scans with exact key sets and the all-drives pins on every page, opaque-cursor continuation and null termination, a fileId-less permissions binding refused before any HTTP, no-pushdown predicates staying local, LIMIT early-stop, empty shared-drive scan, pagination-loop refusal, failure-envelope surfacing with the dotted action id named, UDTF parity, drift-refusal at registration.

## Phase-5 self-review

The review checklist plus the repo's own code-review pass found **no correctness defect in the pack**: all twelve findings were honesty and sync problems in the prose around it, fixed in the last commit. Two would have actively misled a reader. `size_bytes` was documented in three places as null for every native Docs/Sheets/Slides file, with an all-NULL column on a Docs corpus called expected — a phase-3 guess the phase-4 capture contradicts (the native Doc row carries `sizeBytes: 1024`; the null-size row is the folder), so as written it would have talked a future reader out of investigating exactly the always-NULL defect the live pass exists to catch. And the tasks spec's Review notes still announced phase 4 as PENDING — the first thing a reviewer reads, and the opposite of what the 5.9 entry says.

The checks with teeth came back clean, each verified against the code rather than the comments: `owners`' four inner keys are declared `required` under `additionalProperties: false`, so both pluck columns really are inside the fingerprint gate and "only the five `restrictions.*` flags escape it" holds; a missing pluck key fails the page loudly instead of dropping an item, so the two owner columns cannot silently misalign; the loader does refuse a fixed input colliding with a declared resource, as the yaml claims; `max_pages` × 1000 does equal the default `max_rows`; and `validate_action_id` rejects only `/` and a whole-id dot segment, so the two-dot action ids travel safely.

## Caveat carried in the docs

The provider-level OAuth grant requests all three declared scopes including read-write `auth/drive`, though every action this pack maps is read-only. A pack cannot narrow an upstream connection property; documented in the pack doc.

### DIFF
--- a/crates/skardi/src/sources/providers/open_connector/packs/google_drive.rs
+++ b/crates/skardi/src/sources/providers/open_connector/packs/google_drive.rs
@@ -1575,6 +1575,47 @@ bindings:
     }
 
     #[tokio::test]
+    async fn drives_q_resource_forwards_verbatim() {
+        // `drives.list` has its own query language surface (`name` and
+        // `hidden`, rather than the files corpus fields). Pin the binding-to-
+        // wire path independently of `files.q`: the complete expression must
+        // survive verbatim alongside the action's page size and no other key.
+        let gateway = MockGateway::start(|req| {
+            if req.method == "GET" && req.path == "/v1/health" {
+                return MockResponse::ok("{}");
+            }
+            if req.method == "GET" && req.path.starts_with("/v1/actions/") {
+                return google_drive_discovery(&req.path);
+            }
+            if req.method == "POST" && req.path == "/v1/actions/googledrive.drives.list" {
+                return MockResponse::ok(&envelope_ok(
+                    &json!({"drives": [{"id": "d-1", "name": "Engineering"}],
+                            "nextPageToken": null})
+                    .to_string(),
+                ));
+            }
+            MockResponse::new(404, "{}")
+        })
+        .await;
+        let (gateway, ctx) = setup_with_gateway(
+            gateway,
+            "SKARDI_TEST_OC_GDRIVE_DRIVES_Q",
+            "drives",
+            "{q: \"name contains 'x'\"}",
+        )
+        .await;
+
+        let batches = collect(&ctx, "SELECT id FROM saas.gdrive.drives").await;
+        assert_eq!(column_values(&batches, "id"), vec!["d-1"]);
+
+        let inputs = execute_inputs(&gateway, "googledrive.drives.list");
+        assert_eq!(inputs.len(), 1);
+        assert_eq!(inputs[0]["q"], "name contains 'x'");
+        assert_eq!(inputs[0]["pageSize"], 100);
+        assert_eq!(input_keys(&inputs[0]), vec!["pageSize", "q"]);
+    }
+
+    #[tokio::test]
     async fn file_permissions_scan_forwards_its_required_file_id_on_every_page() {
         // The binding names the file, and the value rides EVERY request
         // including continuations — rows carry no file identity of their
--- a/crates/skardi/src/sources/providers/open_connector/packs/google_drive.rs
+++ b/crates/skardi/src/sources/providers/open_connector/packs/google_drive.rs
@@ -209,13 +209,15 @@ mod tests {
         MockResponse::ok(&discovery_ok("{}", output_schema, true, None))
     }
 
-    // ── Contract tests. The row fixtures are SYNTHETIC (phase 3),
-    // provider-shaped key-for-key from the executors' normalizers: the
-    // always-present `?? null` keys appear in every row (null where no
-    // value), the conditionally-spread keys are genuinely absent where
-    // upstream would drop them, and identities/names are obviously
-    // synthetic. Phase 4 replaces them with redacted live captures and
-    // adds the default-deny redaction audit the one_drive pack carries. ──
+    // ── Contract tests. The row fixtures are REDACTED LIVE CAPTURES
+    // (phase 4, 2026-08-25), provider-shaped key-for-key: the
+    // always-present `?? null` keys carry explicit nulls, the
+    // conditionally-spread keys are absent exactly where upstream
+    // dropped them on the wire, and every identity, name and URL is
+    // synthetic — a property enforced by
+    // `fixtures_are_redacted_captures_under_a_default_deny_audit`
+    // rather than asserted by hand. The deliberately-broken
+    // mismatch fixture stays synthetic by design, and says so. ──
 
     fn convert_fixture(table: &SourcePackTable, fixture: &str) -> RecordBatch {
         let page: Value = serde_json::from_str(fixture).expect("fixture parses");
@@ -1766,7 +1768,8 @@ bindings:
         // an error. This is also the shape design record R1 warns about:
         // "empty and correct" and "empty because the account cannot see
         // shared drives at all" are indistinguishable from here, which
-        // is exactly why phase 4 needs an account that can see one.
+        // is exactly why phase 4 needed an account that could see one
+        // — R1, resolved: the phase-4 account created its own.
         let gateway = MockGateway::start(|req| {
             if req.method == "GET" && req.path == "/v1/health" {
                 return MockResponse::ok("{}");
--- a/crates/skardi/src/sources/providers/open_connector/packs/google_drive.rs
+++ b/crates/skardi/src/sources/providers/open_connector/packs/google_drive.rs
@@ -1,0 +1,1967 @@
+//! Google Drive source pack: `files`, `drives` and `file_permissions`
+//! over Open Connector's `googledrive` service, reconciled against a
+//! live gateway (v1.3.4, open-connector at `2410fbe`, 2026-08-25).
+//! Design record:
+//! `docs/superpowers/specs/2026-08-25-open-connector-google-drive-pack-design.md`;
+//! user documentation: `docs/open-connector-google-drive.md`.
+//!
+//! **Status: phases 1–4 complete (live-verified 2026-08-25).** Phase 1
+//! reconciled the contract without provider credentials: action
+//! inventory, all six schema captures byte-exact from discovery, the
+//! input surfaces, the declared page-size bounds and the
+//! undeclared-key 400s (the gateway validates inputs before authz).
+//! Phase 4 then verified every load-bearing claim against a real
+//! Workspace account through a seeded corpus (6 files, 1 shared drive,
+//! 3 grants): the five `restrictions.*` spellings witnessed verbatim on
+//! a real drive row; the pinned page sizes real WIRE bounds (1000/100/
+//! 100 accepted; 1001/101/0 all 400); real multi-page walks at
+//! `pageSize: 1` with ~444-char cursors and terminal-page `null` on
+//! every table; the all-drives pin actually surfacing shared-drive
+//! rows; registration passing the fingerprint gate against LIVE
+//! discovery; a full three-table scan through skardi-server with LIMIT
+//! stopping real pagination one page early; and the row fixtures
+//! re-derived as redacted live captures under a default-deny audit.
+//! Three columns remain live-unwitnessed non-null, each key PRESENT
+//! (null) and correctly spelled on real rows, each structurally out of
+//! reach: `drives.org_unit_id` (reports only under Workspace org
+//! units), `drives.theme_id` (null even when a drive is CREATED with an
+//! explicit theme — the theme materializes as color + background link),
+//! and `file_permissions.expiration_time` (Google 403s expirations on
+//! domain/anyone grants and a user-grant expiration needs a second real
+//! account).
+//!
+//! Design decisions and their rationale. Each is held by a named test
+//! below unless marked as an upstream property no Skardi-side test can
+//! hold:
+//!
+//! - **The service is spelled `googledrive` and its action IDs carry
+//!   TWO dots** (`googledrive.files.list` — the upstream action name is
+//!   itself dotted). New to shipped packs, verified safe end to end in
+//!   phase 1: `validate_action_id` rejects only `/` and empty dot
+//!   segments, the URL path keeps dots verbatim, and the `rsplit('.')`
+//!   short-name lookups read TABLE ids (`google_drive.files`), never
+//!   action ids. Every e2e test below exercises the dotted id against
+//!   the mock gateway, and the drift/error tests pin that it survives
+//!   into error identity too.
+//! - **Rows are NORMALIZED, the opposite of `one_drive`.** Every
+//!   executor rebuilds rows through a `normalize*` function into a
+//!   fixed key set, the declared row objects are `additionalProperties:
+//!   false`, and each executor pins its own provider-side `fields`
+//!   projection — `files.list` does not even declare a `fields` input
+//!   (sending one 400s, probed live). So the wire keys are the
+//!   normalizer's (`sizeBytes`, not Google's string-typed `size`), the
+//!   declared contract is column truth to an unusual degree, and every
+//!   mapped path except the five `restrictions.*` flags resolves inside
+//!   the declared item schema
+//!   (`only_the_five_restriction_flags_escape_the_fingerprint_gate`).
+//! - **Two spellings of "no value", both modeled in the fixtures.** The
+//!   `?? null` normalizer fields are always present (explicit JSON
+//!   null); the conditionally-spread ones (`parents`/`owners`/`shared`/
+//!   `starred`/`trashed` on files, `hidden`/`capabilities`/
+//!   `restrictions` on drives, the three `optionalBoolean` flags on
+//!   permissions) vanish from the row entirely. Both convert to SQL
+//!   NULL; a present `false` stays `false`
+//!   (`file_permissions_fixture_converts_every_grant_shape`). The live
+//!   absence patterns are narrower than the declarations allow:
+//!   shared-drive file rows drop exactly `owners`+`shared` (keeping
+//!   `parents`/`starred`/`trashed`), non-user grants carry
+//!   `allowFileDiscovery` while user grants carry
+//!   `deleted`+`pendingOwner`, and every grant carries
+//!   `permissionDetails` — the fixtures pin these as captured.
+//! - **Three tables, scoped by the operator in phase 2.** `files` (the
+//!   corpus), `drives` (shared-drive inventory — legitimately empty on
+//!   an account with no shared-drive membership, design record R1), and
+//!   `file_permissions` (per-file sharing audit). Design record R1
+//!   resolved in phase 4: the account turned out to be Workspace and
+//!   able to CREATE a shared drive, so the drives table was verified on
+//!   real rows instead of deferred. Deferred at the
+//!   admission gate, recorded in the design record: `changes.list`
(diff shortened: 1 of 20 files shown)


yiyi-wang-skardi:
@yiyi-wang-skardi is now reviewing this PR using Claude Code — opus (reasoning effort max). Target HEAD commit: https://github.com/SkardiLabs/skardi/commit/f763ce85529438b27c39a2b9ffc46c7a2af351fd (`f763ce8`).

crates/skardi/src/sources/providers/open_connector/packs/google_drive.yaml
```diff
@@ -0,0 +1,343 @@
...
+      # "unreported", never false.
+      # Why ship them at all: `domain_users_only` / `drive_members_only`
+      # answer "how open is this drive", squarely the sharing-audit
+      # question `file_permissions` exists for. Its per-caller sibling
+      # `capabilities` stays unmapped — its values change with the OAuth
+      # identity doing the scan ("Capabilities the current user has"),
+      # so the same table under two bindings would disagree; that is a
+      # permission view, not drive data, and no declaration shape would
+      # make it a column.
+      - { name: admin_managed_restrictions, path: restrictions.adminManagedRestrictions, type: boolean, nullable: true }
```
yiyi-wang-skardi, line 264: **Consider adding `restrictions` as a `json` column *alongside* these five booleans**, to make the pack's one acknowledged gap diagnosable.

The design record rejects the json option — "Mapping it as one `json` column was rejected: a blob no SQL predicate can reach without JSON functions is not a usable answer to 'which drives are domain-restricted'" — and as a *replacement* for the typed booleans that reasoning is right. But the record only weighs json-**instead-of**, never json-**in-addition-to**, and the additive version keeps the SQL ergonomics while closing the diagnosis problem:

- `path: restrictions` **is** declared in the contract (`{"type":"object","properties":{},"additionalProperties":{}}`), so a `json` column there sits *inside* the fingerprint gate — `fingerprint_uncovered_columns` would not report it.
- Today, an upstream inner-key rename means all five booleans go silently all-NULL with registration still passing; that is exactly the case-1 gap the coverage test pins as reviewed. With a raw `restrictions` column, the same rename shows five NULLs *next to* a populated JSON blob carrying the new spellings — a one-query diagnosis instead of an invisible failure.
- No engine change needed, and no new pattern: mapping one wire path into two columns is already done in this very PR (`owners` → `owner_display_names` + `owner_email_addresses`), and `type: json` over an open object is established precedent (`notion.properties`, `gmail.criteria`/`action`, `feishu.sender`).

Cost is one column on a table that is legitimately empty for many accounts anyway. Not a merge blocker — the trade as shipped is explicit and phase-4 real rows close the authoring-error class — but it turns "nothing closes the future-drift class" into "drift is visible the first time anyone looks".

crates/skardi/src/sources/providers/open_connector/packs/google_drive.rs
```diff
@@ -0,0 +1,1970 @@
...
+            .expect("row path")
+            .rows(page, 1)
+            .expect("row array");
+        RowConverter::new(table.fields)
+            .expect("converter")
+            .convert(rows, 1)
+            .expect("page converts")
+    }
+
+    fn utf8<'a>(batch: &'a RecordBatch, name: &str) -> &'a StringArray {
```
yiyi-wang-skardi, line 238: **DRY, convention-level:** these Arrow/SQL accessor helpers are mechanical and now duplicated into a ninth pack test module. Measured across `packs/*.rs`: `async fn collect` appears in 9 modules, `fn utf8` in 8, `fn boolean` in 6, and `convert_page` / `column_values` / `input_keys` in 4 each. I diffed this module's `fn utf8` against `one_drive.rs` — byte-identical.

`testutil.rs` is already the home for exactly this kind of shared test scaffolding (`envelope_ok`, `discovery_ok`, `fingerprint_uncovered_columns`, `EnvVarGuard`), so lifting the type-generic accessors there would be a natural fit rather than a new abstraction. `utf8_list` and `list_items` are genuinely new here (no sibling pack has list columns) and would be the first users of a shared list helper.

Explicitly **not** a blocker and not this PR's invention — the PR follows the existing convention faithfully, and changing it is a repo-wide cleanup that does not belong in a pack-addition PR. Flagging it because the duplication is now large enough to be worth a follow-up.

crates/skardi/src/sources/providers/open_connector/packs/google_drive.rs
```diff
@@ -0,0 +1,1970 @@
...
+                "q",
+                "supportsAllDrives"
+            ],
+            "both scopes ride alongside the pins: {}",
+            inputs[0]
+        );
+    }
+
+    #[tokio::test]
+    async fn drives_scan_sends_only_its_page_size_and_terminates_on_null() {
```
yiyi-wang-skardi, line 1533: Small coverage asymmetry: **`drives`' optional `q` resource never reaches the wire in any test.** Every `setup_with_gateway` call binding `drives` passes an empty resource (this test and `scan_of_an_account_with_no_shared_drives_is_clean`), so the only end-to-end `q` evidence is on `files` (`optional_resources_forward_verbatim_and_compose`). Its *declaration* is pinned by `resources_and_fixed_inputs_are_exactly_as_designed`, and it is declared in the captured input contract, so this is a gap in wire evidence rather than in correctness — the two `q` resources go through identical machinery.

Given how exhaustive the rest of the e2e matrix is (per-table exact key sets, cursor continuation, loop refusal, drift refusal, UDTF parity), a one-line variant of this test binding `resource: {q: "name contains 'x'"}` and asserting `input_keys == ["pageSize", "q"]` would close it cheaply. Worth noting that the drives query language differs from the files one (`hidden = true`, `name contains …`), so it is not purely redundant with the `files` case.

yiyi-wang-skardi (approved):
## Audit: Google Drive source pack (milestone 5.9)

Three tables (`files`, `drives`, `file_permissions`) over Open Connector's `googledrive` service. **Zero engine changes** — the shipped code is a YAML asset plus a 20-line `loader::builtin` shim and two registration-roster entries; everything else is tests, captured contracts, fixtures and docs. The declarative surface is where the risk lives, so I machine-checked the declarations against the committed contracts rather than only reading them.

### Verified independently

- **All 38 column declarations** (14 / 13 / 11 — counts match the tests and both docs) resolve inside the contract item schemas and are type-consistent with the declared wire types (`sizeBytes` really is `integer` bounded to ±2^53−1, so `int64` is safe; every `timestamp_ms_utc` path is a string). Exactly the five `restrictions.*` paths fall outside the gate — matching the pinned coverage test, no more and no less.
- **All three row fixtures validate against their contracts** (required keys, `additionalProperties: false`, leaf types). `files_type_mismatch.json` violates exactly one field (`sizeBytes` as a string) — precisely the drift path it exists to exercise.
- **Input surface**: every key the pack can send (resources, fixed inputs, `pageToken`, `pageSize`) is declared in the captured input schemas; the pinned page sizes 1000/100/100 each equal that action's declared `maximum`; `nextPageToken` is `anyOf [string, null]` **and** in `required` on all three, so null-cursor termination is sound.
- **The two-dot action id is safe end to end** (the one genuinely new shape here): `validate_action_id` rejects only `/` and a bare `.`/`..`; `ACTION_ID_SET` preserves `.`, so `googledrive.files.list` travels the path verbatim; every `rsplit('.')` / `split_once('.')` in shared code reads *table* ids. Short-name resolution is scoped per pack, so the new `files` table cannot collide with the existing `slack.files`.
- **No shared-namespace collisions**: the pack's fixed inputs, resources and pagination inputs are mutually disjoint (the loader rejects all four collision pairs, including the pagination-overwrites-resource one).
- **Cache isolation**: `scan_cache_key` includes `connection_alias` and the canonicalized `resource`, so bindings differing only by `q`/`fileId` — or by Google account — cannot serve each other's rows. Worth confirming for this pack specifically, since `q` makes the same table mean different row sets.
- **Fixtures are genuinely redacted** (synthetic ids at live lengths, `example.com` identities, rebuilt URL shapes), enforced by a default-deny audit whose fallthrough arm is `false` — a re-captured fixture carrying a real `nextPageToken` would fail it rather than slip through. Conversion errors carry row identity but never row values (the test asserts `48213` never appears).
- **Checkable doc claims hold**: `max_pages` 100 × `page_size` 1000 = `max_rows` 100 000; the 16 MiB response cap; 25 pack tests + 1 registry test. `scripts/check_markdown_links.py` passes on the new docs.

### Not verified here

No Rust toolchain in this environment, so I could not run `cargo test` — **CI is the arbiter for the 26 tests**. The phase-4 live-gateway claims (real Workspace account, wire bounds, witnessed `restrictions.*` spellings) are unreproducible without credentials; I checked them only for internal consistency across the YAML, module doc, user doc, design record and task list, and found no contradiction.

### Security

No provider credentials enter Skardi; all three actions are read-only `.list`; the `q` resource reaches the gateway as a JSON body value, so there is no URL- or query-injection surface. The one real caveat — the provider-level OAuth grant requests read-write `auth/drive` even though the pack is read-only — is an upstream connection property a pack cannot narrow, and the user doc surfaces it prominently rather than burying it. Correct call.

### Verdict

**No correctness, safety or security defect found.** The one acknowledged gap (five columns outside the fingerprint gate) is documented at every layer and deliberately priced. Three non-blocking suggestions below; the first is the only one I would actually push for.

crates/skardi/src/sources/providers/open_connector/packs/google_drive.yaml
gracexmatin: Verified every claim here against the code before deciding, and all three hold. The design record does only argue against json-*instead-of* — "Mapping it as one `json` column was rejected: a blob no SQL predicate can reach…" — and says nothing about json-*in-addition-to*; you read it correctly. A column at `path: restrictions` would sit inside the gate: `fingerprint_uncovered_columns` descends `["properties"][segment]` per dot segment (`testutil.rs`), so `restrictions` resolves to the declared object and is not reported, while the five `restrictions.*` paths still are — the "exactly those five" pin would not need touching. And the failure mode is as described: the flags are `nullable: true`, and `convert_column` turns `RowPathNotFound` into a NULL cell with no error and no event (`json_to_arrow.rs`), so an inner-key rename is five silent NULLs with registration still passing. The precedents are real too (`type: json` in notion/gmail/feishu/outlook/discord; one path into two columns via `owners` in this PR).

Declining it anyway, on three grounds.

**It detects nothing.** No refusal, no event — the blob pays off only when a human looks. A human who is looking already has `open_connector_scan`, which types every object as an opaque JSON column (`raw_schema.rs`), so `SELECT id, restrictions FROM open_connector_scan(<gateway>, 'googledrive.drives.list', '{}', '$.drives')` shows the current spellings today, no pack change. The added column buys one skipped `raw_action_allowlist` entry, not a capability — and it buys it permanently, on a stable schema, for an occasional diagnostic action.

**It moves the risk downstream.** The five booleans keep the documentation-sourced spellings in ONE reviewed place with a coverage test pinning them. A blob makes every caller re-derive `domainUsersOnly` from Google's docs inside a `LIKE` or a JSON function — unreviewed, unpinned, multiplied by the number of readers. That is the same undeclared-spelling risk, relocated to where nothing checks it.

**"No new pattern" is half true.** `type: json` over an open object has precedent, and one path into two columns has precedent. But I checked all nine `packs/*.yaml`: no json column's path is also the prefix of a typed column anywhere in the repo. Typed scalars *and* the same subtree as a raw blob would be a new shape — fine in principle, but it should be adopted as a decision, not slipped in as reuse.

Two things worth adding to the record, since the comment sharpens both. The silent window is narrower than the caveat reads: it needs upstream to rename an inner key *while still declaring nothing* about the object. If Google starts declaring these properties, the declaration changes, the fingerprint changes, and registration refuses — that direction is already loud. And what would actually *close* the class is neither the blob nor the booleans but a column spelling that fails when a key is missing under a PRESENT parent; the engine already holds exactly that doctrine for object-list plucks (`json_to_arrow.rs`: a dropped pluck key fails the page rather than yielding an all-null list) and has no scalar equivalent. That is engine work for its own milestone.

Both are now written into the design record, along with this additive option and why it was declined, so the next reader does not have to re-derive it — and the pack doc's caveat now ends with the raw-scan query to run when the five ever come back all-NULL, which turns the warning into an action. Thanks for the push on this one; the gap in the record was real even though the column is not going in.

BtXin (approved):

PR #227, opened 2026-08-25
feat(cli): skardi mcp — MCP-over-stdio binding

gracexmatin:
Implements the MCP stdio binding per `docs/superpowers/specs/2026-08-13-mcp-stdio-binding-design.md`: a new `skardi mcp` subcommand serving MCP over stdio as an MCP ⇄ REST bridge, plus the spec's Option A server-side inventory enrichment.

## What changed

- **skardi**: the multi-row `VALUES {name}` detection regex is hoisted into a shared `VALUES_PLACEHOLDER_RE` export (two inline call sites now use it; behavior unchanged).
- **server (Option A)**: `GET /pipelines` items gain `description` + `parameters` with complete per-param `json_schema` fragments; `GET /pipeline/:name` params gain `data_type` + `json_schema` next to the existing Debug-dump `type` field (additive). New `param_schema` module maps Arrow DataTypes → JSON Schema with an unconditional null union; `VALUES` params become array-of-arrays. Params emit in name order for deterministic responses.
- **cli**: new `mcp` module — `projection.rs` (tool-name sanitization `^[a-zA-Z0-9_-]{1,64}$`, sorted-order `_2`/`_3` collision suffixing, `_pipeline` suffix for pipelines shadowing built-ins, schema assembly) and `bridge.rs` (rmcp 3.1.4 `ServerHandler`; inventory re-fetched on every `tools/list`; REST failures surface in-band as `isError` tool results; success bodies pass through verbatim). Built-ins: `query` (optional `purpose` → `ai_context {purpose, session_id}`, one UUID per connection) and `list_data_sources`. Stdin close → exit 0; no bridge-side timeout (host timeout is the backstop); concurrent calls run in parallel.
- **docs**: new `docs/mcp.md` (host setup, tool surface, freshness, auth, lifecycle, troubleshooting); `cli.md` `## MCP` section; `pipelines.md` MCP bullet roadmap → shipped; README §4 "One definition, every binding"; spec's three open decisions moved to Decided (2026-08-25).

## Decisions resolved with Owen (2026-08-25)

- `purpose` param on `query`: **included** (optional; omitted from the request body entirely when absent).
- Built-in tool naming: **unprefixed** `query` / `list_data_sources`.
- Lifecycle: concurrent `tools/call` **yes** (rmcp default), stdin close → **exit 0**, **no bridge request timeout** (documented in docs/mcp.md).

## Notes for review (out-of-spec implementation decisions)

1. **`Capability::Mcp`**: the dispatch pipeline requires every command to declare a capability; cloud contexts refuse `skardi mcp` before serving (the gateway serves neither pipeline execution nor `/data_source`).
2. **Inherent `do_*` methods**: rmcp's `RequestContext` is `#[non_exhaustive]` and unconstructible in tests, so the logic lives in inherent methods with thin trait wrappers.
3. **`eprintln!` instead of tracing**: the CLI has no logging framework; the spec's tracing/env_logger wording is downgraded to stderr prints (stdout is the protocol channel and stays untouched).
4. **Name-sorted parameter emission** on the server, so responses and tests are deterministic.
5. The old `type` Debug-dump field on `GET /pipeline/:name` is kept verbatim; new info rides the added fields.

## Testing

Per house convention nothing was run locally except `cargo fmt` — verification is this PR's CI. Coverage added: param_schema unit tests, extended + 2 new server HTTP tests, 7 projection unit tests, 8 wiremock bridge tests, a cloud-gating test, and a spawned-binary e2e that walks the full MCP handshake over stdio (doubling as the permanent guard for the stdout-is-protocol-only invariant). The local markdown-link checker flags `mcp.md` links only because it validates against `git ls-files` — resolved by these commits.

### DIFF
--- a/crates/cli/src/mcp/bridge.rs
+++ b/crates/cli/src/mcp/bridge.rs
@@ -1,0 +1,349 @@
+//! MCP ⇄ REST bridge: a manual `ServerHandler` whose tools come from the
+//! server's pipeline inventory at list time. All REST I/O goes through the
+//! CLI's `ApiClient`; stdout belongs to the JSON-RPC transport, so nothing
+//! here may print to it.
+//!
+//! The MCP trait methods are thin wrappers over inherent `do_*` methods:
+//! `RequestContext` is `#[non_exhaustive]` and cannot be constructed in
+//! tests, so the wiremock tests below drive the `do_*` methods directly and
+//! the spawned-binary e2e (`tests/mcp_e2e.rs`) covers the trait wiring.
+
+use std::collections::HashMap;
+use std::sync::RwLock;
+
+use rmcp::ServerHandler;
+use rmcp::model::{
+    CallToolRequestParams, CallToolResponse, CallToolResult, ContentBlock, ErrorData,
+    Implementation, JsonObject, ListToolsResult, PaginatedRequestParams, ServerCapabilities,
+    ServerInfo,
+};
+use rmcp::service::{RequestContext, RoleServer};
+use serde_json::Value;
+
+use crate::client::{ApiClient, ApiError, encode_component};
+use crate::mcp::projection;
+
+const INSTRUCTIONS: &str = "Skardi is a federated SQL data plane: operator-defined \
+pipelines plus an ad-hoc SQL engine over the configured data sources. Prefer the \
+pipeline tools for tasks they cover. Before writing ad-hoc SQL with `query`, call \
+`list_data_sources` to see tables, schemas, and their plain-English descriptions.";
+
+pub(crate) struct McpBridge {
+    client: ApiClient,
+    /// Tool name → original pipeline name; rebuilt on every tools/list, so a
+    /// host that re-lists (or reconnects) sees a re-configured server's fresh
+    /// inventory with no bridge restart.
+    tool_map: RwLock<HashMap<String, String>>,
+    /// One UUID per MCP connection, sent as `ai_context.session_id` whenever
+    /// the model provides a `purpose`. This is the v1 carrier of the
+    /// agent-identity passthrough seam.
+    session_id: String,
+}
+
+impl McpBridge {
+    pub(crate) fn new(client: ApiClient) -> Self {
+        McpBridge {
+            client,
+            tool_map: RwLock::new(HashMap::new()),
+            session_id: uuid::Uuid::new_v4().to_string(),
+        }
+    }
+
+    pub(crate) async fn do_list_tools(&self) -> Result<ListToolsResult, ErrorData> {
+        // Fetched on every call — no cache. Hosts list rarely (typically at
+        // connect time) and the hop is usually localhost.
+        let inventory = self
+            .client
+            .get("/pipelines")
+            .await
+            // ApiError::Connect's Display already names the resolved URL and
+            // the three ways to configure it.
+            .map_err(|e| ErrorData::internal_error(e.to_string(), None))?;
+        let (tools, map) = projection::project(&inventory);
+        *self.tool_map.write().expect("tool map lock") = map;
+        Ok(ListToolsResult::with_all_items(tools))
+    }
+
+    pub(crate) async fn do_call_tool(
+        &self,
+        name: &str,
+        args: Option<JsonObject>,
+    ) -> Result<CallToolResult, ErrorData> {
+        let outcome = match name {
+            "query" => self.call_query(args).await,
+            "list_data_sources" => self.client.get("/data_source").await,
+            _ => {
+                let pipeline = self
+                    .tool_map
+                    .read()
+                    .expect("tool map lock")
+                    .get(name)
+                    .cloned();
+                match pipeline {
+                    Some(pipeline) => {
+                        // Arguments pass through as the flat execute body —
+                        // the server is the validator.
+                        let body = Value::Object(args.unwrap_or_default());
+                        let path = format!("/{}/execute", encode_component(&pipeline));
+                        self.client.post(&path, &body).await
+                    }
+                    None => {
+                        // Protocol-level: covers host bugs and a server
+                        // re-configured since the host last listed.
+                        return Err(ErrorData::invalid_params(
+                            format!(
+                                "unknown tool '{name}' — the pipeline inventory may have \
+                                 changed; re-issue tools/list to refresh it"
+                            ),
+                            None,
+                        ));
+                    }
+                }
+            }
+        };
+        Ok(match outcome {
+            // Success: the response JSON verbatim, no client-side reshaping.
+            Ok(value) => CallToolResult::success(vec![ContentBlock::text(value.to_string())]),
+            // Execution errors are for the model to see and react to (fix a
+            // parameter, pick another tool), not protocol errors. ApiError's
+            // Display carries the server's message and error_type.
+            Err(err) => CallToolResult::error(vec![ContentBlock::text(err.to_string())]),
+        })
+    }
+
+    /// The one choke-point that assembles the `/query` body; later versions
+    /// extend this to full identity injection without touching call sites.
+    async fn call_query(&self, args: Option<JsonObject>) -> Result<Value, ApiError> {
+        let args = args.unwrap_or_default();
+        let mut body = serde_json::Map::new();
+        for key in ["sql", "max_rows"] {
+            if let Some(value) = args.get(key) {
+                body.insert(key.to_string(), value.clone());
+            }
+        }
+        // When `purpose` is absent, ai_context is omitted entirely — the
+        // server rejects a partial object.
+        if let Some(purpose) = args.get("purpose").and_then(Value::as_str) {
+            let purpose = purpose.trim();
+            if !purpose.is_empty() {
+                body.insert(
+                    "ai_context".to_string(),
+                    serde_json::json!({"purpose": purpose, "session_id": self.session_id}),
+                );
+            }
+        }
+        self.client.post("/query", &Value::Object(body)).await
+    }
+}
+
+impl ServerHandler for McpBridge {
+    fn get_info(&self) -> ServerInfo {
+        ServerInfo::new(ServerCapabilities::builder().enable_tools().build())
+            .with_server_info(Implementation::new("skardi", env!("CARGO_PKG_VERSION")))
+            .with_instructions(INSTRUCTIONS)
+    }
+
+    async fn list_tools(
+        &self,
+        _request: Option<PaginatedRequestParams>,
+        _context: RequestContext<RoleServer>,
+    ) -> Result<ListToolsResult, ErrorData> {
+        self.do_list_tools().await
+    }
+
+    async fn call_tool(
+        &self,
+        request: CallToolRequestParams,
+        _context: RequestContext<RoleServer>,
+    ) -> Result<CallToolResponse, ErrorData> {
+        self.do_call_tool(&request.name, request.arguments)
+            .await
+            .map(CallToolResponse::from)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::config::ClientConfig;
+    use serde_json::json;
+    use wiremock::matchers::{body_json, method, path};
+    use wiremock::{Mock, MockServer, ResponseTemplate};
+
+    fn test_config(server: &str) -> ClientConfig {
+        ClientConfig {
+            server: server.to_string(),
+            token: None,
+            context: None,
+        }
+    }
+
+    fn bridge_for(server: &MockServer) -> McpBridge {
+        McpBridge::new(ApiClient::new(&test_config(&server.uri())).unwrap())
+    }
+
+    fn inventory() -> Value {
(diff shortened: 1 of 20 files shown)


yiyi-wang-skardi:
@yiyi-wang-skardi is now reviewing this PR using Claude Code — opus (reasoning effort max). Target HEAD commit: https://github.com/SkardiLabs/skardi/commit/ed818e9e97ee54d9d218d9556725052e883ba154 (`ed818e9`).

BtXin:
@codex

gracexmatin:
@codex

crates/cli/src/mcp/projection.rs
```diff
@@ -0,0 +1,316 @@
...
+    let description = match entry["description"].as_str() {
+        Some(d) if !d.trim().is_empty() => format!("{d} (pipeline `{pipeline_name}`)"),
+        _ => format!("Execute pipeline `{pipeline_name}`"),
+    };
+    let mut properties = serde_json::Map::new();
+    let mut required = Vec::new();
+    if let Some(params) = entry["parameters"].as_array() {
+        for param in params {
+            if let Some(name) = param["name"].as_str() {
+                properties.insert(name.to_string(), param["json_schema"].clone());
```
yiyi-wang-skardi: **Version skew produces a malformed tool schema rather than a clear error.**

`param["json_schema"]` yields `Value::Null` when the key is absent, and that null is inserted verbatim as the property's schema. Against a server predating this PR, `GET /pipelines` has no `parameters` key at all, so line 104's `as_array()` returns `None` and every pipeline tool is published with `{"type":"object","properties":{},"required":[],"additionalProperties":false}` — a closed, empty schema the model cannot pass anything through, so every pipeline call fails with "Missing required parameters". And if `parameters` is present but an entry lacks `json_schema`, the result is `"properties": {"brand": null}`, which is not valid JSON Schema (property values must be a schema object or a boolean); strict hosts may reject the whole `tools/list` response.

This matters because mixed versions are an explicitly supported deployment: `docs/mcp.md` says "pointing `--server` at a remote skardi-server also works, making `skardi mcp` a local MCP gateway to a remote deployment" — so the CLI and the server it talks to are not guaranteed to ship together.

A one-line guard keeps the listing well-formed:

```rust
let schema = match &param["json_schema"] {
    v @ (Value::Object(_) | Value::Bool(_)) => v.clone(),
    _ => json!({}),  // unknown/older server: accept anything, let the server validate
};
properties.insert(name.to_string(), schema);
```

The empty-`parameters` case is harder to paper over, but it might be worth one `eprintln!` when an inventory entry carries no `parameters` key at all — that turns a silently-unusable tool into a diagnosable one.

crates/cli/src/mcp/bridge.rs
```diff
@@ -0,0 +1,350 @@
...
+                    .unwrap_or_else(|p| p.into_inner())
+                    .get(name)
+                    .cloned();
+                match pipeline {
+                    Some(pipeline) => {
+                        // Arguments pass through as the flat execute body —
+                        // the server is the validator.
+                        let body = Value::Object(args.unwrap_or_default());
+                        let path = format!("/{}/execute", encode_component(&pipeline));
+                        self.client.post(&path, &body).await
```
yiyi-wang-skardi: **Pipeline tool calls carry no session attribution, so half the MCP surface is invisible to the audit ledger.**

`call_query` attaches `ai_context: {purpose, session_id}` so `/query` calls from one MCP connection group together in the ledger — but pipeline executions go out as a bare `post`, with no `x-skardi-session-id`. An operator looking at the ledger sees the agent's ad-hoc SQL grouped by session and its pipeline runs unattributed, which is the surface an agent is *encouraged* to prefer (`INSTRUCTIONS` says "Prefer the pipeline tools for tasks they cover").

Everything needed is already in hand: `self.session_id` is a hyphenated v4 UUID, which satisfies the server's predicate exactly (`session_id_from_headers`: visible ASCII, no comma, ≤200 chars), and `ApiClient::post_with_headers` exists for precisely this — `skardi run --session-id` already uses it (`commands/run.rs:40`).

```rust
self.client
    .post_with_headers(&path, &body, &[("x-skardi-session-id", &self.session_id)])
    .await
```

Unlike `query`'s `purpose`, this needs no new tool parameter and no model cooperation — the bridge knows its own session id. Worth confirming it was a conscious v1 scope cut: the spec's "Non-goals (v1)" list doesn't mention it, and the "agent identity passthrough seam" paragraph reads as if it should cover both surfaces.

crates/cli/src/mcp/bridge.rs
```diff
@@ -0,0 +1,350 @@
...
+        Ok(ListToolsResult::with_all_items(tools))
+    }
+
+    pub(crate) async fn do_call_tool(
+        &self,
+        name: &str,
+        args: Option<JsonObject>,
+    ) -> Result<CallToolResult, ErrorData> {
+        let outcome = match name {
+            "query" => self.call_query(args).await,
```
yiyi-wang-skardi: **The built-in tool names live in three places with nothing tying them together (DRY).**

1. `projection::RESERVED_NAMES` — seeds the `taken` set so no pipeline can claim the name;
2. `projection::builtin_tools()` — hardcodes `"query"` / `"list_data_sources"` again in the `Tool::new` calls;
3. this dispatch `match` — hardcodes both literals a third time.

All three must agree, and nothing enforces it. The failure mode of a drift between (1) and (2) is the worst one: `assign_tool_names` seeds `taken` from `RESERVED_NAMES`, but `project()` appends `builtin_tools()` unconditionally at the end — so a built-in whose name is missing from `RESERVED_NAMES` would let a pipeline claim that name, and `tools/list` would then emit **two tools with the same name** (an MCP protocol violation), with this `match` silently shadowing the pipeline. A drift between (1)/(2) and (3) is milder: the built-in appears in the listing but every call to it falls through to the `tool_map` lookup and returns "unknown tool".

The existing `builtins_have_the_specified_schemas` test doesn't catch this — it looks each built-in up by string literal, so it stays green under any drift.

Cheapest fix that preserves the current shape: build the built-ins from the constant and assert the invariant once.

```rust
// projection.rs
pub(crate) const QUERY: &str = "query";
pub(crate) const LIST_DATA_SOURCES: &str = "list_data_sources";
pub(crate) const RESERVED_NAMES: [&str; 2] = [QUERY, LIST_DATA_SOURCES];
// ...and use QUERY / LIST_DATA_SOURCES in builtin_tools() and in this match.

#[test]
fn builtin_tool_names_match_the_reserved_set() {
    let names: Vec<&str> = builtin_tools().iter().map(|t| t.name.as_ref()).collect();
    assert_eq!(names, RESERVED_NAMES);
}
```

crates/server/src/pipeline_handlers.rs
```diff
@@ -423,14 +444,20 @@ pub async fn get_pipelines_info(
     })?;
 
     if let Some(pipeline) = config.pipelines.get(&name) {
-        let request_schema = pipeline.request_schema();
-        let params: Vec<Value> = request_schema
-            .fields
-            .iter()
+        let sql = &pipeline.query_definition().sql;
```
yiyi-wang-skardi: **`get_pipelines_info` re-implements `enriched_parameters` inline (DRY).**

These four lines plus the `map` below are a copy of the helper defined 60 lines up at line 384 — same `query_definition().sql`, same sort-by-name, same `name` / `data_type` / `json_schema` construction. The only difference is the extra legacy `"type"` key. Two copies of the *same* enrichment contract is exactly the shape that drifts: a change to the sort, to `data_type`'s formatting, or to how `json_schema` is derived has to be made twice, and only one of the two has a caller (the bridge) that would notice.

Since the added fields are additive, the simplest consolidation is to fold the legacy `"type"` into `enriched_parameters` and call it from both handlers — it costs `/pipelines` a Debug string per parameter but leaves exactly one place where the parameter shape is defined:

```rust
fn enriched_parameters(pipeline: &StandardPipeline) -> Vec<Value> {
    // ...
    serde_json::json!({
        "name": name,
        "type": format!("{field:?}"),   // legacy Debug dump of the whole InferredFieldType
        "data_type": format!("{:?}", field.field_type),
        "json_schema": param_json_schema(&field.field_type, sql, name),
    })
}
```

If carrying `"type"` on the list endpoint is unwanted, the alternative is to extract just the shared part — a `sorted_fields(pipeline) -> Vec<(&String, &InferredFieldType)>` helper — so at minimum the ordering guarantee has one home. Either way, the current duplication is the thing worth removing.

yiyi-wang-skardi (approved):
## `skardi mcp` — MCP-over-stdio bridge

Audited the whole diff (7 commits, ~2.8k lines): the new `crates/cli/src/mcp/` module (bridge + projection), the enriched `GET /pipelines` / `GET /pipeline/:name` server response, the shared `VALUES_PLACEHOLDER_RE` extraction, cloud gating, tests, and docs.

**Overall this is careful, well-scoped work.** The design holds up on the things that matter most for a stdio MCP server:

- **stdout purity** — nothing on the `mcp` path writes to stdout (all warnings go through `eprintln!`), and `tests/mcp_e2e.rs` makes that a permanent guard: any stray print breaks the initialize handshake and fails the test.
- **Lock discipline** — the `std::sync::RwLock` guards in `do_list_tools`/`do_call_tool` are deliberately confined to a single statement (`.get(name).cloned()`), so no `!Send` guard is ever held across an `.await`. Poisoning is recovered rather than panicked on.
- **Path safety** — pipeline names go through `encode_component` before landing in `/{name}/execute`, same as `skardi run`, so a name containing `/`, `?` or `%` cannot re-route the request.
- **Error surface** — REST failures become `isError` tool results (model-recoverable) while only `tools/list` raises a JSON-RPC error; unknown tools return `invalid_params` with a re-list nudge.
- **Name projection** — sanitize → reserved-rename → numeric suffix, driven off a sorted original-name list so the mapping is deterministic; truncation is safe because every kept char is ASCII, and the `_2` path re-truncates the base so 64 is never exceeded. The collision tests (`a.b` / `a_b` / `a_b_2`, empty name, 64-char names) cover the nasty cases properly.
- **Cloud gating** — `Capability::Mcp` is refused whole in a cloud context, before anything is written to the would-be protocol channel, and it's tested.

The server-side enrichment is equally tidy: `param_json_schema` is a pure function with its own tests, parameters are sorted for a deterministic response, `VALUES {name}` detection now shares one `LazyLock` regex with the loader (previously recompiled per call — a nice correctness *and* efficiency win), and the legacy `type` Debug field on `/pipeline/:name` is preserved additively.

**No blocking issues found.** Four non-blocking items are filed inline: one robustness gap against server/CLI version skew, two DRY/drift seams, and one missing audit attribution on pipeline tool calls.

### Notes that aren't findings

- `GET /pipelines` was already unauthenticated and now carries descriptions and full parameter schemas. This widening is explicitly acknowledged in `docs/mcp.md` ("Auth notes") and in the design spec, so it's a recorded decision rather than an oversight — flagging only so it stays on the radar for deployments with sensitive schemas.
- `data_type` being a Rust `Debug` dump is called out in the spec as human/debug-only ("never parses Debug strings"), and the bridge indeed reads only `json_schema`. Deliberate.
- `rmcp` is pinned with `=3.1.4`, so patch releases won't be picked up by `cargo update`. Reasonable for a transport dependency, just worth remembering at bump time.

### Verification caveat

No Rust toolchain is available in this review environment, so I could not run `cargo test` / `cargo clippy`. The review is a static read of the diff plus the surrounding code it integrates with (`ApiClient`, `session_header.rs`, `cloud.rs`, the router). Assertions about test *coverage* come from reading the tests; assertions about them *passing* rely on CI.

crates/cli/src/mcp/projection.rs
```diff
@@ -0,0 +1,316 @@
...
+    )
+}
+
+pub(crate) fn builtin_tools() -> Vec<Tool> {
+    let query_schema = object_schema(
+        json!({
+            "sql": {"type": "string"},
+            "max_rows": {
+                "type": "integer",
+                "description": "Result row cap; server default 1000."
```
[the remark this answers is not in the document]
gracexmatin: Fixed. The MCP schema rejects zero and negative integers now.

crates/server/src/param_schema.rs
```diff
@@ -0,0 +1,138 @@
...
+    match field_type {
+        DataType::Utf8 | DataType::LargeUtf8 => json!({"type": "string"}),
+        dt if dt.is_integer() => json!({"type": "integer"}),
+        dt if dt.is_floating() || dt.is_decimal() => json!({"type": "number"}),
+        DataType::Boolean => json!({"type": "boolean"}),
+        DataType::Date32 | DataType::Date64 => json!({"type": "string", "format": "date"}),
+        DataType::Timestamp(_, _) => json!({"type": "string", "format": "date-time"}),
+        DataType::List(inner) => {
+            json!({"type": "array", "items": base_fragment(inner.data_type())})
+        }
```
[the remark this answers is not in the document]
gracexmatin: Fixed. Add minItems assertions to the two param_schema unit tests (List mapping and VALUES override).

PR #228, opened 2026-08-25
docs(readme): name the agent hosts the skill runs on

Userneima:
## What

The Install section said the skill runs on Claude Code, then sent everyone else
away with "For Cursor and other Agent Skills-compatible hosts, see the
skardi-skills README." A reader on Codex or OpenClaw had no way to tell from
this page whether the skill was for them.

It now names the hosts — Codex, Cursor, Pi, dsh, OpenClaw, Hermes — and still
sends the per-host directories to skardi-skills, so this section does not grow.
Net change is one sentence.

## Companion PR

The install paths themselves land in
SkardiLabs/skardi-skills#29 — that PR carries the per-host
directories and the verification notes. Merging this one first would point at a
section that does not exist yet, so land the skills PR first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,8 +176,8 @@ turns a folder of documents — or a datastore you already run — into governed
 searchable context served over HTTP by `skardi-server`: hybrid search (vector +
 FTS + RRF), defaulting to a local SQLite file the skill creates and owns, or
 pointed at Postgres + pgvector, MongoDB, or Lance. The same skill also runs on
-Codex, Cursor, Pi, OpenClaw, and Hermes; per-host installation instructions are in
-the [skardi-skills README](https://github.com/SkardiLabs/skardi-skills#installation).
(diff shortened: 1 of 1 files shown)


Userneima:
Updated after review: dsh is dropped from the sentence. It ignores any skill whose frontmatter `name` is not kebab-case, and both skardi-skills entries are underscored, so naming it here would have pointed readers at something that loads nothing. Details and the source reference are in SkardiLabs/skardi-skills#29.

Also dropped "anything that speaks the Agent Skills standard" — dsh conforms to the format and still refuses these skills, so format compatibility does not imply installability.

abbccdda (approved):

Userneima:
dsh is back in the sentence — the skills are renamed to kebab-case in SkardiLabs/skardi-skills#29, so it loads them now. Merge order is unchanged: #29 first, then this one.

PR #216, opened 2026-08-16
feat(sources): milestone 5.10 — Dropbox source pack + split-action cursor continuation

abbccdda:
Milestone 5.10: a Dropbox source pack — `dropbox.files`, `dropbox.shared_links`, `dropbox.file_search` — plus the backward-compatible engine extension two of those three tables cannot pass the admission gate without.

Fourteen commits: the design spec and its execution plan, the engine extension, the pack, a self-review pass, the live-verification pass, a second review round, a fix-up that reconciles this branch with the `Keyset` / `single_page` strategies the Discord pack (5.6) landed meanwhile, the milestone renumber the rebases forced, a third review round, a test fix-up for a CI failure that round introduced (see **CI fix-up** below), and a fourth review round.

> **✅ Live-verified 2026-08-18** against a self-hosted Open Connector gateway at commit `a3efa99` and a real free-tier Dropbox account, following [the live-evaluation runbook](docs/superpowers/plans/2026-08-18-dropbox-live-evaluation.md).
>
> The earlier revision of this description warned that **registration was expected to fail the contract gate until the pins were re-captured**. That was wrong, and pleasantly so: all five committed contracts came back **byte-identical** to the live captures, so every fingerprint matched live discovery unchanged and registration passed on the first try. Nothing needed re-pinning. The opener/continuation schema identity — previously flagged as suspicious because each `*_continue.json` was byte-identical to its opener — turns out to be a fact about the wire, not an artifact of how the files were derived.
>
> The pass did find one real defect, in the one place a contract-level check cannot look. See **What the live pass changed**.

## What the wire confirmed

Every claim the pack had labelled as an inference off the provider source:

| Claim | Live result |
|---|---|
| Cursor fed to the *opening* action is a hard 400 | Confirmed — rejected bare and alongside the full input |
| `cursor_only`: the continue action takes the cursor and nothing else | Confirmed — any extra input is refused; its real input schema declares `cursor` as its only property, `required`, `additionalProperties: false`. Committed as `fixtures/dropbox/contracts/inputs/*.json` and checked against the declaring table by a test, so a re-capture that grows a `required` key fails CI |
| `list_folder`'s final page carries `hasMore: false` beside a **non-empty** cursor | Confirmed — 259-char cursor on the last page; following it returns an empty page, so cursor-spelling termination really would refetch and trip the loop guard. `has_more_path` is load-bearing, not decorative |
[...]

PR #229, opened 2026-08-26
refactor(sources): lift the shared pack-test Arrow/SQL accessors into testutil

gracexmatin:
## What

The Open Connector pack test modules each carried their own copy of the same mechanical Arrow/SQL helpers. Measured across `packs/*.rs`: `collect` in 9 modules, `utf8` in 8, `boolean` in 6, `convert_page` / `column_values` / `input_keys` in 4 each, `execute_inputs` in 7 (4 with an `action_path` filter, 3 without), and `ids_of` in 3 plus two inline spellings — all byte-identical within each group, except slack's older `utf8`/`boolean`, which used bare `unwrap()` instead of naming the missing column.

This lifts them all into `open_connector::testutil` — already the home for exactly this kind of shared test scaffolding (`envelope_ok`, `discovery_ok`, `fingerprint_uncovered_columns`, `EnvVarGuard`) — and points all 9 modules at them, google_drive included (rebased onto #226 after it merged mid-review). github's `strings`/`bools` were the same accessors under other names and are renamed at their call sites; outlook's two hand-rolled downcasts (the conversation_id HashSet count and `is_hidden`) now go through `utf8`/`boolean`. Net −436 lines across the two commits; no assertions change.

## Review-driven hardening (beyond the verbatim move)

- `collect` panics now carry the SQL that failed to plan/collect.
- `utf8`/`boolean` mismatch panics name the column **and** the actual Arrow type found.
- `column_values` asserts non-NULL per slot instead of silently reading the buffer as `""` (all current call sites read non-null identity columns, so nothing regresses today); its panic row index counts across batches; and it composes over `utf8` instead of re-implementing the lookup+downcast.
- `convert_page` is renamed `convert_first_page`: it hardcodes page 1, and the old name let a fixture for page 3 report errors as "page 1".
- The shared `execute_inputs` takes `action_path`; a single-action pack passes `""` (always matches, since `ends_with("")` is true) — one function covers both prior shapes.

One CI round-trip found the only non-mechanical wrinkle: `column_values` calls `.len()` on a concrete `StringArray`, which resolves through the `Array` trait — in the pack modules the trait rode along on each module's imports, and the lift had to bring it explicitly (`utf8`/`boolean` didn't need it: their receiver is a `dyn Array` trait object, whose method lookup works without the trait in scope).

## Deliberately not moved

- **Pack-local single-type accessors** (`timestamp`, `timestamps`, `millis`, `int64`, `uint64`, github's `u64s`/`string_list`): one or two copies each, several under pack-specific names. Sharing them is not yet paying for the move.
- **google_drive's `utf8_list` / `list_items`**: genuinely new (no sibling pack has list columns); they stay with the pack until a second user shows up.
- **`convert_fixture`**: three different spellings across packs (some inline the conversion, some delegate to `convert_first_page`, with different expect messages); unifying it would change behavior text, not just location.
- **discord's `sorted_keys`**: same idea as `input_keys` but returns `Vec<String>`; converging it would touch assertion types, not just location.
- **`table_functions.rs`'s `collect`**: same shape but outside `packs/`; can switch to the shared helper in a follow-up if wanted.

## In-flight branches

#216 (dropbox) still carries local copies; after this merges it can drop them for the testutil imports on rebase — a mechanical change on that branch.

## Verification

`cargo fmt --check` clean locally; correctness rides on CI running the full suite — the change moves test helpers (plus the hardening above) and touches no production code.

### DIFF
--- a/crates/skardi/src/sources/providers/open_connector/packs/feishu.rs
+++ b/crates/skardi/src/sources/providers/open_connector/packs/feishu.rs
@@ -110,14 +110,14 @@ mod tests {
     use crate::sources::providers::open_connector::row_path::RowPath;
     use crate::sources::providers::open_connector::source_pack::SourcePackTable;
     use crate::sources::providers::open_connector::testutil::{
-        EnvVarGuard, MockGateway, MockResponse, discovery_ok, envelope_ok,
-        fingerprint_uncovered_columns,
+        EnvVarGuard, MockGateway, MockResponse, boolean, collect, discovery_ok, envelope_ok,
+        fingerprint_uncovered_columns, utf8,
     };
     use crate::sources::providers::open_connector::{
         OpenConnectorConfig, OpenConnectorGateways, register_open_connector_tables,
         register_open_connector_udtfs,
     };
-    use arrow::array::{Array, BooleanArray, StringArray, TimestampMillisecondArray};
+    use arrow::array::{Array, StringArray, TimestampMillisecondArray};
     use arrow::record_batch::RecordBatch;
     use datafusion::prelude::SessionContext;
     use serde_json::{Value, json};
@@ -168,24 +168,6 @@ mod tests {
             .expect("converter")
             .convert(rows, 1)
             .expect("fixture converts")
-    }
-
-    fn utf8<'a>(batch: &'a RecordBatch, name: &str) -> &'a StringArray {
-        batch
-            .column_by_name(name)
-            .unwrap_or_else(|| panic!("column {name}"))
-            .as_any()
-            .downcast_ref()
-            .expect("Utf8 column")
-    }
-
-    fn boolean<'a>(batch: &'a RecordBatch, name: &str) -> &'a BooleanArray {
-        batch
-            .column_by_name(name)
-            .unwrap_or_else(|| panic!("column {name}"))
-            .as_any()
-            .downcast_ref()
-            .expect("Boolean column")
     }
 
     fn millis<'a>(batch: &'a RecordBatch, name: &str) -> &'a TimestampMillisecondArray {
@@ -558,15 +540,6 @@ bindings:
         .expect("gateway registration succeeds");
         register_open_connector_udtfs(&ctx, gateways).expect("UDTF registration succeeds");
         (gateway, ctx)
-    }
-
-    async fn collect(ctx: &SessionContext, sql: &str) -> Vec<RecordBatch> {
-        ctx.sql(sql)
-            .await
-            .expect("plan")
-            .collect()
-            .await
-            .expect("collect")
     }
 
     fn ids_of(batches: &[RecordBatch]) -> Vec<String> {
--- a/crates/skardi/src/sources/providers/open_connector/packs/feishu.rs
+++ b/crates/skardi/src/sources/providers/open_connector/packs/feishu.rs
@@ -110,8 +110,8 @@ mod tests {
     use crate::sources::providers::open_connector::row_path::RowPath;
     use crate::sources::providers::open_connector::source_pack::SourcePackTable;
     use crate::sources::providers::open_connector::testutil::{
-        EnvVarGuard, MockGateway, MockResponse, boolean, collect, discovery_ok, envelope_ok,
-        fingerprint_uncovered_columns, utf8,
+        EnvVarGuard, MockGateway, MockResponse, boolean, collect, column_values, discovery_ok,
+        envelope_ok, execute_inputs, fingerprint_uncovered_columns, utf8,
     };
     use crate::sources::providers::open_connector::{
         OpenConnectorConfig, OpenConnectorGateways, register_open_connector_tables,
@@ -542,34 +542,6 @@ bindings:
         (gateway, ctx)
     }
 
-    fn ids_of(batches: &[RecordBatch]) -> Vec<String> {
-        batches
-            .iter()
-            .flat_map(|batch| {
-                let ids = batch
-                    .column_by_name("id")
-                    .expect("id column")
-                    .as_any()
-                    .downcast_ref::<StringArray>()
-                    .expect("Utf8 ids")
-                    .clone();
-                (0..ids.len()).map(move |i| ids.value(i).to_string())
-            })
-            .collect()
-    }
-
-    fn execute_inputs(gateway: &MockGateway) -> Vec<Value> {
-        gateway
-            .requests()
-            .into_iter()
-            .filter(|r| r.method == "POST")
-            .map(|r| {
-                serde_json::from_str::<Value>(&r.body).expect("request body is JSON")["input"]
-                    .clone()
-            })
-            .collect()
-    }
-
     fn chat_row(id: &str) -> Value {
         json!({"chat_id": id, "name": id})
     }
@@ -605,9 +577,9 @@ bindings:
             setup_with_gateway(gateway, "SKARDI_TEST_OC_FEISHU_CHATS", "chats").await;
 
         let batches = collect(&ctx, "SELECT id FROM saas.ws.chats ORDER BY id").await;
-        assert_eq!(ids_of(&batches), vec!["oc-1", "oc-2", "oc-3"]);
+        assert_eq!(column_values(&batches, "id"), vec!["oc-1", "oc-2", "oc-3"]);
 
-        let inputs = execute_inputs(&gateway);
+        let inputs = execute_inputs(&gateway, "");
         assert_eq!(inputs.len(), 2, "two cursor pages");
         assert!(inputs[0].get("pageToken").is_none(), "{}", inputs[0]);
         assert_eq!(inputs[1]["pageToken"], "tok-2");
@@ -657,9 +629,9 @@ bindings:
         .await;
         // Inexact pushdown: DataFusion re-trims, so only the matching row
         // survives even though the stub returned both.
-        assert_eq!(ids_of(&batches), vec!["om-1"]);
+        assert_eq!(column_values(&batches, "id"), vec!["om-1"]);
 
-        let inputs = execute_inputs(&gateway);
+        let inputs = execute_inputs(&gateway, "");
         assert_eq!(inputs.len(), 1);
         let input = &inputs[0];
         assert_eq!(input["containerIdType"], "chat", "container kind pin");
@@ -738,7 +710,7 @@ bindings:
             .collect();
         assert_eq!(guids, vec!["t-1"], "status filtering happened locally");
 
-        let input = &execute_inputs(&gateway)[0];
+        let input = &execute_inputs(&gateway, "")[0];
         assert_eq!(input["type"], "my_tasks", "population pin");
         let mut keys: Vec<&str> = input
             .as_object()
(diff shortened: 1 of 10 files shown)


BtXin:
@codex

crates/skardi/src/sources/providers/open_connector/testutil.rs
```diff
@@ -64,6 +73,93 @@ pub(crate) fn discovery_ok(
...
+pub(crate) fn column_values(batches: &[RecordBatch], name: &str) -> Vec<String> {
+    batches
+        .iter()
+        .flat_map(|batch| {
+            let column = batch
+                .column_by_name(name)
+                .unwrap_or_else(|| panic!("column {name}"));
+            let values = column
+                .as_any()
+                .downcast_ref::<StringArray>()
```
yiyi-wang-skardi: DRY, right where it counts most: `column_values` re-implements `utf8`'s lookup + downcast + panic text (lines 131-140) instead of calling it. In a PR whose whole point is collapsing duplicate accessors, the shared module shipping two copies of the same seven lines is the one place a reader will notice.

The `.clone()` is only there so the inner `move` closure can own the array, and `utf8` returns a reference, so this composes directly:

```rust
(0..values.len()).map(move |i| { ... })  // unchanged
```
with
```rust
let values = utf8(batch, name).clone();
```

Same semantics, same panic messages (they're already identical strings), one definition of "named Utf8 column" instead of two.

crates/skardi/src/sources/providers/open_connector/testutil.rs
```diff
@@ -64,6 +73,93 @@ pub(crate) fn discovery_ok(
...
+                .unwrap_or_else(|| panic!("column {name} is {:?}, not Utf8", column.data_type()))
+                .clone();
+            (0..values.len()).map(move |i| {
+                assert!(!values.is_null(i), "column {name} is NULL at row {i}");
```
yiyi-wang-skardi: Two notes on this assertion — it's the only real behavior change in the PR beyond slack's convergence, and it's a good one, so it's worth getting the message exactly right.

**The row index is batch-local.** `i` restarts at 0 for each batch, so on a multi-batch result (which is precisely what `column_values` exists to flatten — several of the pagination tests return two or three batches) the panic can say "NULL at row 0" for a row that is actually the 40th in `Vec` order. A running offset keeps the number meaning what the caller thinks it means:

```rust
let mut offset = 0;
// ...per batch: assert!(!values.is_null(i), "column {name} is NULL at row {}", offset + i);
```

Or, if threading state through the `flat_map` isn't worth it, just say "at row {i} of a batch" so the number doesn't over-promise.

**No regression today, for the record.** Previously a NULL slot silently yielded `""` (`StringArray::value` reads the offset buffer regardless of the validity bitmap), so this genuinely changes what happens on a NULL. I checked all 22 call sites across gmail/google_drive/one_drive/outlook — every one reads an identity column (`id`, `message_id`, `thread_id`) that the fixtures always populate, so nothing flips. Worth stating in the doc comment that this is a deliberate tightening rather than a verbatim move, since the rest of the module is verbatim.

crates/skardi/src/sources/providers/open_connector/packs/feishu.rs
```diff
@@ -110,14 +110,14 @@ mod tests {
     use crate::sources::providers::open_connector::row_path::RowPath;
     use crate::sources::providers::open_connector::source_pack::SourcePackTable;
     use crate::sources::providers::open_connector::testutil::{
-        EnvVarGuard, MockGateway, MockResponse, discovery_ok, envelope_ok,
-        fingerprint_uncovered_columns,
+        EnvVarGuard, MockGateway, MockResponse, boolean, collect, discovery_ok, envelope_ok,
```
yiyi-wang-skardi: `column_values` is imported one pack over, but `ids_of` survives here — and it's the same function.

`fn ids_of` at feishu.rs:545, notion.rs:420 and slack.rs:583 are byte-identical to each other (I hashed the three bodies: same digest), and each is exactly `column_values(batches, "id")` with `"id"` hardcoded and different `expect` strings. discord.rs inlines the same `flat_map`/`downcast`/`value(i).to_string()` loop twice more, at lines 598 and 892.

That's five more copies of a helper this PR just centralized, and unlike `timestamp`/`millis`/`int64` they aren't a different *type* of accessor — they're the lifted one. The "deliberately not moved" list in the description doesn't mention them, so I read this as an oversight rather than a call.

The fix is import-and-delete: add `column_values` to the testutil import in these three packs, replace `ids_of(&batches)` with `column_values(&batches, "id")` at the call sites, drop the three definitions. Fine as a follow-up if you'd rather keep this PR's diff purely mechanical.

crates/skardi/src/sources/providers/open_connector/testutil.rs
```diff
@@ -64,6 +73,93 @@ pub(crate) fn discovery_ok(
...
+                values.value(i).to_string()
+            })
+        })
+        .collect()
+}
+
+/// The sorted top-level keys of one recorded executor input, for asserting
+/// exactly which inputs traveled on the wire.
+pub(crate) fn input_keys(input: &Value) -> Vec<&str> {
```
yiyi-wang-skardi, line 168: The largest duplicate left in this family is `execute_inputs`, at **seven** copies — and it would slot in right here next to `input_keys`, which is the helper its callers always pair it with.

Hashing the bodies gives exactly two shapes:
- with `action_path` — gmail.rs:740, google_drive.rs:1327, one_drive.rs:1454, outlook.rs:986 (4 identical)
- without — discord.rs:528, feishu.rs:561, notion.rs:436 (3 identical)

And the second shape is the first one specialized: the only difference is the `r.path.ends_with(action_path)` conjunct, and `ends_with("")` is always true. So a single shared

```rust
pub(crate) fn execute_inputs(gateway: &MockGateway, action_path: &str) -> Vec<Value>
```

covers all seven, with the three single-action packs passing `""`. That's ~70 lines for one definition, on top of this PR's 265.

Same caveat as the `ids_of` note — happy to see it as a follow-up rather than growing this diff.

crates/skardi/src/sources/providers/open_connector/packs/github.rs
```diff
@@ -631,7 +631,7 @@ bindings:
     use crate::sources::hierarchy::HierarchyLevel;
     use crate::sources::providers::open_connector::testutil;
     use crate::sources::providers::open_connector::testutil::{
-        MockGateway, MockResponse, RecordedRequest, discovery_ok, envelope_ok,
+        MockGateway, MockResponse, RecordedRequest, collect, discovery_ok, envelope_ok,
```
yiyi-wang-skardi: github takes `collect` from testutil here but keeps `strings` (line 379) and `bools` (line 397) — which are the lifted `utf8` and `boolean` verbatim, renamed and with the parameter spelled `column` instead of `name`.

The description's "deliberately not moved" rationale covers single-type accessors like `timestamp`/`int64`/`uint64`, whose sharing genuinely isn't paying for itself yet. But `strings`/`bools` aren't that case — they're two of the six helpers this PR is centralizing, hiding under pack-local names, in the one pack that only picked up `collect`. `u64s`/`timestamps`/`string_list` next to them are the real single-type accessors and should stay.

Since the import line is already being touched, this is a rename-only change: add `boolean, utf8` to the import, `s/strings(/utf8(/` and `s/bools(/boolean(/` at the call sites, delete the two definitions. The panic text is already identical, so no assertion output moves.

crates/skardi/src/sources/providers/open_connector/packs/outlook.rs
```diff
@@ -185,8 +185,9 @@ mod tests {
     use crate::sources::providers::open_connector::row_path::RowPath;
     use crate::sources::providers::open_connector::source_pack::{FixedValue, SourcePackTable};
     use crate::sources::providers::open_connector::testutil::{
-        EnvVarGuard, MockGateway, MockResponse, discovery_ok, envelope_err, envelope_ok,
-        fingerprint_uncovered_columns,
+        EnvVarGuard, MockGateway, MockResponse, boolean, collect, column_values,
```
yiyi-wang-skardi, line 188: Nit, free while you're in this file: outlook now imports `utf8` and `boolean`, but two spots a few hundred lines down still hand-roll the same downcast.

- **outlook.rs:388-393** — `batch.column_by_name("conversation_id").expect("column").as_any().downcast_ref::<StringArray>().expect("Utf8")`, three lines after the same column is read via `utf8(&batch, "conversation_id")` at line 385. Collapses to `utf8(&batch, "conversation_id").iter().flatten()...`.
- **outlook.rs:445-450** — `let hidden: &BooleanArray = ...downcast_ref().expect("Boolean column")` becomes `let hidden = boolean(&batch, "is_hidden");`.

Both also pick up the better panic text for free (naming the column and the actual Arrow type, instead of `"column"` / `"Utf8"`). Note that dropping the second one removes the last use of `BooleanArray` in this module, so its arrow import would need trimming too — worth a `cargo check` after.

yiyi-wang-skardi (approved):
## Audit: lift the shared pack-test Arrow/SQL accessors into `testutil`

Test-only change — `testutil` is behind `#[cfg(test)]` (`open_connector/mod.rs:52`), so nothing here reaches a production build. No `unsafe`, no network/credential surface, no new dependency. Six mechanical helpers move into one place and 9 pack suites point at them; net −265 lines.

### Correctness — verified mechanically

I reconstructed every removed helper body from `main` and hashed it against the lifted version:

- `collect`, `convert_page`→`convert_first_page`, `utf8`, `boolean`, `column_values`, `input_keys` — byte-identical across all donor packs, exactly as the PR body claims. Slack's `utf8`/`boolean` were the only divergence (bare `unwrap()`), and converging them onto the named-panic variant is a strict improvement.
- **No shadowing**: no pack still defines a helper it now imports (would be `E0255`).
- **No unused imports**: I stripped every `use` statement and re-scanned each pack body. Every imported testutil symbol is used, and every arrow import kept/dropped matches actual remaining use — `StringArray` correctly dropped from `one_drive` only, `BooleanArray` dropped from exactly the five packs that no longer name it, kept in `github`/`outlook` where local `bools`/inline downcasts still need it. Clean under `-D warnings`.
- **No stray call sites**: zero `convert_page(` references remain; all 20 call sites use `convert_first_page`.
- The `Array`-trait fix in `cd0138f` is the right diagnosis: `len()`/`is_null()` on a concrete `StringArray` need the trait in scope, while `utf8`/`boolean` resolve through `dyn Array`'s object candidates and don't.
- `convert_first_page` keeps the borrow-from-temporary shape verbatim (`RowPath::parse(..).rows(page, 1)`), which is sound because `rows` ties its output lifetime to `page`, not `self`.

The one genuine behavior change beyond slack's convergence is the new NULL assertion in `column_values`. I checked all 22 call sites: every one reads an identity column (`id` / `message_id` / `thread_id`) whose fixtures always carry a value, so nothing regresses. Detail on the panic text inline.

### Simplicity / DRY

The lift itself is clean. What I'd flag is that the sweep stopped a little short of its own thesis — three families of byte-identical duplication in the *same* helper class survive and aren't in the PR's "deliberately not moved" list:

| helper | copies | status |
|---|---|---|
| `ids_of` | 3 identical (feishu, notion, slack) + 2 inlined (discord) | is literally `column_values(batches, "id")` |
| `execute_inputs` | 4 identical + 3 identical | the 3 are the 4 with `action_path == ""` |
| `strings` / `bools` (github) | 1 each | the lifted `utf8`/`boolean`, renamed |

None of this blocks the PR — it's the same follow-up shape the author already scoped for `#216` and `table_functions.rs`. Details inline.

### Verification note

I could not compile locally (no Rust toolchain in this environment), so the above is static analysis. `Docs` and `Markdown links` are green; the `Test` job was still running when I reviewed — this sign-off rides on it going green.

**Verdict: no blocking issues.** Nice mechanical refactor with an honest, well-scoped PR description; the findings below are all cleanups for this PR or a follow-up.

PR #230, opened 2026-08-27
docs(readme): give each Agent Skills host its own install commands

Userneima:
## Why

The Install section already named Codex, Cursor, Pi, dsh, OpenClaw and Hermes, but only Claude Code could be followed without leaving the page — everyone else got a pointer to another repo. Reading the section, it still looks like a Claude Code–only project.

## What changed

- **Per-host commands, inline.** One copy into `~/.agents/skills/` covers Codex, Cursor, Pi and dsh (they share that cross-tool convention). OpenClaw installs through its own CLI; Hermes reads `~/.hermes/skills/`.
- **Fixed a 404.** The `auto_context` link still pointed at the pre-rename path; the directory is `auto-context` now.
- **Added `retrieval`** next to `auto-context`, since both skills ship from the same repo.

Native per-host directories, project-scoped installs and the Hermes `external_dirs` option stay in the [skardi-skills README](https://github.com/SkardiLabs/skardi-skills#installation) — this section gives the shortest path that works, not the full matrix.

## Verification

- `python3 scripts/check_markdown_links.py` → all relative links resolve
- Directory names checked against `skardi-skills@main`: `auto-context`, `retrieval`

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ skardi run weekly-churn -p window='7 days'
 curl -X POST localhost:8080/weekly-churn/execute \
   -H 'Content-Type: application/json' -d '{"window": "7 days"}'
 # same pipeline as an MCP tool — hosts without a shell (Claude Desktop, …)
+# from a source checkout until `skardi mcp` ships in a tagged release
 skardi mcp
 ```
 
--- a/README.md
+++ b/README.md
@@ -171,13 +171,43 @@ you:
 /plugin install auto-context@skardi-skills
 ```
 
-[`auto_context`](https://github.com/SkardiLabs/skardi-skills/tree/main/auto_context)
+[`auto-context`](https://github.com/SkardiLabs/skardi-skills/tree/main/auto-context)
 turns a folder of documents — or a datastore you already run — into governed,
 searchable context served over HTTP by `skardi-server`: hybrid search (vector +
 FTS + RRF), defaulting to a local SQLite file the skill creates and owns, or
-pointed at Postgres + pgvector, MongoDB, or Lance. The same skill also runs on
-Codex, Cursor, Pi, dsh, OpenClaw, and Hermes; per-host installation instructions
-are in the [skardi-skills README](https://github.com/SkardiLabs/skardi-skills#installation).
+pointed at Postgres + pgvector, MongoDB, or Lance.
+[`retrieval`](https://github.com/SkardiLabs/skardi-skills/tree/main/retrieval)
+teaches the agent to answer questions from a `skardi-server` you already run.
+
+**Other Agent Skills hosts** — the same two skills run on Codex, Cursor, Pi,
+dsh, OpenClaw and Hermes. They differ only in where the skill directory goes,
+and all of them install from a checkout:
+
+```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -171,13 +171,43 @@ you:
...
 searchable context served over HTTP by `skardi-server`: hybrid search (vector +
 FTS + RRF), defaulting to a local SQLite file the skill creates and owns, or
-pointed at Postgres + pgvector, MongoDB, or Lance. The same skill also runs on
-Codex, Cursor, Pi, dsh, OpenClaw, and Hermes; per-host installation instructions
-are in the [skardi-skills README](https://github.com/SkardiLabs/skardi-skills#installation).
+pointed at Postgres + pgvector, MongoDB, or Lance.
+[`retrieval`](https://github.com/SkardiLabs/skardi-skills/tree/main/retrieval)
+teaches the agent to answer questions from a `skardi-server` you already run.
+
+**Other Agent Skills hosts** — the same two skills run on Codex, Cursor, Pi,
```
BtXin: Let's not say Claude Code and others, sounds like we value Claude Code over other agents. Put all of them in the same priority.

README.md
```diff
@@ -171,13 +171,43 @@ you:
...
+dsh, OpenClaw and Hermes. They differ only in where the skill directory goes,
```
BtXin: Let's add OpenCode as well.

Userneima:
Both applied in cf3a577.

**Same priority for every host.** The hosts are now one flat list. What the two skills do is stated once up front — it does not depend on the host — and then each host gets the same bold label and its own install line. Claude Code keeps the marketplace commands because that is genuinely how it installs there, but it is no longer introduced as "the fastest path" ahead of a grouped "other hosts" section.

**OpenCode added.** Its [docs](https://opencode.ai/docs/skills/) list `~/.agents/skills/` among the global skill directories it loads (alongside `~/.config/opencode/skills/` and `~/.claude/skills/`), so it joins the four hosts already covered by the single cross-tool copy instead of needing a block of its own.

One thing worth flagging while you look: of the seven hosts named here, only OpenClaw has actually been installed and run by us. The rest follow each host's published documentation. The skardi-skills README states that boundary explicitly; this page currently links there without repeating it. Happy to add the same sentence here if you would rather it be visible on the front page.

BtXin (approved):

README.md
Userneima: Applied in cf3a577. Claude Code was introduced as "the fastest path" with its own section and everything else grouped under "Other Agent Skills hosts" — that ranking is gone. The hosts are now one flat list: what the two skills do is stated once up front (it does not depend on the host), then every host gets the same bold label and its own install line. Claude Code keeps the marketplace commands only because that is genuinely how it installs, not because it comes first.

README.md
Userneima: Added in cf3a577. OpenCode's docs list `~/.agents/skills/` among the global skill directories it loads, alongside `~/.config/opencode/skills/` and `~/.claude/skills/`, so it joins the four hosts already covered by the single cross-tool copy rather than needing its own block — five hosts on that one line now.

PR #231, opened 2026-08-27
docs(skills): graph-source — connect and query a property graph (Cypher over AGE) end-to-end

bakey:
A shared Claude Code skill (`docs/superpowers/skills/graph-source/`, installed per the README's symlink convention) that takes a developer from an Apache AGE instance to queryable SQL tables in Skardi, end to end. Sibling to `source-pack`: that one onboards SaaS/API sources, this one onboards graph sources.

## Structure

- **SKILL.md** — five phases (provision → declare → verify registration → query → pipeline), a view-vs-ad-hoc decision guide, and the working style (verify against the running system; env-var names only; don't widen bounds to hide symptoms).
- **references/provisioning.md** — the least-privilege `kg_reader` recipe, why `LOAD 'age'` is best-effort and what to do instead (never superuser), a local dev AGE.
- **references/configuration.md** — the full `type: graph` contract (verified against `sources/providers/graph/config.rs`: defaults 30 s / 10 000 rows / 4 conns and their valid ranges), the lowercase view-name rule and its DataFusion ident-folding rationale, "a view's bound lives in its Cypher", and the healthy/degraded/refused registration semantics including what recovery does and does not answer.
- **references/querying.md** — positional `columns` binding, node columns as `STRUCT<id, labels, properties>` with `.properties` as the JSON text, the getter-by-stored-type rule, the deliberately-absent `->`/`->>` rewrite, and the one working `{params}` pipeline spelling.
- **references/troubleshooting.md** — symptom → diagnosis table, the three failure families (connectivity / contract / bounds), and the cloud caveat: skardi-cloud's gateway projection deliberately drops the `status` trio (its contexts/login design §7.4.1), so tenant-side `skardi schema` cannot distinguish a degraded graph source from a healthy empty one — diagnose on the operator side.

**Scope is AGE-only on purpose** — the config carries `backend:` for neo4j/kuzu, but writing guidance for unshipped backends would manufacture fake contracts (the same lesson source-pack's history teaches about un-reconciled contracts).

## Evaluation before submission

Three realistic tasks (onboard an AGE graph with two views; diagnose a `RowCapExceeded` on a one-row `WHERE`; build a parameterized pipeline), each run twice by independent agents — once with the skill, once as a no-skill baseline **inside this repo** (the honest baseline: it can and did find `docs/graph.md`). Graded against 15 written assertions:

| | with skill | baseline |
|---|---|---|
| assertions passed | **15/15** | 13/15 |
| mean tokens | 43.2k | 65.2k (−34%) |
| mean wall-clock | 88 s | 134 s (−34%) |

The two baseline failures are the ones that matter: its onboarding YAML shipped **both views with unbounded Cypher** (the production `RowCapExceeded` trap — it knew the rule, warned in prose, and didn't apply it to its own deliverable), and its verification query passed the whole node STRUCT to `json_get_str` instead of `col.properties` (the query fails). On the diagnosis and pipeline tasks both sides were fully correct — there the skill's value is the 2.5× token/time gap on the diagnosis task (42k/72s vs 89k/177s), since the baseline re-derives from source every time.

The eval also improved the skill: the lowercase view-name rule (which the baseline found in `config.rs` and the draft lacked) and the STRUCT `.properties` access spelling were added because the runs surfaced them.

## Not in this PR

- No `.gitignore` change and no tracked `.claude/` — installation stays the README's per-clone symlink, per the repo's convention.
- Eval workspace/transcripts are session-local scratch, not committed; the summary above is the durable record.

### DIFF
--- a/docs/superpowers/skills/README.md
+++ b/docs/superpowers/skills/README.md
@@ -16,3 +16,4 @@ the link points into the repo.
 | Skill | Purpose |
 | --- | --- |
 | [source-pack](source-pack/SKILL.md) | Develop a new Open Connector source pack end-to-end: live contract reconciliation, implementation under the admission gate, self-review against the repo's review standards, PR submission. |
+| [graph-source](graph-source/SKILL.md) | Connect a property graph (Cypher over AGE) to Skardi end-to-end: least-privilege provisioning, `type: graph` declaration and views, registration-state verification, correct query patterns, pipeline parameterization, troubleshooting. |
--- a/docs/superpowers/skills/graph-source/SKILL.md
+++ b/docs/superpowers/skills/graph-source/SKILL.md
@@ -1,0 +1,144 @@
+---
+name: graph-source
+description: >-
+  Connect a property graph (knowledge graph, GraphRAG corpus, Apache AGE /
+  openCypher data) to Skardi and query it through SQL, end to end: provision
+  the AGE backend with a least-privilege role, declare the `type: graph`
+  source and its views in context YAML, verify registration health, write
+  correct queries (`cypher_query`, `graph_schema`, the JSON getters), and
+  wire Cypher parameters into pipelines. Use this skill whenever the user
+  wants graph data queryable in Skardi, mentions graphs, Cypher, AGE,
+  Neo4j-style data, knowledge graphs, GraphRAG, `kg.*` tables,
+  `cypher_query`, or is debugging a degraded graph source, a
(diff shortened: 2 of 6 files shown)


bakey:
Relocating: user-facing skills belong in the skardi-skills marketplace, not the engine repo's internal-dev skill tree — resubmitted as SkardiLabs/skardi-skills#33 with the same content (plus plugin/marketplace manifests) and the same eval evidence.

PR #233, opened 2026-08-28
feat: author-written parameter descriptions (spec.parameters)

gracexmatin:
## Goal

Tool schemas generated for MCP carry only types (`{"type": ["string", "null"]}`) — the model has to guess a parameter's semantics from its name. This adds the missing input channel: a pipeline author writes one line of prose per parameter, and it lands verbatim in front of the model.

## Shape

```yaml
spec:
  parameters:
    query: 'Search terms (web-search syntax: AND, "phrase", or, -not)'
    limit: "Maximum number of results"
  query: |
    SELECT id, content, _score
    FROM pg_fts('documents', 'content', {query}, {limit})
```

- **Documentation, not declaration.** The parameter set still comes from the SQL placeholders; `spec.parameters` is optional per pipeline and per parameter.
- **Strict at load.** A described name the SQL does not declare (a typo, or a doc string gone stale after a query edit) fails the load, with the declared parameter list in the error — the published schema never lies.

## Flow

1. **skardi** — `build_from_parsed` reads the optional `spec.parameters` map into `StandardPipeline.parameter_docs` and validates every key against the inferred request schema.
2. **server** — `enriched_parameters` (the single home of the parameter shape) folds each one-liner into that parameter's `json_schema` as the standard JSON Schema `description` keyword, on both `GET /pipelines` and `GET /pipeline/:name`.
3. **bridge** — zero changes: `skardi mcp` already copies `json_schema` verbatim into tool properties, so descriptions reach MCP hosts on the next `tools/list`.

## Tests

- skardi: descriptions load; undocumented parameters have no entry; unknown described name rejects the file with a useful error.
- server: described `brand` publishes `description` inside `json_schema` on both inventory endpoints; undocumented `max_price` carries no `description` key.

## Docs & demo

`docs/pipelines.md` gains a *Parameter descriptions* section; `docs/mcp.md` points authors at it. One demo (`demo/rag/server/pipelines/search_fulltext.yaml`) converts its `# Parameters:` comment block into the structured form as the exemplar — the ~30 remaining demo comment blocks are left for a follow-up sweep.

### DIFF
--- a/crates/server/src/pipeline_handlers.rs
+++ b/crates/server/src/pipeline_handlers.rs
@@ -392,11 +392,10 @@ fn enriched_parameters(pipeline: &StandardPipeline) -> Vec<Value> {
             // The author's one-liner rides inside the JSON Schema itself
             // (standard `description` keyword), so schema consumers — the
             // MCP bridge copies it verbatim into the tool's properties —
-            // surface it with no changes of their own.
+            // surface it with no changes of their own. The loader already
+            // normalized: entries are trimmed and never blank.
             if let Some(doc) = pipeline.parameter_docs.get(name) {
-                if !doc.trim().is_empty() {
-                    schema["description"] = serde_json::json!(doc);
-                }
+                schema["description"] = serde_json::json!(doc);
             }
             serde_json::json!({
                 "name": name,
--- a/crates/server/src/pipeline_handlers.rs
+++ b/crates/server/src/pipeline_handlers.rs
@@ -388,10 +388,20 @@ fn enriched_parameters(pipeline: &StandardPipeline) -> Vec<Value> {
     fields
         .into_iter()
         .map(|(name, field)| {
+            let mut schema = param_json_schema(&field.field_type, sql, name);
+            // The author's one-liner rides inside the JSON Schema itself
+            // (standard `description` keyword), so schema consumers — the
+            // MCP bridge copies it verbatim into the tool's properties —
+            // surface it with no changes of their own.
+            if let Some(doc) = pipeline.parameter_docs.get(name) {
+                if !doc.trim().is_empty() {
+                    schema["description"] = serde_json::json!(doc);
+                }
+            }
             serde_json::json!({
                 "name": name,
                 "data_type": format!("{:?}", field.field_type),
-                "json_schema": param_json_schema(&field.field_type, sql, name),
+                "json_schema": schema,
             })
         })
         .collect()
--- a/crates/skardi/src/pipeline/pipeline.rs
+++ b/crates/skardi/src/pipeline/pipeline.rs
@@ -490,11 +490,35 @@ impl StandardPipeline {
         };
 
         // Optional author-written parameter descriptions (name → one line).
-        let parameter_docs: HashMap<String, String> = match spec_section.get("parameters") {
+        // Normalized right here so downstream consumers publish the text
+        // as-is: YAML block scalars keep a trailing newline (trimmed away),
+        // and a blank description is an author slip, not a request to
+        // publish nothing.
+        let raw_docs: HashMap<String, String> = match spec_section.get("parameters") {
             Some(section) => serde_yaml::from_value(section.clone())
                 .map_err(|e| anyhow!("Failed to parse spec.parameters section: {}", e))?,
             None => HashMap::new(),
         };
+        let mut parameter_docs = HashMap::with_capacity(raw_docs.len());
+        let mut blank: Vec<String> = Vec::new();
+        for (name, doc) in raw_docs {
+            let doc = doc.trim();
+            if doc.is_empty() {
+                blank.push(name);
+            } else {
+                parameter_docs.insert(name, doc.to_string());
+            }
+        }
+        // All offenders at once, sorted: HashMap iteration order is
+        // arbitrary, and naming one random slip per restart would make the
+        // author fix them one server boot at a time.
+        if !blank.is_empty() {
+            blank.sort();
+            return Err(anyhow!(
+                "spec.parameters gives blank descriptions for {:?} — write one line or drop the key",
+                blank,
+            ));
+        }
 
         let inferencer = SqlSchemaInferrer::new(ctx.clone())?;
         let request_schema = inferencer
@@ -503,18 +527,21 @@ impl StandardPipeline {
 
         // A described parameter the SQL does not declare is a typo or a doc
         // string gone stale after a query edit — refuse to load rather than
-        // publish documentation that lies.
-        for documented in parameter_docs.keys() {
-            if !request_schema.fields.contains_key(documented) {
-                let mut declared: Vec<&String> = request_schema.fields.keys().collect();
-                declared.sort();
-                return Err(anyhow!(
-                    "spec.parameters describes '{}', but the SQL declares no {{{}}} placeholder (declared parameters: {:?})",
-                    documented,
-                    documented,
-                    declared,
-                ));
-            }
+        // publish documentation that lies. Same all-at-once, sorted
+        // reporting as the blank check above.
+        let mut unknown: Vec<&String> = parameter_docs
+            .keys()
+            .filter(|k| !request_schema.fields.contains_key(*k))
+            .collect();
+        if !unknown.is_empty() {
+            unknown.sort();
+            let mut declared: Vec<&String> = request_schema.fields.keys().collect();
+            declared.sort();
+            return Err(anyhow!(
+                "spec.parameters describes {:?}, but the SQL declares no such placeholder (declared parameters: {:?})",
+                unknown,
+                declared,
+            ));
         }
 
         let response_schema = inferencer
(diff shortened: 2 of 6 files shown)


yiyi-wang-skardi:
@yiyi-wang-skardi is now reviewing this PR using Claude Code — opus (reasoning effort max). Target HEAD commit: https://github.com/SkardiLabs/skardi/commit/8392ffc36bdbd0b6f729bd8aebaad6767972c5b3 (`8392ffc`).

crates/server/src/pipeline_handlers.rs
```diff
@@ -388,10 +388,20 @@ fn enriched_parameters(pipeline: &StandardPipeline) -> Vec<Value> {
     fields
         .into_iter()
         .map(|(name, field)| {
+            let mut schema = param_json_schema(&field.field_type, sql, name);
+            // The author's one-liner rides inside the JSON Schema itself
+            // (standard `description` keyword), so schema consumers — the
+            // MCP bridge copies it verbatim into the tool's properties —
+            // surface it with no changes of their own.
+            if let Some(doc) = pipeline.parameter_docs.get(name) {
+                if !doc.trim().is_empty() {
```
yiyi-wang-skardi: Nit (KISS/DRY): the emptiness rule is checked on the trimmed value but the **untrimmed** value is inserted. YAML block scalars keep a trailing newline, so a perfectly reasonable longer description written as

```yaml
parameters:
  brand: >
    Exact brand name as it appears in the catalog; null matches every brand
```

publishes `"description": "Exact brand name …brand\n"` verbatim into every `tools/list` response. (`metadata.description` in the demo already uses `>`, so this form is likely to show up.)

More broadly, this splits `spec.parameters` validation across two crates: `build_from_parsed` is the declared home of the rules ("strict at load"), yet `brand: ""` — exactly the kind of author slip that stance exists to catch — loads clean and then silently vanishes here. Suggest normalizing and validating once in the loader (trim on read, reject a blank description alongside the unknown-key check) and letting this site be a plain `if let Some(doc) = pipeline.parameter_docs.get(name) { schema["description"] = json!(doc); }`. That also removes the currently-untested `!doc.trim().is_empty()` branch rather than needing a test for it.

crates/skardi/src/pipeline/pipeline.rs
```diff
@@ -481,21 +489,47 @@ impl StandardPipeline {
...
+
         let inferencer = SqlSchemaInferrer::new(ctx.clone())?;
         let request_schema = inferencer
             .extract_request_schema(&query_definition.sql)
             .await?;
+
+        // A described parameter the SQL does not declare is a typo or a doc
+        // string gone stale after a query edit — refuse to load rather than
+        // publish documentation that lies.
+        for documented in parameter_docs.keys() {
```
yiyi-wang-skardi: Nit: `parameter_docs` is a `HashMap`, so with two typo'd keys the error names an arbitrary one and a re-run can name the other. The author then fixes them one server restart at a time. `declared` is sorted just below for exactly this determinism reason — the outer loop deserves the same:

```rust
let mut unknown: Vec<&String> = parameter_docs
    .keys()
    .filter(|k| !request_schema.fields.contains_key(*k))
    .collect();
if !unknown.is_empty() {
    unknown.sort();
    let mut declared: Vec<&String> = request_schema.fields.keys().collect();
    declared.sort();
    return Err(anyhow!(
        "spec.parameters describes {:?}, but the SQL declares no such placeholder (declared parameters: {:?})",
        unknown, declared,
    ));
}
```

One shot, stable output, and it also drops the per-iteration rebuild of `declared`.

crates/skardi/src/pipeline/pipeline.rs
```diff
@@ -481,21 +489,47 @@ impl StandardPipeline {
...
+            None => HashMap::new(),
+        };
+
         let inferencer = SqlSchemaInferrer::new(ctx.clone())?;
         let request_schema = inferencer
             .extract_request_schema(&query_definition.sql)
             .await?;
+
+        // A described parameter the SQL does not declare is a typo or a doc
+        // string gone stale after a query edit — refuse to load rather than
```
yiyi-wang-skardi: The fail-fast choice looks right to me, but the comment's stated reason isn't quite what happens: an unknown key can never "publish documentation that lies", because `enriched_parameters` looks descriptions up *by parameter name* — a stale key is simply never read. The actual (and sufficient) justification is catching author typos at load instead of letting a description silently go missing from the published schema.

Worth being precise here because the blast radius is real: `crates/server/src/config.rs:559` `?`-propagates every pipeline load failure, so this makes a typo in a *comment-grade* field the first thing that can abort startup for the entire server, all other pipelines included. Every pre-existing load failure meant "this pipeline cannot work." That's a defensible trade for a loud, precise, deploy-time error — just worth saying out loud in the comment so the next reader doesn't relax the check on the mistaken belief that it only guards published output.

crates/skardi/src/pipeline/pipeline.rs
```diff
@@ -481,21 +489,47 @@ impl StandardPipeline {
             parameters: Vec::new(),
         };
 
+        // Optional author-written parameter descriptions (name → one line).
+        let parameter_docs: HashMap<String, String> = match spec_section.get("parameters") {
```
yiyi-wang-skardi: Heads-up on the shared builder: `crates/skardi/src/jobs/definition.rs:181` also routes through `build_from_parsed`, so a `kind: job` YAML may carry `spec.parameters`, gets it strictly validated (a typo fails the job load), and then never surfaces it anywhere — `GET /jobs` publishes only `request_schema.fields.keys()`, no per-parameter JSON Schema.

Harmless and arguably the right default, but it's a silent no-op an author can't distinguish from a working feature. A clause in the `docs/pipelines.md` section saying descriptions are validated for jobs too but only published for pipelines would settle it, without touching code.

yiyi-wang-skardi (approved):
## Audit — `spec.parameters` author-written parameter descriptions

Small, well-scoped change: 3 commits, 6 files, +171/-13. The design is right — the parameter *set* still comes from the SQL placeholders, `spec.parameters` is pure documentation folded into the existing `json_schema` as the standard JSON Schema `description` keyword, and it lands in exactly one place (`enriched_parameters`) that both inventory endpoints already share.

### What I verified

- **The "zero bridge changes" claim holds.** `crates/cli/src/mcp/projection.rs:135` copies `json_schema` verbatim into tool properties whenever it is an object, and `param_json_schema` returns an object on every branch (including the `json!({})` unknown-type fallback). So `description` reaches MCP hosts unchanged, and `schema["description"] = …` cannot hit serde_json's non-object `IndexMut` panic.
- **The strict load check cannot produce false positives.** `extract_request_schema` → `extract_parameters` → `extract_named_parameters` builds its map 1:1 from `parameter_order`, i.e. every `{name}` the regex finds in the SQL. Validating described keys against `request_schema.fields` is therefore equivalent to validating against the declared placeholder set — a correctly-spelled name can never be rejected because the inferrer dropped it.
- **No backwards-compatibility break.** `spec.parameters` is unused across every YAML in the repo except the one this PR adds, and nothing that deserializes a pipeline `spec` is strict — `MinimalSpec` in `crates/server/src/config.rs:688` (which runs *before* the full load in the startup loop) has no `deny_unknown_fields`, so the new key is ignored there.
- **The converted demo loads.** `demo/rag/server/pipelines/search_fulltext.yaml` describes exactly `{query}` and `{limit}`, which are exactly the placeholders its SQL declares.
- **Test fixture change is safe.** `spec.parameters` was added to the shared `make_app_state` YAML; the only assertions that compare a whole `json_schema` are the four updated ones, and `max_price` keeps good negative coverage (asserts the exact object, so a stray `description` would fail).
- **`#[serde(skip)]` costs nothing.** The only serializer of `StandardPipeline` is `to_yaml()`, an unused placeholder whose `request_schema`/`response_schema` are already skipped.

No correctness, safety, or security defects found. Pipelines are loaded from operator-controlled files on disk (there is no create-pipeline endpoint), so a description reaching a model verbatim sits inside the same trust boundary as the SQL that is already there.

**Caveat:** this environment has no Rust toolchain, so I could not run `cargo build` or the test suite — verification above is static reading only.

The four notes below are all non-blocking polish (KISS/DRY and error-message ergonomics).

docs/pipelines.md
```diff
@@ -66,6 +66,31 @@ Each `POST /<name>/execute` body is a JSON object keyed by placeholder
...
+### Parameter descriptions
+
+An optional `spec.parameters` map attaches a one-line, human-written
+description to any parameter:
+
+```yaml
+spec:
+  parameters:
+    brand: "Exact brand name; null matches every brand"
+    max_price: "Upper price bound in USD, inclusive"
```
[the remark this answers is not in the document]
gracexmatin: Fixed with one word.

crates/skardi/src/pipeline/pipeline.rs
gracexmatin: Fixed. 
Also extended the same treatment to the blank-description check that landed in the previous round: it had the identical shape (first arbitrary HashMap key wins, one fix per server restart), so both validations now share the collect → sort → report-all pattern, and the loader stays deterministic end to end.
The regression test now feeds two typo'd keys and asserts both names land in a single error message, which pins the all-at-once behavior rather than just the rejection.

One small correction for the record: the old code returned on the first offender, so declared was only ever built once — the real wins here are the stable output and the one-pass fix, not the rebuild. Thanks for catching the double standard right next to the line that sorts declared for exactly this reason.

PR #234, opened 2026-08-28
docs(specs): MCP streamable HTTP transport design — /mcp in skardi-server

gracexmatin:
## Summary

Design spec for the follow-up milestone named in the stdio binding spec's non-goals (#227): skardi-server itself speaks MCP over streamable HTTP at `/mcp`, so hosts that cannot spawn a local binary — claude.ai, mobile clients, hosted agent platforms — connect to a Skardi deployment directly. The stdio bridge stays as the local entry point.

## Decisions captured

- **Auth: existing Bearer, forwarded verbatim** — every tool call becomes a synthetic in-process request carrying the `/mcp` caller's own `Authorization` header, so enforcement and audit stay entirely in the REST layer. OAuth is explicitly deferred to the Cloud side.
- **Default-on, no config flag** — `/mcp` exposes no privilege the REST surface doesn't already expose under the same auth.
- **Architecture: tower self-dispatch** — the MCP handler is a protocol adapter holding a clone of the server's own `Router`, dispatching via `tower::ServiceExt::oneshot` (the pattern the HTTP tests already use). Zero refactor of the execution path; one inventory contract (`GET /pipelines` enriched JSON) shared with the bridge.
- **Shared crate `crates/mcp-core` (`skardi-mcp-core`)** — projection, built-in name constants, and `encode_component` move out of the CLI; the dispatch `match` deliberately stays per-binding (different error types, different session carriers).
- **`listChanged` stays a non-goal** — the server has no production config hot-reload, so the inventory only changes across a restart, which tears down every MCP session anyway.

## Open probes (resolve at plan time)

- R1: how rmcp 3.1.4's streamable-HTTP server exposes inbound headers and session ids (fallback design included in the spec).
- R2: axum interop via `nest_service` (low-risk, confirmed by the same probe).

Spec only — no code changes.

### DIFF
--- a/docs/superpowers/specs/2026-08-28-mcp-http-transport-design.md
+++ b/docs/superpowers/specs/2026-08-28-mcp-http-transport-design.md
@@ -107,13 +107,14 @@ is possible. rmcp's streamable-HTTP server service is a tower `Service`, so
 
 The inverse collision exists and is accepted: the nest shadows REST's
 `/:name/execute` for a pipeline literally named `mcp` — the router
-prefers the static `mcp` segment, so inbound `POST /mcp/execute` reaches
-the nested MCP service, never that pipeline. The pipeline stays fully
-callable *through* `/mcp` (the captured router is the pre-nest one — the
-same property that makes recursion impossible) and via the stdio bridge;
-only its direct REST endpoint is lost. The mount point stays `/mcp`
-because that is the endpoint path MCP hosts preconfigure — one shadowed
-name is a smaller cost than a nonstandard endpoint.
+prefers the static `mcp` segment, so every URL-borne caller loses that
+pipeline: the REST endpoint, `skardi run`, and the stdio bridge, which
+builds the same path over real HTTP and — its `tools/list` coming from
+an unaffected `GET /pipelines` — advertises a tool whose calls then
+fail. Only `/mcp` itself still reaches it: the captured router is the
+pre-nest one. Accepted as a known limitation — no data operation is
+plausibly named `mcp`, and the mount point is the path MCP hosts
+preconfigure.
 
 ---
 
--- a/docs/superpowers/specs/2026-08-28-mcp-http-transport-design.md
+++ b/docs/superpowers/specs/2026-08-28-mcp-http-transport-design.md
@@ -26,7 +26,12 @@ The MCP layer added to skardi-server is a **protocol adapter, not a second
 execution path**. The handler holds a clone of the server's own REST
 `Router`; every tool call is translated into a synthetic HTTP request and
 dispatched in-process via `tower::ServiceExt::oneshot` — the same mechanism
-the server's HTTP tests already use. No network hop, no socket, but the
+the server's HTTP tests already use. (`ServiceExt` sits behind tower's
+`util` feature, which the server's default-features `tower = "0.4"`
+dependency gets only transitively from axum/tower-http today, and only
+tests use it; promoting `oneshot` to production code comes with an
+explicit `features = ["util"]` on the server's `tower` entry rather than
+leaning on feature unification.) No network hop, no socket, but the
 request runs the routing table and the unmodified REST handlers (see
 "Transport vs. execution middleware" below for exactly which layers a
 synthetic request does and does not traverse).
@@ -100,6 +105,16 @@ and the final router is the REST router with the MCP service nested at
 is possible. rmcp's streamable-HTTP server service is a tower `Service`, so
 `nest_service` mounts it regardless of rmcp's internal axum version.
 
+The inverse collision exists and is accepted: the nest shadows REST's
+`/:name/execute` for a pipeline literally named `mcp` — the router
+prefers the static `mcp` segment, so inbound `POST /mcp/execute` reaches
+the nested MCP service, never that pipeline. The pipeline stays fully
+callable *through* `/mcp` (the captured router is the pre-nest one — the
+same property that makes recursion impossible) and via the stdio bridge;
+only its direct REST endpoint is lost. The mount point stays `/mcp`
+because that is the endpoint path MCP hosts preconfigure — one shadowed
+name is a smaller cost than a nonstandard endpoint.
+
 ---
 
 ## Shared code: `crates/mcp-core` (package `skardi-mcp-core`)
@@ -119,8 +134,26 @@ Moves from `crates/cli/src/mcp/` into the new crate:
   contract and both dispatchers need it. The CLI re-imports it from the
   crate.
 
-Dependencies of the new crate: `rmcp = "=3.1.4"` (model types only),
-`serde_json`. Both `skardi-cli` and `skardi-server` depend on it.
+Dependencies of the new crate: `rmcp = { version = "=3.1.4",
+default-features = false }` — model types only, no features, preserving
+the intent behind the CLI's deliberate `default-features = false` pin;
+`serde_json`; and `percent-encoding` (`encode_component` is built on
+`utf8_percent_encode` / `AsciiSet`). Both `skardi-cli` and
+`skardi-server` depend on it.
+
+`skardi-server`'s own Cargo changes: `skardi-mcp-core`; `rmcp`
+(`default-features = false`, features `["server",
+"transport-streamable-http-server"]`); `uuid` for the per-request audit
+ids (already in `[workspace.dependencies]`); and the explicit tower
+`util` feature named above. One consequence stated rather than silently
+caused: with `resolver = "2"`, features unify across workspace members
+built together, so a workspace build now compiles rmcp's HTTP stack into
+the CLI binary's graph too. The comment on the CLI's rmcp entry —
+"keeps rmcp's HTTP stack (and any axum version skew) out of the
+dependency graph" (`crates/cli/Cargo.toml`) — stops being true at
+workspace-build granularity and is updated in the same PR; its axum-skew
+worry is separately answered by R2 below (rmcp has no runtime axum
+dependency).
 
 **Deliberately NOT shared:** the dispatch `match` (three arms: `query`,
 `list_data_sources`, pipeline fallthrough) stays duplicated between the
@@ -180,10 +213,27 @@ version, missing credentials would otherwise surface only as tool errors
 inside HTTP 200 responses instead of the transport-level 401 that host
 credential flows key on. A thin tower layer wrapping only the nested
 MCP service closes this: every inbound `/mcp` request passes
-`verify_session(&state, &headers)` — the same single-home check the REST
-handlers call — before reaching rmcp; failures return a transport-level
-401, which is also the signal host credential flows key on. On a no-auth
-deployment `verify_session` always allows and the gate is a no-op.
+`verify_session` — the same single-home check the REST handlers call,
+with one carrier narrowed below — before reaching rmcp; failures return
+a transport-level 401. On a no-auth deployment `verify_session` always
+allows and the gate is a no-op.
+
+The gate accepts Bearer only. On the REST surface `verify_session` takes
+the token from `Authorization` *or* falls back to the better-auth
+session cookie (`crates/server/src/auth/routes.rs`). Honoring that
+fallback here would be a trap: a cookie-only caller would pass the gate,
+but synthetic requests forward only `Authorization`, so every tool call
+would then fail handler-level `require_session` — a 401 surfacing as
+`isError` inside an HTTP 200, the exact failure shape this gate exists
+to avoid, reachable by a caller the gate deliberately admitted. So the
+gate calls `verify_session` on a headers view with `cookie` stripped:
+token validation stays single-home, the accepted carrier narrows to the
+one MCP hosts actually send, and a cookie-only caller gets the
+transport-level 401. The alternative — forwarding `cookie` alongside
+`Authorization` on synthetic requests — was rejected: it would hand a
+browser's ambient session silent access to a JSON-RPC surface no browser
+page has any business driving, and it also keeps the open
+`allowed_origins` posture below honest.
 
 One deliberate parity divergence, worth naming: with auth enabled, the
 MCP inventory (`tools/list`) requires a token even though REST's
@@ -192,6 +242,22 @@ server-side and is gated as a whole; deployments already weigh
 `/pipelines` exposure per the v1 auth notes. Handler-level
 `require_session` (reached via synthetic dispatch) stays as the second
 layer of defense.
+
+A limitation, stated rather than left to the reader: the gate
+authenticates *a* session, not *the* session that owns the MCP session.
+`verify_session` answers "is some active token present", never "which
+principal" — and auth is multi-user (better-auth sign-up,
+`sessions.user_id`). Under `legacy_session_mode`, any valid-token holder
+can therefore attach to an `Mcp-Session-Id` minted for another user: the
+gate passes and rmcp accepts. The blast radius today is small: REST
+grants every authenticated session identical privilege (no handler does
+per-user authorization), so attaching buys no data access the attacher
+did not already have — it is cross-user visibility of that MCP session's
+in-flight server→client traffic, plus corrupted audit grouping (the
+attacher's pipeline runs get the victim's session id). Binding —
+recording the principal at `initialize` and requiring a match on every
+subsequent session request — is the follow-up if per-user authorization
+ever lands.
 
 ### Public host allowlist — the DNS-rebinding guard, configured
 
@@ -218,14 +284,35 @@ default in the wild, stripping the loopback protection the default
 exists for.
 
 `allowed_origins` stays at rmcp's default (no restriction): MCP hosts
-are not browsers, the session gate is the actual barrier, and an open
-Origin posture is consistent with the existing CORS
-`mirror_request` configuration.
+are not browsers, and the session gate is the actual barrier — it
+accepts Bearer only, so a browser's ambient cookie cannot authenticate
+`/mcp` and there is no CSRF-shaped surface for Origin to guard. An open
+Origin posture is consistent with the existing CORS `mirror_request`
+configuration.
+
+### Response mode — SSE with keep-alive
+
+A tool call blocks its `/mcp` POST for the whole synthetic dispatch, and
+v1 sends no progress notifications — a multi-minute pipeline run is a
+connection with no application payload until the result arrives. Reverse
+proxies recycle connections that look idle (nginx's
+`proxy_read_timeout` defaults to 60 s), and the remote deployments this
+transport targets sit behind one almost by definition: skardi-server
+speaks plain HTTP, so TLS for remote hosts means a proxy in front. rmcp
+offers two response modes; the service is configured for SSE responses
+with `sse_keep_alive` pings (interval chosen at plan time) rather than
+`json_response`, so bytes keep flowing and proxy idle timers keep
+resetting during long calls. The stdio bridge's "no request timeout —
+the host's tool-call timeout is the backstop" reasoning stays valid here
+only because of this: the keep-alive is what prevents a proxy from
+firing before the host does. docs/mcp.md still advises checking proxy
+read timeouts, since a keep-alive only helps proxies that count any
+bytes as liveness.
 
 ### Tool call translation
 
-Identical tool surface to the stdio bridge — same projection, same three
-built-ins, same pipeline tools:
+Identical tool surface to the stdio bridge — same projection, same two
+built-ins (`query`, `list_data_sources`), same pipeline tools:
 
 | Tool call | Synthetic request |
 |---|---|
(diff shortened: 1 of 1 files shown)


yiyi-wang-skardi:
@yiyi-wang-skardi is now reviewing this PR using Claude Code — opus (reasoning effort high). Target HEAD commit: https://github.com/SkardiLabs/skardi/commit/8214b53065d074b6268f6f0f1ba803cb7514e9ab (`8214b53`).

docs/superpowers/specs/2026-08-28-mcp-http-transport-design.md
```diff
@@ -0,0 +1,443 @@
...
+`initialize` before any tool call happens, so with auth enabled an
+anonymous caller could mass-create sessions; and on every protocol
+version, missing credentials would otherwise surface only as tool errors
+inside HTTP 200 responses instead of the transport-level 401 that host
+credential flows key on. A thin tower layer wrapping only the nested
+MCP service closes this: every inbound `/mcp` request passes
+`verify_session(&state, &headers)` — the same single-home check the REST
+handlers call — before reaching rmcp; failures return a transport-level
+401, which is also the signal host credential flows key on. On a no-auth
+deployment `verify_session` always allows and the gate is a no-op.
```
yiyi-wang-skardi: **The gate authenticates *a* session, not *the* session that owns the MCP session.** `verify_session` returns `Result<(), Response>` — it answers "is some active token present", never "which principal" (`crates/server/src/auth/routes.rs:129`). Auth here is multi-user: better-auth has sign-up and `sessions.user_id` (`crates/server/src/auth/types.rs:106`). So with `legacy_session_mode` on, user B holding any valid token can attach to an `Mcp-Session-Id` minted for user A: the gate passes, rmcp accepts the session, and B rides A's MCP session — reading whatever the server→client stream for that session carries, and (per the attribution design below) getting A's session id stamped on B's pipeline runs in the ledger.

The blast radius is genuinely small today, and that's worth saying in the spec rather than leaving the reader to work out: REST grants every authenticated session identical privilege (there is no per-user authorization anywhere in the handlers), so this buys no data access B didn't already have — it's cross-user visibility of in-flight MCP traffic plus corrupted audit grouping. Suggest one paragraph in this section stating that MCP session ids are unbound to a principal and that the gate's "governs who may hold an MCP session" is a coarse any-valid-session check, with binding (record the principal at `initialize`, require a match) noted as the follow-up if per-user authorization ever lands.

docs/superpowers/specs/2026-08-28-mcp-http-transport-design.md
```diff
@@ -0,0 +1,443 @@
...
+built-ins, same pipeline tools:
+
+| Tool call | Synthetic request |
+|---|---|
+| `tools/list` | `GET /pipelines` → `projection::project` |
+| `query` | `POST /query`, body `{sql, max_rows?, ai_context?: {purpose, session_id}}` (`ai_context` only when `purpose` is provided, as in the bridge) |
+| `list_data_sources` | `GET /data_source` |
+| pipeline tool | `POST /{encoded-name}/execute`, arguments as the flat body, header `x-skardi-session-id` |
+
+Every synthetic request forwards the inbound `Authorization` header
```
yiyi-wang-skardi, line 325: **Cookie-authenticated callers pass the gate and then fail the handler.** `verify_session` accepts a Bearer token *or* a session cookie fallback (`crates/server/src/auth/routes.rs:145-155`). The gate uses `verify_session`, so a cookie-only caller is admitted to rmcp — but the synthetic request forwards only `Authorization`, so `require_session` inside the handler finds no credential and returns 401, which this design maps to `isError: true` inside an HTTP 200. That's the exact failure shape the gate section says it exists to avoid, reachable by a caller the gate deliberately let through.

Pick one and state it: forward `cookie` alongside `Authorization` on synthetic requests (keeps parity with REST, since REST accepts cookies on these same endpoints), or have the `/mcp` gate accept Bearer only and 401 cookie-only callers at the transport (defensible — MCP hosts send Bearer, and it keeps the ambient-credential surface off this endpoint). The second option also strengthens the `allowed_origins` argument two sections down: "MCP hosts are not browsers" is weaker than it reads while the gate honors a browser-supplied cookie.

docs/superpowers/specs/2026-08-28-mcp-http-transport-design.md
```diff
@@ -0,0 +1,443 @@
...
+into `configure_middleware`, synthetic dispatch would bypass it; the
+session gate would still authenticate `/mcp` callers, but the
+handler-parity argument of this design would need revisiting.
+
+### Router wiring
+
+`configure_routes` builds in two phases: the REST router is built exactly
+as today; the MCP service is then constructed capturing `rest_router.clone()`
+and the final router is the REST router with the MCP service nested at
+`/mcp`. Dispatch only ever targets REST paths, so no recursion into `/mcp`
```
yiyi-wang-skardi, line 104: **"No recursion" is right, but the inverse collision isn't covered.** Nesting at `/mcp` shadows a REST path that exists today: `/:name/execute` (`server.rs:471`) matches `/mcp/execute`, and matchit prefers the static `mcp` segment, so the nested MCP service wins. A pipeline literally named `mcp` (`encode_component("mcp") == "mcp"`) therefore becomes unreachable via inbound `POST /mcp/execute` — while still being callable through `/mcp` itself, because the captured `rest_router` is the pre-nest one, which is exactly what keeps recursion impossible.

Harmless in practice, but it's a silent asymmetry between the two bindings on a name a user is allowed to pick. Worth a sentence here (and it argues mildly for keeping the mount point at a path REST's `/:name/execute` can't produce).

docs/superpowers/specs/2026-08-28-mcp-http-transport-design.md
```diff
@@ -0,0 +1,443 @@
...
+  generation, `builtin_tools()`, `sanitize`/`assign_tool_names`, the
+  `QUERY` / `LIST_DATA_SOURCES` / `RESERVED_NAMES` constants, the
+  version-skew fallbacks (missing `parameters` → open schema, missing
+  `json_schema` → `{}`), and all its unit tests.
+- `encode_component` (currently in `crates/cli/src/client.rs`): the
+  pipeline-name → URL path encoding is part of the tool→REST translation
+  contract and both dispatchers need it. The CLI re-imports it from the
+  crate.
+
+Dependencies of the new crate: `rmcp = "=3.1.4"` (model types only),
```
yiyi-wang-skardi: **The dependency list is short by three, and one entry contradicts a load-bearing comment in the CLI.**

- `encode_component` needs `percent-encoding` (`utf8_percent_encode`, `AsciiSet`, `NON_ALPHANUMERIC` — `crates/cli/src/client.rs:20-66`), which isn't listed.
- `rmcp = "=3.1.4"` here is spelled without `default-features = false`. Both existing rmcp entries pin `default-features = false` deliberately: "keeps rmcp's HTTP stack (and any axum version skew) out of the dependency graph" (`crates/cli/Cargo.toml:23-25`). If `mcp-core` really needs only model types, say `default-features = false` with no features (or none at all) so the intent survives the move.
- skardi-server currently has no `rmcp` and no `uuid` dependency; the per-request UUID v4 and `transport-streamable-http-server` both need adding (`uuid` is already in `[workspace.dependencies]`).

Separately worth naming: with `resolver = "2"`, features unify across workspace members built together, so once the server enables `transport-streamable-http-server`, a `cargo build --workspace` compiles rmcp with the HTTP stack for the CLI binary too. The CLI comment's stated goal stops being true at workspace-build granularity — R2 answers the axum-skew half of it, but the comment should be updated rather than quietly falsified.

docs/superpowers/specs/2026-08-28-mcp-http-transport-design.md
```diff
@@ -0,0 +1,443 @@
...
+exists for.
+
+`allowed_origins` stays at rmcp's default (no restriction): MCP hosts
+are not browsers, the session gate is the actual barrier, and an open
+Origin posture is consistent with the existing CORS
+`mirror_request` configuration.
+
+### Tool call translation
+
+Identical tool surface to the stdio bridge — same projection, same three
```
yiyi-wang-skardi: There are two built-ins, not three: `builtin_tools()` returns `query` and `list_data_sources` (`crates/cli/src/mcp/projection.rs:150-183`), and `RESERVED_NAMES` has length 2. The table below has three rows because `tools/list` is a protocol method, not a built-in tool — and the dispatch section above correctly says "three arms" counting the pipeline fallthrough. s/same three built-ins/same two built-ins/.

docs/superpowers/specs/2026-08-28-mcp-http-transport-design.md
```diff
@@ -0,0 +1,443 @@
...
+- Unknown tool name → JSON-RPC `invalid_params` with the "re-issue
+  tools/list" hint.
+- `tools/list` unable to produce an inventory (the synthetic `GET
+  /pipelines` fails, which in-process means a bug rather than a network
+  condition) → JSON-RPC internal error.
+
+### Response size cap
+
+The server handler collects synthetic-response bodies with the same
+256 MB ceiling the stdio bridge inherits from the CLI client, kept as a
```
yiyi-wang-skardi, line 371: KISS: `axum::body::to_bytes(body, MAX_RESPONSE_BYTES)` already *is* a bounded collect and is already used in this crate (`crates/server/src/auth/routes.rs:196`), so no bespoke collector is needed — worth naming it here so the implementation PR doesn't reach for `http_body_util::BodyExt::collect` (which is a dev-dependency only today, `crates/server/Cargo.toml:85`) and hand-roll the ceiling.

docs/superpowers/specs/2026-08-28-mcp-http-transport-design.md
```diff
@@ -0,0 +1,443 @@
...
+```
+
+---
+
+## Architecture: a protocol adapter over the server's own router
+
+The MCP layer added to skardi-server is a **protocol adapter, not a second
+execution path**. The handler holds a clone of the server's own REST
+`Router`; every tool call is translated into a synthetic HTTP request and
+dispatched in-process via `tower::ServiceExt::oneshot` — the same mechanism
```
yiyi-wang-skardi, line 28: `tower::ServiceExt::oneshot` lives behind tower's `util` feature, and the server declares `tower = "0.4"` with default features (`Cargo.toml:72`, `crates/server/Cargo.toml:67`). It compiles today only because axum/tower-http enable `util` transitively, and `ServiceExt` is currently used from tests alone (`crates/server/tests/pipelines_http.rs:20`). Moving `oneshot` into production code should come with an explicit `features = ["util"]` on the server's tower dependency rather than relying on feature unification.

docs/superpowers/specs/2026-08-28-mcp-http-transport-design.md
```diff
@@ -0,0 +1,443 @@
...
+  the configured router on an ephemeral-port listener inside the test and
+  drive it with rmcp's streamable-HTTP client: `tools/list` contains the
+  seeded pipeline tools plus the built-ins; a pipeline call executes and
+  returns rows; `query` round-trips; unknown tool → `invalid_params`; a
+  failing call (bad SQL) → `isError: true`.
+- **Auth forwarding test**: with a token-required server configuration, a
+  tool call through `/mcp` with the Bearer token succeeds — which proves
+  the header actually reaches the handlers: a dropped `Authorization`
+  would make the handler-level `require_session` reject the synthetic
+  request even though the session gate had already passed the caller.
```
yiyi-wang-skardi, line 423: The auth-forwarding test proves the positive case (Bearer token → success), but nothing pins the negative one, and the gate makes it unreachable end-to-end: with auth enabled a credential-less caller never gets past the transport, so "handler-level `require_session` stays as the second layer of defense" is asserted and untested. Add a `do_*`-level test (the pattern this design already adopts for testability) that dispatches a tool call with *no* `Authorization` and asserts the handler's 401 surfaces as `isError` — that's the regression test for the middleware-boundary constraint the design says it depends on.

Same gap for the transport-middleware claim: nothing here pins that CORS applies exactly once. A test asserting a single `access-control-allow-origin` on an `/mcp` tool-call response would make the pre-middleware capture a tested property rather than a comment.

docs/superpowers/specs/2026-08-28-mcp-http-transport-design.md
```diff
@@ -0,0 +1,443 @@
...
+
+### Request identity and audit correlation
+
+The carriers are unchanged from the bridge — `ai_context.session_id` when
+`query` carries a `purpose`, `x-skardi-session-id` on pipeline executes —
+but the id itself is layered, because the new protocol has no session to
+name:
+
+- A request carrying `Mcp-Session-Id` (legacy-protocol clients under
+  `legacy_session_mode`) uses that value: the session's pipeline runs and
```
yiyi-wang-skardi, line 337: The forwarded id lands in a header with a strict validator: `session_id_from_headers` rejects anything but non-empty visible ASCII, no commas, ≤ 200 chars, and a malformed value *fails the whole request* (`crates/server/src/session_header.rs:47-75`, cap from `query-audit/src/lib.rs:78`). rmcp mints UUIDs so this is fine in practice, but the coupling is invisible and the failure mode is bad (every pipeline call 400s → `isError`). Worth one line stating the constraint and the guard: validate before forwarding and fall back to a minted UUID rather than propagating a 400.

docs/superpowers/specs/2026-08-28-mcp-http-transport-design.md
```diff
@@ -0,0 +1,443 @@
...
+
+- `docs/mcp.md` — new "Remote (streamable HTTP)" section: the `/mcp`
+  endpoint, host configuration for URL-based MCP servers, Bearer-token
+  guidance, and an explicit statement of the local-vs-remote split (stdio
+  for local hosts, `/mcp` for hosts that cannot spawn a binary). The
+  architecture diagram gains the second path. Includes a reverse-proxy
+  subsection: either forward the public `Host` and declare it via
+  `--mcp-allowed-host`, or have the proxy rewrite `Host` to the upstream
+  loopback authority — and how the 403 from a mismatch presents.
+- `docs/pipelines.md` — roadmap note moves to shipped.
```
yiyi-wang-skardi: Both documentation bullets need adjusting against what's actually in the tree:

- `docs/pipelines.md` has no `/mcp` roadmap note to move to shipped — its only roadmap marker is Claude skills, and the MCP bullet reads "via `skardi mcp`" (`docs/pipelines.md:14-19`). The real edit is broadening that bullet to name the remote binding.
- `docs/mcp.md` needs *edits*, not just a new section. Its "Auth notes" currently tell operators that `GET /pipelines` and `GET /data_source` are readable without a token (`docs/mcp.md:123-131`) — true for the bridge, false for `/mcp` once the gate lands. Its "Timeouts & lifecycle" section is stdio-only ("when the host closes stdin, the bridge exits 0"). Also the page title is `# MCP Binding — \`skardi mcp\``, which stops covering the content.

docs/superpowers/specs/2026-08-28-mcp-http-transport-design.md
```diff
@@ -0,0 +1,443 @@
...
+---
+
+## Documentation
+
+- `docs/mcp.md` — new "Remote (streamable HTTP)" section: the `/mcp`
+  endpoint, host configuration for URL-based MCP servers, Bearer-token
+  guidance, and an explicit statement of the local-vs-remote split (stdio
+  for local hosts, `/mcp` for hosts that cannot spawn a binary). The
+  architecture diagram gains the second path. Includes a reverse-proxy
+  subsection: either forward the public `Host` and declare it via
```
yiyi-wang-skardi: The reverse-proxy subsection covers `Host`, but not the other thing that bites remote MCP: long tool calls. Deferring execution timeouts to resource governance is the right call, yet the *transport* consequence is new here — a synthetic dispatch blocks the POST for the whole pipeline run with no progress notifications, and proxies routinely cut idle connections around 60s (the bridge's "no request timeout, the host's timeout fires" reasoning at `docs/mcp.md:134-140` assumed nothing in between). One line on proxy read/idle timeouts, next to the `Host` guidance, would save a confusing first deployment.

docs/superpowers/specs/2026-08-28-mcp-http-transport-design.md
```diff
@@ -0,0 +1,443 @@
...
+legacy-protocol clients rmcp creates and retains session state for an
+`initialize` before any tool call happens, so with auth enabled an
+anonymous caller could mass-create sessions; and on every protocol
+version, missing credentials would otherwise surface only as tool errors
+inside HTTP 200 responses instead of the transport-level 401 that host
+credential flows key on. A thin tower layer wrapping only the nested
+MCP service closes this: every inbound `/mcp` request passes
+`verify_session(&state, &headers)` — the same single-home check the REST
+handlers call — before reaching rmcp; failures return a transport-level
+401, which is also the signal host credential flows key on. On a no-auth
```
yiyi-wang-skardi: Nit: "the transport-level 401 that host credential flows key on" (lines 180-181) and "which is also the signal host credential flows key on" (line 185) say the same thing twice in one paragraph. Drop the second clause.

yiyi-wang-skardi (approved):
## Audit — MCP streamable HTTP transport design spec

Docs-only PR: one new 443-line design spec (`docs/superpowers/specs/2026-08-28-mcp-http-transport-design.md`), no code. I reviewed it as a design: every load-bearing claim about the *current* tree was checked against the code, and the security/correctness reasoning was probed for gaps.

**Verified correct against the tree**

- `create_server` really does split `configure_routes` (bare table) / `configure_middleware` (CORS on the final router), so the "capture the pre-middleware router" argument holds and the double-CORS/double-tracing hazard it avoids is real (`crates/server/src/server.rs:114,464,499`).
- Auth really is handler-level, not middleware: `require_session` is called inside `execute_query` (`query_handlers.rs:152`) and `execute_pipeline_by_name` (`pipeline_handlers.rs:901`) only. `GET /pipelines` and `GET /data_source` are unauthenticated, so the "`/mcp` adds no new privilege" claim is accurate — and the session gate makes the MCP surface strictly *tighter* than REST, which the spec names honestly.
- `verify_session` is `pub`, `require_session` is `pub(crate)` — both reachable from an in-crate `mcp` module. No new auth code needed, as claimed.
- 256 MB = `MAX_RESPONSE_BYTES` in `crates/cli/src/client.rs:40`; the spec's admission that in-process this only bounds a *second* copy is correct (the REST handler materializes the full result first).
- `config.write()` outside tests: none (the one hit, `pipeline_handlers.rs:2010`, is inside `mod tests`), so the `listChanged` deferral is justified.
- Every projection symbol the crate extraction names exists (`project`, `pipeline_tool`, `builtin_tools`, `sanitize`, `assign_tool_names`, `QUERY`/`LIST_DATA_SOURCES`/`RESERVED_NAMES`, and the `builtin_tool_names_match_the_reserved_set` invariant test).
- R2 checks out as far as the lockfile can show it: rmcp 3.1.4's locked dependency set contains no axum, no `http`, no `tower-service`; the axum 0.8.9 and `http 0.2` in `Cargo.lock` both belong to tonic.

**Could not verify locally**: rmcp 3.1.4 is not vendored in this sandbox, so the claims stated as "verified in the 3.1.4 source" — `http::request::Parts` in request extensions, the loopback-only `allowed_hosts` default and `with_allowed_hosts` port semantics, `legacy_session_mode` defaulting to `true`, once-per-request service-factory construction on the stateless path, and MCP protocol `2026-07-28`/SEP-2567 removing sessions — were taken as given. They are all self-consistent and the design degrades safely if the protocol-version detail is off (per-call resolution is correct in both modes either way). Also note enabling `transport-streamable-http-server` will change rmcp's locked dependency set, so R2's lockfile evidence is about today's feature set, not tomorrow's.

**Overall**: a strong, unusually honest spec — the self-dispatch architecture is the right call, the middleware boundary is drawn precisely, and the session gate + host allowlist together mean a default-on `/mcp` is not reachable from off-box until an operator names a hostname. Nothing here should block the merge. The findings below are one security consideration I think the spec should state rather than leave implicit (MCP session ids are not bound to the authenticating principal), one real parity gap (cookie auth passes the gate but not the synthetic dispatch), a route-shadowing case the "no recursion" claim doesn't cover, plus dependency/doc/test corrections to save the implementation PR some rework.

docs/superpowers/specs/2026-08-28-mcp-http-transport-design.md
gracexmatin: Have mentioned in design doc.

docs/superpowers/specs/2026-08-28-mcp-http-transport-design.md
gracexmatin: Noted. However, the possibility of the matter is quite little.

PR #235, opened 2026-08-29
docs(readme): point Install at both skills, and mark skardi mcp unreleased

Userneima:
Three places where the README had drifted from what a reader can actually get today. Docs only.

## The `auto_context` link 404s

Install links to `skardi-skills/tree/main/auto_context`. That directory was renamed to `auto-context` in SkardiLabs/skardi-skills#29 — underscore names are silently dropped by dsh — so the link has been dead since that merge. Fixed, and the inline code label matches the real directory again.

## `retrieval` had no entry point

SkardiLabs/skardi-skills#27 shipped the `retrieval` skill on 2026-08-26 and Install never mentioned it. The only path a reader could see was "turn a folder into a knowledge base" — while the second path this README's own loop depends on, *query a server that is already running*, had nowhere to start from. Added its install line and a short description taken from the skill's own README entry.

The intro line said "A skill"; it now says two, since that is what the block installs.

## `skardi mcp` is presented as available, but is not released

The **4 — The agent has a new tool** block shows `skardi mcp` alongside `skardi run` and the REST call, as one of three bindings you can use now. #227 landed after `v0.5.0`, so anyone on the released CLI gets an unknown-subcommand error there. The `--query-audit-db` line a few paragraphs up already carries exactly this caveat, so this adds the matching one rather than inventing a new convention.

## Verified

- `auto-context/` and `retrieval/` both resolve on `skardi-skills` `main`; `auto_context/` returns 404 (`gh api repos/SkardiLabs/skardi-skills/contents/<path>`)
- `git grep -c mcp v0.5.0 -- crates/cli/src/main.rs` → zero hits; same grep on `main` finds the subcommand
- `python3 scripts/check_markdown_links.py` → all relative markdown links resolve

No behaviour changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ skardi run weekly-churn -p window='7 days'
 curl -X POST localhost:8080/weekly-churn/execute \
   -H 'Content-Type: application/json' -d '{"window": "7 days"}'
 # same pipeline as an MCP tool — hosts without a shell (Claude Desktop, …)
+# from a source checkout until `skardi mcp` ships in a tagged release
 skardi mcp
 ```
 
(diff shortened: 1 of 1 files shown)


Userneima:
Closing in favour of #230, which touches the same section and is further along — it already carries a review round from @BtXin.

All three changes here are preserved:

- **`auto_context` link 404** — fixed in #230 (the directory rename to `auto-context`).
- **`retrieval` had no entry point** — #230 adds it to the Install block alongside `auto-context`.
- **`skardi mcp` presented as available** — moved to #230 in 0f67376, keeping the wording and the convention this PR chose (matching the `--query-audit-db` caveat a few paragraphs up rather than inventing a second one). Re-verified before moving it: `git grep -c mcp v0.5.0 -- crates/cli/src/main.rs` is empty, the same grep on `main` finds the subcommand.

#230 additionally moves the per-host install commands into this README and, per review, puts every host on the same footing instead of leading with Claude Code as "the fastest path" — that part would have been reverted had this branch merged after it.

Issue #197, opened 2026-08-06
feishu pack: document text is unreachable — the content actions are single-object, row_path requires an array


Userneima:
The `feishu` pack can enumerate documents but cannot read them. The gateway has the actions; the pack layer cannot express their shape. Evidence below is pinned to `SkardiLabs/skardi@2291b19` and `oomol-lab/open-connector@bf94ce8`.

## The gap

The `feishu` pack (merged 2026-08-06, `2291b19`) exposes six tables: `chats`,
`messages`, `chat_members`, `tasks`, `wiki_spaces`, `wiki_nodes`. That covers
chat history well. It does not cover document text at all: `wiki_nodes` yields
`node_token` / `obj_token` / `obj_type` / `title` / `url` — the keys you need to
open a document, but no column carrying what the document says.

So today an agent can answer *"what was discussed in this chat"* but not
*"what does this document say"*, even though the second is the request users
keep making.

**This is not an upstream gap.** The gateway has the actions. Verified against
`oomol-lab/open-connector` at `bf94ce8`, and the feishu provider registers every
action group (`src/providers/feishu/actions.ts:201-240`), so all of these exist
on a current gateway:

| Action | Response shape | Mappable as a table today? |
|---|---|---|
| `feishu.get_document_content` | single object `{documentId, content}` — full plain text (`actions.ts:94`) | **No** |
| `feishu.fetch_document` | single object `{document: {…}}`, with `scope` / `detail` / `keyword` / block-range reads (`shared/docs-actions.ts:91`, schema at `:28`) | **No** |
| `feishu.list_document_blocks` | array, same envelope as the six mapped tables (`actions.ts:112`) | Yes, but see below |
| `feishu.search_documents` | array `{results, hasMore, pageToken}` (`shared/docs-actions.ts:192`) | Yes — metadata only, no text |

## Why the two content actions cannot be mapped

A pack table locates its rows with `row_path`, and that path must resolve to an
array:

- `RowPath::parse` rejects `$` outright — "no segments"
  (`crates/skardi/src/sources/providers/open_connector/row_path.rs:31`).
- `RowPath::rows` fails with `RowPathNotArray` on a non-array target
  (`row_path.rs:98`).

Both content actions return one object at the root. There is no nested array to
point at, so no `row_path` can express them.

## Why `list_document_blocks` does not close the gap either

It is array-shaped and needs no engine change — but a Feishu docx block carries
its payload under a key named by the block's own type (`text`, `heading1` …
[...]

PR #239, opened 2026-09-01
feat(open-connector): accept a single response object as one row

Userneima:
Implements the engine half of #197, to the design @bakey set out in the 2026-08-10 comment and confirmed as the contract of record on 2026-08-30.

## What this enables

Point-read actions — `feishu.get_document_content`, Notion's rendered Markdown — return **one object at the root**. No `row_path` can locate a row array in that, so their content has been unreachable from SQL. A pack can now declare `row_shape: object` and say its response *is* the row.

No shipped pack uses it yet; the Feishu `document_content` table is the follow-up this unblocks.

## Against the contract

| Contract | Where |
| --- | --- |
| `RowShape::{Array, Object}` as a typed field on `SourcePackTable` **and the execution target** | `source_pack.rs`; carried on `ScanTarget` alongside `pagination` / `error_path`, since it is the same kind of per-table response contract |
| Root `$` accepted **only** for object rows, not a global relaxation of `RowPath::parse` | separate constructor `RowPath::parse_object_root`; `parse` is untouched and still rejects `$` |
| `object` only with `single_page` | `validate_table`, load time |
| selected value must be an object; null / array / primitive fail loudly | `RowPath::row_object` → `RowPathNotObject` |
| existing packs default to `array`, zero behaviour change | `#[serde(default)]` + `Default` on `RowShape` |
| `std::slice::from_ref` into the existing converter, no second conversion path | `exec.rs` — the object is borrowed in place and handed to the same `RowConverter` as a one-element slice |

## Raw-action schema inference: deferred, with a typed error

You asked for this decision to be written down rather than guessed at later. **Raw scans (`open_connector_scan`) stay array-only**, and asking for `$` fails at planning time with `ObjectRowsUnsupportedForRawAction`.

Reasoning: `derive_raw_columns` descends `row_path.segments()` to find the row type. An object row has zero segments, so it would silently treat the response envelope as the row schema — a different inference rule, not a relaxation of the existing one. Pack tables have a fingerprint pinning their response shape; an ad-hoc action has no such gate. Supporting it later is additive; guessing at it now is not.

## Tests

11 new, all green (`cargo test --workspace`):

- **`row_path`** — root parses only as bare `$`; `$` stays invalid for the array parser; object extraction returns the root; null / array / primitive each fail as `RowPathNotObject` (the "never degrade to zero rows" invariant).
- **`loader`** — object shape parses with root + `single_page`; rejected without `single_page`; rejected with a non-root path; array shape still rejects `$`.
- **`source_pack`** — every shipped table across all ten built-in packs is asserted array-shaped, so no pack can acquire object rows unnoticed. When Feishu `document_content` lands it should name itself here rather than delete the test.
- **`exec`** — end to end against a mock gateway: a single-object response yields exactly one row with correct column values and stops after one call; a null response fails loudly instead of returning an empty scan.
- **`table_functions`** — a raw UDTF scan passing `$` fails at planning with the targeted error and issues no HTTP request.

One caveat worth stating: `unknown_pipeline_request_is_logged` (skardi-server, untouched by this change) failed once during a full-workspace run and passed on the clean tree, in isolation, and on every run since. It looks like parallel-execution sensitivity in that log-capture test rather than anything from here, but I could not reproduce it deliberately, so I am not claiming more than that.

## Docs

The source-pack authoring guide documents `row_shape`, the root-only path, the `single_page` requirement, and the raw-action refusal. README is unchanged — no new published pack or user-queryable table, so the capability tables stay accurate.

## Not in scope

Fingerprint / live-verification gates are untouched, per your note that you would run that half once this merges.

### DIFF
--- a/crates/skardi/src/sources/providers/open_connector/error.rs
+++ b/crates/skardi/src/sources/providers/open_connector/error.rs
@@ -452,6 +452,18 @@ pub enum OpenConnectorError {
         found: String,
     },
 
+    /// A raw-action scan asked for the object row shape. Object rows are a
+    /// pack-declared contract (`row_shape: object`) whose response shape is
+    /// pinned by a fingerprint; an ad-hoc action has no such gate, and
+    /// deriving a row schema from the response ROOT rather than a row array
+    /// is a different inference rule that has not been designed. Deferred
+    /// deliberately, with this error, rather than guessed at.
+    #[error(
+        "Open Connector raw-action scans do not support object rows: '$' selects the response root, \
+         which only pack tables may declare via 'row_shape: object'. Use a row path locating a row array."
+    )]
+    ObjectRowsUnsupportedForRawAction,
+
     /// The row path resolved to a non-array value; relational rows must be an
     /// array of objects.
     #[error(
--- a/crates/skardi/src/sources/providers/open_connector/exec.rs
+++ b/crates/skardi/src/sources/providers/open_connector/exec.rs
@@ -33,7 +33,9 @@ use super::error::OpenConnectorError;
 use super::json_to_arrow::RowConverter;
 use super::pagination::{CursorContinuation, Pagination, PaginationStrategy};
 use super::row_path::RowPath;
+use super::source_pack::RowShape;
 use super::source_pack::{FixedValue, SourcePackTable};
+use std::slice;
 
 /// The scanned collection's identity and pagination contract — the shape
 /// shared by YAML-bound source-pack tables and `open_connector_scan` raw
@@ -89,6 +91,7 @@ pub struct OpenConnectorExec {
     target: ScanTarget,
     converter: Arc<RowConverter>,
     row_path: RowPath,
+    row_shape: RowShape,
     resource: Value,
     filter_inputs: Vec<(String, Value)>,
     projection: Option<Vec<usize>>,
@@ -125,6 +128,7 @@ impl OpenConnectorExec {
         target: ScanTarget,
         converter: Arc<RowConverter>,
         row_path: RowPath,
+        row_shape: RowShape,
         resource: Value,
         filter_inputs: Vec<(String, Value)>,
         projection: Option<Vec<usize>>,
@@ -153,6 +157,7 @@ impl OpenConnectorExec {
             target,
             converter,
             row_path,
+            row_shape,
             resource,
             filter_inputs,
             projection,
@@ -274,6 +279,7 @@ struct ScanState {
     target: ScanTarget,
     converter: Arc<RowConverter>,
     row_path: RowPath,
+    row_shape: RowShape,
     resource: Value,
     filter_inputs: Vec<(String, Value)>,
     projection: Option<Vec<usize>>,
@@ -361,6 +367,7 @@ impl ScanState {
             target: exec.target.clone(),
             converter: exec.converter.clone(),
             row_path: exec.row_path.clone(),
+            row_shape: exec.row_shape,
             resource: exec.resource.clone(),
             filter_inputs: exec.filter_inputs.clone(),
             projection: exec.projection.clone(),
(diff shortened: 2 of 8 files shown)


gracexmatin:
Could the solution be applied to all source-packs? I think the situation is similar for other sources.

gracexmatin:
Engine half looks good, and I think the scope line is right. Not blocking — just want to think through one thing for the follow-up while it's cheap.

How does the document id get to the action? There are three ways into the action input (exec.rs:481-490), and I'm not sure either of the plausible two is ready.

Resource is what `object_row_table()` uses, which is fine for a fixture, but required resources are checked at startup against the binding config (mod.rs:188) and a binding is a schema (config.rs:163) — so a real table would mean one document per binding, hand-edited and restarted.

Filter pushdown feels closer: `WHERE document_id = 'doc_x'` lands in the same input map and composes fine. But FilterMapping (filters.rs:77) can't say a filter is mandatory, so a bare `SELECT *` would send no documentId and eat a gateway 400 at runtime instead of failing at planning.

So maybe what's wanted is "this input must come from a predicate, absent = planning error"? That's an engine surface rather than a YAML field, which is why I'm asking here and not on the table PR.

Curious what you're picturing — and whether the real target is the join case (wiki_nodes -> point-read each obj_token), since that drags in concurrency and partial-failure semantics and might change the answer. No need to resolve it in this PR.

Userneima:
Both good questions. Taking the second one first, since the first depends on it.

Your reading of the two routes is right, and I re-checked both before answering.

The resource route is a dead end for a per-document read, for exactly the reason you give. `required_resources` are validated at startup against `binding.resource` (mod.rs), and each binding becomes a schema in the catalog (config.rs). That means one document per schema, hand-edited, with a restart between reads. It is not a table.

Filter pushdown is the shape I want. The literal lands in the same input map as everything else and composes with pagination and fixed inputs with no special case. The gap you found is real: `FilterMapping` has no way to say "this one is not optional", so a bare `SELECT *` builds a request with no `documentId` and the gateway answers 400. That is a runtime failure for something the planner already knows.

So yes, I think the missing piece is an input that can only be satisfied by a predicate, and it belongs in the engine. The smallest version I can see is a `required` flag on `FilterMapping`, plus a check where the scan translates its filters: if a required mapping produced no input, fail at planning and name the column. It stays in the engine; it is declarable by builtin and YAML packs alike (`convert_filter` in loader.rs already carries every other mapping field across); and loader.rs already validates filters against declared columns, so the load-time half has a home. I have not written it, so treat that as a sketch rather than a design.

On the join case: you are right that it is the real target, and it has moved since you asked. I put the same question on #197 on 2026-09-01, and @bakey answered on 2026-09-04: build the batch version directly, start with conservative limits, and open them up by config once there is real usage. So the direction is settled; what is still open is the layer, engine or a skill above it, and I am following that up on #197. Either way it leaves this PR alone, and the gap you found stands on its own, since a single point read still needs a document id and nothing today makes that predicate mandatory.

On your first question: yes, nothing here is Feishu shaped. `row_shape` is a per-table field on `SourcePackTable`, so any pack whose point read returns one object at the root can declare it, under the same two rules (root path only, `single_page` only). No pack sets it yet because I did not want to change an existing table's behaviour inside this PR. There is a test asserting every table across all ten builtin packs is array shaped, so adding one is a deliberate edit with a visible diff rather than something that can happen quietly.

@bakey when you have a slot, this is the engine piece from the 2026-08-30 split, ready for the review you offered. Owen's points above are follow-ups rather than changes to this one, and the fingerprint and live-verification gates are untouched, per your note that you would run that half after it merges.

bakey:
@codex

Userneima:
Both Codex findings addressed in 9a01ece.

**P2, cache key omits row shape.** Real bug, fixed. `row_shape` now belongs to `ScanKeyParts` and the canonical key, and the call site in `exec.rs` passes `target.row_shape`. Two tables identical in action, pack version, resource, projection, limit and schema but differing in shape now key apart, asserted in `key_distinguishes_alias_projection_and_gateway`. This mirrors the reasoning already in that key list: it holds everything that changes the result set, and shape does.

**P1, `row_object` has no rustdoc example.** Added, covering the root-object success and the array-at-root error. Doctest passes.

`cargo check`, `cargo fmt --check`, the cache and row_path unit tests, and the doctests are green on the crate. The only clippy warning in the crate is a pre-existing one in `mongo/mod.rs`, untouched here.

bakey (approved):

PR #244, opened 2026-09-03
feat(obsidian): read-only Obsidian vault source — notes/links/tags

gracexmatin:
Implements `type: obsidian` — a read-only native source that exposes one Obsidian vault (a local directory or an `s3://` prefix) as three catalog tables, rescanned on every query. Not an Open Connector pack: Obsidian has no cloud API and the gateway has no provider for it, so the vault is read straight off disk.

Design record: `docs/superpowers/specs/2026-09-02-obsidian-source-design.md` (merged in 5713a6e). Implementation plan: `docs/superpowers/plans/2026-09-03-obsidian-source.md` (commit 6257f06 on this branch), executed task by task.

User docs: `docs/obsidian.md`.

## What ships

`<name>.main.notes` / `.links` / `.tags`, behind the new `obsidian` Cargo feature on both crates:

- **`notes`** — one row per `.md` file: `path`, `name`, `folder`, `body`, `frontmatter_json`, `frontmatter_error`, `aliases List<Utf8>`, `size_bytes`, `modified_at`.
- **`links`** — one row per link, already resolved: `from_path`, `to_path`, `target`, `kind ∈ {wikilink, embed, markdown, external}`, `display_text`, `heading`, `block_id`, `resolution ∈ {exact, name, ambiguous, missing, external}`, `source ∈ {body, frontmatter}`, `line`.
- **`tags`** — one row per (note, tag, source), nested tags keep their `/`.

Every schema carries `skardi.obsidian.surface_version = "1"` in its Arrow metadata.

## Shape of the change

- `crates/skardi/src/sources/providers/documents/blob.rs` → `providers/blob.rs`, shared by `documents` and `obsidian` (`#[cfg(any(...))]`). Its API grows `ListOptions` / `ReadOptions` / `BlobEntry`: listings now carry size and mtime, listing can be non-recursive, and symlinks are controllable. `documents` keeps its old behavior by passing `follow_symlinks: true`.
- `providers/obsidian/` — four pure parsing units (`config`, `markdown`, `frontmatter`, `resolve`), one synchronous scanner (`scan`) that only ever runs inside `spawn_blocking`, and one `TableProvider`/`ExecutionPlan` pair parameterized by table kind (`table`). Registration in `mod.rs` mirrors `rss`: catalog-only, read-only, `MemoryCatalogProvider` published last so a failed registration leaves the `SessionContext` untouched.
- Server: `DataSourceType::Obsidian`, the `obsidian` feature (`obsidian = ["skardi/obsidian"]`), an entry in `CATALOG_SUPPORTED_SOURCES`, the registration arm in `config.rs`, and the two exhaustive `DataSourceType` matches.

## Decisions worth a reviewer's attention

- **Link resolution mirrors Obsidian, not the filesystem.** A bare `[[Name]]` is a vault-wide name lookup (`resolution = 'name'`, `ambiguous` when several notes share the name); a path-ish target resolves exactly; Markdown links resolve note-relative first. Aliases are deliberately *not* resolved — Obsidian only offers them in autocomplete — which is what makes the alias-repair query in the docs possible.
- **No cache.** Every query lists, reads and parses the whole vault, so edits are visible immediately and nothing needs reloading; a two-table query scans twice. The cost model (and the S3 egress note) is documented.
- **Symlinks are never followed.** Listing skips them and each read opens with `O_NOFOLLOW`, so a symlink swapped in between listing and reading is refused rather than followed. Non-Unix falls back to a `symlink_metadata` check, which leaves a small race — documented.
- **Wholesale-failure guard:** a non-empty vault where every read fails errors with the first cause instead of returning three empty tables. Oversized notes (`max_file_bytes`, default 16 MiB) are skipped with a `warn` naming the path.
- Registration refuses non-catalog hierarchy, `read_write`, unknown option keys (by name), and an unreachable root — all before anything is registered.

Four spec deviations settled during implementation are recorded in the spec's new "Implementation notes" section (dotted bare wikilink names, RFC 3986 scheme detection, the async `run_scan` wrapper, `display_text = NULL` for autolinks).

## Tests

A hand-written fixture vault (`providers/obsidian/fixtures/vault/`, 15 files) with a row-level oracle: 12 notes, 28 links, 10 tags, covering every `kind`, every `resolution`, both `source` values, CJK content, broken frontmatter, an oversized note, and an excluded `.trash/`. Unit tests per parsing unit; SQL-level integration tests query the fixture through a `SessionContext` (in-degree, orphan anti-join, alias repair, `EXPLAIN`, edits visible between scans on a temp copy); server-side tests cover registration through the config path with and without the feature.

**Nothing was run locally** — per this repo's convention, CI is the verification oracle for this branch. Treat the first CI run as the first real test pass.


gracexmatin:
@codex

crates/skardi/Cargo.toml
```diff
@@ -28,7 +28,11 @@ chunking = ["dep:text-splitter"]
 # RAG umbrella: chunk → embed → write loop in one feature.
 rag = ["embedding", "chunking"]
 # documents source connector: parse a folder of files into queryable (file, page) rows.
-documents = ["dep:liteparse", "dep:glob", "dep:object_store"]
+documents = ["dep:liteparse", "dep:glob", "dep:object_store", "dep:libc"]
```
BtXin: We do we need libc here? Which feature requires libc? Does it have rust implementation replacement?

gracexmatin: Only one function uses `libc`: `read_local_no_follow()` in `providers/blob.rs` — it reads a local file by opening the path one component at a time, starting from the vault root's fd. It uses the single function `libc::openat` plus four constants, `O_NOFOLLOW` / `O_DIRECTORY` / `O_CLOEXEC` / `O_NONBLOCK`:

```rust
let flags = O_RDONLY | O_NOFOLLOW | O_CLOEXEC
    | if is_last { O_NONBLOCK } else { O_DIRECTORY };
let fd = libc::openat(dir.as_raw_fd(), name, flags);
```

`O_NOFOLLOW` makes the open fail with `ELOOP` as soon as any component is a symlink, so a note can never be read from outside the vault; `O_NONBLOCK` plus the following `fstat` means a FIFO named `note.md` is refused instead of blocking on a writer and hanging the query.

If we wanted a Rust replacement, the options are `rustix` or `cap-std`.

gracexmatin: Checked again. For documents, libc is unnecessary. libc is used only for obsidian. Fixed already.

BtXin: rustix seems to be a reasonable choice

gracexmatin: Solved. Use rustix now.

gracexmatin:
Two things I looked at in `blob.rs` while swapping `libc` for `rustix` and deliberately did **not** change in this PR — recording them so they are not lost. Neither is reachable through today's call sites; both are places where the type system does not enforce the contract the code relies on.

### 1. `BlobStore::Remote` drops `Loc::S3`'s bucket

`Remote(Arc<dyn ObjectStore>)` is already bound to one bucket (`AmazonS3Builder::with_bucket_name` inside `BlobStore::resolve`), while `get`/`put` match `Loc::S3 { key, .. }` — the `Loc`'s bucket is neither used nor checked. All five construction sites go through `resolve`, which builds the store from the very same `Loc`, so the two copies agree today.

The hazard is that `(BlobStore, Loc)` is a pair that can be recombined, and two `Remote` values are indistinguishable to the compiler. `documents::parse_source_blocking` holds exactly two of them — the source store and the `image_store` store. Passing the wrong one where a `&BlobStore` is expected compiles cleanly, and a crop would then be written into the *source* bucket under the right key, with no error and no warning.

A fix would be `Remote { store, bucket }` plus a bucket-equality check in the three match arms. That changes the shape of a `pub enum` and touches nine construction sites (mostly tests), so it is its own change rather than part of this one.

### 2. `get` / `list` / `put` are `async fn` but block on their local arms

The `BlobStore::Local` arms call `File::open` + `read_to_end`, `read_dir` + a `metadata` per entry, and `create_dir_all` + `fs::write`. The signature advertises "await me from anywhere"; the implementation requires "call me from a blocking thread".

All three call sites honour that — `documents::parse_source_blocking`, `obsidian::VaultScan::run` and the `llm_extract` image fetch each run under `spawn_blocking` and `block_on` the future — but nothing in the types says so, and the doc comments on these three methods do not mention it either (only `resolve`'s doc talks about threads, and it talks about reqwest's connection pool, not about blocking). A future `store.get(&loc, opts).await` inside a request handler would compile, pass tests, look fine under a single request, and stall a runtime worker for the length of a disk read — very visible once the vault lives on an NFS/SMB mount. AGENTS.md forbids blocking inside async fns.

The cheap mitigation is a `# Blocking` section on each of the three methods. A real fix is either wrapping the local arms in `spawn_blocking` (owned paths, one thread hop per file — a 10k-note vault pays it 10k times) or splitting the API into a blocking-local and an async-remote half. Either way, its own change.

PR #252, opened 2026-09-15
feat(sources): re-land the slack/feishu pack tables on main's RowShape, and register the Postgres FTS UDFs

BtXin:
Two independent things skardi-cloud needs, plus the doc corrections review turned up. **1 of 2 in a stack** — the CLI half is #253. Both are independently mergeable.

## Why

skardi-cloud pins this crate at a rev on `claude/slack-messages-ts-pushdown`, which `main` cannot absorb: both branches independently implement design contract #197 ("a pack table may be one object"), each with its own `RowShape` and `RowShapeDoc` in the same two files. All five merge conflicts were that duplication and nothing else.

Separately, `to_tsvector` / `websearch_to_tsquery` / `ts_rank` existed only as string literals inside `pg_fts`'s hand-built SQL, so any statement naming them failed at SQL to LogicalPlan resolution — a full-text pipeline could not be planned by any server.

## What

**`8358b37` — re-land the pack tables.** Main's #239 implementation wins; the branch's competing copy is dropped, and only what main lacks is re-landed: `slack.messages` with its `conversations.history` time-window pushdown, and feishu's `document_blocks` and document-content tables. The YAML surface was already identical (`rename_all = "snake_case"` vs explicit renames), so the pack declarations port unchanged.

The strongest evidence these were duplicates rather than rival designs: **all 23 fingerprint tests pass with their committed numbers untouched**, feishu's object-shaped `document_content` included. Main's `RowShape` derives a byte-identical output schema to the branch's for the same responses. `git diff <pinned-rev> HEAD` over both packs, both YAMLs, `json_to_arrow.rs`, `testutil.rs` and the entire fixtures tree is **empty**.

One exception to "take main's version": `packs/loader.rs` also carries `ColumnType::TimestampSFracStringUtc`, the YAML half of a fractional-epoch type whose runtime half rides with `json_to_arrow.rs`. `slack.yaml` consumes it, and without it 173 tests across every pack fail. Those ~8 lines touch no row-shape symbol.

**`a3fd292` + `9cc7307` — register the FTS functions as pushdown-safe scalar UDFs.** Their `invoke` deliberately **errors** rather than returning NULL: these exist to be planned and pushed to Postgres, which evaluates them. A NULL-returning stub would turn a failed pushdown into a query answering "no matches" — indistinguishable from a real empty result, and worse than the prior state where it failed loudly. Review demonstrated this by mutation: replacing the error with a typed NULL collapses the search plan to `EmptyRelation: rows=0`.

**`791bfcb` — doc corrections.** Three false statements in the feishu module doc (one contradicted by its own test 500 lines below), the CJK redaction tripwire restored to its strong form with the vendor error envelope exempted by name rather than the scope silently narrowed, and the `@@` test's stated reason corrected against measurement.

## What the pushdown evidence does and does not show

`default_filter_pushdown` is literally `Unparser::expr_to_sql(f).is_ok() && !contains_subquery` yielding `Exact`, so `Exact` means *the expression unparsed* and nothing more. **No Postgres is contacted here and no index is consulted.** What it does establish is that the predicate renders to SQL of the same shape `fts_exec.rs` already builds and sends live. The module doc says so explicitly, because an earlier revision of it overstated exactly this.

Whole-plan federation — which is what would carry a `ts_rank(...)` **projection** to Postgres — is available on **neither** server: nothing here wraps a provider in a `FederatedTableProviderAdaptor`, so `get_leaf_provider` returns `NopFederationProvider` and no `Extension` is ever produced; and skardi-cloud's engine refuses federation at boot. Scan-level pushdown carries the WHERE clause only.

## Verification

`cargo test -p skardi --lib` 1203 passed / 0 failed (baseline 1195 + 8) · `cargo test -p skardi-cli` 302 / 0 · `cargo fmt --all -- --check` clean · clippy: no new findings.


BtXin:
All three findings addressed — `df792a1`, `7e85afb`, `46048e5`.

### P2 — overload arities

Correct, and the diagnosis was exact: the signatures pinned a single arity, so the registration supported only the spelling `pg_fts` generates. `to_tsvector(body)` failed with *no function matches* at **plan** time, before anything could be pushed down.

Now registered with the arities Postgres accepts — 1 and 2 for `to_tsvector`/`websearch_to_tsquery`, and 2/3/4 for `ts_rank` (optional leading `weights float4[]`, optional trailing `normalization`). Nine `rstest` cases, one per overload, including your `to_tsvector(body) @@ websearch_to_tsquery('cats')`. Mutation-verified: restoring the single arity fails five of the nine.

One deliberate choice worth flagging: `TypeSignature::Any` rather than `String`. Postgres' own argument types here are not all text — `ts_rank`'s normalization is an integer and its optional weights are `float4[]` — and these stubs are never evaluated, so the destination does the real type-checking. The **return** types are unchanged and still exact, since that is what lets `@@` coerce and the predicate type-check as a filter.

### P1 — raw Postgres FTS documentation

Added to `docs/postgres/README.md`: the supported WHERE-clause form with a worked example, the overloads, the Postgres-only requirement, and a clearly-labelled **unsupported** example showing the `ts_rank`-in-projection mistake and pointing at `pg_fts` for ranking. It also explains *why* the stub errors rather than returning NULL — a NULL-returning stub makes `NULL @@ NULL` filter every row out and report "no matches", indistinguishable from a genuine empty result. That seemed the single most useful thing a reader can know here.

### P1 — pack table documentation

`slack.messages`, `feishu.document_content` and `feishu.document_blocks` are now documented in the guides, `docs/open-connector.md` and the README capability line, with contracts read off the pack YAML rather than the surrounding prose — stale prose being what the finding was about.

**Your "message history remains gated" catch was right**, and I verified it independently rather than taking it on report: `slack.yaml:33` declares the table, and the pack's own module doc now reads *"No thread table yet. `messages` arrived…"*. That section is rewritten as **No thread table (yet)**, which is still true.

### One caveat I want visible rather than buried

**The OAuth scope names are not verified by anything in this repository.** Grep finds `channels:history`, `groups:history` and `docx:document:readonly` only in these two guides — nothing pins them. They come from vendor documentation, and they sit beside facts that genuinely were measured (the page-size caps carry `live-verified` dates), so a reader could not tell the two apart. `46048e5` marks the provenance explicitly.

The Feishu one matters most: it lands in the bulk-import JSON users paste, where a wrong name fails at **authorization**, not at scan time. If whoever ran the 2026-09-01 / 09-03 live document captures can confirm the scope they actually enabled, that comment should become a plain statement of fact.

Historical spec docs under `docs/superpowers/specs/` still say "six tables" and "message/thread tables stay gated" — left alone deliberately as point-in-time records.

bakey (approved):

BtXin:
@codex

powerfulZ (approved):
Approving — this is what skardi-cloud needs, and the outstanding thread is a docs caveat rather than a reason to hold it.

**The de-duplication argument is the convincing part.** "Both branches independently implement #197, each with its own `RowShape`, and all five conflicts were that duplication" is a claim that could easily be wishful, and the evidence given is the right evidence: **all 23 fingerprint tests pass with their committed numbers untouched**, feishu's object-shaped `document_content` included. Rival designs do not produce identical fingerprints. Taking main's implementation and re-landing only what main lacks is the correct direction for a re-land.

**The UDF half is a real unblock.** `to_tsvector` / `websearch_to_tsquery` / `ts_rank` existing only as string literals inside `pg_fts`'s hand-built SQL means any statement naming them failed at SQL-to-LogicalPlan — not at execution, at *planning* — so no server could plan a full-text pipeline at all. Registering them is what makes the predicate form expressible.

## On the open thread — codex is right, and I verified it

`SqlxPostgresTableProvider::supports_filters_pushdown` maps every `Exact` to `Inexact` unconditionally:

```rust
.map(|verdict| match verdict {
    TableProviderFilterPushDown::Exact => TableProviderFilterPushDown::Inexact,
    other => other,
})
```

That is the read_write wrapper, so under `access_mode: read_write` DataFusion keeps a local copy of the filter, the local copy invokes the error-only UDFs, and the query fails at execution. The documented guarantee in `docs/postgres/README.md` therefore does not hold for the full-CRUD configuration it is offered alongside.

**Not blocking**, and worth being precise about why: the downgrade is pre-existing behaviour, not something this PR introduces, and the PR's own contribution — registration — is what makes the predicate work at all in the read-only case. What is wrong is the scope of the sentence in the docs, which currently promises something guaranteed to fail for one documented configuration.

A clause restricting the guarantee to read-only registrations closes it. If that is easier after merge, an issue is fine — but it should not be dropped, because the failure mode is a user following the documentation and getting an execution error with no hint that their `access_mode` is the reason.

CI fully green including `Test` at 37m49s, and bakey has already approved. I reviewed the re-land argument, the UDF registration and the open thread; I did not read all 2153 lines.

BtXin:
Addressed in `b792cae` — and the finding was right in a way worth naming: **that limitation was already written in the code, at `supports_filters_pushdown`, and never reached the guide anyone actually reads.** Documenting it next to the implementation is not documenting it.

I verified the claim rather than taking it on reading, since #259 merged between the two reviews and I wanted to be sure which behaviour the branch actually has. Against a real `pgvector/pg16`:

- read-write registration → fails with `to_tsvector is evaluated by PostgreSQL; this query was not pushed down`
- read-only registration → returns the expected row

### What the docs now say

The section states the requirement, the symptom, and the **reason** — because "use read-only" without the why invites someone to "fix" the clamp later:

> A read-write source is served by a provider that also implements `UPDATE` and `DELETE`, and DataFusion builds those statements' `WHERE` from the filters it hands those methods — which it only has while the filter node is still in the plan. So that provider reports its filters *inexact*: they still reach PostgreSQL, but DataFusion also keeps a local copy. For an ordinary predicate that costs a redundant re-check. For these functions it is fatal, because the local copy invokes them and they refuse to run. Reporting them *exact* instead would delete the filter node and silently turn `UPDATE … WHERE id = 1` into an `UPDATE` of the whole table.

So: a deliberate trade, not a gap awaiting closure. That last sentence is not hypothetical — it is the bug #259's review caught in my first version of the clamp.

**`pg_fts` works in both modes**, and the guide now says so as the answer for sources that must stay writable. I checked the mechanism rather than assuming: it resolves a pooled connection from the registry and returns its own `PgFtsProvider`, never passing through `SqlxPostgresTableProvider`.

I also cross-referenced it from the catalog-mode `access_mode` notes, so a reader arriving from the configuration side meets the cost before paying it.

### Two tests, not just prose

Documenting a failure mode without observing it is exactly how the unqualified claim got written, so both directions are pinned as `#[ignore]`d container tests beside the existing Postgres suite CI runs with `--ignored`:

- the predicate **fails** against a read-write registration, asserting on the refusal's cause
- and **works** against a read-only one, returning the expected row — without which "use a read-only registration" is advice nothing verifies

**20/20 serial** against a real pgvector/pg16 (the existing 18, plus these two). `cargo test -p skardi --lib` → 1218 passed, 0 failed.

PR #253, opened 2026-09-15
feat(cli): run and pipeline are available against a cloud gateway

BtXin:
**2 of 2 in a stack, based on #252.** Independently mergeable — see "Why this no longer waits on anything".

## Why

`Capability::served_by_gateway` is `matches!(self, Query | Schema)`, and `ensure_available` runs **before** `ApiClient::new` — so `skardi run <name>` and `skardi pipeline list` are refused client-side against a cloud context no matter what the gateway serves. Live:

```
$ skardi health
error: 'health' is not available in a cloud context (internal). Available: query, schema.
```

Everything about transport was already correct: `config.rs` states "`mode` never selects a URL — the gateway mounts skardi-server's own paths", the client already sends `Authorization: Bearer` plus the workspace header on every request, and `commands/run.rs` / `commands/pipeline.rs` already target `/{name}/execute`, `/pipelines` and `/pipeline/{name}` — exactly the paths the cloud gateway now mounts. Only the predicate changed.

## Why this no longer waits on anything

An earlier revision of this PR was marked "do not merge until skardi-cloud deploys", because against a gateway without the pipeline routes the commands 404. That was the wrong call, and the repository owner rejected it: **an OSS change must not be gated on a private deployment's rollout schedule.** The CLI ships to anyone running any skardi deployment.

The concern underneath it was real but narrow, and belonged in this repository anyway. `commands/run.rs` remapped *any* 404 to `"pipeline '{name}' not found"` — so against a server with no pipeline surface, a missing **route** was reported as a missing **pipeline**. That is a wrong statement about a server the CLI can simply ask.

`6707817` fixes it: on a 404 from `POST /{name}/execute`, the CLI probes `GET /pipelines`. If that 404s too, the server does not serve pipelines and says so; if it succeeds, the name genuinely does not exist and the old message stands. `pipeline list`'s own 404 *is* that probe, so it answers directly. One extra request, only on a path that already failed.

The message names no product and no version — the honest statement is about the server that answered, not about a class of deployments. Which is the same reason the client-side capability table was the wrong place to encode this.

## What

**`803f240`** — `served_by_gateway` gains `Run` and `Pipeline`. `ALL_CAPABILITIES` already derives the user-facing "Available:" list from that predicate, so no second list was written. `Job` and `Health` stay refused as engine-local surfaces, pinned by a test.

`mcp` stays refused, and **its reason has changed**. The old comment said `mcp` needs "pipeline execution and `/pipelines`, which the gateway does not mount" — obsolete. The cloud gateway serves its **own** `/mcp` with its own tool catalog, and the console documents that as the cloud agent path, so the CLI's stdio bridge would be a second, redundant surface rather than a missing one. Stale reasoning that reaches the right answer is worse than a wrong answer, because it survives review.

**`6707817`** — distinguishes "this server has no pipeline surface" from "this pipeline does not exist", so the CLI reports what is true of the server in front of it rather than guessing. Five new tests.

**`fe52d14`** — describes the cloud pipeline commands by behaviour rather than by a skardi-cloud sha no reader of this repository can resolve, and drops the present-tense promise that a gateway serves them. The text now reads correctly both before and after the cloud side deploys: the CLI no longer refuses these commands, and whether a given gateway serves them depends on its version.

## Verification

Gate tests mutated **both directions** in review: reverting the predicate fails three tests, widening it to admit `Job` fails three. `cargo test -p skardi --lib` 1203 / 0 · `cargo test -p skardi-cli` 302 / 0 · fmt clean.

One known cosmetic follow-up, not blocking: `the_at_at_operator_type_checks_over_the_two_string_returns` (in #252) over-claims relative to its now-corrected doc comment and wants a rename.


BtXin:
P2 addressed in `c7f0c29`, and you were right that the probe did not cover it.

`serves_pipelines` answers *"does this deployment have a pipeline API"*. A gateway that serves discovery but not execution — the exact case `cloud.rs` admits `Run` for — answers **yes**, so the friendly message was still produced and the gateway's own error was still discarded.

A 404 now resolves by what the server says, using the `error_type` `ApiError::Http` already carries:

- **`pipeline_not_found`** → the route ran and the pipeline is genuinely absent. Friendly message, and no probe at all (the server already told us).
- **any other 404 with a pipeline surface present** → genuinely ambiguous between an absent pipeline and an absent route. I did not swap one unproven claim for another: the message names *both* possibilities and preserves the server's own text, e.g.

  ```
  POST /real-pipeline/execute returned 404: no route for POST /real-pipeline/execute
  This server does serve pipelines, so either 'real-pipeline' does not exist or this
  deployment has no execute route (a gateway predating 'skardi run' answers that way).
  Run 'skardi pipeline list' to see which pipelines it knows.
  ```
- **any 404 with no pipeline surface** → unchanged.

Three existing tests asserted `contains("pipeline 'ghost' not found")` for an **untyped** 404 — the precise claim this change forbids — so I rewrote them rather than patching the assertion:

- the missing-pipeline test now mocks the typed error a real server sends (`pipeline_handlers.rs` emits `pipeline_not_found`);
- the two probe tests keep their real purpose — an inconclusive probe is not proof the server lacks pipelines — and now *also* assert the message does not claim absence, which is exactly what the probe could not establish;
- a new test covers your case: a route-level 404 from a gateway whose discovery works, asserting the gateway's words survive and the pipeline is not declared missing.

Mutation-verified: letting any typed 404 take the friendly arm fails four tests. `cargo test -p skardi-cli` → 287 passed, 0 failed.

bakey (approved):

PR #255, opened 2026-09-15
feat(mcp): let a caller declare the task a query belongs to

Userneima:
## What this is

`ai_context.purpose` describes one statement, because a ledger row is one statement. What a *run* of statements was for has nowhere to go, and it cannot be recovered downstream. Measured on a real day: 69 queries whose purposes were all present and specific still summarized as a list of lookups, because "a month-long cross-repo delivery review" was not a fact in any row.

This adds an optional `task` alongside `purpose`, one line naming the larger piece of work, repeated verbatim on every query that serves it. It lands on both the MCP `query` tool and the CLI.

**The engine needs no change.** `ai_context` is a free-form object it size-checks without constraining its keys, so a caller can send `task` against an unchanged server today.

## The two commits

**`feat(mcp)`** puts `task` in the shared tool schema (`mcp-core/src/projection.rs`) and threads it through **both** bindings. They share that schema but assemble the `/query` body separately, so adding the field on one side only would have advertised it to every host and then silently dropped it on the other. That is exactly the drift the projection module's own header says the shared surface exists to prevent.

**`feat(cli)`** adds `skardi query --task`, so the two ways an agent reaches the server carry the same context.

`task` rides only alongside a `purpose`, in both places. The server takes `ai_context` whole or not at all, so synthesizing a partial object from a task alone would turn a query that would have answered into a 400. clap enforces it on the CLI (`requires = "purpose"`), and the MCP handlers only insert the key inside the existing purpose branch.

## Who reads it

Nothing on this side. The consumer is the daily brief in skardi-cloud, whose companion PR carries it into the digest payload and the prompt. Merging this alone is inert and safe.

## Verification

`cargo test -p skardi-server --lib mcp::` (19), `-p skardi-mcp-core --lib` (12), `-p skardi-cli --bins` (277). Four new tests: the schema advertises the field; the task reaches `ai_context` on the HTTP binding; a task without a purpose builds no partial object; the CLI's `build_ai_context` adds no key when none was given. `cargo fmt --check` clean.


Userneima:
Companion: SkardiLabs/skardi-cloud#647, which reads the field into the digest payload and the prompt. Either can merge alone; this one is inert until something sends `task`, and that one is inert until this ships.

Userneima:
Closing this in favour of an issue first: the shape of the field, and whether it belongs in `ai_context` at all, is worth agreeing before there is code to argue with. The branch `Userneima:feat/mcp-query-task` stays pushed, so this reopens as-is if the issue lands on the same design.

Userneima:
@BtXin 这条和 SkardiLabs/skardi-cloud#647 是一件事的两半:让调用方能声明一批查询是为哪件活做的(这条,引擎侧加 task 字段),再把它连同账本里每条语句的事实一起送进日报提示词(#647)。codex 先过了一轮,查出两个 P1,已经修了。

想请你定方向:字段名 task、以及放 ai_context 合不合适(现在没东西写它,改还来得及),这套要不要做。要正式 review 可以 @codex。背景在 issue #256。

Userneima:
@BtXin This and SkardiLabs/skardi-cloud#647 are the two halves of one change: let a caller declare the task a run of queries serves (this PR, the engine-side `task` field), and carry that plus each statement's own facts into the daily brief (#647).

Codex reviewed twice. The first round found two P1s, since fixed; the second raised two more, both now addressed in `6d796dd`:
- the per-field caps did not bound the whole object, so a long purpose + task could pass client-side and 400 on the server — `build_ai_context` now checks the serialized byte budget locally;
- `--task` and the MCP `query` property were undocumented — README, docs/cli.md and docs/mcp.md now cover them.

For direction: is `task` the right name, and is `ai_context` the right home (nothing writes it yet, so it is still free to change)? Background in #256. Re-review with `@codex review` whenever you want it.

BtXin (approved):

PR #259, opened 2026-09-16
fix(postgres): the read-write provider must forward filter pushdown

BtXin:
`SqlxPostgresTableProvider` — the read-write wrapper — forwards `schema`, `table_type`, `constraints` and `scan` to its read provider, but **not `supports_filters_pushdown`**. `TableProvider`'s default answers `Unsupported` for every filter, so DataFusion never handed the wrapper's `scan` any filters at all: it kept a `Filter` node above the scan and evaluated every predicate locally, after dragging the whole table across the wire.

**A Postgres source registered read-write pushed down nothing** — not `WHERE id = 1`, nothing — while the same source registered read-only (a bare `SqlTable`, which does implement this) pushed down normally. Same source, same SQL, different plan, decided only by whether writes were enabled.

### The fix: forward the verdict, clamped to `Inexact`

The clamp is the whole subtlety, and the first version of this PR got it wrong — see the thread. This provider also serves UPDATE and DELETE through `TableProvider::update` / `delete_from`, and DataFusion builds their `WHERE` from the filters it hands those methods — which it only has while the `Filter` node is still in the plan. An `Exact` verdict tells the optimizer the scan applies the predicate itself, `PushDownFilter` removes that node, and the DML then runs **unfiltered**:

```
-- written:
UPDATE users SET email = '…' WHERE name = 'Alice Smith'
-- reached Postgres as:
UPDATE public.users SET "email" = '…'
```

Every row. `DELETE` degraded identically. `Inexact` keeps the `Filter` node so the DML planner still has its predicate, while the scan still receives the filters and still renders them into the SQL sent — confirmed in the postmaster log:

```sql
SELECT "users"."id", "users"."name" FROM "public"."users" WHERE ("users"."id" = 2)
```

The cost is a redundant local re-check of rows Postgres has already narrowed.

This matches the rule [`dynamodb.rs`](https://github.com/SkardiLabs/skardi/blob/main/crates/skardi/src/sources/providers/dynamodb.rs#L342-L345) already documents, "matching the MongoDB provider". It is a house rule for any provider serving DML through these methods.

### What this does and does not fix

**Does:** read-write registrations now push predicates into the SQL they send, instead of pulling whole tables.

**Does not:** make full-text search work through a read-write registration. Because the `Filter` node survives, a predicate that cannot be evaluated locally still fails — the Postgres FTS scalars are registered to *error* rather than return NULL (a NULL-returning stub would make `NULL @@ NULL` filter everything out and answer "no matches", indistinguishable from a genuine empty result). **FTS needs a read-only registration**, which is what a serving engine uses. The limitation is documented at the method.

### Tests

- The verdict mapping, as three named `rstest` cases: `Exact` → `Inexact` (so DML keeps its predicate), `Inexact` and `Unsupported` unchanged.
- One verdict per filter, in order.
- **The consequence**: a predicate must *reach* the read provider's `scan`. Checking the verdict alone only tests the accessor.
- **The regression guard**: the `FilterExec` must survive in the physical plan. This is the cheap, database-free test that would have caught the full-table `UPDATE`.

All mutation-verified. The live proof is the `#[ignore]`d Postgres suite, which CI runs: **18/18 serial**, matching `main`; ten of them failed under the unclamped version.

`cargo test -p skardi --lib` → **1192 passed, 0 failed**.

Adds `rstest` to this crate's dev-dependencies — the workspace already pins it, but `crates/cli` had been its only consumer.

### Scope

Only the pushdown forwarding. The wrapper's other omissions relative to `TableProvider` (`statistics`, for one) are untouched — each deserves its own argument; this one is a correctness bug with a live symptom.


BtXin:
Both P1s addressed, and the CI failure turned out to be a **serious bug in my own first commit** — worth reading before re-reviewing, because the PR's claim has changed.

### The CI failure was not flaky

Ten of the eighteen `#[ignore]`d Postgres tests failed. They pass on `main` (18/18 serial) and failed at 8/10 here, so I bisected rather than assumed. The postmaster log gave it away:

```
-- the test issues:
UPDATE users SET email = '…' WHERE name = 'Alice Smith'
-- Postgres received:
UPDATE public.users SET "email" = '…'
```

No WHERE. **Forwarding `Exact` turned every filtered UPDATE and DELETE into a full-table one.** This provider serves DML through `TableProvider::update` / `delete_from`, and DataFusion builds their WHERE from the filters it hands those methods — which it only has while the `Filter` node is in the plan. `Exact` tells the optimizer the scan applies the predicate itself, so `PushDownFilter` removes that node and the DML runs unfiltered.

The tests were doing their job; I had read `202 ignored` locally as noise.

### The corrected fix: clamp to `Inexact`

`Inexact` keeps the `Filter` node, so the DML planner still has its predicate, while the scan still receives the filters and still renders them into the SQL sent — confirmed in the log: `SELECT "users"."id", "users"."name" FROM "public"."users" WHERE ("users"."id" = 2)`. Cost is a redundant local re-check of rows Postgres already narrowed.

This is the rule [`dynamodb.rs`](https://github.com/SkardiLabs/skardi/blob/main/crates/skardi/src/sources/providers/dynamodb.rs#L342-L345) already documents for exactly this reason ("matching the MongoDB provider"). It is a house rule for any DML-serving provider, and my first commit broke it.

**The PR's scope is now narrower and I have corrected the description.** It fixes the performance half — read-write registrations pushed down *nothing* at all. It does **not** make full-text search work through a read-write registration: the surviving `Filter` node evaluates the FTS scalars locally, where they error by design. FTS needs a read-only registration, which is what a serving engine uses. That limitation is documented at the method.

### On the two P1s — both correct, both my miss for not reading AGENTS.md first

- **Imports at module scope**: done, no inline paths.
- **rstest cases**: the verdict loop is now three named `#[case::…]` parameters. Note `rstest` was *not* a dev-dependency of `crates/skardi` (only `crates/cli` used it, and its Cargo.toml says it was "the first one to"), so this PR adds it there.

Also added the database-free guard that would have caught the full-table UPDATE: it asserts the `FilterExec` survives in the physical plan. Mutation-verified — removing the clamp fails both it and the clamp case.

`cargo test -p skardi --lib` → 1192 passed, 0 failed. Ignored Postgres suite → 18/18 serial, matching main.

BtXin:
@codex

bakey (approved):

